PR #9, opened 2018-06-14
pin rust version in travis

### LINKED ISSUE #7: decide on a specific version of rust to dev against

thedavidmeister:
fixes #7


zippy:
I did approve this for travis, but just FYI, my emacs toolchain **requires** nightly because the regular version doesn't have enough debugging options.

PR #12, opened 2018-06-14
dna structures and serialization

neonphog:
There is more to do here, and more tests to write, but would love to start getting some feedback!

@lucksus @zippy


lucksus:
Wow, nice work so far!

thedavidmeister:
can we merge this so that other branches can build on top of it?

PR #13, opened 2018-06-15
63 library interface

zippy:
Ok here's a first pass at a public API with just one function `new`. 

The idea is that it wraps the internals of the redux implementation so users of the library don't have to know about that an can just have simple functional access.  The idea is that you instantiate an app with a call to:

`let app = App::new(your_dna).unwrap()`

And then you'll be able to do things like call a function in the app with:
`app.call(fn_name,fn_params)`

feedback appreciated on the approach.


zippy:
I love how the examples in your docs also get run by `cargo test`  really nice!

hc_core/src/error.rs
```diff
@@ -0,0 +1,62 @@
+use std::error::Error;
+use std::fmt;
+
+/// starter module for holding Holochain specific errors
+
+#[derive(Debug, PartialEq)]
+pub struct HolochainError {
+    details: String,
+}
```
neonphog: I've seen errors handled pretty well as enums in rust, which lets us provide composite results, and make better use of the trailing `?` upgrade operator... so, something like:

```rust
pub enum HolochainError {
  ErrorGeneric(String),
  SerdeError(serde_json::error::Error),
}

impl From<serde_json::error::Error> for HolochainError {
  fn from(err: serde_json::error::Error) -> Self {
    HolochainError::SerdeError(err)
  }
}
```

and then our own function like:

```rust
pub fn my_func() -> Result<(), HolochainError> {
  // call to fn that returns serde_json::error::Error can be upgraded to return holochain error
  let result = serde_json::to_string(something)?;
  // if error, we returned already, so we don't have to unwrap or anything
  // now we can just use result
}
```

hc_api/src/lib.rs
```diff
@@ -0,0 +1,78 @@
+/*!
```
neonphog: do we want to standardize on `/*!` or `//!`?

zippy: well for me `//!` feels like a pain because I have to redo it if I want to change line breaks...  But maybe I should have tool to fix that?  Thoughts?

hc_core/src/error.rs
zippy: great idea, I'll change it.

zippy: ok @neonphog just pushed a commit with Enum based implementation.

hc_api/src/lib.rs
neonphog: awesome, let's go with `/*!` - i'll change the dna stuff to that for my next commit

hc_core/src/error.rs
neonphog: nice, looks good :+1:

zippy:
@lucksus I updated `hc_core` and `hc_api` to use the DNA from `hc_dna`, FYI I still haven't removed the other parts of `hc_core/src/nucleus/dna.rs`, but this will need some merging into your wasm stuff...

hc_api/src/lib.rs
ddd-mtl: FYI the Rust API conventions states:
```
Only use inner doc comments //! to write crate and module-level documentation, nothing else. When using mod blocks, prefer /// outside of the block
```

zippy:
Ok folks, I added a `Context` object to the api that allows us to separate out all those things that the core needs to run that come from the external world and would be implemented differently, or used independently.  So for example, 

1) the agent to be used could also be used by an agent management app totally separate from core, so I broke that into it's own crate.
2) the logger that the core should use to output logging info will totally depend on the context, so its implemented as a trait.  I wrapped that in an <Arc<Mutex<>> so that it will work across threads...  I actually want to turn that into a channel receiver too.

feedback appreciated, possibly this bit should go into an ADR.

zippy:
BTW I converted the test_bin app to use the api, check it out at: https://github.com/holochain/holochain-rust/blob/63-library-interface/hc_test_bin/src/main.rs

hc_api/src/lib.rs
```diff
@@ -0,0 +1,259 @@
...
+    pub fn stop(&mut self) -> Result<(), HolochainError> {
+        if !self.active {
+            return Err(HolochainError::InstanceNotActive);
+        }
+        self.active = false;
+        Ok(())
+    }
+
+    /// call a function in a zome
+    pub fn call(&mut self, fn_name: &str) -> Result<(), HolochainError> {
```
lucksus: What about function parameters?

hc_api/src/lib.rs
```diff
@@ -0,0 +1,259 @@
...
+use hc_core::state::Action::*;
+use hc_core::state::State;
+
+impl Holochain {
+    /// create a new Holochain instance
+    pub fn new(dna: Dna, context: Arc<Context>) -> Result<Self, HolochainError> {
+        let mut instance = hc_core::instance::Instance::new();
+        let action = Nucleus(InitApplication(dna.clone()));
+        instance.dispatch(action);
+        instance.consume_next_action()?;
```
lucksus: So this will change in my branch, but I think we should finish and merge this one first and and then I will merge develop into my branch and adjust..

hc_core/src/nucleus/fncall.rs
```diff
@@ -0,0 +1,19 @@
+/// Represents a function call, with the function name and its parameters
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct Params {}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct Call {
```
lucksus: Hm, I have basically the same thing in my ribosome branch.. we should take care not to duplicate work.

hc_core_api/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+    pub fn stop(&mut self) -> Result<(), HolochainError> {
+        if !self.active {
+            return Err(HolochainError::InstanceNotActive);
+        }
+        self.active = false;
+        Ok(())
+    }
+
+    /// call a function in a zome
+    pub fn call(&mut self, fn_name: &str) -> Result<(), HolochainError> {
```
lucksus: What about function parameters?

zippy: yep, just didn't get to it...

zippy: also return value needs to be the encapsulated return value from the function call...

hc_core/src/logger.rs
```diff
@@ -0,0 +1,24 @@
+/// trait that defines the logging functionality that hc_core requires
+pub trait Logger {
+    fn log(&mut self, msg: String);
+    fn read(&self) -> String;
+}
+
+#[derive(Clone, Debug)]
+pub struct SimpleLogger {
+    //    log: Vec<String>,
+}
```
neonphog: I think it would be worth looking into https://crates.io/crates/log

provides some convenience macros like

```rust
trace!("doing a thing: {}", something);
warn!("something bad: {}", something);
```

and has a plugable backend that we could swap out if needed. could start with https://crates.io/crates/simple_logger

zippy: Yah, I looked at that this morning.  It works fine for our global debug logging, and I do think we should use it for that, but it doesn't work for per-app logging.  I created a ticket #21 regarding this distinction.  But it certainly will help us distinguish between container level logging and application level logging.

hc_core/src/instance.rs
```diff
@@ -15,16 +17,18 @@ impl Instance {
         &self.pending_actions
     }
 
-    pub fn consume_next_action(&mut self) {
+    pub fn consume_next_action(&mut self) -> Result<(), HolochainError> {
         if self.pending_actions.len() > 0 {
+            // TODO: convert the unwrap to return a converted error
```
thedavidmeister: followup please!

hc_core/src/nucleus/fncall.rs
thedavidmeister: @lucksus yeah i saw that in your screen share, a little duplication of work is ok while everything is so new, but we should agree on the details :)

thedavidmeister:
@zippy other than the merge conflict and dangling TODO, this looks great

i'd rather ship all the good work you've done, then try to actually use the persistence trait in context to form an opinion, than try to intellectualise it too much up front here

hc_core/src/state.rs
```diff
@@ -9,25 +10,27 @@ pub enum Action {
...
+            nucleus: Rc::new(NucleusState::new()),
             agent: Rc::new(AgentState::new()),
         }
     }
 
-    pub fn reduce(&mut self, action: &Action) -> Self {
-        State {
-            nucleus: ::nucleus::reduce(Rc::clone(&self.nucleus), action),
+    pub fn reduce(&mut self, action: &Action) -> Result<Self, HolochainError> {
+        let nucleus = ::nucleus::reduce(Rc::clone(&self.nucleus), action)?;
```
lucksus: There it is again ;)

hc_core_api/src/lib.rs
lucksus: As I said verbally, I am implementing all the stuff you need so that we can actually call zome functions. For that I have to change the API offered by Instance. So the body of this function will have to change anyway. But I would like to get your changes in first because of the renaming you did. So I would suggest we merge this branch first without having this working actually and then I go and merge this (from develop) into my branch and create a PR for that..

So, what I say is: make the signature of this final but don't bother implementing it in a functional way.

PR #14, opened 2018-06-15
sketch of source chain methods based on go version

thedavidmeister:
very basic, ready for review:

- Defines `Pair`
- Defines `SourceChain` trait
- Defines basic in-memory source chain implementation
- Extends `Entry` to support string content
- Enforces immutable `Entry`
- Extends `Header` to support previous header hash and entry hash
- Enforces immutable `Header`
- Implements default rust hashing for `Entry` and `Header`
- Implements `IntoIter` for source chains, which allows basic loop/map/filter over pairs
- Implements `push()` for source chains, to append new pairs
- Tests for iteration and append to source chain

followups:
- #20
- #23
- #24
- #25


thedavidmeister:
summary of chat with @zippy 

- pluggable/enum based serialize/deserialize
  - plan to support JSON, graph vis, cap'n proto, etc.
- simplify struct to be a sequence of headers, map of entries and reference to agent
- simplify API/methods to remove everything that was only juggling extra internal complexity in go version
- anything that looks like queries, file system read/write, databases, networking, goes somewhere else and uses this API (e.g. the go version has file system and JSON handling in the same place as the chain structure/behaviour definition)

thedavidmeister:
been thinking about this a bit more, some ideas...

- source chain could be a trait with a minimal set of methods, allowing memory/file/db implementations, something like:
  - query
    - based on enum, support all current get/top/nth methods
  - add
  - validate
  - serialize/deserialize
  - agent
- functionality to build headers/pairs for an entry to be added could be a method on the entry struct
- all source chain methods work with valid header/entry pairs for i/o

thedavidmeister:
so the source chain trait should sit on top of core traits like `std::iter::Iterator`

e.g. it should be possible to `chain.iter()` from the top to the bottom, following header references all the way down.

scope of this ticket:

- implement a trait for chains
  - `Iterator` of `Pair`
  - `push` new `Pair` (e.g. vector)
- implement a basic in-memory chain using this trait
- basic tests

followup:

- serialization/deserialization
- header/entry lookup by hash
- queries
- hashing
- validation
- disk backed

thedavidmeister:
looking into what structural sharing options there are for rust...

zippy:
Yes, I think this is correct.  For this ticket we should just implement the in-memory part of the chain. (though at some point we may also implement here the wire-format serialization...)

I have just added a `Persister` trait to the Context object of `hc_core` to abstract out how in different contexts holochain core will persist its state (including the chain data and dht date) in different ways. With that trait we can implement the functions it needs to get the work done, and then the actual persistence can be implemented separately, i.e. as persistence to files, database, cookies, a separate Holcohain app, etc..

Issue #29, opened 2018-06-21
port all chain.go functionality and tests


thedavidmeister:
## Checklist

A checklist to keep track of the `chain.go` basic port to the `source_chain` module:

- [x] a concept of a "chain" of headers and entries
  - implemented as `source_chain::Pair`
- [ ] chains can be saved/loaded from the file system #231 
- [ ] headers and entries use the correct hashing algo #20 
- [x] headers and entries reference each other by hash
  - implemented as `common::entry::Entry` and `common::entry::Header`
- [x] helper fn `top` to get top header (most recently added)
- [ ] helper fn `nth` to get nth most recently added header
- [x] helper fn `TopType` to get most recently added header of given type
- [ ] ability to provide an entry + key and have it appended with correct headers + signatures
- [x] helper fn `get` to lookup header by hash: #30
- [x] helper fn `getEntry` to lookup entry by entry hash: #30
- [x] helper fn `getEntryHeader` to lookup header by entry hash: #30
  - this should be merged with `getEntry` if all lookups return pairs
- [x] ability to serialize/deserialize a chain: #61
- [ ] ability to "dump" a chain to a string
- [x] ability to walk a chain from most recent to least recent
  - implemented move and ref `for` loop as `chain.iter()`, `(&chain).into_iter()` and `chain.into_iter()`
- [x] ability to validate a chain: #31  
- [x] ability to JSON stringify a chain #61 
- [ ] ability to graphvis a chain
- [ ] helper fn `Length` to get the number of entries in a chain
- [ ] helper fn `BundleStarted` to get the start index of the active bundle
- [ ] ability to start/close bundles

## Diff

summary of major differences planned/implemented between the go/rust versions:

- multiple implementations of the source chain trait, so memory/file/db chains can all be independently maintained
- dump/JSON/graphvis functionality will all come under the serialize/deserialize banner
- rust chains:
  - append entries (internally generating headers/pairs)
  - return pairs
  - perform direct lookups by hash
  - prefer composable iteration methods/traits over ad-hoc DSL (e.g.`filter` vs. `TopType` 😬 )
  - leverage immutability much more

## Ongoing

things that will likely come up again in the future, that are not solved in the go version currently.

- source chains that support concurrency

PR #38, opened 2018-06-21
reversed in-memory representation and iterator of SourceChain

### LINKED ISSUE #35: use a collection that allows more efficient inserting of pairs into source chains
currently we insert new pairs at the start of a vector, which is O(n)

other collection types can do O(1) or O(log n) for inserts

https://doc.rust-lang.org/std/collections/#sequences

Slavoj:
This seems to me to be the more natural form. 

Side effect:
resolves #35

### DIFF
--- a/hc_core/src/source_chain/memory/mod.rs
+++ b/hc_core/src/source_chain/memory/mod.rs
@@ -38,11 +38,9 @@ impl super::SourceChain for SourceChain {
     // @TODO - appending pairs should fail if hashes do not line up
     // @see https://github.com/holochain/holochain-rust/issues/31
     fn push(&mut self, pair: &super::Pair) {
-        // @TODO - inserting at the start of a vector is O(n), some other collection could be O(1)
-        // @see https://github.com/holochain/holochain-rust/issues/35
-        self.pairs.insert(0, pair.clone())
+        self.pairs.append(&mut vec![pair.clone()])
     }
-    // returns an iterator referencing pairs from top (most recent) to bottom (genesis)
+    // returns an iterator referencing pairs from bottom (genesis) to top (most recent)
     fn iter(&self) -> std::slice::Iter<super::Pair> {
         self.pairs.iter()
     }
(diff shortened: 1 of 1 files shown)


thedavidmeister:
@Slavoj we're deliberately iterating from top to bottom

from memory, this is how it works in the go version we're porting from, and follows the direction of headers referencing the hashes of previous items in the chain, so is a natural "next" in this sense (you can start at any pair and move to the next pair by looking up the hash #30)

i'd suggest that if we believe it makes sense to support chronological iteration, that we provide iterators for both directions

thedavidmeister:
although, admittedly, it's pretty confusing that the pair calls the "next" pair from iteration the "previous" hash :P

Slavoj:
@thedavidmeister not confusing at all and I understand the reasoning for the iteration in this direction. Makes totally sense.

I just was curious about the code and looked for the issues as an exercise. For issue #35 I could not imagine a better collection other than appending a Pair to Vec (O(1)) and instead of changing the collection, revert the iterator direction to satisfy the old behaviour. But given the existing `rev` method of `Iterator` and my missing insight in your architecture I felt better with skipping the implementation work for the new iterators :-)

I am looking forward to the new rust holochain implementation. Keep up the good work!

thedavidmeister:
@Slavoj ah, well, for example we could do a hashmap with `O(1)~*` for inserts and `O(1)~` for `get` by hash and a single "top" pair, which can then be iterated from... but then we lose efficient `nth` lookup by index...

in fact, the go version has several parallel internal representations of the same data!

https://github.com/holochain/holochain-proto/blob/develop/chain.go#L52

```go
type Chain struct {
	Hashes   []Hash
	Headers  []*Header
	Entries  []Entry
	TypeTops map[string]int // pointer to index of top of a given type
	Hmap     map[Hash]int   // map header hashes to index number
	Emap     map[Hash]int   // map entry hashes to index number
```

that approach gives high theoretical performance but then also higher complexity (bugs) and is coupled to a specific implementation...

the plan moving forward (i've discussed this with @zippy ) is to focus on a simple, implementation independent API for working with a source chain. We can always layer performance optimisations (e.g. caches and indexes) on top of that for specific implementations (e.g. file system, databases, etc.) _after_ we have some evidence that there is a problem somewhere.

actually i'd love a hand with things @Slavoj if you're interested i've put together a simple roadmap for the `chain.go` port:

https://github.com/holochain/holochain-rust/issues/29

Slavoj:
@thedavidmeister I am happy to help, but can not commit too much time or rust experience. So my first priority would be to not slow you down.

But if you create a simple roadmap anyway or if you maybe put some TODOs in the code when the trait/API grows, I will try to follow the development and come back to you if I see a chance to help.

PR #44, opened 2018-06-23
62-client-side-wrapper, closes org:62

zippy:
Here's a spike on the beginning of org:#62

the parameters to be passed into the zome function are now actually a param to nucleus::ribosome.call() and are set into the memory.

the return val of the zome function called is assumed to be a serialized vector starting at a particular constant offset in memory, and the return value of the wasm call itself is the length that vector,  currently I'm just reading that vector and turning it into a string.

This protocol won't quite work because it means if the input parameters are long they might overlap the constant starting place of the return value(s) so you couldn't start writing there till you've consumed the input.

@lucksus does this look like the approach you were thinking?  Next step would be implementing the de-/serialization on both sides, i.e. rust/wasm.


demo/wasm/src/lib.rs
```diff
@@ -11,16 +20,68 @@ pub extern "C" fn _call(input_data: *mut u8, input_len: usize) -> i32 {
...
+    len as i32
+    //unimplemented!()
+}
 
-    return 0;//a * a;
+#[no_mangle]
+pub extern "C" fn test_dispatch(data: *mut c_char, params_len: usize) -> i32 {
+    let input = make_internal(data);
+    let output = test(input);
+    make_external(data, params_len, output)
```
lucksus: Hm, so you are basically just writing over the same memory that you got the input through... which in our case starts at 0. And you're not checking the length, which could be be longer for the output than the input.. I would not expect this to work in the general case.. I believe we will have to implement a malloc() function that at least just increments an index by the number of allocated bytes..

zippy: yes added #65 to account for this.

zippy:
Hey folks, I think this is now far enough along to count as completed for the purposes of the sprint, would love your review of it so we can merge.

Issue #32, opened 2018-06-21
tests to show that entry/header/pair are immutable


thedavidmeister:

thedavidmeister:
@sphinxc0re is this actually possible?

i was expecting to use a `should_panic` but the private field errors are compile time, not panics

sphinxc0re:
There are no private fields or methods in tests. All tests can access everything

sphinxc0re:
Okay, then I don't get the question.

thedavidmeister:
this issue is to write a test that shows that Entry cannot be accidentally modified, i.e. the fields are private

there are private fields in tests, just not in sub modules, so we could test Entry as long as we put the test outside Entry

the issue i'm facing is that private fields are compile time errors, not panics, so `should_panic` doesn't cut it here

sphinxc0re:
You can't test for that. Rust ensures, that all private fields are immutable from the outside. So you just want to ensure that nobody is changing the visibility?

thedavidmeister:
yes, that i what i want, but understand it isn't possible

PR #51, opened 2018-06-26
validate pushes and support hash lookups in source chain

### LINKED ISSUE #30: source chain should support index lookups by header/entry hash as per get/getEntry/getEntryHeader from golang

### LINKED ISSUE #31: source chain should error if attempting to push pairs with invalid hashes

thedavidmeister:
fixes #30 
fixes #31 

changes:

- validation for headers/entries is simply `true` atm, to be revisited in #32
- `push` to a source chain now validates the pair, that the header references the old top pair, and the whole proposed new chain passes validation
- added `validate` fn for chains, pairs, entries, headers
- added `get` and `get_entry` to chain to lookup pairs by header/entry hash respectively
- tests
- pairs now validate that the header entry hash references the hash of the entry in the pair
- pairs now validate that their headers and entries are valid

followup:

- #52

### DIFF
--- a/hc_core/src/common/entry.rs
+++ b/hc_core/src/common/entry.rs
@@ -88,3 +88,31 @@ impl Header {
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct Hash {}
+
+#[cfg(test)]
+mod tests {
+    use super::Entry;
+    use super::Header;
+
+    #[test]
+    fn new_entry() {
+        let c = String::from("foo");
+        let e = Entry::new(&c);
+
+        assert_eq!(e.content(), c);
+        assert_ne!(e.hash(), 0);
+        assert!(e.validate());
+    }
+
+    #[test]
+    fn new_header() {
+        let e = Entry::new(&String::from("foo"));
+        let h = Header::new(None, &e);
+
+        assert_eq!(h.entry(), e.hash());
+        assert_eq!(h.previous(), None);
+        assert_ne!(h.hash(), 0);
+        assert!(h.validate());
+    }
+
+}
--- a/hc_core/src/common/entry.rs
+++ b/hc_core/src/common/entry.rs
@@ -32,6 +32,10 @@ impl Entry {
     pub fn content(&self) -> String {
         self.content.clone()
     }
+
+    pub fn validate(&self) -> bool {
+        true
+    }
 }
 
 #[derive(Clone, Debug, PartialEq)]
@@ -73,6 +77,10 @@ impl Header {
 
     pub fn hash(&self) -> u64 {
         self.hash
+    }
+
+    pub fn validate(&self) -> bool {
+        true
     }
 }
 
--- a/hc_core/src/common/entry.rs
+++ b/hc_core/src/common/entry.rs
@@ -34,6 +34,7 @@ impl Entry {
     }
 
     pub fn validate(&self) -> bool {
+        // always valid iff immutable and new() enforces validity
         true
     }
 }
@@ -80,6 +81,7 @@ impl Header {
     }
 
     pub fn validate(&self) -> bool {
+        // always valid iff immutable and new() enforces validity
         true
     }
 }
--- a/hc_core/src/source_chain/memory/mod.rs
+++ b/hc_core/src/source_chain/memory/mod.rs
@@ -39,8 +39,10 @@ impl super::SourceChain for SourceChain {
     // @see https://github.com/holochain/holochain-rust/issues/31
     fn push(&mut self, pair: &super::Pair) {
 
+        let previous_hash_lookup = pair.header.previous().and_then(|h| self.get(h));
+
         // smoke test this pair in isolation, and check the hash reference against the top pair
-        if !(pair.validate() && self.pairs.first() == pair.header.previous().and_then(|h| self.get(h)).as_ref()) {
+        if !(pair.validate() && self.pairs.first() == previous_hash_lookup.as_ref()) {
             // we panic because no code path should attempt to append an invalid pair
             panic!("attempted to push an invalid pair for this source chain");
         }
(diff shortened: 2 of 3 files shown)


hc_core/src/common/entry.rs
```diff
@@ -32,6 +32,11 @@ impl Entry {
     pub fn content(&self) -> String {
         self.content.clone()
     }
+
+    pub fn validate(&self) -> bool {
+        // always valid iff immutable and new() enforces validity
```
sphinxc0re: iffffff

hc_core/src/common/entry.rs
```diff
@@ -74,7 +79,40 @@ impl Header {
     pub fn hash(&self) -> u64 {
         self.hash
     }
+
+    pub fn validate(&self) -> bool {
+        // always valid iff immutable and new() enforces validity
```
sphinxc0re: iffffffff

hc_core/src/source_chain/memory/mod.rs
```diff
@@ -36,6 +36,25 @@ impl super::SourceChain for SourceChain {
     // @TODO - appending pairs should fail if hashes do not line up
     // @see https://github.com/holochain/holochain-rust/issues/31
     fn push(&mut self, pair: &super::Pair) {
+        let previous_pair = match pair.header.previous() {
+            Some(h) => self.get(h),
+            None => None,
+        };
```
sphinxc0re: Please take a look at `Option::and_then`. You could write the same in one line:
```rust
let previous_pair = pair.header.previous().and_then(|h| self.get(h));
```

hc_core/src/source_chain/memory/mod.rs
```diff
@@ -36,6 +36,25 @@ impl super::SourceChain for SourceChain {
...
+        let mut validation_chain = self.clone();
+        validation_chain.pairs.insert(0, pair.clone());
+        if !validation_chain.validate() {
+            // we panic because no code path should ever invalidate the chain
+            panic!("adding this pair would invalidate the source chain");
+        }
+
         // @TODO - inserting at the start of a vector is O(n), some other collection could be O(1)
         // @see https://github.com/holochain/holochain-rust/issues/35
         self.pairs.insert(0, pair.clone())
```
sphinxc0re: I don't think the clone is necessary here. Or is it? Am I missing something?

hc_core/src/common/entry.rs
thedavidmeister: if and only if

hc_core/src/source_chain/memory/mod.rs
thedavidmeister: this was already in the code, not sure if it's relevant to this branch?

hc_core/src/source_chain/memory/mod.rs
thedavidmeister: thanks, this is exactly what i was reaching for

hc_core/src/common/entry.rs
thedavidmeister: if and only if

hc_core/src/common/entry.rs
sphinxc0re: 😅 Sorry. Lesson learned!

thedavidmeister:
@sphinxc0re i don't mind empty lines, whatever you think is more idiomatic i'm happy to implement

please review my latest updates :)

hc_core/src/source_chain/memory/mod.rs
sphinxc0re: Yeah... this can probably wait

thedavidmeister: also i do get the compiler complaining at me about references when i drop it, so followup at #52

hc_core/src/common/entry.rs
thedavidmeister: it certainly is an odd word that looks like a mistake 😬

hc_core/src/source_chain/mod.rs
```diff
@@ -25,11 +33,27 @@ impl Pair {
...
+    pub fn validate(&self) -> bool {
+        self.header.validate() && self.entry.validate() &&
+        self.header.entry() == self.entry.hash()
+    }
+
 }
 
 pub trait SourceChain: IntoIterator {
+
+    // append a pair to the source chain if the pair and new chain are both valid, else panic
```
sphinxc0re: Please use doc comments here:
```rust
/// Some documentation...
fn bar_foo_bar();
```
This ensures that we will be able to generate documentation for those functions

thedavidmeister:
@sphinxc0re kk, give it a whirl now

sphinxc0re:
Again, good job! We should definitely look into removing all the `panic!`s from the repository and switch to `Result` types.

thedavidmeister:
@sphinxc0re why?

panic looks correct for an invalid source chain to me, from docs:

https://doc.rust-lang.org/book/second-edition/ch09-03-to-panic-or-not-to-panic.html#guidelines-for-error-handling

> It’s advisable to have your code panic when it’s possible that your code could end up in a bad state. In this context, a bad state is when some assumption, guarantee, contract, or invariant has been broken, such as when invalid values, contradictory values, or missing values are passed to your code—plus one or more of the following:

> - The bad state is not something that’s expected to happen occasionally.
> - Your code after this point needs to rely on not being in this bad state.
> - There’s not a good way to encode this information in the types you use.

> If someone calls your code and passes in values that don’t make sense, the best choice might be to call panic! and alert the person using your library to the bug in their code so they can fix it during development. Similarly, panic! is often appropriate if you’re calling external code that is out of your control and it returns an invalid state that you have no way of fixing.

sphinxc0re:
Are you up for a quick meet in the tealroom about this?

thedavidmeister:
sure

PR #55, opened 2018-06-26
Hc core api c bindings

lucksus:
I need this stuff for https://github.com/holochain/holosqape.


zippy:
I did approve this but conflicts need to be fixed before merge.

hc_core_api_c_binding/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+#[derive(Clone, Debug)]
+struct NullLogger {}
+
+impl Logger for NullLogger {
+    fn log(&mut self, _msg: String) {}
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn hc_new(ptr: *mut Dna) -> *mut Holochain {
+    let agent = Agent::from_string("c_bob");
```
neonphog: rust ffi docs recommend catching panics as they result in undefined behavior... although panics seem pretty bad anyway, so may not be a big issue?

https://doc.rust-lang.org/book/first-edition/ffi.html#ffi-and-panics

PR #58, opened 2018-06-27
Refactored naming and directory structure

sphinxc0re:
## Summary
- dropped the `hc_` from all directories
- renamed all crates from `hc_*` to `holochain_*` for clarity
- fixed c bindings

My reasoning here is that our crates are in a `holochain/%modulename%` folder. The corresponding crate should be named `holochain_%modulename%` accordingly


sphinxc0re:
I will merge after the build succeeds

PR #60, opened 2018-06-27
sit Dockerfile.test on top of the docker hub ubuntu image

### LINKED ISSUE #59: sit test docker on top of ubuntu docker (use docker hub image)

thedavidmeister:
fixes #59

### DIFF
--- a/docker/Dockerfile.ubuntu
+++ b/docker/Dockerfile.ubuntu
@@ -24,5 +24,3 @@ WORKDIR /holochain
 
 # run a build here to cache dependencies in a docker layer
 RUN cargo build
-
-CMD make test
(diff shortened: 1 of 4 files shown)


docker/Dockerfile.ubuntu
```diff
@@ -24,5 +24,3 @@ WORKDIR /holochain
 
 # run a build here to cache dependencies in a docker layer
 RUN cargo build
```
sphinxc0re: Why does this have to happen in this image? Do I have to recompile the whole image when running tests?

thedavidmeister: @sphinxc0re this make sure all the dependencies are already in the image that gets pushed to docker hub, so when you use `Dockerfile.test` to run tests, you don't have to wait for cargo

sphinxc0re: Nice! 😄

Issue #48, opened 2018-06-25
decide and implement a system for support/deprecation of new/old rust versions


sphinxc0re:
child of #117 

As the README says, we are currently targetting Rust 1.26.2, which is very specific. I think we should be able to always target the current stable version of Rust since it is developed in a backwards compatible manner.

thedavidmeister:
@sphinxc0re thanks for opening this, there are a lot of opinions/discussion floating around here and it's good to have a place to summarise where things are at:

- as of today this was set to `nightly-2018-06-01` as some clippy failures from the other day were breaking travis builds, but i suppose i forgot the README, sorry!
- from what i've seen, the stable version of rust is missing things we want:
  - WASM support/tooling being developed on top of by @lucksus 
  - clippy/linting tools are tracking nightly and don't seem to work well against stable
- "backwards compatible" isn't all we might want, is it really OK to go to the length of hashing DNA/zome code so that even adding a comment "forks" a DHT/network, but then allow the rust version/dependencies of hc_core itself to change multiple times per month without being pinned/reviewed?
- for the future, we've discussed using something like nix to not just pin a few versions here and there, but allow/enforce fully deterministic builds for the whole hc stack, so that rust dependencies would have integrity checks analogous to zome integrity checks
- if we want to support multiple versions and environments, we can relatively easily enumerate the versions we support so that we can throw them at travis
- nightly versions of rust come out so fast and libraries tracking nightly lag enough that i saw travis breaking almost every morning last week, which hurt our productivity while the team had a lot of balls in the air, and it led to tests being temporarily turned off or ignored on travis twice, which is bad
- stable versions of rust come out infrequently enough that it's not hard to imagine us implementing a process where we bump and deprecate versions manually as they come and go
- we can definitely test against both stable and nightly on travis using the `allow_failures` flag on travis, so then at least we know ahead of time what we're up against :)
  - in some sense, we'd then be "targetting" both stable and nightly, but with the understanding that failures in a new version don't "stop the line" until we decide we're ready to bump officially supported versions

sphinxc0re:
My reasoning here is that the whole project doesn't need to use nightly builds. We are perfectly able to build and agent binary on stable Rust. The code that actually has to compile to Wasm could use nightly as a separate build path.

jamesray1:
[M4-core] P5-sometimesoon m0-build

thedavidmeister:
@jamesray1 wow, this has progressed a lot since then

[...]

PR #64, opened 2018-06-27
JSON round trip support for source chains

### LINKED ISSUE #61: serialize/deserialize for source chains

thedavidmeister:
fixes #61


core/src/lib.rs
```diff
@@ -1,6 +1,13 @@
 #![deny(warnings)]
+
+#[macro_use]
+extern crate serde_derive;
+extern crate serde;
+extern crate serde_json;
+
 extern crate holochain_dna;
 extern crate wabt;
```
sphinxc0re: I would order the import this order:
```rust
extern crate foo;
extern crate bar;

mod blah;
mod blahblah;
mod blahblahblah;

use foo::Bar;
use foo::Bang;
use foo::Boom;
use foo::Bang;
```
**But** we have *external* crates and *internal* crates, the ones that are in the repo. SO I'm thinking of having the external ones above the internal one with a gap in between.

core/Cargo.toml
```diff
@@ -9,4 +9,7 @@ holochain_agent = { path = "../agent" }
 chrono = "0.4"
 wasmi = { git = "https://github.com/paritytech/wasmi.git"}
 wabt = "0.2.0"
-snowflake = "1.2"
\ No newline at end of file
+snowflake = "1.2"
+serde = "1.0"
+serde_derive = "1.0"
+serde_json = { version = "1.0", features = ["preserve_order"] }
```
sphinxc0re: Wouldn't it be better to change the versions to just `"1"`?

thedavidmeister: well wouldn't that unpin the version and allow the minor version to float?

core/src/lib.rs
thedavidmeister: what are you wanting me to change here?

sphinxc0re: For example, `holochain_dna` is *internal*, so I would move it below the *external* crates with an empty line between

core/Cargo.toml
thedavidmeister: this was copied from elsewhere in the repo, i think this convo lives here: https://github.com/holochain/holochain-rust/issues/48

Issue #63, opened 2018-06-27
review whether it still makes sense to panic in source chain pushing once we have more sophisticated validation


thedavidmeister:
currently validation is _only_ based on hashes - pushing invalid hashes onto a source chain is a panicable offense, because hashes should never fail to line up at that point, that's _always_ a bug in core.

however, once validation can be farmed out to zomes, this may no longer be the case so we should return `Result` instead.

thedavidmeister:
@sphinxc0re FYI ^^

jamesray1:
It seems like validation can now be pushed out to zomes, as shown e.g. here:

https://github.com/holochain/holochain-rust/blob/fb13b213a1b3c5c15814aa00435c71428dbbf7d2/doc/holochain_101/src/zome/entry_validation.md

And it seems like validation functions now return a `ValidationResult` e.g. as shown here:

`pub fn validate_provenances(validation_data: &ValidationData) -> ValidationResult {`
https://github.com/holochain/holochain-rust/blob/aa53657ee25bd35304c177a8d5560318644b0ccf/crates/core/src/nucleus/validation/provenances.rs

Is this out of date?

[M4-core][Z3-stale]

thedavidmeister:
@jamesray1 yeah validation has been rewritten several times since this was opened, thanks

Issue #20, opened 2018-06-18
implement correct hashing algos


thedavidmeister:
e.g. https://doc.rust-lang.org/std/hash/trait.Hash.html

looks like rust natively supports data hashing itself

also this hashing is used by the native hashmap implementation IIRC, which would make lookups quite neat

thedavidmeister:
https://github.com/multiformats/rust-multihash looks interesting as a direct port of how the go version works

PR #69, opened 2018-06-28
entry types

### LINKED ISSUE #67: add type to headers in source chain
in the go version all headers had a type and there were helper fns like `TopType` that gave types first class support

i think we can add the type, but get away without forcing `TopType` into the source chain trait

thedavidmeister:
fixes #67 

changes:

- rename `source_chain` to `chain`
- dedicated module for `chain::SourceChain` trait
- moved and dedicated module for `chain::entry::Entry` and added tests
- moved and dedicated module for `chain::header::Header` and added tests
- moved and dedicated module for `chain::pair::Pair` and added tests
- added all the properties for `chain::header::Header` from the original JSON schema
  - note the names don't line up, this will be handled at the serialization layer in #75 
  - several properties only have string placeholders atm
- extended getters for `Header` properties to be more rustlike
- `Header::hash()` is now a method rather than a property
- `Entry::hash()` is now a method rather than a property
- `chain.push()` now takes entry type and an entry, rather than a pair
- `chain.push()` now auto generates headers from passed entries and _returns_ the pair
- added `SourceChain::top()`
- added `SourceChain::top_type()`

followup:

- #70
- #71
- #73
- #74 
- #75 
- #76


core/src/agent/mod.rs
```diff
@@ -1,16 +1,16 @@
...
+use chain::memory::MemChain;
 use state;
 use std::sync::mpsc::Sender;
 use std::sync::Arc;
 
 #[derive(Clone, Debug, PartialEq, Default)]
 pub struct AgentState {
     keys: Option<Keys>,
-    source_chain: Option<Box<SourceChain>>,
+    source_chain: Option<Box<MemChain>>,
```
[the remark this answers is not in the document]
thedavidmeister: @sphinxc0re this is a hangover from the very start of this repo, at some point i'll get you to help me clean it up :)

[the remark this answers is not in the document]
sphinxc0re: We'll probably want to have a generic `AgentState` over some specific `T: SourceChain`.

```rust
pub struct AgentState<T: SourceChain> {
    keys: Option<Keys>,
    source_chain: Option<T>,
}
```

core/src/chain/header.rs
```diff
@@ -0,0 +1,108 @@
...
+/// Properties defined in HeadersEntrySchema from golang alpha 1
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[allow(non_snake_case)]
+pub struct Header {
+    /// the type of this entry
+    /// system types may have associated "subconscious" behavior
+    Type: String,
+    /// ISO8601 time stamp
+    Time: String,
+    /// optional link to the immediately preceding header in the chain
```
zippy: this is optional only because it could be the last entry in the chain.  It would be good to include that in a comment here just for making the reason  for optionality transparent.

core/src/chain/header.rs
```diff
@@ -0,0 +1,108 @@
...
+    HeaderLink: Option<u64>,
+    /// mandatory link to the entry for this header
+    EntryLink: u64,
+    /// optional link to the most recent header of the same type in the chain
```
zippy: same, but just if it's first entry of that type.

core/src/agent/mod.rs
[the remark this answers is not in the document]
thedavidmeister: yeha, now that you showed me how this works it makes perfect sense :)

core/src/chain/header.rs
thedavidmeister: fair

core/src/agent/mod.rs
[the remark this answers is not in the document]
thedavidmeister: @sphinxc0re followup for this opened at #74

core/src/chain/header.rs
```diff
@@ -0,0 +1,108 @@
...
+    TypeLink: Option<u64>,
+    Signature: String,
```
sphinxc0re: All these should be formatted `snake_case`

core/src/chain/header.rs
```diff
@@ -0,0 +1,108 @@
...
+    Type: String,
+    /// ISO8601 time stamp
+    Time: String,
+    /// link to the immediately preceding header, None is valid only for genesis
+    HeaderLink: Option<u64>,
+    /// mandatory link to the entry for this header
+    EntryLink: u64,
+    /// link to the most recent header of the same type, None is valid only for the first of type
+    TypeLink: Option<u64>,
+    Signature: String,
```
sphinxc0re: All these fields should be formatted `snake_case`

sphinxc0re: @thedavidmeister I accidentally approved 😅 but I think you know what to do

thedavidmeister: @sphinxc0re this is intentional as per https://github.com/holochain/holochain-proto/blob/4d1b8c8a926e79dfe8deaa7d759f930b66a5314f/entry_headers.go#L7

all the getters are snake case, but i'm trying to keep the struct that is serialized by serde closer to the original for now

i extended the doc comment to explain this

sphinxc0re: If it's only about serialization, then serde is able to rename the fields on serialization.

core/src/chain/entry.rs
```diff
@@ -17,16 +16,14 @@ impl Entry {
...
-        _Hash::hash(&e, &mut hasher);
-        e.hash = hasher.finish();
         e
     }
 
     pub fn hash(&self) -> u64 {
-        self.hash
+        let mut hasher = DefaultHasher::new();
+        _Hash::hash(&self, &mut hasher);
+        hasher.finish();
```
sphinxc0re: 😄 You have to lose the last `;`

core/src/chain/header.rs
thedavidmeister: OK then, i'll make the property names match the methods and i put a tx up to configure serde like that - https://github.com/holochain/holochain-rust/issues/75

thedavidmeister:
@sphinxc0re @zippy OK - now ready for review :D

core/src/chain/header.rs
```diff
@@ -0,0 +1,276 @@
...
+impl _Hash for Header {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.entry_type.hash(state);
+        self.time.hash(state);
+        self.next.hash(state);
+        self.entry.hash(state);
+        self.type_next.hash(state);
+        self.signature.hash(state);
+    }
+}
```
sphinxc0re: I believe, if you just need this you can just `#[derive(Hash)]` for header.
https://doc.rust-lang.org/std/hash/index.html

sphinxc0re: And you don't even need to import `Hash` or `Hasher` for this.

thedavidmeister: yup, whether or not we use the native hashing trait, there's already a ticket up to review all of this and swap it out for our hashing algo

https://github.com/holochain/holochain-rust/issues/20

the point here is just to get tests in place around the _intent_ of hashes (e.g. this hash points to that thing, or changing this property changes the hash)

and i don't want to just derive Hash because:
- need a network friendly hashing algo that works on a network with non-rust hc
- i might want to put properties on entries for convenience that don't end up in the hash/serialized form (e.g. https://github.com/holochain/holochain-rust/issues/73)
- we may also move away from implementing the native hashing and do our own thing anyway

core/src/chain/header.rs
```diff
@@ -0,0 +1,276 @@
...
+    /// hashes the header
+    pub fn hash(&self) -> u64 {
+        let mut hasher = DefaultHasher::new();
+        _Hash::hash(&self, &mut hasher);
+        hasher.finish()
+    }
+
+    /// returns true if the header is valid
+    pub fn validate(&self) -> bool {
+        // always valid iff immutable and new() enforces validity
```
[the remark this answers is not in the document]
thedavidmeister: @zippy yes, the intent is just a placeholder for conscious/subconscious validation that can be run at chain/pair/header/entry level, but at the moment just returns `true` for some of them

core/src/chain/header.rs
```diff
@@ -0,0 +1,276 @@
...
+    /// a Header is immutable, but the chain is mutable if chain.push() is used.
+    /// this means that a header becomes invalid and useless as soon as the chain is mutated
+    /// the only valid usage of a header is to immediately push it onto a chain in a Pair.
+    /// normally (outside unit tests) the generation of valid headers is internal to the
+    /// chain::SourceChain trait and should not need to be handled manually
+    /// @see chain::pair::Pair
+    /// @see chain::entry::Entry
+    pub fn new<'de, C: SourceChain<'de>>(chain: &C, entry_type: &str, entry: &Entry) -> Header {
+        Header {
+            entry_type: entry_type.to_string(),
```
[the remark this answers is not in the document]
thedavidmeister: that's a neat trick, i'd actually like to avoid even that as what i ultimately want is just:

`Header::new(&chain, &agent, &entry)`
`Pair::new(&chain, &agent, &entry)`
`Chain::push(&agent, &entry)`

which would mean putting things like time and type on the entry within rust, which would be copied into the `Header` for serialization and hashing for interop and protocol compatibility.

https://github.com/holochain/holochain-rust/issues/73

core/src/chain/memory/mod.rs
```diff
@@ -0,0 +1,368 @@
...
+}
+
+/// SouceChain trait implementation
+impl<'de> SourceChain<'de> for MemChain {
+    // appends the current pair to the top of the chain
+    fn push(&mut self, entry_type: &str, entry: &Entry) -> Pair {
+        let pair = Pair::new(self, entry_type, entry);
+
+        if !(pair.validate()) {
+            panic!("attempted to push an invalid pair for this source chain");
```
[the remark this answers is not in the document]
thedavidmeister: yar, i know, but atm it can only happen at dev time

there's a ticket up to reassess this once zome validation is a thing - https://github.com/holochain/holochain-rust/issues/63

PR #72, opened 2018-06-28
Standard refactor

sphinxc0re:
## Changes
- Reordered imports to fit community defaults
- Removed `#![deny(warnings)]` since it might effect backwards compatibility. CI takes over.
- Unified code comments to fit community defaults
- Made docker test script mount current working directory

Good read on doc-comments:
https://doc.rust-lang.org/book/second-edition/ch14-02-publishing-to-crates-io.html#making-useful-documentation-comments

### DIFF
--- a/docker/run
+++ b/docker/run
@@ -1,2 +1,2 @@
 #!/bin/bash
-docker run -v "`pwd`:/holochain" --rm -p 3141:3141 -it holochain/rust-ubuntu:latest bash
+docker run --rm -p 3141:3141 -it holochain/rust-ubuntu:latest bash
--- a/docker/run-test
+++ b/docker/run-test
@@ -1,2 +1,2 @@
 #!/bin/bash
-docker run -v "`pwd`:/holochain" --security-opt seccomp=unconfined holochain/test:latest
+docker run --security-opt seccomp=unconfined holochain/test:latest


core/src/logger.rs
```diff
@@ -1,11 +1,8 @@
-//use std::sync::{Arc, Mutex};
-
-/*!
-This logger is the logger that's attached to each Holochain application
-which is separate from standard logging via the log crate warn! info! debug! logging that
-gets emitted globaly from the container.
-*/
+//! This logger is the logger that's attached to each Holochain application
```
zippy: why do people like this //! style instead of /*! ... */  This doesn't make sense to me as it means you have to keep fixing the line prefix every time you update the text?

core_api/src/lib.rs
```diff
@@ -1,67 +1,62 @@
...
-    // do some other stuff with the state here
-    // ...
-}
-
-// stop the app
-hc.stop().expect("couldn't stop the app");
-
-```
-*/
+//! holochain_core_api provides a library for container applications to instantiate and run holochain applications.
```
zippy: ARG, this is to me so much uglier!

core/src/logger.rs
sphinxc0re: In Rust, there are these doc comments. They are actually converted into attributes and serve as a hint for tools like rustdoc. Also they make it possible to check for missing documentation.
So this:
```rust
/// Lorem ipsum...
fn aasdfasdf();
```
actually gets converted into:
```rust
#[doc = "Lorem ipsum..."]
fn aasdfasdf();
```
Also the `!` in attributes means that the enclosing module is affected:
```rust
pub mod something {
    //! This is documentation for the module

    /// This is documentation for the function
    fn blah() {}
}
```
this would be:
```rust
pub mod something {
    #![doc = "This is documentation for the module"]

    #[doc = "This is documentation for the function"]
    fn blah() {}
}
```

core/src/state.rs
```diff
@@ -93,15 +92,13 @@ impl State {
...
+// TODO: write macro for DRY reducer functions
+// macro_rules! reducer {
+//     ($func_name:ident) => (
+//         fn reducer(old_state: Rc<$state_type>, action: &_Action) -> Rc<$state_type>  {
+//             // The `stringify!` macro converts an `ident` into a string.
+//             println!("You called {:?}()",
+//                      stringify!($func_name));
+//         }
+//     )
+// }
```
ddd-mtl: Shouldn't we just delete this dead code and make a ticket out of that TODO?

core/src/logger.rs
ddd-mtl: To make it less painful, the IDE can automate line prefixing. There are also shortcuts for commenting/uncommenting blocks of code.

core/src/state.rs
sphinxc0re: Good idea 👍

thedavidmeister:
once you guys make a decision on comments, and the tests pass, i think :shipit:

PR #77, opened 2018-06-29
61 low level api

ddd-mtl:
This branch is getting huge so I'll not finish all three functions here.
I did lay out some work for COMMIT and VALIDATE, but this branch is focused on GENESIS

Genesis is more complex than expected because it is called during the InitApplication Action and generates more Actions to complete. Thus Init is finished asynchronously so I added a NucleusStatus member to properly keep track of that.

Here is what a stack of Actions might look like when calling InitApplication on a app with two Zomes with each one there genesis() that commits something to the source chain.
```
InitApplication
	ExecuteZomeFunction - Zome0.Genesis
	ExecuteZomeFunction - Zome1.Genesis
	...
		Commit - Zome0
			ValidateEntry
			...
		Commit - Zome1
			ValidateEntry
			...
	...
	ReturnZomeFunctionResult - Zome0.Genesis
	ReturnZomeFunctionResult - Zome1.Genesis
	...
ReturnInitializationResult
`````
I copied some of Instance's methods out of Instance because those methods cannot be called in NucleusState because there is no instance object available.
 
What's missing in my current implementation is handling error cases where not all of the Zomes have done their genesis() for some reason and we would want a second call to InitApplication to complete the Init and do genesis() only on the remaining Zomes.

### CODE REVIEW

File: Cargo.lock.bak [hunk 1]
```diff
@@ -1,447 +0,0 @@
-[[package]]
```
thedavidmeister: don't forget to clean this up before merging

core/src/lib.rs
```diff
@@ -28,15 +28,21 @@ pub mod test_utils {
...
-    pub fn create_test_dna_with_wasm() -> Dna {
-        let wasm_binary = Wat2Wasm::new()
-          .canonicalize_lebs(false)
-          .write_debug_names(true)
-          .convert(
-              r#"
+    use std::fs::File;
+
+    pub fn create_wasm_from_file(fname: &str) -> Vec<u8> {
+        use std::io::prelude::*;
```
thedavidmeister: is this normal, sticking a `use` inside a fn?

core/src/lib.rs
```diff
@@ -28,15 +28,21 @@ pub mod test_utils {
...
-    // Create DNA containing WASM code that returns 1337 as integer
-    pub fn create_test_dna_with_wasm() -> Dna {
-        let wasm_binary = Wat2Wasm::new()
-          .canonicalize_lebs(false)
-          .write_debug_names(true)
-          .convert(
-              r#"
+    use std::fs::File;
+
+    pub fn create_wasm_from_file(fname: &str) -> Vec<u8> {
```
thedavidmeister: can you put some doc comments `///` on each fn?

core/src/lib.rs
```diff
@@ -28,15 +28,21 @@ pub mod test_utils {
...
+        use std::io::prelude::*;
+        let mut file = File::open(fname).unwrap();
+        let mut buf = Vec::new();
+        file.read_to_end(&mut buf).unwrap();
```
thedavidmeister: `unwrap()` panics, right? i think we're trying to avoid that for recoverable errors (e.g. the file might not exist)

core/src/lib.rs
```diff
@@ -48,100 +54,86 @@ pub mod test_utils {
                     (export "memory" (memory 0))
                 )
             "#,
-          )
-          .unwrap();
+            nucleus::ribosome::RESULT_OFFSET
+        );
+        let wat_str = match wat {
```
thedavidmeister: i think `unwrap_or` could help here https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or

core/src/lib.rs
```diff
@@ -48,100 +54,86 @@ pub mod test_utils {
...
+            None => default_wat.as_str(),
+            Some(w) => w,
+        };
+
+        let wasm_binary = Wat2Wasm::new()
+            .canonicalize_lebs(false)
+            .write_debug_names(true)
+            .convert(wat_str)
+            .unwrap();
```
thedavidmeister: unwrap could be bad because of panic

core/src/lib.rs
```diff
@@ -48,100 +54,86 @@ pub mod test_utils {
...
 
-        return create_dna("test_zome".to_string(), "test_cap".to_string(), wasm_binary);
+        return create_test_dna_with_wasm(zome_name, cap_name, wasm_binary.as_ref().to_vec());
+    }
+
+
+    // Prepare valid DNA struct with that WASM in a zome's capability
```
thedavidmeister: doc comment is triple slash `///`

core/src/lib.rs
```diff
@@ -48,100 +54,86 @@ pub mod test_utils {
...
+        capability.name = cap_name.to_string();
+        capability.code = DnaWasm { code: wasm };
+        zome.name = zome_name.to_string();
+        zome.capabilities.push(capability);
+        dna.zomes.push(zome);
+        dna
     }
 
 
+    // Prepare valid DNA struct helper
```
thedavidmeister: what's this? still needed?

Cargo.lock.bak
ddd-mtl: I understand that @lucksus added this file in develop 3 days ago, probably by mistake. I'm erasing it here. So I'm not sure what you are asking. Should I leave it?

core/src/lib.rs
ddd-mtl: Not in tests. We want tests to panic.
And anyway those unwraps are already in develop. They come from PR#44.

Cargo.lock.bak
thedavidmeister: oh! yes i see, good job then 👍

lucksus: Oh! Ooops.. sorry. Yes, this needs deleted..

Makefile
```diff
@@ -27,7 +27,7 @@ C_BINDING_CLEAN = $(foreach dir,$(C_BINDING_DIRS),$(dir)Makefile $(dir).qmake.st
 # apply formatting / style guidelines, and build the rust project
 main:
 	cargo fmt
-	cargo clippy -- --deny clippy
+	cargo clippy -- --warn clippy
```
sphinxc0re: Why would you do that? We want the process to fail if it doesn't comply

core/src/nucleus/mod.rs
```diff
@@ -104,77 +171,208 @@ impl FunctionResult {
...
+
+/// Reduce ReturnInitializationResult Action
+/// On initialization success, set Initialized status
+/// otherwise reset the nucleus
+fn reduce_rir(nucleus_state : & mut NucleusState, has_succeeded : bool) {
+    assert!(nucleus_state.status == NucleusStatus::Initializing);
+    if has_succeeded {
+        (*nucleus_state).status = NucleusStatus::Initialized
+    } else {
+        *nucleus_state = NucleusState::new();
```
zippy: I'm not sure I totally understand what you were thinking here, but it seems like if the initialization didn't succeed we can't just reset the state, to uninitialized, because a client of the state object actually might want to know that the initialization failed and distinguish between these states.

ddd-mtl: My thought was that if init fails, then undo everything that was done in the InitApplication Action reducer, which basically means resetting the state. In the absence of clarity for the whole App init procedure (#86) it seemed like the safest thing to do for now. I agree about needing some callback to notify that the init failed. I am not satisfied with the wait loop on history counts.

Makefile
ddd-mtl: I want to be able to put returns when it makes the code more understandable and clippy is preventing me from doing that, so I'm trying to make clippy less annoying. To my knowledge, there is no agreement to what the code is suppose to comply to.

zippy:
I do agree with @sphinxc0re that we will have to refactor `reduce()` in `nucleus/mod.rs` but also that it's fine for now.

ddd-mtl:
Yes, we can easily pull out functions for each Action reducer like I did for the `ReturnInitializationResult` reducer. Definitely the next step here.

Issue #73, opened 2018-06-29
passing entry_type around alongside entry feels awkward


thedavidmeister:
see `chain::chain::SourceChain::push()` for an example of what i'm talking about...

PR #79, opened 2018-06-30
WIP: 73 inline entry type

thedavidmeister:
#73

### DIFF
--- a/core/src/chain/entry.rs
+++ b/core/src/chain/entry.rs
@@ -7,7 +7,7 @@ pub struct Entry {
     entry_type: String,
 }
 
-impl<'a> _Hash for Entry<'a> {
+impl _Hash for Entry {
     fn hash<H: Hasher>(&self, state: &mut H) {
         self.content.hash(state);
         // do NOT include the entry_type in the entry hash
@@ -16,16 +16,16 @@ impl<'a> _Hash for Entry<'a> {
     }
 }
 
-impl<'a> Entry<'a> {
+impl Entry {
     /// build a new Entry from passed content
     /// an Entry is immutable, this is important for absolutely everything downstream
     /// an entry is not valid until paired with a header and included in a chain.
     /// @see chain::header::Header
     /// @see chain::pair::Pair
-    pub fn new(entry_type: &'a str, content: &'a str) -> Entry<'a> {
+    pub fn new(entry_type: &str, content: &str) -> Entry {
         Entry {
-            entry_type,
-            content,
+            entry_type: entry_type.to_string(),
+            content: content.to_string(),
         }
     }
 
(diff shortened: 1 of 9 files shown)


core/src/chain/entry.rs
```diff
@@ -4,11 +4,18 @@ use std::hash::{Hash as _Hash, Hasher};
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct Entry {
     content: String,
+
+    // @TODO do NOT serialize entry_type in Entry as it should only be in Header
+    // @see https://github.com/holochain/holochain-rust/issues/80
+    entry_type: String,
 }
 
 impl _Hash for Entry {
```
sphinxc0re: Since the standard `Hash` trait doesn't clash with anything in this context, could you remove the alias for now?

thedavidmeister: it clashes with `Entry::hash()` doesn't it?

sphinxc0re: It shouldn't. Not in this particular context.

thedavidmeister: @sphinxc0re it causes recursion issues if i try `self.has()` but i can drop the alias, check out the latest commit

core/src/chain/header.rs
```diff
@@ -24,7 +24,7 @@ pub struct Header {
     signature: String,
 }
 
-impl _Hash for Header {
+impl Hash for Header {
```
sphinxc0re: You could just `#[derive(Hash)]` this.

thedavidmeister: let's continue this in #20

core/src/chain/entry.rs
```diff
@@ -47,52 +69,106 @@ impl Entry {
...
+        let t1 = "a";
+        let t2 = "b";
+
+        // same type and content is equal
+        assert_eq!(Entry::new(t1, c1), Entry::new(t1, c1));
+
+        // same type different content is not equal
+        assert_ne!(Entry::new(t1, c1), Entry::new(t1, c2));
+
+        // same content different type is equal
```
zippy: This deserves a little bit more in the comment to explain why, because at first glance you might think not.  The reason is that what's hashed is JUST the content.  i.e. the type is just a convenience value for the implementation, not part of what's getting hashed.

thedavidmeister: heh, i'm not even sure if this is desirable behaviour!

part of the reason for writing this test was to highlight something unintuitive about the code :/

i opened up https://github.com/holochain/holochain-rust/issues/85 because intuitively the hash of an entry, and also its concept of equality, would contain its type

thedavidmeister: @zippy i added an `@see` to #85 to provide a pointer to the reasoning

PR #84, opened 2018-07-02
83 volume mount

### LINKED ISSUE #83: use volume instead of copy for docker files
setting a volume rather than COPY means we can do edits and watch tests instead of build/run being a one shot process in docker

sphinxc0re:
closes #83


thedavidmeister:
@zippy tradeoffs with both approaches

as i understand it:

COPY
- is a build time thing
  - files inside docker will be frozen at build time
- invalidates steps that come after it in the build, breaking subsequent layer caches
- can be a little slow

VOLUME
- is a run time thing
  - files inside docker will update at run time
- doesn't break layer caches
- allows writing back to the host, within the volume
  - does not allow writing to directories outside the docker project
- can be set to read only with the `:ro` flag (but this breaks tests from memory)

setting a volume was most important for the mdbook docker because otherwise the live reload was kind of pointless. for other docker builds, either copy or volume is fine i think, but this just makes everything consistent.

if you're worried about docker leaving traces on your HD... well try running the following commands!

`docker container prune`
`docker image prune`
`docker volume prune`

this usually cleans up many GB for me 😬

thedavidmeister:
@sphinxc0re i fixed the merge conflicts, i'll merge it for you when tests pass

PR #107, opened 2018-07-04
94 genesis return val

### LINKED ISSUE #94: genesis()
child of #89

zippy:
closes #94 

return value from genesis() now expected to be an empty string if genesis succeeds or an error message.

### DIFF
--- a/core/src/nucleus/mod.rs
+++ b/core/src/nucleus/mod.rs
@@ -193,11 +193,17 @@ pub enum Action {
 /// On initialization success, set Initialized status
 /// otherwise set the failed message
 fn reduce_rir(nucleus_state: &mut NucleusState, result: &Option<String>) {
-    //assert!(nucleus_state.status == NucleusStatus::Initializing);
-    match result {
-        None => (*nucleus_state).status = NucleusStatus::Initialized,
-        Some(err) => (*nucleus_state).status = NucleusStatus::InitializationFailed(err.clone()),
-    };
+    if nucleus_state.status != NucleusStatus::Initializing {
+        (*nucleus_state).status = NucleusStatus::InitializationFailed(
+            "reduce of ReturnInitializationResult attempted when status != Initializing"
+                .to_string(),
+        );
+    } else {
+        match result {
+            None => (*nucleus_state).status = NucleusStatus::Initialized,
+            Some(err) => (*nucleus_state).status = NucleusStatus::InitializationFailed(err.clone()),
+        };
+    }
 }
 
 fn return_initialization_result(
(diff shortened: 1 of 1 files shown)


core/src/nucleus/mod.rs
```diff
@@ -251,17 +262,14 @@ pub fn reduce(
...
 
-                                    // genesis returns a i32 as a string
-                                    // 0 == success
+                                    // genesis returns a string
+                                    // "" == success, otherwise error value
                                     match call_result {
-                                        // not okay if genesis returned an errorCode
-                                        Ok(ref s) if s != "0" => {
+                                        // not okay if genesis returned an value
+                                        Ok(ref s) if s != "" => {
```
thedavidmeister: is this where we should use an enum to represent the possible return strings and what the error codes might be?

core/src/nucleus/mod.rs
```diff
@@ -182,24 +183,34 @@ impl FunctionResult {
...
-/// otherwise reset the nucleus
-fn reduce_rir(nucleus_state: &mut NucleusState, has_succeeded: bool) {
-    assert!(nucleus_state.status == NucleusStatus::Initializing);
-    if has_succeeded {
-        (*nucleus_state).status = NucleusStatus::Initialized
-    } else {
-        *nucleus_state = NucleusState::new();
+/// otherwise set the failed message
+fn reduce_rir(nucleus_state: &mut NucleusState, result: &Option<String>) {
+    //assert!(nucleus_state.status == NucleusStatus::Initializing);
```
thedavidmeister: why is this commented out?

core/src/nucleus/mod.rs
```diff
@@ -182,24 +183,34 @@ impl FunctionResult {
...
+    match result {
+        None => (*nucleus_state).status = NucleusStatus::Initialized,
+        Some(err) => (*nucleus_state).status = NucleusStatus::InitializationFailed(err.clone()),
     };
 }
 
+fn return_initialization_result(
+    result: Option<String>,
```
thedavidmeister: i think (haven't tried) that if this was `Option<&str>` you wouldn't need to `.to_string()` as much below

core/src/nucleus/mod.rs
zippy: I don't think so because these may be application specific.  There's no predefined set.  Of course there may be some pre-defined values so we could go with and ErrorGeneric as we did in the errors.rs

core/src/nucleus/mod.rs
lucksus: hmm.. yeah, why?

core/src/nucleus/mod.rs
thedavidmeister: that makes sense to me

i'm imagining something like this https://doc.rust-lang.org/std/io/enum.ErrorKind.html

that has an `Other` and you also provide a text message to fit your type

PR #113, opened 2018-07-05
Cleaned up DNA spec

sphinxc0re:


doc/specs/dna.md
```diff
@@ -66,124 +66,121 @@ They are available in keyword specific Capabilities and function names
 
 ## Example DNA Tree
 
-``` yaml
```
zippy: Just FYI, JSON is technically a subset of YAML.  So if you want JSON with comments, you can use yaml, which is why I did that, but javascript of course has comments too, so what you did is fine...

PR #114, opened 2018-07-05
Switched to stable Rust

sphinxc0re:
Since only the top-level parts of the API that compile to Wasm need to use the nightly channel, I figured it would be awesome to switch the rest of the API to stable Rust


sphinxc0re:
I made changes to the base image. It seems like it hasn't been updated. @thedavidmeister do you known, what to do?

thedavidmeister:
@sphinxc0re base image is only pushed to docker hub when merged into develop

should be able to build the image locally for testing though

sphinxc0re:
I took the liberty to remove the lock file. We don't need it, since we are supplying a library

sphinxc0re:
@lucksus @thedavidmeister could you take another look at this?

PR #115, opened 2018-07-05
core_api Holochain::new new returns genesis error, closes #87

### LINKED ISSUE #87: add MVP gen phase that calls the genesis() callback
child of #86

As a container developer I want to know whether and how genesis failed/succeeded after a call to `Holochain::new()`

- [ ] add some tests to `core_api`for successful and failed genesis

zippy:


core/src/nucleus/mod.rs
```diff
@@ -429,7 +440,8 @@ pub fn reduce(
 #[cfg(test)]
 mod tests {
     use super::{
-        super::{nucleus::Action::*, state::Action::*}, *,
+        super::{nucleus::Action::*, state::Action::*},
+        *,
     };
```
sphinxc0re: could you make this import absolute and also add a `use super::*;` above?

zippy: ok done.

core_api/src/lib.rs
```diff
@@ -79,14 +79,35 @@ impl Holochain {
...
+                    .send(nucleus_state.status())
+                    .expect("test channel must be open");
+                true
+            } else {
+                false
+            }
+        });
+
+        // is unwrap ok here?
+        let status = receiver.recv().unwrap();
```
ddd-mtl: Don't know for unwrap, but wouldn't you want to use `recv_timeout` instead in case nucleus' status never reach `initialized` or `initializedFail`

core_api/src/lib.rs
```diff
@@ -203,6 +225,38 @@ mod tests {
         };
     }
```
ddd-mtl: You should be able to uncomment:
                // TODO #61 - Should instantiation also initialize?
                // assert!(hc.instance.state().nucleus().has_initialized());

zippy: done

core_api/src/lib.rs
```diff
@@ -194,15 +224,49 @@ mod tests {
...
+        let result = Holochain::new(dna.clone(), context.clone());
+
+        match result {
+            Ok(_) => assert!(false),
+            Err(err) => assert_eq!(err, HolochainError::ErrorGeneric("fail".to_string())),
+        };
+    }
+
+    #[test]
+    fn fails_instantiate_if_genesis_times_out() {}
```
lucksus: Do want to implement this test?

Either implement or remove, please.

zippy: oh yah, will do.

zippy: done

PR #116, opened 2018-07-06
CI revamp

### LINKED ISSUE #49: build travis against `stable` and `nightly` unpinned, but allow failures
child of #117

sphinxc0re:
## Summary
This PR should simplify the whole CI process. The base image for testing is built on DockerHub. `make` shouldn't be run from the host machine. Instead, everything should be done from inside the docker container unless you have a good reason not to do so. I updated the README.md to include the new workflow.

Since the `make` is still being used by the docker containers, I added a few handy shortcuts for commands. An environment variable inside the Makefile pins down the nightly Rust version we are using.

~I also added docker image caching to reduce build time even further.~
Docker caching only adds time in our case because Travis doesn't support it natively

## Problems
As of this PR, we are only building against stable. This is because having multiple Rust versions would require each build to build the Docker image for that specific build. Since we are only using stable Rust I figured it might be a waste of build time to do this.

Also I wasn't able to make tarpaulin work on any other Rust version.

## Misc

fixes #49 

child of #117


thedavidmeister:
oh actually, there's some panics to to cleanup:

```
thread '<unnamed>' panicked at 'Failed to instantiate module: Function("host module doesn\'t export function with name cosf")', libcore/result.rs:945:5
```

@sphinxc0re

sphinxc0re:
@thedavidmeister @zippy I need some more feedback. I wrote a summary in the description. The war is finally over...

sphinxc0re:
@lucksus please take a look as well

PR #125, opened 2018-07-10
add tests and ::new() for agent keys

### LINKED ISSUE #123: implement new for key and keys
child of #88 

currently there's no real interface for working with agent keys, a basic set of `new` fns can help

thedavidmeister:
fixes #123 

this fleshes out the agent keys just enough to make them useful while mocking things elsewhere...

### DIFF
--- a/core/src/agent/keys.rs
+++ b/core/src/agent/keys.rs
@@ -18,11 +18,11 @@ pub struct Keys {
 
 impl Keys {
 
-    pub fn new(pub_key: &Key, priv_key: &Key, node_id: &str) -> Keys {
+    pub fn new<S: Into<String>> (pub_key: &Key, priv_key: &Key, node_id: S) -> Keys {
         Keys {
             pub_key: pub_key.clone(),
             priv_key: priv_key.clone(),
-            node_id: String::from(node_id),
+            node_id: node_id.into(),
         }
     }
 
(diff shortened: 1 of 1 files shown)


core/src/agent/keys.rs
```diff
@@ -1,9 +1,78 @@
...
 #[derive(Clone, Debug, PartialEq)]
 pub struct Keys {
     pub_key: Key,
     priv_key: Key,
     node_id: String,
 }
+
+impl Keys {
+
+    pub fn new(pub_key: &Key, priv_key: &Key, node_id: &str) -> Keys {
```
sphinxc0re: I think it would be easier to use a `S: Into<String>`.

core/src/agent/keys.rs
```diff
@@ -1,9 +1,78 @@
...
+        Keys {
+            pub_key: pub_key.clone(),
+            priv_key: priv_key.clone(),
+            node_id: String::from(node_id),
```
sphinxc0re: Then, you could just do `node_id.into()` here.

core/src/agent/keys.rs
```diff
@@ -1,9 +1,78 @@
...
+    use super::Keys;
+
+    /// generates a new key suitable for testing
+    pub fn test_key() -> Key {
+        Key::new()
+    }
+
+    /// generates a new node id suitable for testing
+    pub fn test_node_id() -> String {
+        String::from("test node id")
```
sphinxc0re: you can just do:
```rust
"test node id".into()
```

core/src/agent/keys.rs
thedavidmeister: @sphinxc0re yar, it works, just seems like a lot of type juggling to support string arguments

PR #129, opened 2018-07-12
91 commit func

ddd-mtl:
First working version of Commit function.
There is a lot of data handling code in the wasm code. I don't know if there is a way around that.
Also I'm using the same memory buffer as the one used for the result of the Zome call.

### DIFF
--- a/core_api/wasm-test/commit/src/lib.rs
+++ b/core_api/wasm-test/commit/src/lib.rs
@@ -1,0 +1,122 @@
+extern crate serde;
+#[macro_use]
+extern crate serde_derive;
+extern crate serde_json;
+extern crate libc;
+
+use serde::{Deserialize, Serialize};
+use std::{ffi::CStr, os::raw::c_char, slice};
+
+extern {
+  // fn commit(data: *mut c_char, params_len: usize) -> *mut c_char;
+  fn commit(params_len: i32) -> i32;
+}
+
+
+//-------------------------------------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


core_api/wasm-test/commit/src/lib.rs
```diff
@@ -70,25 +60,25 @@ fn hc_commit(data: *mut c_char, entry_type_name: &str, entry_content : &str)
...
-// Convert a data struct into json memory buffer
-fn serialize<T: Serialize>(data: *mut c_char, internal: T) -> i32 {
-    let json = serde_json::to_string(&internal).unwrap();
-    let bytes = json.as_bytes();
-    let len = bytes.len();
-    let mem = unsafe { slice::from_raw_parts_mut(data, len) };
+// Write a data struct into a memory buffer as json string
+fn serialize<T: Serialize>(ptr_data: *mut c_char, internal: T) -> i32 {
+    let json_str = serde_json::to_string(&internal).unwrap();
+    let json_bytes = json_str.as_bytes();
```
sphinxc0re: https://docs.serde.rs/serde_json/fn.to_vec.html
This should do the trick

ddd-mtl: done

lucksus:
Besides my comment on ```chain.puhs(entry);``` it looks good though.

ddd-mtl:
I added Julian's change request so for me that counted as 2nd approval.

PR #149, opened 2018-07-18
Upgraded docker and Makefile

sphinxc0re:
## Summary

While I was moving the docker automated build to the `holochain` org, the build of our Dockerfile errored while installing clippy. It seems like clippy is now officially supported by the rust core team and inside the component collection of `rustup`. This means better support but also that I had to do some adjustments. As of this update we will be using:

 - `nightly-2018-07-17` for our tools (`rustfmt`, `clippy`)
 - `nightly-2018-06-01` for Wasm compilation
 - `stable` for everything else

The upgrade of `rustfmt` again will result in some reformatting. I intentionally didn't include the changes of that because the changes will still have to go though the docker hub build. Also as I suspected, the resulting reformatting won't be too heavy since it only touches some imports.

I also did some changes to the Makefile. I purposely removed the `--verbose` flag from the regular Rust build to (drum roll) reduce verbosity.

~Another thing I changed was the call to clippy. It now is called without any flags to prevent unwanted behavior. We should always stick to the default community standards and this helps a lot with doing so.~ 

*The CI will probably fail because it will be using the old docker image*

## What reviewers can do
You can check out the branch locally, run the build script (`. docker/build-ubuntu`), run the formatting script (`. docker/run-fmt`) and check whether the tests run using `. docker/run-test`

Issue #99, opened 2018-07-02
Network Abstraction Layer


lucksus:
As a Core Dev I want a Networking Abstraction Layer so that I can connect HC to Promether or other net libs.

- [ ] minimal notion of a Node/Peer that can send and receive to other nodes
- [ ] minimal notion of Node Address that wraps the nodes logical (i.e. hash) & transport addresses.
- [ ] minimal notion of Node/Peer Store that allows us to resolve node logical address to transport address.

zippy:
This can @neonphog's https://github.com/holochain/lib3h 

Also found that the libp2p folks have a rust version available too now: https://github.com/libp2p/rust-libp2p/tree/cace5bf577c53809aad514db40e72f7290e030d4 (not complete but looks pretty solid)

neonphog:
# Transport Abstraction Feedback

## Serialization Types

### SerializedAddress

(Note after-the-fact: some clarity is added in the "new_node / from param send/deliver" section below.)

Is this meant to uniquely identify a node? If so, maybe we should call it `SerializedIdentity` or something? The following comment assumes this is the intention:

- I'd suggest using `Vec<u8>` as the type for SerializedAddress, this will allow us to gain a slight efficiency by pluging in the raw bytes of the agent's public key (not sure how dpki fits into this).
- The base58 or whatever representation we want to use would only be applied when interaction with logs or user i/o.

### SerializedMessage
- I'd suggest using `Vec<u8>` as the type for SerializedMessage, to allow the use of lower-level, more efficient message packing schemes, whether we roll our own, or use something like https://crates.io/crates/rmp

### TransportAddress
- This one is tricky, I'm tempted to say it should be a `Vec<u8>` as well and let the underlying transport define it however we want, but it also seems like we want it to be human readable for debugging issues. We could go with a url scheme, and either leave it as a String, or explicitly make it a https://crates.io/crates/url, but what of transports that may have multiple paths to reach the same node?
- Perhaps this should actually just be a `Display + PartialEq` trait object that can be defined internally by the specific transport?

## TransportAddress Needed?

See the Config and Persistence notes below, but do we actually need the TransportAddresses to be accessible from the HC code at all? There's currently no apis that take them as input, and I can't think of a strict need if the transport lib is handling the routing / p2p network.

## `new_node` / `from` param send/deliver

Ahh... I see now some differences in how we've been thinking about this.

[...]

PR #150, opened 2018-07-18
#99 network abstraction layer

zippy:
Hey folks, there's still a bit to do, but I wanted to get this up as a WIP PR to start getting feedback.

Please see the discussion @neonphog & I have been having on #99 

The big thing left is about clarifying how topological routing works within one transport and across transports.  I'm working on that next.

BTW fmt is busted again until we complete #149 so don't expect CI to be green.


net/src/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+            if let Some(_node) = self.nodes.iter().find(|node| *node.get_address() == *addr) {
+                true
+            } else {
+                false
+            }
+        }
+    }
+
+    fn to_str(vec: &Vec<u8>) -> String {
+        str::from_utf8(vec).unwrap().to_string()
```
sphinxc0re: Please use `String::from_utf8` here

sphinxc0re:
Formatting isn't busted. The CI seems to fail because of some different issue

net/src/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+type SendResponseClosure = Box<FnMut(Result<SerializedMessage, Error>) -> Option<Error> + Send>;
+
+/// this closure type gets called when a new message arrives, you can respond with a Message or an error.
+type ReceiveClosure =
+    Box<FnMut(&SerializedAddress, &SerializedMessage) -> Result<SerializedMessage, Error> + Send>;
+
+pub trait Node {
+    fn get_address(&self) -> &SerializedAddress;
+    fn get_transport_address(&self) -> TransportAddress;
+}
```
sphinxc0re: Why can't `Node` be a struct?

net/src/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+
+pub struct Handler {
```
sphinxc0re: This wrapper seems unnecessary and to cause unneeded complexity

net/src/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+        }
+    }
+}
+
+pub trait Transport {
+    /** initialize the transport
+     * this might be used for example in a TCP based transport to specify a listening port for
+     * for the transport.
+     */
+    fn initialize(&mut self, config: Option<String>) -> Result<(), Error>;
```
sphinxc0re: Shouldn't initialize return the initial Transport? In mean, you don't want something to be initialized twice or more often.

net/src/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+        transport_addr: u32,
+    }
+
+    impl Node for SimpleNode {
+        fn get_address(&self) -> &SerializedAddress {
+            &self.hc_addr
+        }
+
+        fn get_transport_address(&self) -> TransportAddress {
+            format!("{}", self.transport_addr).into()
```
sphinxc0re: isn't there a `to_string()` method on `u32`?

net/src/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+        }
+    }
+
+    pub struct SimpleTransport {
+        config: String,
+        nodes: Vec<Arc<SimpleNode>>,
```
sphinxc0re: Unless you don't want for those nodes to be mutable you should put them into a `Vec<Arc<Mutex<SimpleNode>>>`

net/src/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+                handlers: HashMap::new(),
+            }
+        }
+        pub fn exists(&self, addr: &SerializedAddress) -> bool {
+            if let Some(_node) = self.nodes.iter().find(|node| *node.get_address() == *addr) {
+                true
+            } else {
+                false
+            }
+        }
```
sphinxc0re: If you are only looking for whether the value exists, you can just as easily call `is_some()` on an `Option<T>`.

net/src/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+    impl Transport for SimpleTransport {
+        fn initialize(&mut self, config: Option<String>) -> Result<(), Error> {
+            if let Some(cfg) = config {
+                if cfg == "".to_string() {
+                    bail!("null config!");
+                }
+                self.config = cfg;
+            } else {
+                self.config = self.get_default_config();
+            }
```
sphinxc0re: We can do this much slicker 😁 

```rust
self.config = if let Some(cfg) = config {
    if cfg.is_empty() {
        bail!("null config");
    }
    cfg
} else {
    self.get_default_config()
}
```

net/src/lib.rs
zippy: can't because then I can't borrow because String::from_utf8 takes a Vec not a &Vec.

net/src/lib.rs
zippy: The original reason was to allow every Transport implementation to also have it's own Node implementation.   This may change, but I think it's still valid.

net/src/lib.rs
zippy: Yes, I think you are right.  I ended up adding the wrapper in the phase when I was trying to get Transport to work as a trait object, and that matches how @lucksus wrapped up the closure in instance.rs.   I'll try simplifying back down...

net/src/lib.rs
zippy: You can't make initialize return an object that doesn't exist yet, and I don't want this to turn into a trait object, because that's what caused all the problems first time around.  Perhaps it could return self for chaining, but it some Transports I could imagine that re-initializing is semantically ok.  We would need to declare an expected behavior if it's not

net/src/lib.rs
zippy: there is, but I saw `into()` and it seems to do the same thing, and I like it.  Is there a reason not to use it?  The compiler picks the right type I think.

net/src/lib.rs
zippy: For a real transport implementation yes, but because this is just for the tests I left it simple.

net/src/lib.rs
zippy: nice, I keep forgetting that.

net/src/lib.rs
zippy: nice.

zippy:
@sphinxc0re Re CI, the issue is that when I run `. docker/build-ubuntu` from this branch it fails on clippy install, so I think it also fails on cargo fmt install, so I can't run `. docker/run-fmt`:

```
$ . docker/build-ubuntu

Step 1/15 : FROM ubuntu
 ---> 8b72bba4485f
Step 2/15 : ENV DEBIAN_FRONTEND noninteractive

... snip ...

Step 12/15 : RUN cargo +${PINNED_NIGHTLY} install clippy --version ${CLIPPY_VERSION}
 ---> Running in a1a1b87a5184
    Updating registry `https://github.com/rust-lang/crates.io-index`
 Downloading clippy v0.0.206
  Installing clippy v0.0.206
error: binary `cargo-clippy` already exists in destination
Add --force to overwrite
The command '/bin/sh -c cargo +${PINNED_NIGHTLY} install clippy --version ${CLIPPY_VERSION}' returned a non-zero code: 101
```

net/src/lib.rs
```diff
@@ -0,0 +1,348 @@
...
+        from: &SerializedAddress,
+        to: &SerializedAddress,
+        msg: SerializedMessage,
+        callback: SendResponseClosure,
+    ) -> Result<(), Error>;
+
+    /** deliver a message to a given recipient
+     * assumes that recipient address was registered locally with new_node
+     * this function may be called by the routing layer to bridge across transports
+     */
```
sphinxc0re: These comments should all be adjusted to follow the Rust idioms

sphinxc0re:
@zippy Yes, I'm trying to fix this right now

net/src/lib.rs
sphinxc0re: Maybe we want to store nodes in a `HashMap` ?

net/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+
+    /// deliver a message to a given recipient
+    /// assumes that recipient address was registered locally with new_node
+    /// this function may be called by the routing layer to bridge across transports
+    fn deliver(
+        &mut self,
+        from: &SerializedAddress,
+        to: &SerializedAddress,
+        message: SerializedMessage,
+    ) -> Result<SerializedMessage, Error>;
```
neonphog: Semantic, but in the ipc code, I had ended up calling the fire-and-forget message type `send`, and the procedure call style message type `call`. Anyone have quick vote, or should we name department this?

(p.s. I liked `send` and `call` because they are both four letters)

net/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+
+    /// return a default config with best configuration guesses for the transport
+    fn get_default_config(&self) -> String;
+
+    /// register a peer as a node in the transport
+    fn new_node(
+        &mut self,
+        addr: SerializedAddress,
+        handler: Option<ReceiveClosure>,
+    ) -> Result<(), Error>;
```
neonphog: Per our lucidchart discussion, should all of the SerializedAddress references be moved to a layer above this? I.e. instead of `new_node` we would have:

```rust
fn get_transport_address(&self) -> TransportAddress;
```

1. Then we would layer a DHT per app on top of the strict p2p transport layer (that is, each app would have a unique SerializedAddress, and use that address to discover / track the opaque p2p TransportAddresses of neighbors and index DHT buckets)
2. If a specific app wanted to message another agent within that same app, a higher level abstraction would allow them to just specify a SerializedAddress. It would then make a high-level DHT query to discover the opaque TransportAddress that app agent is currently connected to, bundle up a layered message like:

```javascript
{
  "to": <TransportAddress>,
  "message": {
    "to": <SerializedAddress>,
    "message": { /* ... */ }
  }
}
```

Now that I've written all this, perhaps THIS is the higher level abstraction, and what we really need to do is define the lower-level one?

thedavidmeister:
@zippy what needs to happen here?

lucksus:
Do we intend to merge this still at some point or is this deprecated and superseded by https://github.com/holochain/holochain-rust/pull/163 and the designing we are doing in https://hackmd.io/_OIEfnTOSzWtW_7sUw6lrg?view, @zippy, @neonphog?

neonphog:
Yeah, I think I'm going to take this over @lucksus (correct me if you had other ideas @zippy) - Even if the api is very different, I can build on this for scaffolding.

thedavidmeister:
@neonphog thanks for picking this up, i was wondering what was happening here :)

thedavidmeister:
12k LOC in the diff :/

the good parts of this should be teased out and landed separately.

neonphog:
egads yes, something went horribly wrong with the develop merge... uhh... i've cherry picked the actual changes into https://github.com/holochain/holochain-rust/compare/99-net-abstraction-david - i'll work from there, but leave this PR open as a placeholder, and close / open a new one when ready

thedavidmeister:
@neonphog thanks for cleaning this up, it's much easier to see what is going on

do you think anything in here could benefit from sitting on top of riker, or should it all be traits still?

neonphog:
@thedavidmeister 

> do you think anything in here could benefit from sitting on top of riker, or should it all be traits still?

My initial thought is that this touchpoint is intended to be as simple as possible so that we can iterate quickly on the api as we gain experience. It may very well make a lot of sense to create a riker actor that takes a trait object of this type and works with it like that. Thoughts?

neonphog:
@thedavidmeister @zippy @lucksus - This is not a full implementation of https://hackmd.io/_OIEfnTOSzWtW_7sUw6lrg, but in the interest of being iterative, I think this set of functions gives a good cross-section of how we would need to implement the various APIs.

Let me know what you think.

lucksus:
But I'd like to see this merged asap. @ddd-mtl might want to use that trait in the DHT module already..

ddd-mtl:
It seems a bit different from what I need since it does all the app tracking, which fits more in the container/ holochain-system layer than the holochain instance layer.
Its missing `dht_publish_meta()` unless we are suppose to also use `dht_publish()` for that?
I wouldn't add any network stuff in my first PR for DHT anyway but it does help me see how it will fit in afterwards. 👍 
I don't feel qualified enough to approve this, but I like it :)

neonphog:
@ddd-mtl - thanks for the feedback!

> It seems a bit different from what I need since it does all the app tracking, which fits more in the container/ holochain-system layer than the holochain instance layer.

This is still designed to go at the holochain layer, in which case it will only ever track a single app, but I still need to get the identity information, and there is the potential that multiple holochain instances are connecting to the same network process that itself is tracking multiple apps, so I'll need that distinction on the back end.

> Its missing dht_publish_meta() unless we are suppose to also use dht_publish() for that?

We are totally going to be adding `dht_publish_meta()`, I just wanted to get feedback / have a smaller PR before I fleshed it out with the full API.

neonphog:
@lucksus 

> Looks really good! I'm just worried about the changes to the ADR that we have just approved. Can you move those changes out into a different PR, and best into another ADR since ADR7 is accepted?

Yeah, these are changes @zippy did in this original work. I can remove them if they are not needed.

ddd-mtl:
@neonphog Okay, I thought it was the full API. No problem.
I forgot to ask:  `Dna.hash()` from `holochain_dna` crate should output a `GenomeHash`, right?

net/src/p2p_network.rs
```diff
@@ -0,0 +1,494 @@
...
+
+/// callback function type for api methods transporting binary data
+pub type ApiFnBin = Box<FnMut(&[u8]) -> Result<Vec<u8>, Error>>;
+
+/// when the network is requesting we store data
+/// this callback will be invoked, expecting the data to be validated
+pub type DhtHoldCallback = Box<FnMut(&str) -> Result<bool, Error>>;
+
+/// the identifier for an application
+pub type GenomeHash = [u8; 32];
```
zippy: Just flagging it here for @thedavidmeister and @lucksus and @ddd-mtl that currently we are just switching to the term `Address` for the content addressable store and the assumption is that that's a SHA256 hash.   What ever this ends up being, these two data structures have to match up in the end!

This doesn't stop my approval of this PR.  Just making sure everybody is aware of the issue.

PR #152, opened 2018-07-19
ADR:13 for events API, resolves #org-92

zippy:
Here's a first draft of an events model building on what we talked about at the beginning of the week.


doc/specs/dna.md
```diff
@@ -156,6 +156,15 @@ They are available in keyword specific Capabilities and function names
             },
             // ...
           ],
+          "evt_declarations": [
```
sphinxc0re: Why not just `events`? It's so much simpler.

Connoropolous:
Or why put it into the DNA at all? Just to make it more consistent with typed languages by defining a contract?

doc/specs/dna.md
zippy: fixed

ddd-mtl:
I was going to suggest "subscribe" or "follow" but i'm fine with "listen". It comes from the fact that there is going to be an "unfollow" / "unsubscribe" function. What I find important is that "emit", "listen" and "unlisten" follows the same mental model.

As to what Connor said, it seems inevitable to me that we are going to allow external observation of the Actions being processed by the Holochain Instance, at least for debugging and testing reasons (Commit is one type of Action). It feels natural to me that the container would also want that, at least locally.

doc/architecture/decisions/0013-events-model-and-api.md
```diff
@@ -0,0 +1,48 @@
...
+``` javascript
+postHash = commit("Post",{content:"foo"})
+emit("Post",postHash)
+```
+
+3. Finally, just as you can call any function using the `core_api::call()`, you can register an observer with `core_api::observe()`
+
+## Consequences
+
+- We need to clarify when you can use a `emit()` call, i.e. is it valid during validations and genesis?
```
ddd-mtl: Also, we would maybe have to do some safeguarding from infinite loops when an observation triggers a call() which triggers an observation...

doc/architecture/decisions/0013-events-model-and-api.md
```diff
@@ -0,0 +1,48 @@
...
+
+We will extend the API in a way that's roughly equivalent to the [Signal-slot pattern](https://en.wikipedia.org/wiki/Signals_and_slots), i.e. in the following ways:
+
+1. In the DNA you can declare observable events anywhere you can declare functions.  You can think of this almost identically to declaring a function except that it "goes the other way," i.e. a function def exposes an entry point where an event def exposes and exit point.  It would look like this:
+
+``` javascript
+          "event_declarations": [
+            {
+              "name": "Post",
+              "params": null,
```
ddd-mtl: I suggest adding a "description" attribute so to have some info on what the event is about, and possibilities when or how it is triggered.

doc/architecture/decisions/0013-events-model-and-api.md
```diff
@@ -0,0 +1,48 @@
...
+
+2. App developers can emit events from their code via a new `emit()` function to be added to the api, e.g. like this:
+
+``` javascript
+postHash = commit("Post",{content:"foo"})
+emit("Post",postHash)
+```
+
+3. Finally, just as you can call any function using the `core_api::call()`, you can register an observer with `core_api::observe()`
+
```
ddd-mtl: "and unregister an observer with `core_api::unobserve()`

doc/architecture/decisions/0013-events-model-and-api.md
```diff
@@ -0,0 +1,48 @@
...
+In the go code we used polling from the UI to discover local state changes, be they in the source-chain or in the DHT.  We want to change this to an event based model where UI (or other clients, like bridged apps) can become observers of events.  Additionally there are system level events (like warrants) as well as other app level definable events that we want consumers of exposed holochain app APIs to be able to access on a push (i.e. observed) basis.
+
+## Decision
+
+We will extend the API in a way that's roughly equivalent to the [Signal-slot pattern](https://en.wikipedia.org/wiki/Signals_and_slots), i.e. in the following ways:
+
+1. In the DNA you can declare observable events anywhere you can declare functions.  You can think of this almost identically to declaring a function except that it "goes the other way," i.e. a function def exposes an entry point where an event def exposes and exit point.  It would look like this:
+
+``` javascript
+          "event_declarations": [
```
lucksus: I'd suggest actually calling it ```signals```.

doc/architecture/decisions/0013-events-model-and-api.md
```diff
@@ -0,0 +1,48 @@
...
+emit("Post",postHash)
+```
+
+3. Finally, just as you can call any function using the `core_api::call()`, you can register an observer with `core_api::observe()`
+
+## Consequences
+
+- We need to clarify when you can use a `emit()` call, i.e. is it valid during validations and genesis?
+
+- We need to make sure that app developers understand that this doesn't happen globally, but rather just to the clients that subscribed to the event from the container.
```
lucksus: add: "Only clients that got permission to access the capability the signal is can register a listener/slot."

doc/specs/dna.md
lucksus: I still see "evt_declarations". Need to push?
As stated above, I would name this property ```signals```.

doc/architecture/decisions/0013-events-model-and-api.md
sphinxc0re: I'm suggesting calling them just `events`

doc/specs/dna.md
zippy: heh, sorry I fixed it in the adr, not the dna spec

doc/architecture/decisions/0013-events-model-and-api.md
zippy: I'm fine with (and like) `signals` because we listen for signals.  I really don't like `slots` because it doesn't evoke anything strong for me as a term and doesn't seem very agent centric where as listening does.

zippy:
hey @sphinxc0re @lucksus @artbrock please see if the recent changes address your review requests and change status to approved if so.  Thanks.

doc/architecture/decisions/0013-signals-listeners-model-and-api.md
```diff
@@ -0,0 +1,55 @@
...
+Note that this is only about signals that can be listened to by container/composer context, i.e. by the client of the core_api, NOT by the app nodes themselves.
+
+## Decision
+
+We will extend the API in a way that's roughly equivalent to the [Signal-slot pattern](https://en.wikipedia.org/wiki/Signals_and_slots), i.e. in the following ways:
+
+1. In the DNA you can declare observable signals anywhere you can declare functions.  You can think of this almost identically to declaring a function except that it "goes the other way," i.e. a function def exposes an entry point where an signal def exposes an exit point.  It would look like this:
+
+``` javascript
+          "signal_declarations": [
```
sphinxc0re: The term `signal` is something originates in the Qt world. I would use a more widely used and adopted term like `event`.

zippy: Well, not just the Qt world.  It's also a big term in information theory and muti-agent systems and complex adaptive systems worlds, which we are (and I would say) want to be connected to.

Connoropolous:
I am in support way more of Julian's suggestion on this particular naming
controversy :p

Connoropolous:
Also, sorry if this is getting too nitty gritty, but why use the word 'declarations'? That feels really redundant in there?

doc/architecture/decisions/0013-signals-listeners-model-and-api.md
lucksus: Signals/Slots is a design pattern that was spearheaded by Qt but there are several other implementations as well, like for instance in boost: https://www.boost.org/doc/libs/1_63_0/doc/html/signals.html.

There is a subtle difference between events and signals/slots:
- with events, the focus is basically on the event object which can exist on its own
- signals/slots is a way to do dynamic binding of function calls, i.e. the use of ```emit``` is actually to be seen as a call to another (set of) function(s).

The reason I prefer the signal/slot frame over events is that it emphasizes the notion of a strong object-model: you connect slots to signals where both belong to a given entity (subclasses of QObject in the case of Qt). There is no signal dangling in the event-loop that does not have a clear source ("sender").

Signal/slots imply an underlying event-loop though on which they are implemented, but are a more elaborate kind of thing.

Since we do have a strong representation of the sender of an event (an app instance), we can make advantage of this more specific concept which I think communicates the right contextual ideas.

lucksus: I agree with @Connoropolous's comment in the main thread: get rid of "_declarations"

doc/architecture/decisions/0013-events-model-and-api.md
lucksus: Indepent of the wording slot, my suggestion was to talk about capabilities and permissions here.

doc/architecture/decisions/0013-signals-listeners-model-and-api.md
```diff
@@ -0,0 +1,55 @@
...
+Note that in the example above we are using the attribute method of declaring the signature, and it declares what the signal will send to the listeners and what must be passed in as "params" on the listen request which may be useful for qualifying some aspect of what to listen for.  See [#134](https://waffle.io/holochain/org/cards/5b4cd03d0df367001d6d12a6) for details.
+
+2. App developers can emit signals from their code via a new `emit()` function to be added to the api, e.g. like this:
+
+``` javascript
+postHash = commit("Post",{content:"foo"})
+emit("Post",postHash)
+```
+
+3. Finally, just as you can call any function using the `core_api::call()`, you can register a listener with `core_api::listen()` and you and unregister a listener with `core_api::unlisten()`
```
lucksus: I'd also change ```listen()``` to ```connect()```, since you listen to events but connect signals.

doc/architecture/decisions/0013-signals-listeners-model-and-api.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Consequences
+
+- We need to clarify when you can use a `emit()` call, i.e. is it valid during validations and genesis?
+
+- We need to make sure that app developers understand that this doesn't happen globally, but rather just to the clients that subscribed to the signal from the container.
```
lucksus: Since my comment in the other thread is not shown anymore in the overview:

add: "Only clients that got permission to access the capability the signal is in can register a listener/slot."

doc/architecture/decisions/0013-signals-listeners-model-and-api.md
zippy: connect signals doesn't make much sense to me and introduces confusion with regular networking connections.

doc/architecture/decisions/0013-signals-listeners-model-and-api.md
zippy: that's been added...

zippy:
Hey @lucksus you are the only outstanding change request person.  When you get a chance please confirm that things look ok.

doc/architecture/decisions/0013-signals-listeners-model-and-api.md
```diff
@@ -0,0 +1,55 @@
...
+
+1. In the DNA you can declare observable signals anywhere you can declare functions.  You can think of this almost identically to declaring a function except that it "goes the other way," i.e. a function def exposes an entry point where an signal def exposes an exit point.  It would look like this:
+
+``` javascript
+          "signals": [
+            {
+              "name": "Post",
+              "description: "signal emmited when a post is committed",
+              "params": null,
+              "sends": {
```
lucksus: Reading through this again, I wonder why there is ```params``` and ```sends```. I understood ```sends``` to be the arguments of the signal.. what is ```params``` then? Can we coalesce both to ```arguments```?

zippy: Nope, actually sends is the equivalent of "returns" i.e. it's what the signal will send to you when triggered.  I meant for "params" to be what you as the listener can pass in when you register as a listener.  The idea was to create a space for parameterizing the listening, e.g. this would be a way to listen for Posts only from a specific person, and that parameter gets passed into the handler for the signal so it can do the filtering.

Perhaps a different name to be less confusing?

zippy:
@thedavidmeister I think they are very close.  @lucksus based the naming and thinking on the qt signals/slots pattern, and I found this article on the difference between the two: https://stackoverflow.com/questions/9903334/are-qts-signals-and-slots-a-form-of-publish-subscribe

The paper they reference in that article seems interesting but I didn't dive in deep.

thedavidmeister:
> signals/slots implement the observer pattern, where the producer has a reference to its subscribers and is responsible of notifying them

> the topic handler, which decouples producers and consumers (the producers does not know who will consume messages)

> A main consequence is that in the p/s paradigm you can have multiple producers on the same topic.

wouldn't we want to decouple topics from producers?

doc/architecture/decisions/0013-signals-listeners-model-and-api.md
lucksus: Well, I don't fully understand your ```params``` then. Or I should rather say, I think to build what you want it gets too complex. I would suggest removing ```params``` and renaming ```sends``` to arguments, because for the listener/slot/handler it is what comes in as parameter / function argument.

ddd-mtl: `params`  are like configuration variables. It is input information on how you want to receive the output (`sends`). For exemple when you subscribe to a newsletter you might be able to indicate the frequency with which you want ot receive the emails.

Connoropolous:
Couldn't we setup a model, more simply, of just letting the subscriber
filter events?

thedavidmeister:
@Connoropolous yes i'm thinking that a standard pub/sub with topics/content filtering makes sense

- widely used pattern
- decoupled
- simple
- flexible

@zippy said last night in scrum that we could implement topics (multiple publishers and subscribers) and change over the names later without it impacting this ADR

ddd-mtl:
ADR is in `proposed` state so I don't understand why it is being blocked.

Letting subscriber filter the events might functionaly be the same but technically there is a difference in performance, since your are wasting CPU / network bandwith for processing and sending discarded events. This might not be a real issue locally so it depends how much 'performance' is important.

doc/architecture/decisions/0013-signals-listeners-model-and-api.md
zippy: Exactly what @ddd-mtl said.  Implementing this doesn't seem hard because all it means is that you have to provide the signaler the value passed in at registration time.  That doesn't seem hard.  Am I missing something think @lucksus?  

It is possible that this is a YAGNI, i.e. and I'm overgeneralizing, and we can implement it later if we want it.

lucksus: Hm, ok. What do you think about changing ```params``` to ```config-params``` to make this more clear? And ```sends``` -> ```arguments```? would make it clearer to me upon first read. I will go and approve this PR now nevertheless. don't want to block it any further.. ;)

doc/architecture/decisions/0013-signals-listeners-model-and-api.md
```diff
@@ -23,15 +23,16 @@ We will extend the API in a way that's roughly equivalent to the [Signal-slot pa
             {
               "name": "Post",
               "description: "signal emmited when a post is committed",
```
sphinxc0re: You are missing a quote there

doc/specs/dna.md
```diff
@@ -156,14 +156,24 @@ They are available in keyword specific Capabilities and function names
             },
             // ...
           ],
+          "signals": [
+            {
+              "name": "Post",
+              "params": null,
+              "description: "signal emmited when a post is committed",
```
sphinxc0re: Another missed quote

PR #153, opened 2018-07-19
Proposal: P2P IPC Abstraction

neonphog:


doc/architecture/decisions/0013-p2p-ipc-abstraction.md
```diff
@@ -0,0 +1,32 @@
+# 13. P2P - IPC Abstraction
```
zippy: pls use 14, I have used 13 for events in #152

doc/architecture/decisions/0013-p2p-ipc-abstraction.md
```diff
@@ -0,0 +1,32 @@
...
+
+It's frustrating to admit, but I'm still pretty slow at writing rust code. Oh, I can hack something together quickly (gogo `unwrap()/clone()`), but when I sit down to write efficient production code, I get bogged down in cyclic lifetime reference issues or any number of other arcane rustisisms.
+
+In this prototyping / proof-of-concept phase, I don't want to take the time to write efficient rust code, but I don't want to leave us with a bunch of unmaintainable inefficient spaghetti if we *do* decide to go forward with the prototype.
+
+## Decision
+
+Abstract the p2p library at the process level.
+
+- Host a [ZeroMQ](http://zeromq.org/) REP socket for the p2p process. This process can be in any language that supports zmq, and using any of the transports it supports. (Likely start with unix domain sockets for their high throughput).
```
zippy: what are the consequences for Windows users?  Will they have to install a separate component?

neonphog: zeromq in general works on windows, and the rust specific zmq crate does too:
- https://github.com/erickt/rust-zmq/pull/124
- https://github.com/erickt/rust-zmq/issues/125

doc/architecture/decisions/0013-p2p-ipc-abstraction.md
neonphog: updated to 14

neonphog:
@zippy - updated per comments

> I see the value of this approach for short term development. I guess the point of having a network abstraction layer (#99) is that it could wrap the connection to the IPC process anyways, and it could be brought into the main process.

Yep, exactly, I see the value of both abstractions, hosts could even decide if it makes sense to use the exact same p2p module either as a builtin component of holochain, or through the ipc mechanism.

doc/architecture/decisions/0014-p2p-ipc-abstraction.md
```diff
@@ -0,0 +1,32 @@
...
+Abstract the p2p library at the process level.
+
+- Host a [ZeroMQ](http://zeromq.org/) REP socket for the p2p process. This process can be in any language that supports zmq, and using any of the transports it supports. (Likely start with unix domain sockets for their high throughput).
+- Connect to the Host with a REQ socket from within the holochain rust code using the [zmq](https://crates.io/crates/zmq) crate.
+- Create a Holochain IPC transport type that fits within the proposed [network-abstraction](0007-abstraction-for-network-layer.md) framework.
+
+## Consequences
+
+### Pros
+* Ability to prototype and experiment with p2p solutions in other languages (including the more mature nodejs version of [libp2p](https://github.com/libp2p/js-libp2p)).
```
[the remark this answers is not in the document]
neonphog: I'm planning on using the native bindings to the c ZeroMQ implementation: https://www.npmjs.com/package/zeromq

neonphog:
@lucksus 

> But I'm not so sure about creating a network protocol so we can talk to the network process via a network socket.. That's what this means right? Or is it that with ZeroMQ we can stay at a quite abstract level and just define the possible RPCs HC core needs the network layer to implement?

Yeah, exactly, I think of domain sockets as an IPC mechanism similar to pipes, and in fact, many cross platform ipc libs use named pipes for windows (https://nodejs.org/api/net.html#net_identifying_paths_for_ipc_connections). ZeroMQ makes it easy to not worry about the details. Someone setting up the system can just say listen on `ipc://*:1234` on linux and `tcp://*:1234` on windows.

> imo, the network layer (as seen from Holochain core) must be implemented in the container ... I guess I just agree that we need to cleanly separate those modules in a way that could have the network layer run in its own process. :)

Yeah, we could go with api injection, or subprocess/fork-based ipc, but I feel like doing the ipc through ZeroMQ will give us a lot of ancillary benefits like:
- not having to worry about the timing on which process starts first
- multiple container/composers being able to use the same p2p connection
- simple api documentation: hey, use zeromq/messagepack and send these message types...

doc/architecture/decisions/0014-p2p-ipc-abstraction.md
```diff
@@ -0,0 +1,32 @@
...
+
+Date: 2018-07-19
+
+## Status
+
+Proposed
+
+## Context
+
+It's frustrating to admit, but I'm still pretty slow at writing rust code. Oh, I can hack something together quickly (gogo `unwrap()/clone()`), but when I sit down to write efficient production code, I get bogged down in cyclic lifetime reference issues or any number of other arcane rustisisms.
```
thedavidmeister: this doesn't seem like relevant context for an architecture design decision, it's just a comment about our personal situations at this point in time

doc/architecture/decisions/0014-p2p-ipc-abstraction.md
thedavidmeister: 👍

doc/architecture/decisions/0014-p2p-ipc-abstraction.md
```diff
@@ -0,0 +1,32 @@
...
+## Consequences
+
+### Pros
+* Ability to prototype and experiment with p2p solutions in other languages (including the more mature nodejs version of [libp2p](https://github.com/libp2p/js-libp2p)).
+* Separation of concerns in terms of security and stability.
+* Multiple Holochain processes can share the overhead (cpu, memory, bandwidth) of a single p2p network connection.
+
+### Cons
+* Work overhead for creation of additional abstraction layer.
+* Additional complexity for system setup / maintanence for both developers and end-users.
```
thedavidmeister: what additional complexity for developers and end-users?

neonphog:
@thedavidmeister - Addressed your comments, let me know your thoughts. Thanks!

doc/architecture/decisions/0014-p2p-ipc-abstraction.md
```diff
@@ -8,17 +8,17 @@ Proposed
 
 ## Context
 
-It's frustrating to admit, but I'm still pretty slow at writing rust code. Oh, I can hack something together quickly (gogo `unwrap()/clone()`), but when I sit down to write efficient production code, I get bogged down in cyclic lifetime reference issues or any number of other arcane rustisisms.
+Rust as a language, in my opinion, does not lend itself to rapid prototyping. Writing efficient Rust code is a problem domain unto itself. E.g. Is it worth returning borrowed references in this case to save memory? Am I able to craft the lifetimes appropriately given all the other api usages of my struct? What about synchronization? These questions are implementation details, not architectural. If we are still working out the details with the underlying architecture, we don't want to be spending brainpower on * how * we accomplish our experiments.
```
sphinxc0re: Rust code is no less efficient than C/C++ code. Do you mean that writing Rust code efficiently is difficult?

neonphog:
@sphinxc0re 

> Rust code is no less efficient than C/C++ code. Do you mean that writing Rust code efficiently is difficult?

I agree it's absolutely possible to write efficient Rust code, updated my wording to reflect this. I don't even necessarily mean more difficult, maybe more time consuming? Especially in experimentation phase.

Rust, by design, won't even compile unless I get the implementation details just right (there are many benefits to this!). But in C, if I'm trying to prototype quickly, at least I can compile and experiment while I'm still sorting out the details, and then valgrind once I settle on an approach.

doc/architecture/decisions/0014-p2p-ipc-abstraction.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Consequences
+
+### Pros
+* Ability to prototype and experiment with p2p solutions in other languages (including the more mature nodejs version of [libp2p](https://github.com/libp2p/js-libp2p)).
+* Separation of concerns in terms of security and stability.
+* Multiple Holochain processes can share the overhead (cpu, memory, bandwidth) of a single p2p network connection.
+
+### Cons
+* Work overhead for creation of additional abstraction layer.
```
thedavidmeister: this seems to imply the opposite of "rapid prototyping" that the context says we need?

doc/architecture/decisions/0014-p2p-ipc-abstraction.md
```diff
@@ -0,0 +1,32 @@
...
+* Additional complexity for system setup / maintanence for both developers and end-users. That is, they will have to start up / configure two processes instead of one.
```
thedavidmeister: this seems like a pretty big deal to me, so you want to prototype in a different language, but then require end users to use the prototype as a standalone system moving forward into production?

doc/architecture/decisions/0014-p2p-ipc-abstraction.md
neonphog: This part only has to be done once, and is complete on the javascript side:
- https://github.com/holochain/n3h/tree/master/lib/ipc
and functional if not finalized on the rust side:
- https://github.com/holochain/holochain-rust/pull/163

doc/architecture/decisions/0014-p2p-ipc-abstraction.md
neonphog: I think we all agree that we need the ultimate version of the p2p lib to be available as a built-in module for the rust client/container/composer whatever we are calling it.

But there are also many benefits to being able to run the p2p connection part as a separate process, that may make it worth-while to have, as an option, even moving forward in production:

- ability to share the p2p connection among multiple composers
- ability to permission sandbox the networking part of the code
- zmq even provides us the option to run the p2p connection process on a completely separate machine instance that could be optimized for networking with a simple config change

All I am proposing is that we start with this solution during the prototyping phase.

neonphog:
@sphinxc0re - merging this is blocked by your "requested changes" flag. Is there anything else you'd like to see done before this can be merged?

thedavidmeister:
i merged, @sphinxc0re ping me if you wanted something changed, we can do a followup

PR #163, opened 2018-07-24
WIP: IPC net connector

neonphog:
Probably not ready for review yet, but putting this up here in case anyone has thoughts on my approach.

### DIFF
--- a/net_ipc/Cargo.toml
+++ b/net_ipc/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "net_ipc"
+version = "0.1.0"
+authors = ["neonphog <neonphog@gmail.com>"]
+
+[dependencies]
+failure = "0.1.1"
+lazy_static = "1.0"
+rmp-serde = "0.13.7"
+serde = "1.0"
+serde_bytes = "0.10.4"
+serde_derive = "1.0"
+zmq = "0.8.2"
+
+[dev-dependencies]
+libc = "0.2.42"
--- a/net_ipc/src/errors.rs
+++ b/net_ipc/src/errors.rs
@@ -1,0 +1,28 @@
+use ::std;
+use failure;
+
+#[derive(Debug, Fail)]
+pub enum IpcError {
+    #[fail(display = "NoneError")]
+    NoneError,
+    #[fail(display = "Timeout")]
+    Timeout,
+    #[fail(display = "IpcError: {}", error)]
+    GenericError { error: String },
+}
+
+#[macro_export]
+macro_rules! gerr {
+    ($e:expr) => {
+        return Err(IpcError::GenericError {
+            error: $e.to_string(),
+        }.into());
+    };
+    ($fmt:expr, $($arg:tt)+) => {
+        return Err(IpcError::GenericError {
+            error: format!($fmt, $($arg)+),
+        }.into());
+    };
+}
+
+pub type Result<T> = std::result::Result<T, failure::Error>;
--- a/net_ipc/src/errors.rs
+++ b/net_ipc/src/errors.rs
@@ -1,5 +1,5 @@
-use ::std;
 use failure;
+use std;
 
 #[derive(Debug, Fail)]
 pub enum IpcError {
--- a/net_ipc/src/lib.rs
+++ b/net_ipc/src/lib.rs
@@ -1,0 +1,20 @@
+#[macro_use]
+extern crate failure;
+#[macro_use]
+extern crate lazy_static;
+extern crate rmp_serde;
+extern crate serde;
+extern crate serde_bytes;
+#[macro_use]
+extern crate serde_derive;
+extern crate zmq;
+
+pub mod msg_types;
+#[macro_use]
+pub mod errors;
+mod util;
+pub mod context;
+pub mod message;
+
+mod ipc_client;
+pub use ipc_client::IpcClient;
--- a/net_ipc/src/lib.rs
+++ b/net_ipc/src/lib.rs
@@ -12,9 +12,9 @@ extern crate zmq;
 pub mod msg_types;
 #[macro_use]
 pub mod errors;
-mod util;
 pub mod context;
 pub mod message;
+mod util;
 
 mod ipc_client;
 pub use ipc_client::IpcClient;
--- a/net_ipc/tests/echo_ipc_test.rs
+++ b/net_ipc/tests/echo_ipc_test.rs
@@ -1,0 +1,84 @@
+extern crate libc;
+extern crate net_ipc;
+#[macro_use]
+extern crate failure;
+
+use net_ipc::IpcClient;
+
+fn prep () -> std::process::Child {
+    assert!(std::process::Command::new("git")
+        .args(&["submodule", "update", "--init", "--recursive"])
+        .status()
+        .expect("failed updating git submodules")
+        .success(), "failed updating git submodules");
+    assert!(std::process::Command::new("npm")
+        .args(&["install", "--production"])
+        .current_dir("./tests/n3h")
+        .status()
+        .expect("failed running npm install")
+        .success(), "failed running npm install");
+    std::process::Command::new("node")
+        .args(&["./tests/n3h/examples/ipc/echo-server.js"])
+        .spawn()
+        .expect("failed running npm install")
+}
+
+#[test]
+fn it_can_send_and_call () {
+    let mut n3h_server = prep();
+    println!("n3h_server pid: {}", n3h_server.id());
+
+    let mut cli = IpcClient::new().unwrap();
+    cli.connect("ipc://echo-server.sock").unwrap();
+
+    let did_send = std::sync::Arc::new(std::sync::Mutex::new(false));
+    let did_call = std::sync::Arc::new(std::sync::Mutex::new(false));
+    let did_call_resp = std::sync::Arc::new(std::sync::Mutex::new(false));
+
+    let did_send_oth = did_send.clone();
+    let did_call_oth = did_call.clone();
+    let did_call_resp_oth = did_call_resp.clone();
+
+    cli.send(b"ab12", b"hello:send", Box::new(move |r| {
+        match did_send_oth.lock() {
+            Ok(mut s) => *s = true,
+            Err(_) => bail!("brains"),
+        }
+        println!("send result: {:?}", r);
+        Ok(())
+    })).unwrap();
+    cli.call(b"ab12", b"hello:call", Box::new(move |r| {
+        match did_call_oth.lock() {
+            Ok(mut s) => *s = true,
+            Err(_) => bail!("brains"),
+        }
+        println!("call result: {:?}", r);
+        Ok(())
+    }), Box::new(move |r| {
+        match did_call_resp_oth.lock() {
+            Ok(mut s) => *s = true,
+            Err(_) => bail!("brains"),
+        }
+        println!("call resp result: {:?}", r);
+        Ok(())
+    })).unwrap();
+    loop {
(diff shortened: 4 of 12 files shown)


zippy:
So far this looks great, and I love all tests and MockIPCSocket.  I was originally thinking that I would try and test #99 against rust-libp2p but now I'm wondering if I shouldn't test it against this branch instead, assuming that, for starters anyway, the P2P IPC might connect in with the javascript version.

net_ipc/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "net_ipc"
```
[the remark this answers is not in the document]
neonphog: done

net_ipc/src/errors.rs
```diff
@@ -0,0 +1,44 @@
+/*!
+This module holds net_ipc custom error types.
+*/
+
```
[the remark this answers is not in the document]
neonphog: done

net_ipc/src/errors.rs
```diff
@@ -0,0 +1,44 @@
...
+    /// Otherwise undefined error message
+    #[fail(display = "IpcError: {}", error)]
+    GenericError { error: String },
+}
+
+/**
+Macro akin to `bail!()` but returns an IpcError::GenericError.
+*/
+#[macro_export]
+macro_rules! gerr {
```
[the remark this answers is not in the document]
neonphog: good call : ), updated

neonphog:
@zippy @sphinxc0re @thedavidmeister - I think this is actually ready for review now. Thanks!

neonphog:
@thedavidmeister / @sphinxc0re - know why my integration test is showing up / throwing off the coverage metrics?
```
net_ipc/tests/echo_ipc_test.rs | 12.5% <12.5%> (ø)
```

net_ipc/src/errors.rs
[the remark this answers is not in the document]
ddd-mtl: Doesn't `bail!` obviously send a generic error?
`bail_ipc!` seems more meaningful, IMO.

thedavidmeister:
@neonphog code coverage _is_ the test runner

i think you can configure it to run the test and ignore the coverage https://github.com/xd009642/tarpaulin

net_ipc/tests/echo_ipc_test.rs
```diff
@@ -0,0 +1,200 @@
...
+        Ok("srv-got: `Error: IpcError: echo: srv-error`".to_string()),
+        frame.execute(b"call-error")
+    );
+
+    println!("### TEST SUITE FINISHED - CLEANUP ###");
+
+    frame.destroy();
+
+    println!("### END TEST SUITE ###");
+}
```
lucksus: Serious integration testing, awesome!

neonphog: @lucksus 

Awesome! Thanks for the feedback : ) Just a couple responses:

> why do we need all of this instead of just opening a websocket

Using unix sockets we can get up to 7X the throughput of TCP (https://github.com/rigtorp/ipc-bench). And with ZMQ we don't loose the flexibility of being able to fall back to TCP if the OS doesn't support them.

> something that is easy to se-/deserialize, like strings and JSON

I figured we'd often be sending raw binary data to other peers on the network, so I went with binary / https://crates.io/crates/rmp... if we'd rather go with json, that's a super simple switch I could make in about 10 minutes, but it would add overhead of base64 if we are sending binary stuff.

neonphog:
omgwtfpitatravis

hehey, I got my green checky :tada:

net_ipc/src/errors.rs
```diff
@@ -0,0 +1,38 @@
...
+
+use failure;
+use std;
+
+/// net_ipc-specific error types
+#[derive(Debug, Clone, Fail)]
+pub enum IpcError {
+    /// Translate an Option<_> unwrap into a Result::Err
+    #[fail(display = "NoneError")]
+    NoneError,
```
ddd-mtl: Can we change this to a better error name that is more explicit on what went wrong like "MissingData" ?

.travis.yml
```diff
@@ -7,6 +7,12 @@ branches:
     - master
     - develop
 
+before_script:
```
thedavidmeister: please move this into make to ensure it also ends up in docker and on the machine of anyone using make without docker

.travis.yml
```diff
@@ -28,8 +34,10 @@ jobs:
         - rm -rf $HOME/.cargo/registry/index
 
     - name: "Standard tests + code coverage"
-      # https://github.com/travis-ci/travis-ci/issues/5175#issuecomment-215016943
-      language: bash
+      language: node_js
```
thedavidmeister: what overhead does this add to build times?
why is it needed?

net_ipc/src/lib.rs
```diff
@@ -0,0 +1,29 @@
+//! Networking / P2P IPC Abstraction
+//!
+//! This crate allows holochain to connect to a running P2P client node over ZeroMq-based socket connection. The recommended ZeroMQ configuration is to use the `ipc:// ` protocol, which will make use of unix domain sockets in a linux or macOs environment. You may need to fall back to `tcp://` for other operating systems.
```
thedavidmeister: very long comment

split over a few lines?

.travis.yml
neonphog: This system library install is specific to Ubuntu Trusty. I feel like the Makefile should be platform agnostic. Most newer systems will have a system package that is viable for zmq (you can probably just `apt-get install libzmq3-dev`) It's just the old travis environment that needs the special install.

I'm up for suggestions on how to do this. I could potentially add something to the Makefile that tries to do a pkg-config for libzmq, and throws an error if it doesn't find it on the system... but it probably wouldn't work on Windows..

.travis.yml
neonphog: Super hard to compare travis build times. this one is 7 minutes with my change: https://travis-ci.org/holochain/holochain-rust/builds/433737994, develop looks like most builds are 12 - 13 minutes...

I'm using a nodejs package as the target for my [Integration Test](https://github.com/holochain/holochain-rust/pull/163/files/86e1b3d23d0f774e827a84650495fc704f421b97#diff-a6f8a312ae4afe529bb9b0b217a00e06) ... i tried installing the apt nodejs package... but again, the old version included with trusty isn't viable.

Again, open to other suggestions... I'm no travis guru by any means : )

net_ipc/src/lib.rs
neonphog: :+1: done

net_ipc/src/errors.rs
neonphog: :+1: done

.travis.yml
thedavidmeister: it's this bit:

screen shot 2018-09-27 at 9 01 41 am

so ~3 seconds. i think we can live with that @neonphog

.travis.yml
thedavidmeister: @neonphog the makefile only supports ubuntu right now so don't worry about windows, at least not for this merge

neonphog: @thedavidmeister - Ok, moved into Makefile, checks pkg-config so doesn't try to install if the lib exists already, mostly it will just do a system `sudo apt-get install libzmq3-dev`, or if `apt-get` doesn't exists it'll give you a warning that the build probably won't work but proceed. Two things to note:

- if the `TRAVIS` environment variable exists, it will use the custom updated version package
- requires `sudo` on all the travis tasks now

.travis.yml
```diff
@@ -65,6 +73,11 @@ jobs:
     - name: "Formatting & clippy"
       # https://github.com/travis-ci/travis-ci/issues/5175#issuecomment-215016943
       language: bash
+
+      # sudo is needed for codecov, && system library install which makes things 20-50s slower
+      # https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system
+      sudo: true
```
thedavidmeister: @neonphog do we absolutely need sudo? as the comment says, this adds 20-50s to each build you put it on

this extra time doesn't show up in the build logs, it's "boot time"

neonphog: @thedavidmeister  – I have to say I’m getting pretty frustrated with this process. I super appreciate your dedication to having a solid CI pipeline, we absolutely need something that is quick and accurate. At the same time, though, I’ve now spent more work hours going back and forth with Travis than I did writing the original code. We aren’t going to be able to reach our goals if we continue at this pace. Any thoughts or suggestions you have would be much appreciated!

To install the libzmq3-dev system library we need `sudo`. * Somehow * when that command was in the  `.travis.yml` file, it worked without the sudo required. I don’t know if travis has a special exception for `before_script` that allows sudo in there, or if the coverage task was somehow caching the library for the other tasks, or what… In any case, when I moved the install into the Makefile, the three jobs that did not have sudo started failing.

thedavidmeister: @neonphog yes, we've been moving things through more quickly as smaller PRs

is it possible for this to be smaller than 1k LOC?

i spent quite a few hours speeding travis up

neonphog: @thedavidmeister - are you asking me to break this up into multiple pull requests?

github is saying `+1,018 −7`, the Makefile changes alone are `35` LOC... so I could split out the stuff for getting `libzmq3-dev` installed and that would bring the remainder below 1000.

Do you have any suggestions about the sudo? I could make my own binary build for Ubuntu Trusty in a separate repo, link it to holochain-rust as a git submodule and only in the TRAVIS case set up environment variables that cause the zmq rust crate to link against that submodule binary... that would not require sudo, though I'm not sure how reliable it would be.

neonphog: Yeah, on second thought, i cannot do that. It's actually 5 libs we'd need to maintain, that's not the business we want to be in....

```
Get:2 http://download.opensuse.org/repositories/network:/messaging:/zeromq:/release-stable/xUbuntu_14.04 ./ libnorm1 1.5r6+dfsg1-9 [219 kB]
Get:3 http://download.opensuse.org/repositories/network:/messaging:/zeromq:/release-stable/xUbuntu_14.04 ./ libpgm-5.2-0 5.2.122~dfsg-2 [157 kB]
Get:4 http://download.opensuse.org/repositories/network:/messaging:/zeromq:/release-stable/xUbuntu_14.04 ./ libsodium18 1.0.11-1 [154 kB]
Get:5 http://download.opensuse.org/repositories/network:/messaging:/zeromq:/release-stable/xUbuntu_14.04 ./ libzmq5 4.2.5 [217 kB]
Get:6 http://download.opensuse.org/repositories/network:/messaging:/zeromq:/release-stable/xUbuntu_14.04 ./ libzmq3-dev 4.2.5 [398 kB]
```

neonphog: aha, @thedavidmeister - according to this https://github.com/travis-ci/travis-ci/issues/8877 - if travis detects a sudo command in the .travis.yml file, then sudo is auto-enabled... that explains why it was working in the before_script at least... it * was * actually enabling sudo for all the tasks

lucksus: So you need `sudo` for the `sudo apt-get install -y libzmq3-dev;` down there in Makefile, right? I think that should go into our image anyway: https://github.com/holochain/holochain-rust/blob/develop/docker/Dockerfile.ubuntu

right? @neonphog @thedavidmeister

neonphog: Docker is no longer used in travic-ci @lucksus - I * did * add that line to the docker image. What makes it worse is that Travis is still using a really old version of Ubuntu, so we cannot just apt-get the system lib (the version is incompatible with the rust crate) it's the four TRAVIS lines to install the binary packages built by the libzmq folks

Issue #93, opened 2018-07-02
debug()


zippy:
child of #89

Calls the log function.
Almost same as print.

PR #164, opened 2018-07-25
65 wasm memory manager + #93 debug()

ddd-mtl:
Simple memory manager for WASM.
Works as a stack. Limited to one page for now, hence the name: SinglePageStack
Added a wasm_utils crate, holding all the utils you would need in the rust-wasm nucleus.
Check `core_api\wasm-test\round_trip\src\lib.rs` to see how much easier it is now to write rust app.

How it works:
Host and WASM pass around only a i32.
That i32 is the last memory allocation on the stack: a i16 offset and a i16 length (which fits with the 64KiB sized of a memory Page).
Complex Input arguments should be stored on the latest allocation on the stack.
Complex Output arguments can be stored anywhere on stack.
ErrorCode passing is also made possible by convention: using i16 offset as error code and i16 length to zero to indicate its an error code.

In the future we could do same with i32 -> i64 and handle multiple memory Pages.

Edit:
I added #93-debug() in the PR since its a small addition that is coherent with this PR
debug() is still called print() for now.


core_api/src/lib.rs
```diff
@@ -478,4 +480,66 @@ mod tests {
...
+
+        // Call the exposed wasm function that calls the Commit API function
+        let result = hc.call("test_zome", "test_cap", "print_hello", r#"{}"#);
+
+        // TODO #165 - check runtime.print_output instead
+        // Expect normal OK result
+        match result {
+            Ok(result) => assert_eq!(result,""),
+            Err(_) => assert!(false),
+        };
```
sphinxc0re: You could write this way cleaner:
```rust
result.and_then(|res| {
    assert!(res.is_empty());
    Ok(())
}).unwrap();
```

core_api/src/lib.rs
```diff
@@ -478,4 +480,66 @@ mod tests {
...
+
+        // Call the exposed wasm function that calls the Commit API function
+        let result = hc.call("test_zome", "test_cap", "print_multiple", r#"{}"#);
+
+        // TODO #165 - check runtime.print_output instead
+        // Expect normal OK result
+        match result {
+            Ok(result) => assert_eq!(result,""),
+            Err(_) => assert!(false),
+        };
```
sphinxc0re: Same here

core_api/src/lib.rs
sphinxc0re: Perfect!

core/src/nucleus/ribosome.rs
```diff
@@ -38,35 +36,51 @@ enum HcApiFuncIndex {
...
+/// args: [0] encoded MemoryAllocation as u32
+/// Expecting a string as complex input argument
+/// Returns an HcApiReturnCode as I32
 fn invoke_print(runtime: &mut Runtime, args: &RuntimeArgs) -> Result<Option<RuntimeValue>, Trap> {
-    let arg: u32 = args.nth(0);
-    runtime.print_output.push(arg);
+    assert!(args.len() == 1);
+
+    // Read complex argument serialized in memory
+    let encoded_allocation: u32 = args.nth(0);
```
thedavidmeister: this whole section, asserting the args, getting the allocation, reading it as utf8, seems like a potentially nice util function, but i'm happy to do that in my branch :)

rustfmt.toml
```diff
@@ -1,2 +1,3 @@
 imports_indent = "block"
 merge_imports = true
+blank_lines_upper_bound = 2    # Clearer seperation helps scan-ability
```
thedavidmeister: don't change global formatting rules inside PRs

thedavidmeister:
@ddd-mtl also could you please put your explanation from your PR description somewhere in the code? it is useful context

core/src/nucleus/ribosome.rs
ddd-mtl: +1

rustfmt.toml
ddd-mtl: Why not? How am I supposed to change them? This setting is not going to change existing code.

wasm_utils/src/lib.rs
```diff
@@ -83,6 +84,8 @@ pub struct SinglePageStack {
 }
 
 
+#[allow(unknown_lints)]
+#[allow(cast_lossless)]
```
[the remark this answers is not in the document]
ddd-mtl: Otherwise Clippy complains about casting `u16 as u32`, which are non-issues...

PR #184, opened 2018-08-01
convert print() to debug(), #93

### LINKED ISSUE #93: debug()

zippy:
This branch is about ready to go with converting `print()` to `debug()`.  It required a bit of reworking of a bunch of sections of the code to pass around the `context` (which contains the logger) so that it could properly be used.  I'm pretty sure we will need access to the context for other api commands too.

The one thing that's missing is that I couldn't figure out how to remove duplication TestLogger as used in the core_api crate because `use holochain_core::instance::tests::TestLogger` from the core_api crate fails.  So I stuck a copy of those functions in `test_utils` but note that you can't use those functions from inside the tests in the `holochain_core` crate because of weird naming issues with crates.  @sphinxc0re any idea how to fix this?

closes #93


agent/src/lib.rs
```diff
@@ -18,6 +18,9 @@ impl Agent {
             content: text.to_string(),
         })
     }
+    pub fn to_string(&self) -> String {
```
sphinxc0re: Please insert an empty line above this one.

core/src/instance.rs
```diff
@@ -218,17 +224,70 @@ pub fn dispatch_action(
...
+
+    impl Logger for TestLogger {
+        fn log(&mut self, msg: String) {
+            self.log.push(msg);
+        }
+    }
+
+    // trying to get a way to print out what has been logged for tests without a read function.
+    // this currently fails
+    impl fmt::Debug for TestLogger {
```
sphinxc0re: You could just `#[derive(Debug)]` this

zippy: done.

agent/src/lib.rs
zippy: done

zippy:
in standup we decided to just add a ticket for the issue of cross-crate testing code #185 leave the duplicate code in place and get on with this ticket.

thedavidmeister:
@zippy dumb question: why isn't the context in the state tree?

zippy:
you know @thedavidmeister I started wondering that as I was coding this... I think the answer is that the context is meta stuff and also stuff that would be not be persisted.  For example, part of what's in the context is the Persister that persists the state.  No sure how you could persist the thing that persists you!

PR #190, opened 2018-08-06
 add default toolchain to mdbook dockerfile

Connoropolous:


sphinxc0re:
@Connoropolous this shouldn't be needed since the base image already has the default toolchain installed.

sphinxc0re:
You probably didn't rebuild the mdbook docker image and that let to the confusion of not having a default toolchain

Connoropolous:
Ah, ok, thanks :)

PR #191, opened 2018-08-07
Mdbook publishing automation

Connoropolous:
When new commits are added to `develop` branch, it automatically builds the MDBOOK and commits and pushes those changes to the github pages (`gh-pages`) branch on the repo.
For now, those pages publish at 
https://holochain.github.io/holochain-rust , but we may end up adding a custom domain name.

What's here is based off of https://gist.github.com/domenic/ec8b0fc8ab45f39403dd

- [x] set up deploy file
- [x] create deploy key (need @zippy to help with this one, for settings level access to the repo)
- [x] add deploy key label to .travis.yml
- [x] setup .travis.yml

It will likely need a follow up step of testing it live on travis and finalizing any little things


_config.yml
```diff
@@ -0,0 +1 @@
+theme: jekyll-theme-cayman
```
Connoropolous: This file is for github pages

travis-update-book.sh
```diff
@@ -56,10 +59,10 @@ ENCRYPTED_KEY_VAR="encrypted_${ENCRYPTION_LABEL}_key"
 ENCRYPTED_IV_VAR="encrypted_${ENCRYPTION_LABEL}_iv"
 ENCRYPTED_KEY=${!ENCRYPTED_KEY_VAR}
 ENCRYPTED_IV=${!ENCRYPTED_IV_VAR}
-openssl aes-256-cbc -K $ENCRYPTED_KEY -iv $ENCRYPTED_IV -in ../../../deploy_key.enc -out ../../../deploy_key -d
-chmod 600 ../../../deploy_key
+openssl aes-256-cbc -K $ENCRYPTED_KEY -iv $ENCRYPTED_IV -in ../../../build_docs_key.enc -out ../../../build_docs_key -d
```
sphinxc0re: I would rather like the keys to be injected via travis env variables

Connoropolous: Do you mean eliminating the encrypted key? It was encrypted with travis and set up an environment variable which is used to decrypt it. So it seems to me that it is injecting via travis env variables

sphinxc0re: No, I mean you can inject the key directly without encryption via travis. Travis will hide the key form the logs

sphinxc0re: screen shot 2018-08-08 at 14 46 38

Connoropolous: Would you be able to hop on a call briefly today to show me what you mean? I could fumble at trying to ask about it, but I think that might be easier

PR #211, opened 2018-08-10
Agreement on Rustfmt

ddd-mtl:
Some time ago I looked into rustfmt and took the time to figure out what would be my prefered rustfmt configuration.
I have separated the settings in categories of importance and put a rationale for each one as a line comment.

I'm not expecting us to agree on this but maybe there are some settings in there that everyone would actually already agree to?

My most important criteria (besides the obvious) is scan-ability. 
Why? Because [we don’t read, we scan](https://uxplanet.org/10-small-design-mistakes-we-still-make-1cd5f60bc708#7de8).
Another note on [unsustainable spacing](http://www.youtube.com/watch?v=ZsHMHukIlJY&t=22m5s).


zippy:
Well, I ran format with all of these options enabled, and commented out various ones to see the actual difference they made and how I feel about it.  The the truth about my particular preferences is that they don't actually make that much of a difference to me in terms of scan-ability in any absolute sense, because (for me) it mostly has to do with having a template in mind that things match to, rather than any particular version of the template.  I get that there are subtle differences that might improve things from a psychologial/physiological perspective, but I don't feel in a position to analyze that or make a recommendation based on that.  For my personal preferences I like more compact code so I can see more at one time, so I like:
```
fn_single_line               = true # For accessors
tab_spaces          = 2 # Helps save screen space and keep code below max_width
```
and I don't like:
```
control_brace_style = "AlwaysNextLine" # scan-ability - don't have to search for opening braces at end of lines
brace_style         = "AlwaysNextLine" # scan-ability - don't have to search for opening braces at end of lines
```

But for me the biggest issue here isn't our personal preferences, it's the question about how much it matters to follow the defaults i.e. the templates in other people's minds.  I don't believe that we necessarily have to blindly go by the defaults if they actually affect our productivity and happiness.  But lacking that, I would.

My suggestion is to generalize the [linter decision card](https://trello.com/c/Pk9f9j7F/18-the-clippy-linter-enforces-requirements-that-not-all-buy-into) in our trello to include formatting, and talk about this at the next retrospective, and then put the issue to bed.

ddd-mtl:
What about `tab_spaces = 2` ?
What is the reason for not doing two spaces tabs? (apart from "non idiomatic community standard" reasons)

sphinxc0re:
Rusts syntax is not very smooth. I worked with dynamic languages a lot in the past and those where the syntax is very smooth like ruby for example, 2 spaces are almost unrecognizable and make the code more readable. The main reason why Rust has it is that it pulls expressions apart and makes it more readable. In ruby, lines are easy to distinguish because of it's syntax. With Rust you need something "stronger" than 2 spaces.

PR #225, opened 2018-08-16
Split `fn start_action_loop` into multiple functions (see #169)

### LINKED ISSUE #169: isolate side effects of instance action loop so they can be unit tested
currently `start_action_loop` starts the loop and also processes each iteration of the loop inline

if we moved the iteration logic out into a standalone method(s) then we could:

- test it
- call it in other tests to avoid waiting X milliseconds and ending up with awkward/flaky logic like we had in the go version
- turn instance into an iterator just for fun

timotree3:
fixes #169 

I think that this resolves #169 although there are some next steps
with this solution that we may want to make:
- Create tests for various actions that use the `fn process_action` API or something better
- Create tests for the action processing code that use a test action type as well a test state type (requires state trait)

I'm just giving this a shot. Totally open to feedback or suggestions. It's possible I misunderstood something and this is a negative change. I figure it's better for me to put something out there and get feedback than for me to stew on a pull request asking a lot of questions until I feel sure it's good.

One thing I'm not sure about is why `#[derive(Clone)]` was
commented out and if there is some flaw with using it.

### DIFF
--- a/core/src/instance.rs
+++ b/core/src/instance.rs
@@ -4,7 +4,7 @@ use context::Context;
 use state::State;
 use std::{
     sync::{
-        mpsc::{channel, Sender, Receiver},
+        mpsc::{channel, Receiver, Sender},
         Arc, RwLock, RwLockReadGuard,
     },
     thread,
@@ -120,7 +120,8 @@ impl Instance {
     ) -> Vec<Observer> {
         // Mutate state
         {
-            let mut state = self.state
+            let mut state = self
+                .state
                 .write()
                 .expect("owners of the state RwLock shouldn't panic");
             *state = state.reduce(
@@ -136,7 +137,8 @@ impl Instance {
 
         // Run all observer closures
         {
-            let state = self.state
+            let state = self
+                .state
                 .read()
                 .expect("owners of the state RwLock shouldn't panic");
             let mut i = 0;
--- a/core/src/instance.rs
+++ b/core/src/instance.rs
@@ -13,7 +13,7 @@ pub const REDUX_DEFAULT_TIMEOUT_MS: u64 = 2000;
 
 /// Object representing a Holochain app instance.
 /// Holds the Event loop and processes it with the redux state model.
-//#[derive(Clone)]
+#[derive(Clone)]
 pub struct Instance {
     state: Arc<RwLock<State>>,
     action_channel: Sender<ActionWrapper>,
@@ -74,14 +74,21 @@ impl Instance {
         )
     }
 
-    /// Start the Event Loop on a seperate thread
-    pub fn start_action_loop(&mut self, context: Arc<Context>) {
+    /// Returns recievers for actions and observers that get added to this instance
+    fn start_listening(&mut self) -> (Receiver<ActionWrapper>, Receiver<Observer>) {
         let (tx_action, rx_action) = channel::<ActionWrapper>();
         let (tx_observer, rx_observer) = channel::<Observer>();
         self.action_channel = tx_action.clone();
         self.observer_channel = tx_observer.clone();
 
-        let state_mutex = self.state.clone();
+        (rx_action, rx_observer)
+    }
+
+    /// Start the Event Loop on a seperate thread
+    pub fn start_action_loop(&mut self, context: Arc<Context>) {
+        let (rx_action, rx_observer) = self.start_listening();
+
+        let sync_self = self.clone();
 
         thread::spawn(move || {
             let mut state_observers: Vec<Box<Observer>> = Vec::new();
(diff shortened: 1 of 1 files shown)


ddd-mtl:
I think this needs comments from @lucksus and @thedavidmeister.
From what I can tell, this PR seems to go in the right direction.
The loop revolving around a timeout is kind of awkward and would still need some refactoring I would say.

timotree3:
I agree. I'm not sure if upgrading that code should be in this PR or a different one though.

core/src/instance.rs
```diff
@@ -13,7 +13,7 @@ pub const REDUX_DEFAULT_TIMEOUT_MS: u64 = 2000;
 
 /// Object representing a Holochain app instance.
 /// Holds the Event loop and processes it with the redux state model.
-//#[derive(Clone)]
+#[derive(Clone)]
```
thedavidmeister: yeah i don't know why this was commented out, probably some legacy reason

core/src/instance.rs
```diff
@@ -74,14 +74,21 @@ impl Instance {
         )
     }
 
-    /// Start the Event Loop on a seperate thread
-    pub fn start_action_loop(&mut self, context: Arc<Context>) {
+    /// Returns recievers for actions and observers that get added to this instance
+    fn start_listening(&mut self) -> (Receiver<ActionWrapper>, Receiver<Observer>) {
```
thedavidmeister: "start listening" doesn't mean anything to me

"add/init/build channels" is what we're doing here right?

core/src/instance.rs
```diff
@@ -92,41 +99,59 @@ impl Instance {
...
                     }
                     Err(ref _recv_error) => {}
                 }
             }
         });
     }
 
+    /// Calls the reducers for an action and calls the observers with the new state
+    /// returns the new vector of observers
+    fn process_action(
```
thedavidmeister: can we get some basic tests around this to show that it is compatible with testing now?

core/src/instance.rs
timotree3: I'll change this.

core/src/instance.rs
```diff
@@ -349,6 +375,52 @@ pub mod tests {
...
+    /// to the state and that no observers or actions
+    /// are sent on the passed channels.
+    pub fn can_process_action() {
+        let mut instance = Instance::new();
+
+        let context = test_context("jane");
+        let (rx_action, rx_observer) = instance.initialize_channels();
+
+        let aw = test_action_wrapper_get();
+        let new_observers = instance.process_action(
```
thedavidmeister: this is great @timotree3 

much nicer than having to run the loop and wait in tests. deterministic. no chance for a test to hang.

timotree3:
I'm gonna update this with the conflicts. Thanks for the approval :smiley:

timotree3:
Regarding #169 and it's `@TODO` should I remove the TODO and add a commit that "fixes" #169?

core/src/instance.rs
```diff
@@ -76,54 +76,77 @@ impl Instance {
...
 
-        let state_mutex = Arc::clone(&self.state);
+        (rx_action, rx_observer)
+    }
+
+    /// Start the Event Loop on a seperate thread
+    pub fn start_action_loop(&mut self, context: Arc<Context>) {
+        let (rx_action, rx_observer) = self.initialize_channels();
+
+        let sync_self = self.clone();
```
lucksus: I just stumbled across this line in develop and was amazed that this did not break something yet.
Cloning self means that we have two different instances of Holochain, one that is accessible from the outside (the container) and one that holds and mutates the state inside the action loop. That is bad and I think the only reason this could get through tests is that we don't have integration tests yet that actually run an app - exactly what we are adding right now with the app-spec-rust repo and adding holoconsole (and thus a whole container instance) to the CI.

This PR was merged with only one approval. Sorry for having been late on reviewing this, but even if that happens we should not break our rules and merge without sufficient review, @timotree3, @thedavidmeister.

I will change this in the context of https://github.com/holochain/holochain-rust/issues/256 and an upcoming PR. Just FYI.

thedavidmeister: @lucksus it could also be that the original implementation had no tests for consistency of state.

it's ironic because this branch is a step towards making it easier to test for that type of thing by allowing us to step through "the loop" one iteration at a time

actually this is a good lead to chase up re: why tests are randomly broken at the moment!

timotree3: I didn't know about that policy of two reviews. Thanks for letting me know. 

I think that there is a misunderstanding here about what calling `.clone()` on an `Instance` actually does. 
If my understanding is correct (see explanation below), then cloning this `Instance` actually just creates new references to the same structures, and I don't see how it would have a problem with other accesses to the `Instance`.

So here's my explanation. 

The `impl` of `Clone` for `Instance` that is currently provided by `#[derive(Clone)]` is equivalent to this code.
```rust
impl Clone for Instance {
    fn clone(&self) -> Self {
        Instance {
            state: Arc::clone(&self.state), // creates a new reference to the same mutex
            action_channel: self.action_channel.clone(), // creates a new producer for the same channel
            observer_channel: self.observer_channel.clone(), // creates a new producer for the same channel
        }
    }
}
```
The code above (as mentioned in the comments) just creates new accesses to the same underlying concurrent data structures.
The call to `<Instance as Clone>::clone` that you're referring to has the following code:
https://github.com/holochain/holochain-rust/blob/4426c88a89d62dbf0d02136e3eeaef0a4fd11ce6/core/src/instance.rs#L93
If we substitute that `impl` of `Clone` we get this code.
```rust
let sync_self = Instance {
    state: Arc::clone(&self.state),
    action_channel: self.action_channel.clone(),
    observer_channel: self.observer_channel.clone(),
};
```
The reason that we're creating a "new" `Instance` here is so that we can call `process_action` with fewer explicit arguments than if we defined it as:
```rust
fn process_action(
        state: Arc<RwLock<State>>,
        action_channel: Sender<ActionWrapper>,
        observer_channel: Sender<Observer>,
        action_wrapper: ActionWrapper,
        mut state_observers: Vec<Observer>,
        rx_observer: &Receiver<Observer>,
        context: &Arc<Context>,
    ) -> Vec<Observer>
```
Before, we just didn't give the thread an `Instance` and copied the fields of an `Instance` individually and gave them to the thread.
https://github.com/holochain/holochain-rust/blob/5fdf4d11b1155eaaf2eea3777789b873d24926e8/core/src/instance.rs#L83-L86

Thanks for bringing this up. I agree that the usage of `clone` is a bit misleading and we should probably have some comments or change it to the explicit version where we construct a new `Instance` from cloning it's fields.
Does this clarification resolve your concerns or am I missing something, @lucksus @thedavidmeister?

thedavidmeister: i think the hanging tests are due to incorrect history length checking.... again...

lucksus: You are right, @timotree3. Thanks for making this clearly visible! 👍 
I got fooled by my precondition concepts reading that line.. :)

Issue #194, opened 2018-08-09
standardise reduce function signatures


thedavidmeister:
see `action::AgentReduceFn` and `action::NucleusReduceFn`

these are so close that they could be merged into a single `action::ReduceFn` type, which would automatically guarantee a standard context available to all reduce functions, making additions/refactors easier to plan

PR #227, opened 2018-08-19
Transition to use more common rust idioms

timotree3:
This is a pretty big changelog. I'm open to feedback and I hope this doesn't come across as critical. I love this code.

***Once I submit this PR, I'm going to write a review explaining each change that's not a pure formatting change.***

**REVIEW IN PROGRESS. MAY BE DELAYED IF I LOSE INTERNET ON MY PLANE**

Here are some 
### Transition `unwrap()` -> `expect("how it should be")`

This change is pretty pedantic.

The benefit of it is that if one of these `expect`s panics you'll get a descriptive error message that could be the difference between breaking your flow or not.

I figure that if I'm going to do all the work to convert these unwraps then we may as well convert them.

### Make functions own the arguments that they'd otherwise clone

Cloning takes resources and if the caller can choose whether to clone or just to move ownership to the function, that can be more efficient.

For example
```rust
#[derive(Clone, Debug)]
struct MyStruct(pub u64);

fn clones_and_prints(arg: &MyStruct) {
    println!("{:?}", arg.clone());
}

fn takes_and_prints(arg: MyStruct) {
    println!("{:?}", arg)
}

fn worse_main() {
    let to_print = MyStruct(4);
    clones_and_prints(&to_print); // 1 clone
}

fn better_main() {
    let to_print = MyStruct(4);
    takes_and_prints(to_print); // no clones
}
```

### Make functions return references to the objects that they'd otherwise clone

This is similar to above, gives the caller the choice of whether or not to clone. This helps in cases where cloning is unnecessary.

### Formatting differences done by `cargo fmt`

OOPS! I wish I hadn't run that command. :P Now the changes list is diluted.

### Shifting certain patterns to their more common counterparts

Examples:

```rust
// old way
get_some_option().and_then(|v| Some(v+1))
// new way
get_some_option().map(|v| v+1)

// old way
get_some_result().or_else(|e| panic!("some message"))
// new way
get_some_result().expect("some message")

// old way
let result = get_some_result();
if result.is_err() {
    return Err("some error")
}
let value = result.unwrap();
// new way
let value = get_some_result().or(Err("some error"))?;
```

### DIFF
--- a/core/src/action.rs
+++ b/core/src/action.rs
@@ -90,6 +90,8 @@ pub enum Action {
 }
 
 /// function signature for action handler functions
+// @TODO merge these into a single signature
+// @see https://github.com/holochain/holochain-rust/issues/194
 pub type AgentReduceFn = ReduceFn<AgentState>;
 pub type NucleusReduceFn = ReduceFn<NucleusState>;
 pub type ReduceFn<S> =
--- a/core/src/action.rs
+++ b/core/src/action.rs
@@ -10,13 +10,13 @@ use std::{
     sync::{mpsc::Sender, Arc},
 };
 
-#[derive(Clone, Debug)]
 /// Wrapper for actions that provides a unique ID
 /// The unique ID is needed for state tracking to ensure that we can differentiate between two
 /// Action dispatches containing the same value when doing "time travel debug".
 /// The standard approach is to drop the ActionWrapper into the key of a state history HashMap and
 /// use the convenience unwrap_to! macro to extract the action data in a reducer.
 /// All reducer functions must accept an ActionWrapper so all dispatchers take an ActionWrapper.
+#[derive(Clone, Debug)]
 pub struct ActionWrapper {
     action: Action,
     id: snowflake::ProcessUniqueId,
@@ -25,22 +25,22 @@ pub struct ActionWrapper {
 impl ActionWrapper {
     /// constructor from &Action
     /// internal snowflake ID is automatically set
-    pub fn new(a: &Action) -> Self {
+    pub fn new(a: Action) -> Self {
         ActionWrapper {
-            action: a.clone(),
+            action: a,
             // auto generate id
             id: snowflake::ProcessUniqueId::new(),
         }
     }
 
     /// read only access to action
-    pub fn action(&self) -> Action {
-        self.action.clone()
+    pub fn action(&self) -> &Action {
+        &self.action
     }
 
     /// read only access to id
-    pub fn id(&self) -> snowflake::ProcessUniqueId {
-        self.id
+    pub fn id(&self) -> &snowflake::ProcessUniqueId {
+        &self.id
     }
 }
 
@@ -90,12 +90,10 @@ pub enum Action {
 }
 
 /// function signature for action handler functions
-// @TODO merge these into a single signature
-// @see https://github.com/holochain/holochain-rust/issues/194
-pub type AgentReduceFn =
-    fn(Arc<Context>, &mut AgentState, &ActionWrapper, &Sender<ActionWrapper>, &Sender<Observer>);
-pub type NucleusReduceFn =
-    fn(Arc<Context>, &mut NucleusState, &ActionWrapper, &Sender<ActionWrapper>, &Sender<Observer>);
+pub type AgentReduceFn = ReduceFn<AgentState>;
+pub type NucleusReduceFn = ReduceFn<NucleusState>;
+pub type ReduceFn<S> =
+    fn(Arc<Context>, &mut S, &ActionWrapper, &Sender<ActionWrapper>, &Sender<Observer>);
 
 #[cfg(test)]
 pub mod tests {
@@ -113,21 +111,21 @@ pub mod tests {
 
     /// dummy action wrapper with test_action()
     pub fn test_action_wrapper() -> ActionWrapper {
-        ActionWrapper::new(&test_action())
+        ActionWrapper::new(test_action())
     }
 
     /// dummy action wrapper with commit of test_entry()
     pub fn test_action_wrapper_commit() -> ActionWrapper {
-        ActionWrapper::new(&Action::Commit(test_entry()))
+        ActionWrapper::new(Action::Commit(test_entry()))
     }
 
     /// dummy action for a get of test_hash()
     pub fn test_action_wrapper_get() -> ActionWrapper {
-        ActionWrapper::new(&Action::Get(test_hash()))
+        ActionWrapper::new(Action::Get(test_hash()))
     }
 
     pub fn test_action_wrapper_rzfr() -> ActionWrapper {
-        ActionWrapper::new(&Action::ReturnZomeFunctionResult(test_function_result()))
+        ActionWrapper::new(Action::ReturnZomeFunctionResult(test_function_result()))
     }
 
     #[test]
@@ -158,7 +156,7 @@ pub mod tests {
         let aw2 = test_action_wrapper();
 
         assert_eq!(aw1.action(), aw2.action());
-        assert_eq!(aw1.action(), test_action());
+        assert_eq!(aw1.action(), &test_action());
     }
 
     #[test]
@@ -178,7 +176,7 @@ pub mod tests {
         let aw1 = test_action_wrapper();
         let aw2 = test_action_wrapper();
 
-        assert_ne!(calculate_hash(&aw1), calculate_hash(&aw2),);
+        assert_ne!(calculate_hash(&aw1), calculate_hash(&aw2));
     }
 
 }
--- a/core/src/context.rs
+++ b/core/src/context.rs
@@ -15,13 +15,8 @@ pub struct Context {
 impl Context {
     // helper function to make it easier to call the logger
     pub fn log(&self, msg: &str) -> Result<(), HolochainError> {
-        let result = self.logger.lock();
-        match result {
-            Err(_) => return Err(HolochainError::LoggingError),
-            Ok(mut logger) => {
-                logger.log(msg.to_string());
-            }
-        }
+        let mut logger = self.logger.lock().or(Err(HolochainError::LoggingError))?;
+        logger.log(msg.to_string());
         Ok(())
     }
 }
--- a/core/src/error.rs
+++ b/core/src/error.rs
@@ -105,6 +105,6 @@ mod tests {
     fn can_return_result() {
         let result = raises_holochain_error(false);
 
-        assert_eq!(result, Ok(()))
+        assert!(result.is_ok());
     }
 }
--- a/core/src/error.rs
+++ b/core/src/error.rs
@@ -1,3 +1,4 @@
+use self::HolochainError::*;
 use std::{error::Error, fmt};
 
 /// module for holding Holochain specific errors
@@ -38,8 +39,6 @@ impl fmt::Display for HolochainError {
     }
 }
 
-use self::HolochainError::*;
-
 impl Error for HolochainError {
     fn description(&self) -> &str {
         match self {
@@ -72,37 +71,32 @@ mod tests {
     /// test that we can convert an error to a string
     fn to_string() {
         let err = HolochainError::new("foo");
-        assert_eq!("ErrorGeneric(\"foo\")", err.to_string(),);
+        assert_eq!(r#"ErrorGeneric("foo")"#, err.to_string());
     }
 
     #[test]
     /// test that we can convert an error to valid JSON
     fn test_to_json() {
         let err = HolochainError::new("foo");
-        assert_eq!("{\"error\":\"foo\"}", err.to_json());
+        assert_eq!(r#"{"error":"foo"}"#, err.to_json());
     }
 
     #[test]
     /// smoke test new errors
     fn can_instantiate() {
         let err = HolochainError::new("borked");
-        if let HolochainError::ErrorGeneric(err_msg) = err {
-            assert_eq!(err_msg, "borked")
-        } else {
-            assert!(false)
-        }
+
+        assert_eq!(HolochainError::ErrorGeneric("borked".to_string()), err);
     }
 
     #[test]
     /// test errors as a result and destructuring
     fn can_raise_holochain_error() {
-        let result = raises_holochain_error(true);
-        match result {
-            Ok(_) => assert!(false),
-            Err(err) => match err {
-                HolochainError::ErrorGeneric(msg) => assert_eq!(msg, "borked"),
-                _ => assert!(false),
-            },
+        let err = raises_holochain_error(true).expect_err("should return an error when yes=true");
+
+        match err {
+            HolochainError::ErrorGeneric(msg) => assert_eq!(msg, "borked"),
+            _ => panic!("raises_holochain_error should return an ErrorGeneric"),
         };
     }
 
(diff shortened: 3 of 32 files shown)


core/src/error.rs
```diff
@@ -72,48 +71,40 @@ mod tests {
...
     /// test errors as a returned result
     fn can_return_result() {
         let result = raises_holochain_error(false);
-        let result = match result {
-            Ok(x) => x,
-            Err(_) => panic!(),
-        };
-        assert_eq!(result, ())
+
+        assert_eq!(result, Ok(()))
```
sphinxc0re: You could just assert for `result.is_ok()`

core/src/hash_table/entry.rs
```diff
@@ -67,7 +67,7 @@ impl Entry {
 
     /// returns true if the entry is valid
     pub fn validate(&self) -> bool {
-        // always valid iff immutable and new() enforces validity
+        // always valid if immutable and new() enforces validity
```
sphinxc0re: I stumbled upon this as well. `iff` stands for `if and only if`

core/src/hash_table/pair.rs
```diff
@@ -10,40 +11,44 @@ pub struct Pair {
...
     /// Header is generated automatically
+    ///
     /// a Pair is immutable, but the chain is mutable if chain.push() is used.
+    ///
     /// this means that if two Pairs X and Y are generated for chain C then Pair X is pushed onto
     /// C to create chain C' (containing X), then Pair Y is no longer valid as the headers would
     /// need to include X. Pair Y can be regenerated with the same parameters as Y' and will be
     /// now be valid, the new Y' will include correct headers pointing to X.
+    ///
+    /// panics if entry is somehow invalid
```
sphinxc0re: I would put this under the `# Panics` section

core/src/instance.rs
```diff
@@ -149,45 +138,46 @@ impl Default for Instance {
...
+/// panics if called before start_action_loop
 pub fn dispatch_action_and_wait(
     action_channel: &Sender<ActionWrapper>,
     observer_channel: &Sender<Observer>,
-    action_wrapper: &ActionWrapper,
+    action_wrapper: ActionWrapper,
 ) {
     // Create blocking channel
-    let (sender, receiver) = channel::<bool>();
+    let (sender, receiver) = channel::<()>();
```
timotree3: We don't actually care what value gets sent so it's the perfect use for the unit type.

core/src/instance.rs
```diff
@@ -338,9 +327,9 @@ pub mod tests {
             .history
             .iter()
             .find(|aw| match aw.action() {
-                Action::ReturnZomeFunctionResult(_) => true,
+                Action::ReturnInitializationResult(_) => true,
```
timotree3: typo

core/src/instance.rs
```diff
@@ -373,17 +362,21 @@ pub mod tests {
...
+            &ActionWrapper::new(Action::InitApplication(dna.clone())),
             move |state: &State| match state.nucleus().dna() {
                 Some(dna) => {
-                    sender.send(dna).expect("test channel must be open");
-                    return true;
+                    sender
+                        .send(dna)
+                        // the channel stays connected until the first message has been sent
+                        // if this fails that means that it was called after having returned done=true
+                        .expect("observer called after done");
```
timotree3: see the comments

core/src/logger.rs
```diff
@@ -3,10 +3,9 @@
 //! gets emitted globaly from the container.
 
 use chrono::Local;
-use std::fmt;
 
 /// trait that defines the logging functionality that holochain_core requires
-pub trait Logger: fmt::Debug + Send {
+pub trait Logger: Send {
```
timotree3: Why would a logger implement `Debug`? I'm curious why this was there to begin with.

core/src/nucleus/ribosome/api/commit.rs
```diff
@@ -25,71 +25,74 @@ pub fn invoke_commit(
...
+    let entry_input: CommitArgs = match serde_json::from_str(&args_str) {
+        Ok(entry_input) => entry_input,
+        // Exit on error
+        Err(_) => {
+            // Return Error code in i32 format
+            return Ok(Some(RuntimeValue::I32(
+                HcApiReturnCode::ErrorSerdeJson as i32,
+            )));
+        }
+    };
```
timotree3: Unfortunately we can't use `.or()?` here because we want to return an `Ok` and `?` only returns `Err`s.

core/src/nucleus/ribosome/api/get.rs
```diff
@@ -26,37 +26,40 @@ pub fn invoke_get(runtime: &mut Runtime, args: &RuntimeArgs) -> Result<Option<Ru
...
-            let actions = state.agent().actions().clone();
-            if actions.contains_key(&action_wrapper) {
-                // @TODO never panic in wasm
-                // @see https://github.com/holochain/holochain-rust/issues/159
-                let v = &actions[&action_wrapper];
-                sender.send(v.clone()).expect("local channel to be open");
-                true
-            } else {
-                false
+            let mut actions_copy = state.agent().actions();
```
timotree3: 1 clone saved

core/src/nucleus/ribosome/api/get.rs
```diff
@@ -26,37 +26,40 @@ pub fn invoke_get(runtime: &mut Runtime, args: &RuntimeArgs) -> Result<Option<Ru
...
         &runtime.action_channel,
         &runtime.observer_channel,
         &action_wrapper.clone(),
         move |state: &::state::State| {
-            let actions = state.agent().actions().clone();
-            if actions.contains_key(&action_wrapper) {
-                // @TODO never panic in wasm
-                // @see https://github.com/holochain/holochain-rust/issues/159
-                let v = &actions[&action_wrapper];
-                sender.send(v.clone()).expect("local channel to be open");
```
timotree3: 2 clones saved

dna/src/lib.rs
```diff
@@ -132,11 +132,11 @@ impl Dna {
...
     ///
     /// let dna = Dna::new();
-    /// println!("json: {}", dna.to_json().unwrap());
+    /// println!("json: {}", dna.to_json());
     ///
     /// ```
-    pub fn to_json(&self) -> serde_json::Result<String> {
-        serde_json::to_string(self)
+    pub fn to_json(&self) -> String {
+        serde_json::to_string(self).expect("DNA should serialize")
```
timotree3: Everywhere I looked called `unwrap` on the result that this returns. This is one of the few situations where panicking does make sense because it's supposed to be guaranteed that it can serialize and it failing to is unrecoverable.

dna/src/lib.rs
```diff
@@ -181,12 +180,10 @@ impl Dna {
...
         let capability_name = capability_name.into();
 
-        let zome = self.zomes.iter().find(|z| z.name() == zome_name)?;
-        let capability = zome
-            .capabilities
+        self.zomes
             .iter()
-            .find(|c| c.name == capability_name)?;
-        Some(&capability.code)
+            .find(|z| z.name() == zome_name)
```
timotree3: Now that I look at it this should probably be replace with `self.get_zome`

dna_c_binding/src/lib.rs
```diff
@@ -58,10 +58,7 @@ pub extern "C" fn holochain_dna_to_json(ptr: *const Dna) -> *mut c_char {
             &*ptr
         };
 
-        let json = match dna.to_json() {
-            Ok(s) => s,
-            Err(_) => return std::ptr::null_mut(),
-        };
```
timotree3: This is the one case that I saw use `dna.to_json` and not call `unwrap`. In this case the behavior is still the same on failure though. It will panic and due to the `catch_unwind` it will still return `null_mut()`

core/src/hash_table/pair.rs
timotree3: There are several other cases of adding a panics documentation comment. Do you think they should all be in `# Panics` sections?

core/src/hash_table/entry.rs
timotree3: I'll update the PR with a rephrased version then.

core/src/error.rs
timotree3: I don't actually think that's clearer. 
Consider [this case](http://play.rust-lang.org/?gist=1520c29b94800ec83018a60a59ef6ec0&version=stable&mode=debug&edition=2015) which outputs:
```
thread 'main' panicked at 'assertion failed: `(left == right)`
 left: `Err(4)`,
 right: `Ok(())`', src/main.rs:3:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```
versus
[this other case](http://play.rust-lang.org/?gist=f16b58a2c17f755c650be6237fe9db55&version=stable&mode=debug&edition=2015) which outputs:
```
thread 'main' panicked at 'assertion failed: result.is_ok()', src/main.rs:3:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```
I think that the first error message is more descriptive.

If you disagree I can change this though.

core/src/hash_table/pair.rs
sphinxc0re: Yes, definitely. Most of the `std` documentation follows this pattern

core/src/error.rs
sphinxc0re: We might be more flexible with the latter. The actual return value has no real significance here. But we want to know, when the method call failed.

core/src/hash_table/entry.rs
thedavidmeister: iff is totally a word, haha

timotree3:
I've added a new commit addressing your concerns @sphinxc0re. Thank you for the review! Does it look like this solves #194 or should I add back the TODO?

dna/src/lib.rs
```diff
@@ -165,25 +165,24 @@ impl Dna {
...
         capability_name: &str,
     ) -> Option<&'a wasm::DnaWasm> {
-        zome.capabilities
+        let capability = zome
+            .capabilities
             .iter()
-            .find(|c| c.name == capability_name)
-            .map(|capability| &capability.code)
+            .find(|c| c.name == capability_name)?;
+        Some(&capability.code)
```
sphinxc0re: I believe calling `map` on an `Option<T>` is more idiomatic.

sphinxc0re:
@timotree3 no, #194 is about reducer functions. This is already being addressed by the refactor to the actor system

core/src/action.rs
```diff
@@ -90,12 +90,10 @@ pub enum Action {
...
-// @TODO merge these into a single signature
-// @see https://github.com/holochain/holochain-rust/issues/194
-pub type AgentReduceFn =
-    fn(Arc<Context>, &mut AgentState, &ActionWrapper, &Sender<ActionWrapper>, &Sender<Observer>);
-pub type NucleusReduceFn =
-    fn(Arc<Context>, &mut NucleusState, &ActionWrapper, &Sender<ActionWrapper>, &Sender<Observer>);
+pub type AgentReduceFn = ReduceFn<AgentState>;
+pub type NucleusReduceFn = ReduceFn<NucleusState>;
+pub type ReduceFn<S> =
+    fn(Arc<Context>, &mut S, &ActionWrapper, &Sender<ActionWrapper>, &Sender<Observer>);
```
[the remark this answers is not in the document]
sphinxc0re: This is being addressed by the refactor to the Actor system provided by `riker`

dna/src/lib.rs
timotree3: It seems to me that `get_wasm_for_capability`
https://github.com/holochain/holochain-rust/blob/587f2335daf028441e1de8c45ee127b753cbdc9c/dna/src/lib.rs#L176-L186
and `get_validation_bytecode_for_entry_type`
https://github.com/holochain/holochain-rust/blob/587f2335daf028441e1de8c45ee127b753cbdc9c/dna/src/lib.rs#L189-L201
are perfect cases to use the `?` operator on `Option`s. Given the context of that these are the two functions next to `get_capability`, it seems to make sense to me to continue the `?` pattern. If you disagree I can change these to `and_then`s and `map`s.
https://github.com/holochain/holochain-rust/blob/587f2335daf028441e1de8c45ee127b753cbdc9c/dna/src/lib.rs#L162-L201

thedavidmeister:
@timotree3 FYI you _can't_ run `cargo fmt` and push, it will fail travis

this is because the formatting rules change over time and travis is locked at a specific version

to get the correct formatting rules, you need to run it from docker

`. docker/run-fmt`

dna/src/lib.rs
sphinxc0re: No, sure leave it

core/src/context.rs
```diff
@@ -15,13 +15,8 @@ pub struct Context {
...
     // helper function to make it easier to call the logger
     pub fn log(&self, msg: &str) -> Result<(), HolochainError> {
-        let result = self.logger.lock();
-        match result {
-            Err(_) => return Err(HolochainError::LoggingError),
-            Ok(mut logger) => {
-                logger.log(msg.to_string());
-            }
-        }
+        let mut logger = self.logger.lock().or(Err(HolochainError::LoggingError))?;
```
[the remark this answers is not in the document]
ddd-mtl: I disagree. A return statement is obfuscated into a single character at end of line which looks like an upside down semi-colon. It makes it hard to find a function's exit points when scanning code.
Also I find `or`  misleading and should have been something like `unwrap_or` so we can see that an unwrap is done.
So I agree this is probably more idiomatic Rust, but idiomatic Rust is not always great and I don't see it as a priority when writing code.

timotree3:
> One disturbing thing for me is changing all the unwraps into expects. You seem to imply that unwrap is not idiomatic, but its part of Result so if it weren't what is this function for?

The way I see it, the panicking utilities, `unwrap`, `expect`, `assert!`, etc. all serve similar purposes: to guarantee a certain invariant about an implementation.

For example if I published a crate with a parser that used regular expressions internally, it would make sense to call `Regex::new(...).unwrap()` on a pattern that is written by the programmer. That is, the unwrap failing is always a bug. It doesn't make sense to be returning errors for every conceivable bug. Panicking on a violation of an invariant is perfectly acceptable. 

The difference between `unwrap` and `expect` is that `unwrap` is quicker and `expect` provides more helpful error messages if you put in more effort. I believe that it's perfectly fine and idiomatic to use `unwrap` in a program that's being rapidly prototyped or when it's not worth putting effort into an error message for code that will soon become irrelevant.

As someone who tries to write very high-quality code from the start, I prefer to use `expect` over `unwrap` or find a way to statically guarantee my invariant, but I think that that perspective has many weaknesses as well.

*Example taken from [this post](https://www.reddit.com/r/rust/comments/8t8hp6/how_do_you_feel_about_using_expect_and_unwrap_in/e15otl2/).*

core/src/context.rs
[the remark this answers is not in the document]
timotree3: Good point, fair enough. I don't find `?`s that hard to see especially when they're highlighted in my editor.

timotree3:
Thank you all for your reviews! Now that you've approved, is there anything else I can do for this PR or will it just get merged at some point?

ddd-mtl:
Done. Merged.
The main weakness I see with writing "very high-quality code from the start" is speed.
Quality is just one aspect of the quality-cost-delivery triangle of project management.
Emphasizing one side is detrimental to the other sides of the triangle.
And sadly in a business setting quality is usually the least important of the 3.

PR #230, opened 2018-08-19
218 wasm globals

ddd-mtl:
Added `init_globals` zome api function which is automatically called by ribosome at start of a call to a zome wasm function.
Haven't managed to make the `run_start` work,  but lazy_statics makes it work anyway somehow.

Implemented `app_name` and `app_agent_id_str` globals.

Added `hc_` prefix to zome api functions otherwise it conflicts with hdk api.
Also propagated new zome api function names accord to same PR in `hdk-rust`.


core/src/nucleus/ribosome/api/get.rs
```diff
@@ -87,7 +90,7 @@ mod tests {
     #[test]
     /// test that we can round trip bytes through a get action and it comes back from wasm
     fn test_get_round_trip() {
-        let (runtime, _) = test_zome_api_function_runtime("get", test_args_bytes());
+        let (runtime, _) = test_zome_api_function_runtime("hc_get_entry", test_args_bytes());
```
thedavidmeister: nice, i've been heading in this direction on the hash table side of things too

a flat "get" is just not that helpful when we have so many similar things floating around >.<

ddd-mtl:
@Connoropolous I updated the mdbook for zome api stuff.

core/src/nucleus/ribosome/api/get.rs
sphinxc0re: I would like to see all of these function names in consts

Issue #195, opened 2018-08-09
abstraction for inspecting the state history for specific actions, don't inspect history state


thedavidmeister:
currently it is quite fiddly to find a specific action in the history, it looks something like this:

```rust
        while instance
            .state()
            .history
            .iter()
            .find(|aw| match aw.action().signal() {
                Signal::ReturnZomeFunctionResult(_) => true,
                _ => false,
            }) == None
        {
            println!("Waiting for ReturnZomeFunctionResult from genesis");
            sleep(Duration::from_millis(10))
        }
```

it seems like this could be wrapped up into some tooling, with or without the loop

the alternative is checking for unreliable properties of the state, like the length of history

the history length is being read in multiple places in the codebase making everything fragile

it is too easy for some action logic to change slightly somewhere and leave the test suite hanging with no explanation in some seemingly unrelated part of the codebase

<img width="614" alt="screen shot 2018-08-07 at 11 11 59 pm" src="https://user-images.githubusercontent.com/629710/43881659-405bb2a4-9bf0-11e8-918d-c37215d38e75.png">

maackle:
#732 is a large WIP that includes a utility function `expect_action` that can help clean these tests up

thedavidmeister:
@maackle would it be worth popping the history length utils out into a PR that we can get through more easily than the full signals branch?

maackle:
@thedavidmeister the test util depends on the signal implementation which still needs to be tested and vetted, so it can't really be isolated at the moment

Issue #75, opened 2018-06-29
seralization of headers should match JSON schema from go version


thedavidmeister:

thedavidmeister:
no need

PR #238, opened 2018-08-22
186 system entry types abstraction

ddd-mtl:
### System Entry Types
Added `enum EntryType` which lists all entry types. `Data` is the enum variant name for the non-system entry type (entries defined by app dna).
This enum can be converted to a string and has a system-prefix like in the Go Version.

System entry types uses the `Entry` struct. Its just the content that changes and must follow a specific schema (yet to be defined by system entry type).
So I added a `ToEntry` trait, which is a `to_entry()` method that returns an `Entry`.
`Dna` and `Agent` implements that trait, which converts them to their system entry equivalent.

### Genesis Pair at InitApplication 
I added the genesis pair commit in the InitApplication reducer. 
The genesis pair holds the DNA system entry. 

### Renamed functions
When trying to understand the code on Chains and Entries, it seemed important to me to rename a couple of methods to make it clearer and coherent with previous holochain documentation like [this picture](https://holochain.org/assets/images/holochain/private_source_chain_diagram.png). The renamed methods are in `Chain` and `Header`.

The most controversial might be `next` changed to `prev`
I understand that `next` is for coherence with the Iterator pattern, but for me coherence with the mental modal of an append-only chain is more important. `prev` means going back in time which is what we are doing when going through chain from top to bottom.

I also find `Pair` confusing but I didn't find a better alternative than `ChainLink` which I also don't like very much...

### DIFF
--- a/agent/src/lib.rs
+++ b/agent/src/lib.rs
@@ -11,7 +11,6 @@ impl Identity {
     }
 }
 
-
 #[derive(Clone, Debug, PartialEq)]
 pub struct Agent {
     identity: Identity,
--- a/agent/src/lib.rs
+++ b/agent/src/lib.rs
@@ -7,7 +7,7 @@ pub struct Identity {
 
 impl Identity {
     pub fn new(content: String) -> Self {
-        Identity { content: content }
+        Identity { content }
     }
 }
 
--- a/agent/src/lib.rs
+++ b/agent/src/lib.rs
@@ -4,6 +4,14 @@
 pub struct Identity {
     content: String,
 }
+
+impl Identity {
+    pub fn new(content: String) -> Self {
+        Identity { content: content }
+    }
+}
+
+
 #[derive(Clone, Debug, PartialEq)]
 pub struct Agent {
     identity: Identity,
--- a/core/src/agent/state.rs
+++ b/core/src/agent/state.rs
@@ -284,7 +284,7 @@ pub mod tests {
         );
 
         assert_eq!(
-            "{\"header\":{\"entry_type\":\"testEntryType\",\"time\":\"\",\"next\":null,\"entry_hash\":\"QmbXSE38SN3SuJDmHKSSw5qWWegvU7oTxrLDRavWjyxMrT\",\"same_next\":null,\"signature\":\"\"},\"entry\":{\"content\":\"test entry content\",\"entry_type\":\"testEntryType\"}}",
+            "{\"header\":{\"entry_type\":\"testEntryType\",\"timestamp\":\"\",\"prev\":null,\"entry_hash\":\"QmbXSE38SN3SuJDmHKSSw5qWWegvU7oTxrLDRavWjyxMrT\",\"entry_signature\":\"\",\"prev_same\":null},\"entry\":{\"content\":\"test entry content\",\"entry_type\":\"testEntryType\"}}",
             ActionResponse::Get(Some(test_pair())).to_json(),
         );
         assert_eq!("", ActionResponse::Get(None).to_json(),);
--- a/core/src/agent/state.rs
+++ b/core/src/agent/state.rs
@@ -140,7 +140,7 @@ fn reduce_get(
     // @see https://github.com/holochain/holochain-rust/issues/167
 
     let result = chain
-        .get_entry(&key)
+        .entry(&key)
         .expect("should be able to get entry that we just added");
     state
         .actions
@@ -286,7 +286,7 @@ pub mod tests {
         );
 
         assert_eq!(
-            "{\"header\":{\"entry_type\":\"testEntryType\",\"timestamp\":\"\",\"prev\":null,\"entry_hash\":\"QmbXSE38SN3SuJDmHKSSw5qWWegvU7oTxrLDRavWjyxMrT\",\"entry_signature\":\"\",\"prev_same\":null},\"entry\":{\"content\":\"test entry content\",\"entry_type\":\"testEntryType\"}}",
+            "{\"header\":{\"entry_type\":\"testEntryType\",\"timestamp\":\"\",\"link\":null,\"entry_hash\":\"QmbXSE38SN3SuJDmHKSSw5qWWegvU7oTxrLDRavWjyxMrT\",\"entry_signature\":\"\",\"link_same_type\":null},\"entry\":{\"content\":\"test entry content\",\"entry_type\":\"testEntryType\"}}",
             ActionResponse::Get(Some(test_pair())).to_json(),
         );
         assert_eq!("", ActionResponse::Get(None).to_json());
--- a/core/src/agent/state.rs
+++ b/core/src/agent/state.rs
@@ -108,7 +108,7 @@ fn reduce_commit(
 
     state.actions.insert(
         action_wrapper.clone(),
-        ActionResponse::Commit(chain.push(&entry)),
+        ActionResponse::Commit(chain.push_entry(&entry)),
     );
 }
 
@@ -134,7 +134,7 @@ fn reduce_get(
     // drop in a dummy entry for testing
     let mut chain = Chain::new(Rc::new(MemTable::new()));
     let e = Entry::new("testEntryType", "test entry content");
-    chain.push(&e).unwrap();
+    chain.push_entry(&e).unwrap();
 
     // @TODO if the get fails local, do a network get
     // @see https://github.com/holochain/holochain-rust/issues/167
@@ -284,7 +284,7 @@ pub mod tests {
         );
 
         assert_eq!(
-            "{\"header\":{\"entry_type\":\"testEntryType\",\"time\":\"\",\"next\":null,\"entry\":\"QmbXSE38SN3SuJDmHKSSw5qWWegvU7oTxrLDRavWjyxMrT\",\"type_next\":null,\"signature\":\"\"},\"entry\":{\"content\":\"test entry content\",\"entry_type\":\"testEntryType\"}}",
+            "{\"header\":{\"entry_type\":\"testEntryType\",\"time\":\"\",\"next\":null,\"entry_hash\":\"QmbXSE38SN3SuJDmHKSSw5qWWegvU7oTxrLDRavWjyxMrT\",\"same_next\":null,\"signature\":\"\"},\"entry\":{\"content\":\"test entry content\",\"entry_type\":\"testEntryType\"}}",
             ActionResponse::Get(Some(test_pair())).to_json(),
         );
         assert_eq!("", ActionResponse::Get(None).to_json(),);
--- a/core/src/chain/mod.rs
+++ b/core/src/chain/mod.rs
@@ -34,7 +34,7 @@ impl<T: HashTable> Iterator for ChainIterator<T> {
     fn next(&mut self) -> Option<Pair> {
         let previous = self.current.take();
         self.current = previous.as_ref()
-                        .and_then(|p| p.header().prev())
+                        .and_then(|p| p.header().link())
                         // @TODO should this panic?
                         // @see https://github.com/holochain/holochain-rust/issues/146
                         .and_then(|h| self.table.get(&h).expect("getting from a table shouldn't fail"));
@@ -112,7 +112,7 @@ impl<T: HashTable> Chain<T> {
         }
 
         let top_pair = self.top().as_ref().map(|p| p.key());
-        let next_pair = pair.header().prev();
+        let next_pair = pair.header().link();
 
         if top_pair != next_pair {
             return Err(HolochainError::new(&format!(
@@ -151,12 +151,12 @@ impl<T: HashTable> Chain<T> {
     }
 
     /// get a Pair by Pair/Header key from the HashTable if it exists
-    pub fn get_pair(&self, k: &str) -> Result<Option<Pair>, HolochainError> {
+    pub fn pair(&self, k: &str) -> Result<Option<Pair>, HolochainError> {
         self.table.get(k)
     }
 
     /// get an Entry by Entry key from the HashTable if it exists
-    pub fn get_entry(&self, entry_hash: &str) -> Result<Option<Pair>, HolochainError> {
+    pub fn entry(&self, entry_hash: &str) -> Result<Option<Pair>, HolochainError> {
         // @TODO - this is a slow way to do a lookup
         // @see https://github.com/holochain/holochain-rust/issues/50
         Ok(self
@@ -349,7 +349,7 @@ pub mod tests {
             .expect("pushing a valid entry to an exlusively owned chain shouldn't fail");
         assert_eq!(
             Some(&p),
-            c.get_pair(&p.key())
+            c.pair(&p.key())
                 .expect("getting an entry from a chain shouldn't fail")
                 .as_ref()
         );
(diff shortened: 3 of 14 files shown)


core/src/chain/mod.rs
```diff
@@ -469,11 +480,12 @@ pub mod tests {
...
+        let expected_json = "[{\"header\":{\"entry_type\":\"testEntryType\",\"timestamp\":\"\",\"prev\":\"QmPT5HXvyv54Dg36YSK1A2rYvoPCNWoqpLzzZnHnQBcU6x\",\"entry_hash\":\"QmbXSE38SN3SuJDmHKSSw5qWWegvU7oTxrLDRavWjyxMrT\",\"entry_signature\":\"\",\"prev_same\":\"QmawqBCVVap9KdaakqEHF4JzUjjLhmR7DpM5jgJko8j1rA\"},\"entry\":{\"content\":\"test entry content\",\"entry_type\":\"testEntryType\"}},{\"header\":{\"entry_type\":\"testEntryTypeB\",\"timestamp\":\"\",\"prev\":\"QmawqBCVVap9KdaakqEHF4JzUjjLhmR7DpM5jgJko8j1rA\",\"entry_hash\":\"QmPz5jKXsxq7gPVAbPwx5gD2TqHfqB8n25feX5YH18JXrT\",\"entry_signature\":\"\",\"prev_same\":null},\"entry\":{\"content\":\"other test entry content\",\"entry_type\":\"testEntryTypeB\"}},{\"header\":{\"entry_type\":\"testEntryType\",\"timestamp\":\"\",\"prev\":null,\"entry_hash\":\"QmbXSE38SN3SuJDmHKSSw5qWWegvU7oTxrLDRavWjyxMrT\",\"entry_signature\":\"\",\"prev_same\":null},\"entry\":{\"content\":\"test entry content\",\"entry_type\":\"testEntryType\"}}]"
+        ;
```
sphinxc0re: Same here

core/src/hash_table/pair.rs
```diff
@@ -161,7 +164,8 @@ pub mod tests {
     #[test]
     /// test JSON roundtrip for pairs
     fn json_roundtrip() {
-        let json = "{\"header\":{\"entry_type\":\"testEntryType\",\"time\":\"\",\"next\":null,\"entry\":\"QmbXSE38SN3SuJDmHKSSw5qWWegvU7oTxrLDRavWjyxMrT\",\"type_next\":null,\"signature\":\"\"},\"entry\":{\"content\":\"test entry content\",\"entry_type\":\"testEntryType\"}}";
+        let json = "{\"header\":{\"entry_type\":\"testEntryType\",\"timestamp\":\"\",\"prev\":null,\"entry_hash\":\"QmbXSE38SN3SuJDmHKSSw5qWWegvU7oTxrLDRavWjyxMrT\",\"entry_signature\":\"\",\"prev_same\":null},\"entry\":{\"content\":\"test entry content\",\"entry_type\":\"testEntryType\"}}"
+        ;
```
sphinxc0re: same here

core/src/agent/state.rs
```diff
@@ -284,7 +284,7 @@ pub mod tests {
         );
 
         assert_eq!(
-            "{\"header\":{\"entry_type\":\"testEntryType\",\"time\":\"\",\"next\":null,\"entry\":\"QmbXSE38SN3SuJDmHKSSw5qWWegvU7oTxrLDRavWjyxMrT\",\"type_next\":null,\"signature\":\"\"},\"entry\":{\"content\":\"test entry content\",\"entry_type\":\"testEntryType\"}}",
+            "{\"header\":{\"entry_type\":\"testEntryType\",\"timestamp\":\"\",\"prev\":null,\"entry_hash\":\"QmbXSE38SN3SuJDmHKSSw5qWWegvU7oTxrLDRavWjyxMrT\",\"entry_signature\":\"\",\"prev_same\":null},\"entry\":{\"content\":\"test entry content\",\"entry_type\":\"testEntryType\"}}",
```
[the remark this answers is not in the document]
ddd-mtl: I don't think it is easier because its more of a regression test.
That string with the escape characters is a copy-past from the console's output.
So its actually easier keep it in this output format which has the escape characters, and copy-past again when we consciously change things.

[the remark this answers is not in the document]
sphinxc0re: Why does the console output have escape characters? Also, what is the correlation to a regression test?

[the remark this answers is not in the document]
ddd-mtl: Because this is what `serde_json::to_string()` outputs, which is called in the `assert_eq!`
When you copy-past the output into your test as the the thing to test against your are doing a regression test, i.e making sure the output stays the same.

agent/src/lib.rs
```diff
@@ -4,6 +4,13 @@
 pub struct Identity {
     content: String,
 }
+
+impl Identity {
+    pub fn new(content: String) -> Self {
+        Identity { content: content }
```
thedavidmeister: nitpick: `Identity { content }` should work if you don't need a clone or similar

core/src/hash_table/header.rs
```diff
@@ -44,49 +47,39 @@ impl Header {
...
                 .map(|p| p.header().hash()),
             // @TODO implement signatures
             // https://github.com/holochain/holochain-rust/issues/71
-            signature: String::new(),
+            entry_signature: String::new(),
         }
     }
 
-    /// entry_type getter
+    /// getters
```
thedavidmeister: this isn't going to play nice with the generated API docs is it?

i have been adding a comment to every function individually as each one has it's own page/section in the generated docs

core/src/hash_table/sys_entry.rs
```diff
@@ -0,0 +1,194 @@
...
+    fn can_commit_dna() {
+        // Create Context, Agent, Dna, and Commit AgentIdEntry Action
+        let context = test_context("alex");
+        let dna = test_utils::create_test_dna_with_wat("test_zome", "test_cap", None);
+        let dna_entry = dna.to_entry();
+        let commit_action = ActionWrapper::new(Action::Commit(dna_entry));
+
+        // Set up instance and dispatch action
+        let mut instance = Instance::new();
+        instance.start_action_loop(context);
```
thedavidmeister: keep an eye on https://github.com/holochain/holochain-rust/pull/225

if that lands before this, we should use `instance.process_action` to show a 1:1 relationship between the action and result, rather than an opaque loop

core/src/hash_table/sys_entry.rs
```diff
@@ -0,0 +1,194 @@
...
+                    assert_eq!(
+                        EntryType::from_str(&entry.entry_type()).unwrap(),
+                        EntryType::Dna
+                    );
+                    count += 1;
+                    true
+                }
+                _ => false,
+            });
+        assert_eq!(count, 1);
```
sphinxc0re: Sonce you aren't using the iterator afterwards, you clould just do a `.count()` on it.

core/src/hash_table/sys_entry.rs
```diff
@@ -0,0 +1,194 @@
...
+                    assert_eq!(
+                        EntryType::from_str(&entry.entry_type()).unwrap(),
+                        EntryType::AgentId
+                    );
+                    count += 1;
+                    true
+                }
+                _ => false,
+            });
+        assert_eq!(count, 1);
```
sphinxc0re: Same here

core/src/hash_table/sys_entry.rs
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    // Committing a DnaEntry to source chain should work only if first entry
+    #[test]
+    fn cannot_commit_dna() {
+        // FIXME
```
thedavidmeister: followup tx?

core/src/hash_table/sys_entry.rs
```diff
@@ -0,0 +1,194 @@
...
+    #[test]
+    fn can_commit_agent() {
+        // Create Context, Agent and Commit AgentIdEntry Action
+        let context = test_context("alex");
+        let agent_entry = context.agent.to_entry();
+        let commit_agent_action = ActionWrapper::new(Action::Commit(agent_entry));
+
+        // Set up instance and dispatch action
+        let mut instance = Instance::new();
+        instance.start_action_loop(context);
```
thedavidmeister: same as above re: processing actions

core/src/hash_table/sys_entry.rs
```diff
@@ -0,0 +1,194 @@
...
+                }
+                _ => false,
+            });
+        assert_eq!(count, 1);
+    }
+
+    // Committing an AgentIdEntry to source chain should work only as second entry
+    #[test]
+    fn cannot_commit_agent() {
+        // FIXME
```
thedavidmeister: followup tx?

core/src/hash_table/sys_entry.rs
```diff
@@ -0,0 +1,194 @@
...
+                    );
+                    count += 1;
+                    true
+                }
+                _ => false,
+            });
+        assert_eq!(count, 1);
+    }
+
+    // Committing an AgentIdEntry to source chain should work only as second entry
```
thedavidmeister: should this be `///` for doc comment?

core/src/instance.rs
```diff
@@ -438,7 +460,7 @@ pub mod tests {
 
         // @TODO don't use history length in tests
         // @see https://github.com/holochain/holochain-rust/issues/195
-        assert_eq!(instance.state().history.len(), 4);
+        assert_eq!(instance.state().history.len(), 5);
```
thedavidmeister: haha, this pattern is sooo bad >.<

can't wait to clean these length checks out

core/src/nucleus/mod.rs
```diff
@@ -183,72 +186,85 @@ fn reduce_ia(
...
+    state.status = NucleusStatus::Initializing;
+
+    // Set DNA
+    state.dna = Some(dna.clone());
+
+    // Create & launch thread
+    let genesis_action_channel = action_channel.clone();
+    let genesis_observer_channel = observer_channel.clone();
+    let dna_clone = dna.clone();
+    // let action_wrapper_clone = action_wrapper.clone();
```
thedavidmeister: clean up this commented code

core/src/nucleus/mod.rs
```diff
@@ -183,72 +186,85 @@ fn reduce_ia(
...
+
+    thread::spawn(move || {
+        // Send Commit Action for Genesis Entry
+        {
+            // Create Commit Action for Genesis Entry
+            let genesis_entry = dna_clone.to_entry();
```
thedavidmeister: to_entry() is cool 😍

core/src/nucleus/mod.rs
```diff
@@ -183,72 +186,85 @@ fn reduce_ia(
...
+                    &genesis_action_channel,
+                    &genesis_observer_channel,
+                    &zome.name(),
+                    &CallbackParams::Genesis,
+                )
+            })
+            .collect();
+
+        // pad out a single pass if there are no zome results
+        // @TODO is this really OK?
```
thedavidmeister: oops i forgot to put a followup for this it seems, could you throw one in for me?

core/src/hash_table/sys_entry.rs
ddd-mtl: doc comment for tests? hmmm okay...

core/src/instance.rs
ddd-mtl: same :)

core/src/nucleus/mod.rs
ddd-mtl: done

core/src/hash_table/sys_entry.rs
ddd-mtl: removed

core/src/hash_table/sys_entry.rs
ddd-mtl: done

agent/src/lib.rs
ddd-mtl: done

core/src/nucleus/mod.rs
ddd-mtl: I think there was already a card for that: https://waffle.io/holochain/org/cards/5b3644a6eaa711001ce607ce
Does that work for you? You can edit the card if you want.

core/src/hash_table/memory/mod.rs
```diff
@@ -10,6 +10,7 @@ use hash_table::{
     HashTable,
 };
 
+/// Struct Implemenating the HashTable Trait which stores the HashTable in memory
```
lucksus: typo

core/src/hash_table/sys_entry.rs
lucksus: #225 is merged now.

core/src/instance.rs
lucksus: Why is this needed anyway? Can it go since you've added the while-loop above that checks for the Commit action?

thedavidmeister:
@lucksus there's a lot of history length checking https://github.com/holochain/holochain-rust/issues/195

core/src/chain/mod.rs
```diff
@@ -34,14 +34,16 @@ impl<T: HashTable> Iterator for ChainIterator<T> {
     fn next(&mut self) -> Option<Pair> {
         let previous = self.current.take();
         self.current = previous.as_ref()
-                        .and_then(|p| p.header().next())
+                        .and_then(|p| p.header().prev())
```
thedavidmeister: this does bother me, putting `prev` in a `next` call is confusing :/

maybe `link` or `header_link`, which avoids directionality

core/src/chain/mod.rs
```diff
@@ -149,7 +151,7 @@ impl<T: HashTable> Chain<T> {
     }
 
     /// get a Pair by Pair/Header key from the HashTable if it exists
-    pub fn get(&self, k: &str) -> Result<Option<Pair>, HolochainError> {
+    pub fn get_pair(&self, k: &str) -> Result<Option<Pair>, HolochainError> {
```
thedavidmeister: could change this to just `pair()` as per our discussions on #243

ddd-mtl: yes!

core/src/chain/mod.rs
ddd-mtl: Changed to `link`.
I like it because it fits the Chain metaphore.
I would recommend we stick to using `link` as a noun and not a verb in all the API, otherwise it might cause confusion.

zippy: This is fine, the only issue here is that it's a slightly different nomenclature to what we used in the go version which was: 
```
	HeaderLink Hash // link to previous header
	EntryLink  Hash // link to entry
	TypeLink   Hash // link to header of previous header of this type
```
But I think this is fine, and given the context won't cause confusion.

thedavidmeister: @zippy i expect we'll still use that nomenclature, but it would land in the serde config rather than rust methods right?

ddd-mtl: @zippy I wasn't aware of that naming in Go, and I would have indeed proposed that instead. However I'm wondering how much do we want to stick to the Go names?
I think Go names are a starting point, and than we make use of the opportunity to improve them if we see a clear possible improvement based on our experience with the Go code.
Since I don't actually have experience with the Go code, I'm looking at it with new/fresh eyes.

thedavidmeister: @ddd-mtl @zippy we did actually discuss this exact thing in a meeting a while back and agreed that breaking changes are OK here

that said, i agree it makes sense to use what we already have as a starting point

there's this old thing from june https://github.com/holochain/holochain-rust/issues/75

Issue #134, opened 2018-07-16
decide on format of arguments in DNA


zippy:
we could use either list of arguments by name/type pairs

or an object where the key is the name, and the type is value,

e.g. 

```json
          {
              "name": "newPost",
              "signature" :
              {
                  "inputs" : [
                      {
                          "name": "post",
                          "type": "string"
                      }
                  ],
                  "outputs" : [
                      {
                          "name": "hash",
                          "type": "string"
                      }
                  ]
              },
```

or 

```json
{
  "name": "newPost",
  "params" : {
    "post": "string",
  },
  "returns" : {
    "hash": "hash",
  },
}
```

zippy:
What ever way we go, the way we do it for signals should match.

neonphog:
### arrays of type / name pairs

pro: easier to deserialized into strongly typed structs

### compact `"name": "type"`

pro: much more human readable / writable

---

I guess I'd lean toward the more machine-oriented first option, as we're hoping to have fancy tools for generating / working with DNA in any case.

zippy:
Ok, I'm going to recommend a middle ground where we remove the "signature" object (because it's unnecessary) but use the typed values as per @neonphog's leaning and @lucksus's from our call, hence:

```
          "functions": [
            {
              "name": "newPost",
              "inputs" : [
                  {
                      "name": "post",
                      "type": "string"
                  }
              ],
              "outputs" : [
                  {
                      "name": "hash",
                      "type": "string"
                  }
              ],
            },
            // ...
          ],
          "signals": [
            {
              "name": "Post",
              "description": "signal emmited when a post is committed",
              "config-params": null,
              "arguments": {
                  "name": "hash",
                  "type": "hash"
              },
            // ...
          ],
```

zippy:
To complete this requires just minor changes to the `dna` crate

lucksus:
Cool, so this is decided, and this ticket is actually done, right?

zippy:
Yes, but I don't want to get the fact that this needs to be updated in the `dna` crate.  @lucksus It seems like you would need to update it there to get your example app crate actually working correctly right?  If not then there's dependency problem.

PR #253, opened 2018-08-27
DNA improvements

### LINKED ISSUE #221: Implement improvements on DNA spec
There are some changes to the DNA structure that we should implement:

- [x] Switching from arrays to objects for zomes and capabilities, with the (zome/cap) name being the property name within the encompassing object (the way hcdev currently creates JSON from a folder tree)
- [x] Getting rid of "signature" in function definitions as decided here: https://github.com/holochain/holochain-rust/issues/134
- [x] Renaming "fn_declarations" to "functions"
- [ ] Adding links ("links_to" and "linked_from") as specified in https://github.com/holochain/app-spec-rust/pull/1

Not sure if while at it we would also do:
- [ ] Adding signals as specified in https://github.com/holochain/holochain-rust/blob/develop/doc/architecture/decisions/0013-signals-listeners-model-and-api.md
..since the ADR is not accepted yet..

The main intention behind this ticket is to get the spec app to be at least parsable by Holochain.

lucksus:
closes #221 dna improvements

# Change DNA spec to match app spec and hcdev

- [x] "fn_declarations" -> "functions"
- [x] get rid of "signature" (#134)
- [x] change arrays to maps for zomes, capabilites and entry_types and extract name from inner struct to mapping (matching with https://github.com/holochain/holochain-cmd)
- [x] add "linked_from" as used in the app spec (https://github.com/holochain/app-spec-rust/pull/1/files#diff-4f4d6fab1698c4be58c54cb3c2e563aeR10)


lucksus:
A HoloSqape build with this branch of holochain-rust can read the DNA in the app-spec-rust dist directory: https://github.com/holochain/app-spec-rust/pull/1/files#diff-c30b6726492851bac2ff0ab8ea3fb01d
screen shot 2018-08-28 at 00 14 24

That file is almost purely the output of `hcdev package --strip-meta` with manually added `"code": { .. }` around WASM.
@sphinxc0re

lucksus:
@zippy, @sphinxc0re has added such meta fields to the output produced by `hcdev`, which I had to avoid with `--strip-meta`.. but yeah, maybe we want to include that in the DNA spec.

core_api/src/lib.rs
```diff
@@ -100,7 +100,11 @@ impl Holochain {
             }
         });
 
-        match receiver.recv_timeout(Duration::from_millis(1000)) {
+        // TODO: what is the right timeout?
```
thedavidmeister: haha "the right timeout"

dna/src/lib.rs
```diff
@@ -373,14 +363,14 @@ pub mod tests {
     fn parse_with_defaults_zome() {
         let dna = Dna::new_from_json(
             r#"{
-                "zomes": [
-                    {}
-                ]
+                "zomes": {
+                    "0": {}
```
thedavidmeister: i won't reject the PR for this, but it is confusing to have strings that look like indexes

PR #268, opened 2018-08-29
Zome call without dispatch

lucksus:
The awesome new `zome_functions!` macro that @sphinxc0re built for the HDK can plug the de-/serialize code directly into the function itself - no need for a _dispatch wrapper that delegates to the real funciton. Also, Rust macros can't compose function names, so we can't create a *_dispatch function automatically.

-> just don't call fn-name_dispatch, but fn-name instead.


Connoropolous:
I changed the reference to the macro in the PR description to match what it's called in the https://github.com/holochain/hdk-rust/pull/6 PR.

PR #246, opened 2018-08-26
231 file table

thedavidmeister:
fixes #231 
fixes #193 
fixes #168 
fixes #25 

## overview

adds the bare minimum needed to get pairs onto the local file system.

nothing is done in memory or otherwise memoized. 
upserts serialize and write to a file.
lookups read and parse JSON from a file.
the file name is the lookup key.
scans (e.g. all metas for a pair) read and parse every file of the same type (e.g. all metas).

this won't scale or win any benchmarks but it does get us persistent data.

this is not ready for prime time just yet. concurrent file system read/writes are not supported. this needs to sit behind an actor or equivalent #135 e.g. riker guarantees that each actor handles only one message at a time http://riker.rs/actors/#sending-messages.

a simple performance improvement would be creating a hybrid memory/file table that treats the memtable as a cache. scalability (especially for meta data) could be improved by implementing something with indexes, e.g. sqlite.

`FileTable` passes the same test suite as `MemTable` + some extras for internals.

## changes

- add the `walkdir` crate
- add the `tempfile` crate
- rename `Protocol::Meta` to `Protocol::PairMeta`
- rename `Protocol::PairMeta` to `Protocol::MetasForPair`
- add the `json::ToJson` trait for things that can be converted to json (not necessarily via. serde)
- moved everything in core with a `to_json()` method under the `ToJson` trait
- renamed `HcApiReturnCode::ErrorSerdeJson` to `ErrorJson`
- add the `json::FromJson` trait for things that can be parsed from json (not necessarily serde)
- moved most things in core with a `from_json()` method under the `FromJson` trait
- add `key::Key` trait for things that have a lookup key
- moved everything with a `key()` method under the `Key` trait
- rename `get` to `pair` on `HashTable`
- rename `commit` to `commit_pair` on `HashTable`
- implement `From` for a few other error types for `HolochainError`
- add `IoError` to `HolochainError` enum
- add `SerializationError` to `HolochainError` enum
- add `hash_table::file::FileTable` implementing `HashTable`
- factored all `MemTable` tests into a standard suite in `hash_table::test_util` and used for both `MemTable` and `FileTable` and hash table actors
- tests for `FileTable` internals
- add default no-op implementations for `setup` and `teardown` for `HashTable`
- add default implementations for `modify` and `retract` for `HashTable`
- rename `assert_meta` to `assert_pair_meta` for `HashTable`
- rename `get_meta` to `pair_meta` for `HashTable`
[...]

PR #269, opened 2018-08-29
WIP: 188-linking

ddd-mtl:
# Linking

## Intent
 Current implementation intention
### Adding a Link
Have a `Action::LinkAppEntries` which does the following:
1. Validate the Link
  a. Get Linked Entry Types from Dna
  b. Check with Dna if they are linkable
  c. Run link validation wasm code
2. Create and Commit a LinkEntry on source chain
 a. Do I send a `Action::Commit` or should I do it directly??
3. Add LinkListEntry to HashTable
 a. Create&Commit or Update in HashTable a LinkListEntry with `key = base-entry-hash + tag`
 b. Create/Update metadata for base entry: 
    - attribute-name = `"link:" + base + ":" + tag`
    - attribute-value = hash of LinkListEntry
4. Return hash of LinkEntry?

### Getting Links
Have a `Action::GetLinks` which does the following:
1. Looks up for metadata attribute named `"link:" + base + ":" + tag` on entry in HashTable
2. Gets the LinkListEntry in HashTable
3. Return List of targets from each Link in the LinkListEntry

## Details

New actions: `LinkAppEntries` and `GetLinks`
Renamed some other Actions to avoid confusion.

# Refactored Pair out of HashTable
Had to rewrite HashTable to not use Pairs, according to https://github.com/holochain/holochain-rust/issues/274

Header now has ToEntry trait, which converts it into a header system entry type.
When commiting a Pair, it converts the Header into a header system entry type and adds it to the HashTable.
Header.hash() is now the value of the header.to_entry().key()

Moved `Header::new()` to `chain::create_next_header()`
Moved `Pair::new()` to `chain::create_next_pair()`

Renamed PairMeta to Meta.
Renamed file pair_meta.rs to meta.rs
Hash of Meta is now just entry_hash and attribute_name.
Not value and source otherwise search is not possible.

Renamed `push` to `commit` when appropriate.
Renamed `commit` to `put` for HashTable.

# HashString type
Started adding a HashString to make use of Rust's strong typing.

# Renames

`ZomeAPIFunction` to `ZomeApiFunction`
`CRUDStatus` to `CrudStatus`

## AppEntry
`Entry` is a generic entry, i.e any thing can be an entry stored in the HashTable.
`AppEntry` is the name for non-system entries, i.e. the entries defined in the DNA.

<br/>


lucksus:
Steps look basically like the right steps.
I believe you really need chained/composed actions! I guess we should sync on this, @ddd-mtl...

core/src/hash_table/header.rs
```diff
@@ -384,4 +408,19 @@ mod tests {
...
+        let t = "foo";
+        let c = "bar";
+        let e = Entry::new(t, c);
+        let h = chain.create_next_header(&e);
+
+        let h_entry = h.to_entry();
+        let h_trip = Header::new_from_entry(&h_entry);
+
+        assert_eq!(h, h_trip);
+    }
```
lucksus: So we have lots of tests that actually test the chain and that are located in hash_table?

core/src/hash_table/mod.rs
```diff
@@ -25,26 +32,43 @@ pub trait HashTable: Send + Sync + Clone + 'static {
...
-    /// set the status of a Pair to DELETED
-    fn retract(&mut self, keys: &Keys, pair: &Pair) -> Result<(), HolochainError>;
-
-    // meta
+//    fn commit(&mut self, pair: &Pair) -> Result<(), HolochainError>;
+
+    fn put(&mut self, entry: &Entry) -> Result<(), HolochainError>;
+
+    /// lookup an Entry from the HashTable
+    fn entry(&self, key: &str) -> Result<Option<Entry>, HolochainError>;
```
lucksus: Should this be ˋgetˋ or ˋget_entryˋ?

core/src/hash_table/pair.rs
```diff
@@ -169,7 +158,7 @@ pub mod tests {
         let t = "fooType";
 
         let e1 = Entry::new(t, "bar");
-        let p1 = Pair::new(&chain, &e1);
+        let p1 = chain.create_next_pair(&e1);
 
         assert!(p1.validate());
     }
```
lucksus: Same as with header. This should not be here, but in chain.

core/src/nucleus/ribosome/api/commit_app_entry.rs
```diff
@@ -7,25 +7,26 @@ use nucleus::ribosome::{
...
 }
 
-/// HcApiFuncIndex::COMMIT function code
+/// ZomeApiFunction::CommitAppEntry function code
 /// args: [0] encoded MemoryAllocation as u32
-/// expected complex argument: r#"{"entry_type_name":"post","entry_content":"hello"}"#
+/// Expected complex argument: CommitArgs
 /// Returns an HcApiReturnCode as I32
-pub fn invoke_commit_entry(
+pub fn invoke_commit_app_entry(
```
lucksus: What is an „app entry“?

core/src/agent/state.rs
```diff
@@ -65,54 +81,145 @@ impl ActionResponse {
...
+            },
         }
     }
 }
 
+/// Do the LinkAppEntries Action against an agent state:
+/// 1. Validate Link
+/// 2. Commit LinkEntry
+/// 3. Add Link metadata in HashTable
+fn reduce_lap(
```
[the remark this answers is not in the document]
ddd-mtl: Yeah, that experiment was not conclusive.

core/src/hash_table/actor.rs
```diff
@@ -169,13 +215,13 @@ pub mod tests {
     fn round_trip() {
         let mut table_actor = test_table_actor();
 
-        assert_eq!(table_actor.pair(&test_hash()).unwrap(), None,);
+        assert_eq!(table_actor.entry(&test_hash()).unwrap(), None);
 
-        table_actor.commit(&test_pair()).unwrap();
+        table_actor.put(&test_pair().entry()).unwrap();
```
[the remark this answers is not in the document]
ddd-mtl: Fixed

core/src/hash_table/header.rs
```diff
@@ -1,7 +1,11 @@
-use chain::{Chain, SourceChain};
```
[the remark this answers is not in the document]
ddd-mtl: Fixed

core/src/hash_table/header.rs
ddd-mtl: header.rs moved to chain mod

core/src/hash_table/pair.rs
ddd-mtl: Fixed

core/src/chain/mod.rs
```diff
@@ -131,82 +204,109 @@ pub trait SourceChain {
...
+                top_pair, prev_pair,
             )));
         }
 
-        self.table.commit(&pair.clone())?;
+        // 2. pushing
+        // 3. putting
+        let header_entry = &pair.clone().header().to_entry();
+        // println!("Chain.commit_pair() header_entry = {:?}", header_entry);
+        self.table_actor.put(header_entry)?;
```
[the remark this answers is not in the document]
ddd-mtl: For iterating through the chain, header by header. See `ChainIterator.next()`
How am I suppose to do that otherwise?

core/src/agent/state.rs
```diff
@@ -46,15 +57,20 @@ impl AgentState {
...
 
 #[derive(Clone, Debug, PartialEq)]
 /// the agent's response to an action
 /// stored alongside the action in AgentState::actions to provide a state history that observers
 /// poll and retrieve
 pub enum ActionResponse {
-    Commit(Result<Pair, HolochainError>),
-    Get(Option<Pair>),
+    CommitEntry(Result<Pair, HolochainError>),
+    GetEntry(Option<Pair>),
```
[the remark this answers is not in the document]
ddd-mtl: Fixed: Get and Commit returns Entry now.
GetEntry() is here because Agent is the Chain. Chain doesn't have its own state.

core/src/chain/mod.rs
```diff
@@ -131,82 +204,109 @@ pub trait SourceChain {
...
+    fn top_pair_of_type(&self, t: &str) -> Option<Pair> {
         self.iter().find(|p| p.header().entry_type() == t)
     }
 
-    fn push_pair(&mut self, pair: &Pair) -> Result<Pair, HolochainError> {
+    /// Whole process of authoring an entry.
+    /// 1. `validation` of the new entry using the ribosome and validation WASM code
+    /// 2. `pushing` the new entry onto the source chain, if valid
+    /// 3. `putting` the entry into the (distributed) hash table, if defined as public
+    fn commit_pair(&mut self, pair: &Pair) -> Result<Pair, HolochainError> {
```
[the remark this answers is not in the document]
ddd-mtl: For a follow up PR?

core/src/action.rs
```diff
@@ -65,15 +68,18 @@ impl Hash for ActionWrapper {
 pub enum Action {
     /// entry to Commit
     /// MUST already have passed all callback checks
-    Commit(Entry),
-    /// hash to Get
-    Get(String),
-
+    CommitEntry(Entry),
+    /// hash of Entry to get
+    GetEntry(HashString),
```
thedavidmeister: are we doing `Entry` and dropping the `Get` in our naming conventions?

core/src/action.rs
```diff
@@ -65,15 +68,18 @@ impl Hash for ActionWrapper {
...
+    /// entry to Commit
+    /// MUST already have passed all callback checks
+    LinkAppEntries(Link),
+    /// hash to Entry to get links from
+    GetLinks(GetLinksArgs),
```
thedavidmeister: ditto for `Links`

core/src/action.rs
```diff
@@ -65,15 +68,18 @@ impl Hash for ActionWrapper {
...
-    Commit(Entry),
-    /// hash to Get
-    Get(String),
-
+    CommitEntry(Entry),
+    /// hash of Entry to get
+    GetEntry(HashString),
+    /// entry to Commit
+    /// MUST already have passed all callback checks
+    LinkAppEntries(Link),
```
thedavidmeister: hmmm `link` is a noun and a verb :/

core/src/actor.rs
```diff
@@ -30,37 +37,51 @@ pub enum Protocol {
     /// HashTable::modify()
     Modify {
         keys: Keys,
-        old_pair: Pair,
-        new_pair: Pair,
+        old_entry: Entry,
```
thedavidmeister: doesn't meta _need_ to be for a pair? entries aren't even unique across types...

e.g. can't i retract system entries like genesis if i send the retract meta for any other entry with the same content as gensis?

core/src/chain/mod.rs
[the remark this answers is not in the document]
thedavidmeister: @lucksus i disagree that you should put headers in the chain

that's so out of scope here...

core/src/chain/mod.rs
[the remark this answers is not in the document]
thedavidmeister: @lucksus the header has been in the HT literally always

core/src/chain/mod.rs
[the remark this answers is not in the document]
lucksus: The reason I asked Damien to look at the refactor ticket before doing the linking is that he would produce more code that we would need to change if we don't do the refactor first.

core/src/chain/mod.rs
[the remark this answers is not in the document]
lucksus: @ddd-mtl: Why does the HT iterate through the chain??

@thedavidmeister: Just because it was that way before (which I pointed out as not fitting the mental model already) doesn't make it right, right?. ;) 

I don't see why the HT should deal with headers from a logical stand point and I would like to understand if there is an implementation reason for it. I am concerned that we produce more and more code that reproduces this wrong pattern. Like this line which gets added in here. Although HT deals with entries now (in put and get), it still needs to have a header entry putted to it explicitly? Why would you do that? My feeling is that we are building something here that is not really the thing we need..

core/src/action.rs
lucksus: Why dropping the `Get`? Isn't that almost a stronger characteristic of this action than `Entry`?

ddd-mtl: For this part I was sticking to the API in HDK rust:  `hdk::get_entry()`

core/src/action.rs
ddd-mtl: Was sticking to hdk: `hdk::get_links()` but maybe we should change to `hdk::links` in the HDK also?

core/src/action.rs
ddd-mtl: Yes, but its not just `Link` is has more context so its good enough for me.
You would prefer something like `CreateLinkAppEntries` ?

core/src/actor.rs
ddd-mtl: For that, my thought was that certain operations, like retract, are not aloud on system entry types.

core/src/chain/mod.rs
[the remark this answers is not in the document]
ddd-mtl: I took some more time read through @lucksus's initial comment.
I like the non orchestrated version. 
I find it more understandable if we rename the actions like this:
`Commit` to `ValidateForCommit`
`VailidationResult` to `Commit`.

So the story goes:
- `invoke_commit()` issues a `ValidateForCommit(Entry)` action
- Nucleus reduces it and on success issues a `Commit(Entry)` action
- Chain/Agent reduces it and on success issues a `Put(Entry)` action
- DHT reduces it
- `invoke_commit()` listens for a succesful Put of that entry in the DHT.

The story for receiving an Entry from the network goes:
- Network issues a `ValidateForPut(Entry)` action
- Nucleus reduces it and on success issues a `Put(Entry)` action
- DHT reduces it

I believe the main misunderstanding is that for @lucksus the HashTable is the DHT.
But thats not the case in the current code architecture. Currently the HashTable is the Chain.
Meaning, it is the actual data structure holding the Chain's data. The Chain is just a way to browse the HashTable in a certain sequence.
But since we have a HashTable for the Chain, we could also use it to store the DHT.
Eric was okay with that idea. I think its better to separate and have a second HashTable for the DHT, because you would want to store/persist them differently.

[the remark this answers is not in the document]
thedavidmeister: @ddd-mtl "because you would want to store/persist them differently" this is the bit i want to get deeper into because it's not obvious to me why you'd want to change the storage mechanism based on the content of the data being stored?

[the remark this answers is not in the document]
ddd-mtl: @thedavidmeister Source chain data and DHT data are of different nature. So they will most likely have different functional requirements and constraints. I have Holo and also data security in mind for now. Source chain data is more critical for example. You would want to backup that in a more robust way. This will most likely translate into different technical requirements. Especially over time. So I'm thinking ahead of this a bit and thinking its best to separate them right now.

core/src/action.rs
```diff
@@ -83,7 +89,7 @@ pub enum Action {
 
     /// ???
     // @TODO how does this relate to validating a commit?
-    ValidateEntry(EntrySubmission),
+    ValidateAppEntry(EntrySubmission),
```
sphinxc0re: Is there something else than an AppEntry?

core/src/chain/header.rs
```diff
@@ -33,34 +38,29 @@ impl PartialEq for Header {
...
+            entry_type: entry_type.to_string(),
+            timestamp: timestamp.to_string(),
+            link: link,
+            entry_hash: entry_hash.to_string(),
+            entry_signature: entry_signature.to_string(),
+            link_same_type: link_same_type,
         }
     }
 
+    pub fn new_from_json(header_str: &str) -> serde_json::Result<Self> {
```
sphinxc0re: Better name this `from_json_str`. The `new_` is implicit when calling `Header::from_json_str("..")`.

core/src/chain/pair.rs
```diff
@@ -12,37 +14,27 @@ pub struct Pair {
...
-    /// # Panics
-    ///
-    /// Panics if entry is somehow invalid
-    ///
-    /// @see chain::entry::Entry
-    /// @see chain::header::Header
-    pub fn new(chain: &Chain, entry: &Entry) -> Pair {
-        let header = Header::new(chain, entry);
+    /// Reconstruct Pair from Header stored in a HashTable
+    pub fn new_from_header(table: &ActorRef<Protocol>, header: &Header) -> Option<Self> {
```
sphinxc0re: Same as above. I think `from_header` is enough. This is a very common naming scheme for constructors.

core/src/action.rs
ddd-mtl: I'd be happy to know if you have a better name for non-system entry types.
I didn't have a clear preference between: Data, App, User, Custom, Zome

core/src/chain/header.rs
ddd-mtl: I was following pattern in Dna but I'll just change it everywhere then.

core/src/chain/pair.rs
ddd-mtl: fixed

core/src/action.rs
lucksus: Why not call them `Entry`?
And then we also have SystemEntries..

core/src/chain/mod.rs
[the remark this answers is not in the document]
lucksus: Well yes, one thing we haven't mentioned in this discussion so far is that Holo wants to hold a users source chain on the client in the browser's local storage, but not the DHT.

Though I guess that should be possible by having different implementations of the trait, right?

lucksus:
I've approved but as we discussed today, https://github.com/holochain/holochain-rust/pull/246 should be merge first.

core/src/action.rs
ddd-mtl: I'll try explaining with pseudo Math:
`Entry = SystemEntry u AppEntry`
So if you call AppEntry 'Entry' you get:
`Entry = SystemEntry u Entry`
Which is confusing.

I'm going under the assumption that some operations/actions are done on AppEntries only, like Validation.
And some operations are done on any Entry, like Put in HashTable.
So I would like the name of the operations to reflect that.

lucksus: Hm, SystemEntries have to be validated as well, no? @zippy
Just differently than AppEntries.

I'd say in pseudo Math:
`SystemEntry c Entry`

core/src/action.rs
```diff
@@ -66,14 +69,16 @@ pub enum Action {
     /// entry to Commit
     /// MUST already have passed all callback checks
     Commit(Entry),
-    /// hash to Get
-    Get(String),
-
+    /// hash of Entry to get
+    GetEntry(HashString),
```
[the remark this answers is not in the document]
thedavidmeister: discussed this with @lucksus and @zippy and we're happy with `GetEntry`

core/src/nucleus/ribosome/api/link_entries.rs
```diff
@@ -0,0 +1,75 @@
...
+    let action_wrapper = ActionWrapper::new(Action::AddLink(input));
+    // Send Action and block for result
+    let (sender, receiver) = channel();
+    ::instance::dispatch_action_with_observer(
+        &runtime.action_channel,
+        &runtime.observer_channel,
+        action_wrapper.clone(),
+        move |state: &::state::State| {
+            let mut actions_copy = state.agent().actions();
+            match actions_copy.remove(&action_wrapper) {
```
lucksus: Why is this remove()? We are working on a copy anyway, so this is an awkward way to test if action_wrapper was in...

core/src/hash_table/mod.rs
```diff
@@ -34,54 +38,170 @@ pub trait HashTable: Send + Sync + Clone + 'static {
...
-    /// lookup a PairMeta from the HashTable by PairMeta key
-    fn pair_meta(&mut self, key: &str) -> Result<Option<PairMeta>, HolochainError>;
-    /// lookup all PairMeta for a given Pair
-    fn metas_for_pair(&mut self, pair: &Pair) -> Result<Vec<PairMeta>, HolochainError>;
+                // Create Meta
+                new_meta = EntryMeta::new(
+                    &keys_fixme.node_id(),
+                    &base_entry.key(),
+                    &link.to_attribute_name(),
+                    &new_entry.key(),
```
lucksus: So, first I was wondering why you are creating keys here. I see that EntryMeta needs them as source, which makes sense. But it does not make sense at all to just create some keys here for that.

When I glanced over this code some weeks back and approved, I was still under the impression that "hash_table" is meant to become the DHT module.

Now, taking into account that this hash_table is a building block for building chains and DHT shards.. (a CAS), I am very skeptical about putting this code in here.. and the fact that you have to conjure up a key for the source of this entry I take as confirmation for that doubt.

Which makes want to check the reducer for all of this. But there is none?!

core/src/nucleus/ribosome/api/link_entries.rs
```diff
@@ -0,0 +1,75 @@
...
+        Ok(entry_input) => entry_input,
+        // Exit on error
+        Err(_) => {
+            // Return Error code in i32 format
+            return Ok(Some(RuntimeValue::I32(HcApiReturnCode::ErrorJson as i32)));
+        }
+    };
+
+    // Create AddLink Action
+    let action_wrapper = ActionWrapper::new(Action::AddLink(input));
```
lucksus: Links have to be validated first before added!

core/src/nucleus/ribosome/api/link_entries.rs
ddd-mtl: I copied this form `get.rs`

lucksus: Yeah, I saw this before when I've replaced this kind of observer code in the futures action refactor.. This PR gave me a place to complain about it ;)

core/src/nucleus/ribosome/api/get_entry.rs
```diff
@@ -9,17 +10,21 @@ use std::sync::mpsc::channel;
...
 pub fn invoke_get_entry(
     runtime: &mut Runtime,
     args: &RuntimeArgs,
 ) -> Result<Option<RuntimeValue>, Trap> {
     // deserialize args
     let args_str = runtime_args_to_utf8(&runtime, &args);
-    let res_entry: Result<GetArgs, _> = serde_json::from_str(&args_str);
+    let res_entry: Result<GetAppEntryArgs, _> = serde_json::from_str(&args_str);
     // Exit on error
     if res_entry.is_err() {
```
AshantiMutinta: I saw this earlier in scrum, I was just wondering if there was any reason we are using this pattern of error handling instead of match?

ddd-mtl: I find code a lot clearer and easier to debug if you exit early when the pre-conditions are not met.
The problem with doing a match early in a function is that the rest of the function code will be within the match arm and thus indented. If you keep doing that you end up with 'pyramids' of indented code, which gets really hard to read.

ddd-mtl: We could do a `match` that does not nest code by doing it in the `let`, but I find this a bit more complicated to understand when reading it. Matter of taste I guess.

core/src/agent/actions/add_link.rs
```diff
@@ -0,0 +1,158 @@
...
+    // Add Link to HashTable (adds to the LinkListEntry Meta)
+    let res = state.chain().table().add_link(link);
+    if res.is_err() {
+        response = Err(res.err().unwrap());
+    }
+
+    // Insert reponse in state
+    state.actions().insert(
+        action_wrapper.clone(),
+        ActionResponse::LinkEntries(response),
```
AshantiMutinta: Do you think it would be benficial to use an and_then function here? I think it has the potential to eliminate some code

core/src/agent/state.rs
```diff
@@ -77,7 +83,10 @@ impl ToJson for ActionResponse {
                 None => Ok("".to_string()),
             },
             ActionResponse::GetLinks(result) => match result {
-                Ok(hash_list) => Ok(json!(hash_list).to_string()),
+                Ok(hash_list) => Ok(json!(hash_list)
+                    .as_str()
+                    .expect("should jsonify")
```
AshantiMutinta: Will this ever panic?

core/src/agent/actions/add_link.rs
AshantiMutinta: Awesome! Looks good to me. Will approve

Connoropolous:
@ddd-mtl can we close this one?

ddd-mtl:
yep

PR #279, opened 2018-09-03
Clippy lints and refactorings

sphinxc0re:


wasm_utils/src/lib.rs
```diff
@@ -172,25 +172,23 @@ pub fn deserialize<'s, T: Deserialize<'s>>(ptr_data: *mut c_char) -> T {
 pub fn deserialize_allocation<'s, T: Deserialize<'s>>(encoded_allocation: u32) -> T {
     let allocation = SinglePageAllocation::new(encoded_allocation);
     let allocation = allocation.expect("received error instead of valid encoded allocation");
-    return deserialize(allocation.offset as *mut c_char);
+    unsafe { deserialize(allocation.offset as *mut c_char) }
```
ddd-mtl: For the third time, I would like to keep return statements when it makes the code more readable, like in this case.

wasm_utils/src/lib.rs
```diff
@@ -212,7 +210,7 @@ pub fn serialize_into_encoded_allocation<T: Serialize>(
     internal: T,
 ) -> i32 {
     let allocation_of_output = serialize(stack, internal);
-    return allocation_of_output.encode() as i32;
+    allocation_of_output.encode() as i32
```
ddd-mtl: Same.

wasm_utils/src/lib.rs
```diff
@@ -45,7 +45,7 @@ pub fn u32_high_bits(i: u32) -> u16 {
 
 /// returns the u16 low bits from a u32
 pub fn u32_low_bits(i: u32) -> u16 {
-    (i as u16 % std::u16::MAX)
```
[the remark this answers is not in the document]
sphinxc0re: It's just that `std::u16::MAX` is a constant at module level whether `u16::max_value()` is at type level. It makes more sense semantically.

wasm_utils/src/lib.rs
```diff
@@ -88,7 +86,9 @@ impl SinglePageAllocation {
 
         // should never happen
         // we don't panic because this needs to work with wasm, which doesn't support panic
-        if (allocation.offset as u32 + allocation.length as u32) > std::u16::MAX as u32 {
+        if (u32::from(allocation.offset) + u32::from(allocation.length))
+            > u32::from(u16::max_value())
```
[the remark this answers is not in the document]
sphinxc0re: It's not about readability. This method gives us the safety that the conversion is not going to get silently lossy someday.

wasm_utils/src/lib.rs
sphinxc0re: Since Rust is an expression based language, it is generally frowned upon to use `return` statements. "The Rust programming language" even says so.

sphinxc0re: The "official" notion is that `return` statements should only be used for *early* returns. It took a while for me to get used to this, but eventually this enables you to write more streamlined code.

wasm_utils/src/lib.rs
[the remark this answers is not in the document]
sphinxc0re: I think it's more readable *now*. See, the `type as something_else` syntax drags code apart, whereas `something_else::from(type)` pulls it together. For me, it is easier to see which parts belong together

thedavidmeister:
@ddd-mtl @sphinxc0re i've left a bit of a rant on the trello card

i feel we're lost in the weeds here and need to get the basic tooling/processes back on track before deep diving individual lints

thedavidmeister:
i'm closing this. i think the work is good but unfortunately not grounded in a clear strategy for how we want to manage clippy long term.

let's come at this fresh next week when we decide what to do at a high level.

we can either dust off the branch and put a PR back up, or something else.

PR #280, opened 2018-09-03
key managment ADR

### LINKED ISSUE #254: Create ADR for DPKI and Key Management
- [ ] update drafts and to repo (https://hackmd.io/VIVvRD9gQ5KnDy_C-y6leA)
- [ ] connect with Secure Scuttlebut Folks

Here are some of the existing docs & draft ADRs on DPKI:

 - Usage Story: https://hackmd.io/87H6J0pZSaG8MgbpnbjlIA
 - Long ADR draft: https://hackmd.io/VIVvRD9gQ5KnDy_C-y6leA
 - Required Holochain functions: https://hackmd.io/mlZ4AChdStCGOVixO4kR3Q

zippy:
closes #254 

This pull request adds an ADR for holochain_core_api functions that implement how we create and derive keys for multi-device identification, and key revocation.

Please note links to lots of contextualizing discussion.


doc/architecture/decisions/0015-key-management.md
```diff
@@ -0,0 +1,222 @@
...
+    fn seed_to_mnemonic(seed: &[u8; 32]) -> String;
+    fn mnemonic_to_seed(mnemonic: &str) -> [u8; 32];
+}
+
+/// struct for hd seed derivation handling
+impl DpkiHd {
+    pub fn new_random() -> Self;
+    pub fn new_from_primary_seed(primary_seed: &[u8; 32]) -> Self;
+    pub fn get_primary_seed(&self) -> [u8; 32];
+    pub fn get_revoke_seed(&self) -> [u8; 32];
```
lucksus: I get confused by these sub-keys also being called `seed`. Maybe I'm misunderstanding - we are talking about keys in all instances, right? The root key that is used to create the sub-keys (in a deterministic manner) is called the *seed* because it is what other keys are derived from via the derivation paths listed below. But then the keys that are created this way are keys, **not seeds.** Right?

doc/architecture/decisions/0015-key-management.md
```diff
@@ -0,0 +1,222 @@
...
+
+// publish to the dpki dht
+let app_pub_data = DpkiApp::new_defaults(hd);
+publish_to_dpki_dht(app_pub_data);
+
+// initialize a device structure with device seed number 1
+let device = DpkiDevice::new_from_seed(hd.get_device_seed(1));
+// (optional) display device seed to user for backup
+display_to_user(Dpki::seed_to_mnemonic(hd.get_device_seed(1)));
+// blink the device pin on the holoport circle
```
lucksus: ?

doc/architecture/decisions/0015-key-management.md
```diff
@@ -0,0 +1,222 @@
...
+display_to_user(device.get_device_pin());
+
+// reboot will require passphrase re-entry
+save_to_disk(device.get_saved(prompt_user_for_passphrase()));
+```
+
+##### 5 - new holo identity
```
lucksus: I'm not clear what a *holo identity* is.
Seems we are assuming a (centralized?) Holo account, managed by DpkiHolo and connected to our Dpki identity??

doc/architecture/decisions/0015-key-management.md
zippy: I think David's idea here is that that would be a way to display a pin by blinking it on the Holo-port circle, but I don't think that's serious...

doc/architecture/decisions/0015-key-management.md
zippy: Seeds are not keys.  Seeds are 32bit numbers from which you can create an actual private key, using what ever algorithm we use for keys be that signing keys (network/or authoring) or encryption keys.  You can't use a seed as an actual key, i.e. it's not a value you can pass into the code library as a key.  Its a value you use to create the key.  But in some sense you can think of it as a key because it deterministically generates one.  So you get hierarchical deterministic keys because of the hierarchy of seeds.

lucksus: Right. So why are the names of this functions get_*_seed(). What they return is a derived thing, which is a key. No?

doc/architecture/decisions/0015-key-management.md
artbrock: It is serious. Following the principle that private keys (and seeds for key pairs) should never be sent over the wire/network, how do you propose to "display" keys from a HoloPort which people do not have a monitor plugged into? 

We are targeting HoloPorts as turnkey plug/play devices for non-techies. Non-techies often do not have spare keyboards and monitors laying around their house anymore. Most people primarily use consumer grade devices like phones/tablets/laptops now. If we want to give people a way to record critical information to regenerate their system, we can use the Holo LED to flash numbers to them, changing colors for each next number, so they can write down codes to regenerate keys.

Sending it to a phone or computer over the network goes against basic rules of private key management. It COULD be done if people want to take that risk, but we should allow people the more secure option to have codes flashed to them.

doc/architecture/decisions/0015-key-management.md
artbrock: I don't think there's anything **centralized** about it, but if you are creating a DPKI account that you want to use to bind some of your devices and apps together under a set of seeds so you can regenerate the keys, that is a Holo identity.  We have to be able to have web browsing users also generate Holo identities even when they don't run Holochain or DPKI locally, if they want to be able to WRITE to an app. They need to generate their authoring keys, and authorize hosts with network keys, to operate on their behalf.

What makes you think it's centralized though? Each person still manages their keys/seeds... the software provides for them to bridge across multiple apps or devices they use to access Holo apps.

doc/architecture/decisions/0015-key-management.md
artbrock: LibSodium lets you use 32 byte (not bit ... 256 bits) seeds as well as 32 byte keys. However, a seed is not a key, it is the source of entropy for generating key pairs.  I believe a seed doesn't have its counterpart pub/private pairs. It is used with an (elliptical curve) algorithm for selecting the points on the curve which give you a pairing.

doc/architecture/decisions/0015-key-management.md
sphinxc0re: Three words: Integrated mini printer. 😂

doc/architecture/decisions/0015-key-management.md
lucksus: Ok, so you are both confirming that key != seed.
Isn't the names of these functions misleading then? They are creating something that is derived from a seed. So it must be key.

Or are you saying, @artbrock, that since both are 32bit we just use the generated key as a seed to introduce another level in the hierarchy? Is that how you do that? I don't know if I'm asking a dumb question or if I spotted a mistake.

neonphog: In libsodium, derivation takes an input seed buffer and produces another seed buffer in a deterministic and non-reversible manner. This can be done as many times as desired.

Any of those seed buffers can be used to produce asymmetric keypairs for signing or encryption. The same seed will always produce the same keypair, making them equivalent from a security exposure perspective to a private key.

The suggested api works with the seeds for simplicity. DpkiDevice and DpkiApp will use the seed to generate the appropriate keypairs for use in encryption and signing.

doc/architecture/decisions/0015-key-management.md
neonphog: Yeah, so a "default" public identity would consist of:
- revocation public key
- 5 device signing keys that are all authorized to add additional device signing keys
which behind the scenes are managed with an HD hierarchy

A "holo" public identity would consist of:
- a single public key that is authorized to revoke and add additional device signing keys
which behind the scenes is generated from a combination of public salt and password hash

doc/architecture/decisions/0015-key-management.md
neonphog: it is a "holo"-port... where's the holographic projector?

doc/architecture/decisions/0015-key-management.md
```diff
@@ -0,0 +1,222 @@
...
+```rust
+/// utilities for working with mnemonic word lists
+pub mod Dpki {
+    fn seed_to_mnemonic(seed: &[u8; 32]) -> String;
+    fn mnemonic_to_seed(mnemonic: &str) -> [u8; 32];
+}
+
+/// struct for hd seed derivation handling
+impl DpkiHd {
+    pub fn new_random() -> Self;
```
ddd-mtl: Rust naming convention for constructor is:

Item | Convention
-- | --
General constructors | `new` or `with_more_details`
Conversion constructors | `from_some_other_type`

So I would rename `new_random()` to `with_random()`

doc/architecture/decisions/0015-key-management.md
```diff
@@ -0,0 +1,222 @@
...
+    pub fn new_from_primary_seed(primary_seed: &[u8; 32]) -> Self;
```
ddd-mtl: `with_primary_seed()`

doc/architecture/decisions/0015-key-management.md
```diff
@@ -0,0 +1,222 @@
...
+    pub fn get_primary_seed(&self) -> [u8; 32];
+    pub fn get_revoke_seed(&self) -> [u8; 32];
+    pub fn get_device_seed(&self, index: u64) -> [u8; 32];
+}
+
+/// struct for managing keypair seeds with salt / passphrase
+impl DpkiHolo {
+    pub fn new_random_from_passphrase(passphrase: &str) -> Self;
```
ddd-mtl: `with_random_from_passphrase`

doc/architecture/decisions/0015-key-management.md
```diff
@@ -0,0 +1,222 @@
...
+    pub fn new_from_salt_passphrase(
```
ddd-mtl: `with_salt_passphrase`

doc/architecture/decisions/0015-key-management.md
```diff
@@ -0,0 +1,222 @@
...
+        salt: &[u8; 32], passphrase: &str) -> Self;
+    pub fn get_salt(&self) -> [u8; 32];
+}
+
+/// struct for an individual device's usage of crypto
+impl DpkiDevice {
+    pub fn new_from_holo(holo: &DpkiHolo) -> Self;
```
ddd-mtl: `with_holo()`

doc/architecture/decisions/0015-key-management.md
```diff
@@ -0,0 +1,222 @@
...
+    pub fn new_from_seed(device_seed: &[u8; 32]) -> Self;
```
ddd-mtl: `with_seed()`

doc/architecture/decisions/0015-key-management.md
```diff
@@ -0,0 +1,222 @@
...
+    pub fn new_from_seed_and_pin(device_seed: &[u8; 32], device_pin: &str) -> Self;
```
ddd-mtl: `with_seed_and_pin()`

doc/architecture/decisions/0015-key-management.md
```diff
@@ -0,0 +1,222 @@
...
+    pub fn new_from_saved(data: &[u8], passphrase: &str) -> Self;
```
ddd-mtl: `with_saved()`

doc/architecture/decisions/0015-key-management.md
```diff
@@ -0,0 +1,222 @@
...
+    pub fn get_sign_author_public_key(&self) -> [u8; 32];
+    pub fn get_sign_network_public_key(&self) -> [u8; 32];
+    pub fn get_encrypt_public_key(&self) -> [u8; 32];
+    // ... additional functions for actual signing / encryption
+}
+
+/// struct for managing the distributed public data
+/// associated with a user's identity
+impl DpkiApp {
+    pub fn new_defaults(hd_priv: &DpkiHd) -> Self;
```
ddd-mtl: `with_defaults()`

doc/architecture/decisions/0015-key-management.md
```diff
@@ -0,0 +1,222 @@
...
+    pub fn new_holo(username: &str, holo: &DpkiHolo) -> Self;
```
ddd-mtl: `with_holo()`

doc/architecture/decisions/0015-key-management.md
sphinxc0re: `with_passphrase`

doc/architecture/decisions/0015-key-management.md
zippy: From the point of view of code in the **Holochain** library I'm a little worried about having a structure with the name Holo.  How about if the structure were actually: 
`DpkiIdentity` (not `DpkiHolo`) and the `new` functions take a string based parameter that identifies an identity class that actually gets factored into the determinism, i.e. `new_random_from_passphrase("HOLO","some passphrase here")`  That way we aren't tightly binding Holochain to Holo's use of DPKI and allowing/encouraging other uses DPKI solutions too.

lucksus: The reason I asked about centralized ID is that it initialized only form a password.. Not really understanding what's going on and why this is different the other cases...

So you say that creates a key from a password? Instead of using seed words you take the password, salt it and treat that as seed to get a key pair from? And the reason is I guess to have an identity that you can recover just with that password?

doc/architecture/decisions/0015-key-management.md
thedavidmeister: can we mimic how ledger approaches initialization of seeds and pins?

doc/architecture/decisions/0015-key-management.md
thedavidmeister: @zippy visibility that i asked a question re: "market research" that was never responded to ^^

thedavidmeister:
@zippy what is the status of this one?

Connoropolous:
@zippy perhaps this is just too out of date to salvage now, should this get closed?

thedavidmeister:
i'm closing due to extreme staleness

feel free to reopen @zippy if you feel there is value to it

PR #288, opened 2018-09-04
add naming conventions to mdbook

### LINKED ISSUE #287: document naming conventions
because 👿 💥 🤔 

as i understand it:

- `Action` is `VerbNoun` e.g. `GetEntry`
- `ActionResponse` is `ActionName` e.g. `Action::GetEntry` results in `ActionResponse::GetEntry`
- reducer name is `reduce_action_name` e.g. `reduce_get_entry`
- Actor `Protocol` is `VerbNoun` or `Verb` e.g. `PutEntry` or `Setup`
- Result of a `Protocol` is `VerbNounResult` e.g. `PutEntryResult`
- method names that get something are the name of the thing being got, e.g. `entry()`
- method names that have side effect other than get are `verb_noun()` e.g. `put_entry()`
- avoid short names like `t`, `e`, `h` when `table`, `entry`, `header` would be clearer (almost always)

intent is to align with rust naming conventions https://doc.rust-lang.org/1.0.0/style/style/naming/README.html then fill in the gaps where our own code might be ambiguous

if in doubt refer to the rust conventions

if there is no clear fit then make something up, but make it "self documenting", i.e. explicit, descriptive and indicative of what it does/is/will happen

names should be as "boring" as possible

thedavidmeister:
fixes #287 

i won't merge this until _everyone_ approves it!


doc/holochain_101/src/naming_conventions.md
```diff
@@ -0,0 +1,55 @@
+# Naming things
+
+> There are only two hard things in Computer Science: cache invalidation and naming things.
```
sphinxc0re: You forgot "off-by-one errors" 😁

thedavidmeister:
@ddd-mtl back to you

PR #291, opened 2018-09-04
Fixed FileTable not working on windows. 

ddd-mtl:
Made it platform agnostic.

### DIFF
--- a/core/src/hash_table/file/mod.rs
+++ b/core/src/hash_table/file/mod.rs
@@ -188,7 +188,11 @@ pub mod tests {
     fn test_dir() {
         let (table, _dir) = test_table();
         let re = |s| {
-            let regex_str = format!(r".*\.tmp.*\{}{}", MAIN_SEPARATOR, s);
+            let regex_str = if  MAIN_SEPARATOR == '\\'  {
+                format!(r".*\.tmp.*\{}{}", MAIN_SEPARATOR, s)
+            } else {
+                format!(r".*\.tmp.*{}{}", MAIN_SEPARATOR, s)
+            };
             Regex::new(&regex_str).expect("failed to build regex")
         };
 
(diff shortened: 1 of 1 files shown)


core/src/hash_table/file/mod.rs
```diff
@@ -207,7 +209,10 @@ pub mod tests {
         let (table, _dir) = test_table();
 
         let re = |s, k| {
-            let regex_str = format!(r".*\.tmp.*{}{}{}{}{}{}\.json", MAIN_SEPARATOR, MAIN_SEPARATOR, s, MAIN_SEPARATOR, MAIN_SEPARATOR, k);
+            let regex_str = format!(
+                r".*\.tmp.*{}{}{}{}{}{}\.json",
+                MAIN_SEPARATOR, MAIN_SEPARATOR, s, MAIN_SEPARATOR, MAIN_SEPARATOR, k
```
sphinxc0re: Shouldn't there be a backlash in front of the separators? Because, what does this do?
```ruby
/.*\.tmp.*//something//something/
```

This makes sense:
```ruby
/.*\.tmp.*\\something\\something/
/.*\.tmp.*\/something\/something/
```

core/src/hash_table/file/mod.rs
```diff
@@ -61,7 +62,7 @@ impl FileTable {
 
     /// given a Table enum, ensure that the correct sub-directory exists and return the string path
     fn dir(&self, table: Table) -> Result<String, HolochainError> {
-        let dir_string = format!("{}/{}", self.path, table.to_string());
+        let dir_string = format!("{}{}{}", self.path, MAIN_SEPARATOR, table.to_string());
```
sphinxc0re: This can be done much easier and cleaner (I love the path API):

First I'd make `self.path` a `Pathbuf` which is normally used for anything path related. Then you can use:

```rust
let dir = self.path.join(table.to_string());
```
This makes `dir` a `Pathbuf` which is an owned `Path`. I would rather return this than a `String`

core/src/hash_table/file/mod.rs
```diff
@@ -205,7 +213,18 @@ pub mod tests {
         let (table, _dir) = test_table();
 
         let re = |s, k| {
```
sphinxc0re: I don't know what these closures are for. Could you rename them and also rename the params to something more expressive. + some docs would be nice

core/src/hash_table/file/mod.rs
ddd-mtl: I can't do this because self.path is a String.
Can you write a ticket for this please?
The object of this PR is just to fix it so it works on windows.

core/src/hash_table/file/mod.rs
ddd-mtl: I don't know either. I did not code this. It is not part of this PR.

PR #294, opened 2018-09-05
panic if attempting to access capability for zome api functions

thedavidmeister:
delivers zome api panic for capabilities from #269


ddd-mtl:
Yes this needs a TODO follow up

thedavidmeister:
i added a todo @ddd-mtl @zippy

PR #296, opened 2018-09-05
40% faster travis

### LINKED ISSUE #45: tests actually get run twice on travis
once for `make test` and once for `cargo tarpaulin --out Xml` in codecov handling

thedavidmeister:
fixes #45 

## changes

- 4x builds (adds build overhead, see followup discussion for next steps here)
- only run standard tests once, as code coverage (reduces total testing time)
- more fine grained makefile to facilitate CI needs
- standardise casing for makefile

### before

2x builds

- mdbook
- everything else

here are a few recent builds with their durations:

- https://travis-ci.org/holochain/holochain-rust/builds/425097960 27:21
- https://travis-ci.org/holochain/holochain-rust/builds/424862399 26:08
- https://travis-ci.org/holochain/holochain-rust/builds/424778182 27:24

note that duration and total time are different if the build was waiting in the queue to start

the good news: our build times are pretty consistent, +/- 1 minute atm

the bad news: it takes nearly 30 minutes!

### after

4x builds, 16 minutes total = 40% faster!

- mdbook
- c tests
- standard tests + code coverage
- formatting + clippy

https://travis-ci.org/holochain/holochain-rust/builds/425113605?utm_source=github_status&utm_medium=notification

tradeoff: each PR will be faster but we can process fewer PRs at one once

justification: typically we aren't maxing our parallel builds, but we are struggling with PR turnaround times to incorporate feedback

## followup ideas/discussions

- most of the time spent is on build rather than test atm, which really limits the benefits of parallel jobs as every job must re-do the baseline build work
  - most of this build time is cargo downloading and compiling dependencies
- currently we have a "one size fits all" ubuntu box
  - pulling the docker image takes 108 seconds
  - bakes in several versions of rust, does this increase the download time?
  - when cargo runs it is a cold start so it downloads and compiles a _lot_
    - clippy is ~330s https://travis-ci.org/holochain/holochain-rust/jobs/425104644#L543
    - fmt is ~135s https://travis-ci.org/holochain/holochain-rust/jobs/425104644#L535
    - if we pre-ran some cargo commands on docker hub, could this warm things a bit for travis?
  - is ubuntu the best choice? could we find a lighter alternative?
  - could we have a few dedicated docker images targeted at the different testing goals?
- clippy is very heavy at 5+ minutes, yet we don't enforce or review any of the clippy warnings
- docker images are not cached, they are all pulled every time, apparently there is a `--cache-from` option in docker that can be used to hook travis into AWS for BYO caching to speed things up

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,12 +8,12 @@ jobs:
     - name: "C tests"
       install: docker pull holochain/holochain-rust:develop
       script:
-        - . docker/run make build test-c-ci
+        - . docker/run make build test_c_ci
 
     - name: "Standard tests + code coverage"
       install: docker pull holochain/holochain-rust:develop
       script:
-        - . docker/run make wasm-build
+        - . docker/run make wasm_build
         - . docker/run-cov
       after_success:
         - bash <(curl -s https://codecov.io/bash)
@@ -21,7 +21,7 @@ jobs:
     - name: "Formatting & clippy"
       install: docker pull holochain/holochain-rust:develop
       script:
-        - . docker/run make fmt-check
+        - . docker/run make fmt_check
         - . docker/run make clippy
 
     - name: "Mdbook Build"
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,15 @@ dist: trusty
 jobs:
   include:
     - name: "C tests"
-      install: docker pull holochain/holochain-rust:develop
-      script: . docker/run make test_c_ci
+      install:
+        - docker pull holochain/holochain-rust:develop
+        - make wasm-build
+      script: . docker/run make test-c-ci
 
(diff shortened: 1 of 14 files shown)


.travis.yml
```diff
@@ -5,13 +5,25 @@ dist: trusty
...
 jobs:
   include:
-    - stage: "All"
-      name: "Tests"
+    - name: "C tests"
       install: docker pull holochain/holochain-rust:develop
-      script: . docker/run-test
-      after_success:
+      script:
+        - . docker/run make build test_c_ci
```
lucksus: The C tests need Holochain Rust build first and that is what is happening here. Extracting that and only building it once would be the next optimization I guess..

thedavidmeister: @lucksus aight, i'll look into that :)

Issue #305, opened 2018-09-07
speed up travis build


thedavidmeister:
ideas at #296

target is 10 mins

PR #306, opened 2018-09-07
spike 1: dedicated codecov docker

thedavidmeister:
child of #305 

there's no way these tests can pass

this dockerfile needs to be available to docker hub so it can be built there before travis can pull it down to use

~i want to merge this into develop, get it building on docker hub, then make a followup ticket to tweak travis-side and benchmark~

i'll manually push to docker hub


thedavidmeister:
well this makes me sad:

https://travis-ci.org/holochain/holochain-rust/builds/426554302?utm_source=github_status&utm_medium=notification

really no noticeable improvement at all when using the dedicated box, cargo on travis still insists on re-downloading and compiling everything

certainly nothing that justifies the hassle of maintaining parallel docker boxes

i wonder what the story is with native travis builds on top of `make` using the travis cargo cache...

Issue #159, opened 2018-07-23
should we return an error rather than assert! in zome API fns?


thedavidmeister:
see `invoke_commit` for an example, first line is `assert!` but fn signature returns a result

plenty of other things that panic throughout...

zippy:
To me the distinction is based on whether the error is truly a runtime error that could happen at runtime, in which case it MUST be an error message.  A panic! or an assert! would be fine if it's something that would happen only in the development cycle of an app.  i.e. the library can panic on something that would tell the app developer that they have screwed up in a way detectable during app dev time, but should never panic in a way that will stop a production container/composer that's holding the app instance.

I can't tell which case this assert! falls under.

thedavidmeister:
@zippy potentially new information: _any_ panic inside anything that is called from WASM cannot be debugged, all you get is `Err(ErrorGeneric("Trap: Trap { kind: Unreachable }"))` and backtrace does not work because WASM doesn't have the machinery

zippy:
Right!  That's interesting, I hadn't thought about how that would affect calls out from wasm.  I kind of like us keeping the same distinction in mind however, and figuring out our own "panic" style error that would die during dev time for errors like this.  Maybe that's what this ticket turns into:

"As an app developer I want fatal errors in my use of api calls (like incorrect number of arguments passed to a commit, etc) to throw early in dev cycle with clear error messages so I can debug them"

thedavidmeister:
@zippy yeah that's more or less the user story

the context is that i did a quick google and found that directly calling anything that panics in rust is fundamentally not supported in wasm, but of course we can return errors and handle them outside wasm, which should allow debugging like what you're talking about

@ddd-mtl already made a good start on this with the  `HcApiReturnCode`

here's some info on the wasm side of the story:

- https://github.com/rust-lang/rust/issues/46996
- https://github.com/WebAssembly/design/blob/master/Semantics.md#unreachable

my thoughts here are that, even if we don't have a sophisticated alternative, we should never call anything that has the option of panicking in wasm because it's not supported by wasm. even in the case that a panic might be impossible today (there are a few examples where some earlier conditional logic makes the later panic impossible), that doesn't mean it won't become tomorrow's debugging nightmare during some future refactor.

the wasm utils module could give us convenience methods for all this if we find a lot of boilerplate cropping up

zippy:
Next step is to identify all the places where this might happen.

ddd-mtl:
[...]

PR #307, opened 2018-09-07
assemblyscript build needs wasm start call

Connoropolous:
All three memory allocator modules implemented for assemblyscript need the `start` method in wasm to be called, so `assert_no_start` doesn't work in this use case

If start doesn't exist it doesn't get called
https://github.com/paritytech/wasmi/blob/0409913a26ad16a622f49b7f885fc6415e80f26d/src/module.rs#L676

We get WAT code that looks like this out of assemblyscript builds (section of it)
```
(func $start (; 1 ;) (type $v)
  (set_global $~lib/allocator/arena/startOffset
   (i32.and
    (i32.add
     (get_global $HEAP_BASE)
     (get_global $~lib/internal/allocator/AL_MASK)
    )
    (i32.xor
     (get_global $~lib/internal/allocator/AL_MASK)
     (i32.const -1)
    )
   )
  )
  (set_global $~lib/allocator/arena/offset
   (get_global $~lib/allocator/arena/startOffset)
  )
 )
)
```

### DIFF
--- a/core/src/nucleus/ribosome/api/mod.rs
+++ b/core/src/nucleus/ribosome/api/mod.rs
@@ -25,7 +25,7 @@ use std::{
     sync::{mpsc::Sender, Arc},
 };
 use wasmi::{
-    self, Error as InterpreterError, Externals, FuncInstance, FuncRef, ImportsBuilder,
+    self, Error as InterpreterError, Externals, NopExternals, FuncInstance, FuncRef, ImportsBuilder,
     ModuleImportResolver, ModuleInstance, RuntimeArgs, RuntimeValue, Signature, Trap, TrapKind,
     ValueType,
 };
@@ -266,10 +266,10 @@ pub fn call(
     let mut imports = ImportsBuilder::new();
     imports.push_resolver("env", &RuntimeModuleImportResolver);
 
-    // Create module instance from wasm module, and without starting it
+    // Create module instance from wasm module, and start it if start is defined
     let wasm_instance = ModuleInstance::new(&module, &imports)
         .expect("Failed to instantiate module")
-        .assert_no_start();
+        .run_start(&mut NopExternals)?;
 
     // write input arguments for module call in memory Buffer
     let input_parameters: Vec<_> = parameters.unwrap_or_default();


Connoropolous:
@thedavidmeister and @lucksus I worked on adding an "abort" import so that I could full stack test my assemblyscript zome builds... 

consequently, the code is kinda scrappy atm, but it runs :) I need some help to fix it, either through reviews, or could someone make improvements and just commit 'em.

core/src/nucleus/ribosome/api/mod.rs
```diff
@@ -41,8 +41,11 @@ use wasmi::{
 #[repr(usize)]
 #[derive(FromPrimitive, Debug, PartialEq)]
 pub enum ZomeApiFunction {
+    /// Abort
+    Abort = 0,
```
ddd-mtl: Any reason for putting it in first place?
I would like to keep `MissingNo` on top if possible.

core/src/nucleus/ribosome/api/mod.rs
```diff
@@ -216,6 +221,13 @@ pub fn call(
             index: usize,
             args: RuntimeArgs,
         ) -> Result<Option<RuntimeValue>, Trap> {
+            // Abort
+            if index == 0 {
+                // println!("message, file, line, column");
```
ddd-mtl: dead code

ddd-mtl:
If the existing wasm-tests still works, I'm fine with this.

lucksus:
Please do a `. docker/run-fmt` so the fmt test is green as well - test still works.

core/src/nucleus/ribosome/api/mod.rs
Connoropolous: No prob, will do. I thought maybe it should go there, since it's not really "our thing" ... it's there for the sake of assemblyscript. But it doesn't matter

core/src/nucleus/ribosome/api/mod.rs
Connoropolous: as mentioned, @ddd-mtl and @thedavidmeister I just copied what `NopExternals` does here, and throws 'Unreachable' but that's not actually right. I feel in the dark about rust error handling still. 

@AshantiMutinta maybe you have a few to help me in here on this?

AshantiMutinta: absolutely, ping me on mattermost @Connoropolous or send me an email. If you want I can give you a crash course just on how everything works!

thedavidmeister:
@Connoropolous i'm not sure, when would abort be called? an example might help me offer a suggestion for what we should do.

Connoropolous:
@thedavidmeister seems to be a function of the memory management utils that they throw in, it wants to throw a useful error, with the file, message, line and column... See examples of it from their repo https://github.com/AssemblyScript/assemblyscript/search?utf8=✓&q=abort&type=

thedavidmeister:
@Connoropolous hmmm, does that work with rust?

my understanding was that any kind of rust panic in wasm is an undebuggable mess, https://github.com/holochain/holochain-rust/issues/159

if not, and it allows us to get real debugging, is there a way to utilise it in the above issue more generally? ^^

core/src/nucleus/ribosome/api/mod.rs
```diff
@@ -237,18 +257,30 @@ pub fn call(
...
-            // Take the canonical name and find the corresponding ZomeApiFunction index
-            let index = ZomeApiFunction::str_to_index(&field_name);
-            match index {
-                index if index == ZomeApiFunction::MissingNo as usize => {
-                    return Err(InterpreterError::Function(format!(
-                        "host module doesn't export function with name {}",
-                        field_name
-                    )));
-                }
+            let api_fn = ZomeApiFunction::from_str(&field_name)?;
```
Connoropolous: This is the line causing a compile error

screen shot 2018-09-14 at 12 45 42 pm

Has to do with this, but I'm not sure how to fix:
https://paritytech.github.io/wasmi/wasmi/trait.HostError.html

Connoropolous:
@thedavidmeister I looked at the other issue mentioned, truth be told, I don't know. 

I found 
"Under some conditions, wasm execution may produce a `Trap`, which immediately aborts execution.
Traps can't be handled by WebAssembly code, but are reported to the embedder."

thus, the "abort" host function gets called.

core/src/nucleus/ribosome/api/mod.rs
AshantiMutinta: Don't know much about the codebase yet but it seems like there is specific type of trait that string should implement for the str type. I would 1) check if it something is derivable 2) write an implementation definition of the trait for str 3)check if an implementation already exists and bring it into scope.

Connoropolous: Thanks Ashanti
As it turns out, I was misled by the error... the problem really had to do with the fact that `from_str` was returning a "Result" type value, with an Ok or an Err, whereas before it had been something different. That was messing up the `match` statement underneath

PR #309, opened 2018-09-10
spike 2: attempt to run codecov on native travis

thedavidmeister:
child of #305


thedavidmeister:
23 mins for cold cache

thedavidmeister:
11 mins warm cache

https://travis-ci.org/holochain/holochain-rust/jobs/426711870

Makefile
```diff
@@ -46,30 +48,80 @@ test: test_non_c c_binding_tests ${C_BINDING_TESTS}
 test_non_c: main
 	RUSTFLAGS="-D warnings" $(CARGO) test
 
-test_c_ci: c_binding_tests ${C_BINDING_TESTS}
+c_build:
+	cd dna_c_binding && $(CARGO) build
```
lucksus: why is `c_build` only building `dna_c_binding`? I would have assumed `core_api_c_binding` cause that actually includes holochain and is what is needed by the c-tests...

thedavidmeister: @lucksus i can add another item, cherry pick what makes sense, the main thing was not building _everything_

thedavidmeister:
@lucksus is that in scope here? to add new tests to CI?

PR #312, opened 2018-09-11
Action Refactoring - Pre Work 1

lucksus:
# Put reference to state into context

The Futures I'm using in the action refactoring replace the observers living *in* the state with their own implementation of `poll()`. For that to work they need acccess to the *one, global* state.

I've decided to put that in the context since it makes sense to assume the state being always there and accessible and thus part of the *context* of any code running inside Holochain.

The state is private and has a public getter function that only returns a read-lock because nobody other than the reducers should be able to write to it. `set_state()` should only be called once by the Instance (or in tests).


core/src/context.rs
```diff
@@ -2,21 +2,120 @@ use error::HolochainError;
...
+        }
+    }
     // helper function to make it easier to call the logger
     pub fn log(&self, msg: &str) -> Result<(), HolochainError> {
         let mut logger = self.logger.lock().or(Err(HolochainError::LoggingError))?;
         logger.log(msg.to_string());
         Ok(())
     }
+
+    pub fn set_state(&mut self, state: Arc<RwLock<State>>) {
```
ddd-mtl: Use something like pub(self) so its not public to everyone, as intended from PR description.

core/src/context.rs
```diff
@@ -2,21 +2,120 @@ use error::HolochainError;
...
+    impl Logger for TestLogger {
+        fn log(&mut self, msg: String) {
+            self.log.push(msg);
+        }
+    }
+
+    /// create a test logger
+    pub fn test_logger() -> Arc<Mutex<TestLogger>> {
+        Arc::new(Mutex::new(TestLogger { log: Vec::new() }))
+    }
```
ddd-mtl: Duplicate code from `instance.rs`.
Just use `use instance::tests::test_logger;` instead, no?

core/src/context.rs
```diff
@@ -2,21 +2,120 @@ use error::HolochainError;
...
+        let mut context = Context::new(
+            holochain_agent::Agent::from_string("Terence".to_string()),
+            test_logger(),
+            Arc::new(Mutex::new(SimplePersister::new())),
+        );
+
+        match context.state() {
+            None => assert!(true),
+            _ => assert!(false),
+        }
```
ddd-mtl: replace `match` with `assert!(context.state().is_none());`

core/src/context.rs
```diff
@@ -2,21 +2,120 @@ use error::HolochainError;
...
+
+        let global_state = Arc::new(RwLock::new(State::new()));
+        context.set_state(global_state.clone());
+
+        {
+            let _read_lock = global_state.read().unwrap();
+            match context.state() {
+                Some(_read_lock) => assert!(true),
+                _ => assert!(false),
+            };
```
ddd-mtl: replace `match` with `assert!(context.state().is_some());`

core/src/instance.rs
```diff
@@ -86,11 +86,18 @@ impl Instance {
         (rx_action, rx_observer)
     }
 
+    pub fn initialize_context(&self, context: Arc<Context>) -> Arc<Context> {
```
ddd-mtl: `pub` doesn't seem necessary.

core/src/instance.rs
```diff
@@ -116,16 +123,32 @@ impl Instance {
...
+            {
+                // Only get a read lock first so code in reducers can read state as well
+                let state = self
+                    .state
+                    .read()
+                    .expect("owners of the state RwLock shouldn't panic");
+
+                // Create new state by reducing the action on old state
+                new_state = state.reduce(
+                    context.clone(),
```
ddd-mtl: This is kind of weird since `context` also holds a state...
maybe `reduce` doesn't need to be a method anymore, and is better off as a function that gets current state from `context` argument?

core/src/context.rs
```diff
@@ -2,21 +2,120 @@ use error::HolochainError;
...
+    }
     // helper function to make it easier to call the logger
     pub fn log(&self, msg: &str) -> Result<(), HolochainError> {
         let mut logger = self.logger.lock().or(Err(HolochainError::LoggingError))?;
         logger.log(msg.to_string());
         Ok(())
     }
+
+    pub fn set_state(&mut self, state: Arc<RwLock<State>>) {
+        self.state = Some(state);
```
thedavidmeister: how do we stop uncoordinated processes clobbering each other's state?

core/src/context.rs
lucksus: Since I use it in Instance, I can only restrict it to the crate level: `pub(crate)`.
But yes.

core/src/context.rs
lucksus: Ah, thanks! I did try using the one from test_utils but that didn't work and I didn't know why. Couldn't find the Logger implementation which was there.. But using the one from instance works.

core/src/context.rs
thedavidmeister: @lucksus if A and B do concurrent reads, then each compute a new state then both call `set_state`, what happens?

lucksus:
@ddd-mtl, hm, well we could have had `reduce()` being an unassociated pure function from the beginning. I think having it a method of state makes sense because:
* it is only ever used with the state and is logically tightly coupled to this concept
* we don't need to put the thread synchronization into `reduce()` if we keep it a method (Context::state() has to get a read() lock).

But I don't have a strong opinion.
Does that need to block this PR?

core/src/context.rs
lucksus: The state gets only changed in `process_action()` through the main event loop. The whole point of this pattern is that you get a clear sequence of state changes. If we have two threads running an event loop on the same instance, we have violated this architecture and need to fix that.

thedavidmeister: i'm asking about concurrent reads not writes

state = 1
-- concurrent  --
A reads state 1
B reads state 1
A calculates state + 1 = 2
B calculates state + 2 = 3
-- sequential --
A sets state to 2
B sets state to 3

final state is 3

shouldn't state be 4?

core/src/instance.rs
ddd-mtl: In my understanding of Rust, tests in same crate don't care about visibility, so no need for `pub` for tests.

lucksus: But all tests that create an instance and process actions manually (or call reducer functions, etc.) without running the redux loop the regular why need to construct a context and thus need to call this function. Take a look at #315.

core/src/context.rs
lucksus: After our conversation during stand-up, can we resolve this here for now, @thedavidmeister?

core/src/context.rs
```diff
@@ -2,21 +2,99 @@ use error::HolochainError;
...
     }
+
+    pub(crate) fn set_state(&mut self, state: Arc<RwLock<State>>) {
+        self.state = Some(state);
+    }
+
+    pub fn state(&self) -> Option<RwLockReadGuard<State>> {
+        match self.state {
+            None => None,
+            Some(ref s) => Some(s.read().unwrap()),
```
AshantiMutinta: Is there a possibility of this panicking if the lock is poisoned?

ddd-mtl:
Non-blocking but will keep on eye on this.

core/src/context.rs
lucksus: Yes, I've even added a test that is testing that.

But this is related to @thedavidmeister's comment, since there is and should really only be one spot where we write to the state: Instance::process_action(), and that can and should not interfere with other treads reading the state.

AshantiMutinta: Awesome! Looks good to me

PR #313, opened 2018-09-11
Action Refactoring - Pre Work 2

lucksus:
Sits on top of #312 - Action Refactoring - Pre Work 1, merge that one first.

# Actually used Action::ValidateEntry
and its reducer in nucleus.

Added Action::ValidationResult that puts the result that comes back from the validation thread into the state.

### DIFF
--- a/core/src/nucleus/mod.rs
+++ b/core/src/nucleus/mod.rs
@@ -5,6 +5,7 @@ pub mod state;
 use action::{Action, ActionWrapper, NucleusReduceFn};
 use context::Context;
 use error::HolochainError;
+use holochain_dna::{wasm::DnaWasm, zome::capabilities::Capability, Dna};
 use instance::{dispatch_action_with_observer, Observer};
 use nucleus::{
     ribosome::callback::{
@@ -22,6 +23,7 @@ use std::{
 };
 
 use hash_table::sys_entry::ToEntry;
+use nucleus::ribosome::api::call::reduce_call;
 
 /// Struct holding data for requesting the execution of a Zome function (ExecutionZomeFunction Action)
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
@@ -44,6 +46,12 @@ impl ZomeFnCall {
             fn_name: function.to_string(),
             parameters: parameters.to_string(),
         }
+    }
+
+    pub fn same_fn_as(&self, fn_call: &ZomeFnCall) -> bool {
+        self.zome_name == fn_call.zome_name
+            && self.cap_name == fn_call.cap_name
+            && self.fn_name == fn_call.fn_name
     }
 }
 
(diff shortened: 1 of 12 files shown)


core/src/nucleus/state.rs
```diff
@@ -17,6 +18,7 @@ impl Default for NucleusStatus {
     }
 }
 
+pub type ValidationResult = Result<(), String>;
```
ddd-mtl: More reason to rename action. This type shares the same name...

core/src/nucleus/mod.rs
```diff
@@ -366,23 +380,63 @@ fn reduce_validate_entry(
...
+    match zome_name_for_entry_type(&state.dna.clone().unwrap(), entry.entry_type()) {
+        None => {
+            let error = format!("Unknown entry type: '{}'", entry.entry_type());
+            state
+                .validation_results
+                .insert(action_wrapper.clone(), Err(error.to_string()));
         }
-    }
+        Some(zome_name) => {
+            state.validations_running.push(action_wrapper.clone());
```
ddd-mtl: `validations_running` doesn't seem to be useful for anything.
Maybe you are planning to use this later to make sure a validation of an entry is not pushed twice?

lucksus: Yeah, you're right, and I was thinking about not committing it.
The main reason I did is to communicate context. This is how, I think, we should think about the state and this pattern.

I would like us to have a way of formatting or even visualizing the state soon, so that we can reason about state changes easily when debugging. This is a first hint to that...

ddd-mtl: I see two possible changes then:
1. Add `#[cfg(debug)]` in front of `validations_running`. So we know its for debugging only.
2. Panic or return error if we add the same validateEntry twice in that vector, to make `validations_running` actually useful.

lucksus: right, thanks. I will `#[cfg(debug)]` then. Ok?

core/src/nucleus/mod.rs
```diff
@@ -358,6 +360,18 @@ fn reduce_execute_zome_function(
     }
 }
 
+fn get_zome_name_for_entry_type(dna: &Dna, entry_type: String) -> Option<String> {
```
ddd-mtl: I think this should actually be a method of `Dna`.
Would you mind moving this to `dna/src/lib.rs` ?

core/src/nucleus/mod.rs
ddd-mtl: 👍

core/src/nucleus/mod.rs
lucksus: I agree. Will do.

lucksus: Done, and also added a test.

core/src/nucleus/state.rs
thedavidmeister: yeah this does bug me a bit about rust, that aliases are used commonly without namespaces, so your local unique names are sort-of global unique names

ddd-mtl: Would renaming it `HcValidationResult` solve this issue?

PR #315, opened 2018-09-11
Action Refactoring

### LINKED ISSUE #256: spike for co-ordinating actions that need the result of other actions - "Action Refactor"
## problem

we want to be able to dispatch new actions from reducers and use the result of the new action for the result of the first action.

e.g.

commit action ->
commit reduce ->
validate action ->
validate reduce ->
validate result ->
commit result ->

the action loop doesn't start new actions until all the current action reducers return

this means we can't block inside reducers on new actions

since we can't block, we might try splitting the outer action into two actions:

commit start action ->
commit start reduce ->
validate action ->
commit start result ->
validate reduce ->
commit end action ->
validate result ->
commit end reduce ->
commit end result ->

but then how does the commit start observer sensor looking for the unique ID of commit start find the commit end result?

the commit end will have some unique ID that the observer sensor can't know ahead of time

## potential solutions to spike

- futures
  - will be in repo if/when riker lands
  - has a `poll` function that is designed to be nested/composed in this way
  - doesn't have an obvious way to achieve the action reducing pattern

- some kind of wrapper/meta/co-ordinator
  - bundle related actions
  - a bundle would have a shared id that a sensor can track

- actors
  - hybrid future/queue/state option

lucksus:
closes #256 

# Action chaining with Futures FTW

The gist of `invoke_commit_entry()` now reads like this:

```rust
    // Create Chain Entry
    let entry =
        ::hash_table::entry::Entry::new(&entry_input.entry_type_name, &entry_input.entry_content);

    // Wait for future to be resolved
    let task_result: Result<ActionResponse, String> = block_on(
        // First validate entry:
        validate_entry(entry.clone(), &runtime.action_channel, &runtime.context)
            // if successful, commit entry:
            .and_then(|_| commit_entry(entry.clone(), &runtime.action_channel, &runtime.context)),
    );

    let json = match task_result {
```

This works because both commit and validate are implemented adhering to our redux architecture, which means that we have reducers that listen for Action::ValidateEnry and Action::CommitEntry and that start these processes in separate threads and have those threads create actions (`Action::ValidationResult`) that change the state through reducers.

The futures now poll the state for the results to be there:

```rust
impl Future for ValidationFuture {
    fn poll(
        &mut self,
        cx: &mut futures::task::Context<'_>,
    ) -> Result<Async<Self::Item>, Self::Error> {
        cx.waker().wake();
        if let Some(state) = self.context.state() {
            match state.nucleus().validation_result(&self.action) {
                Some(Ok(())) => Ok(futures::Async::Ready(self.action.clone())),
                Some(Err(e)) => Err(e),
                None => Ok(futures::Async::Pending),
            }
        } else {
            Ok(futures::Async::Pending)
        }
    }
}
```

We need something that creates these futures for us. So far, we haven't had Action Creators which are a thing in redux. Now we do:

```rust
pub fn validate_entry(
    entry: Entry,
    action_channel: &Sender<ActionWrapper>,
    context: &Arc<Context>,
) -> ValidationFuture {
    let action_wrapper = ActionWrapper::new(Action::ValidateEntry(entry));
    dispatch_action(action_channel, action_wrapper.clone());
    ValidationFuture {
        context: context.clone(),
        action: action_wrapper,
    }
}
```

## File ordering
I've put these action creators together with the futures in their own file under `actions` whithin the module that also defines the state that the future observes and defines the reducer that processes the according actions (so: `agent` for commit and `nucleus` for validate). I would also like to move back the actions from core/src/actions.rs into that same file and pull them in in action. That way we are fully modular.

Maybe we can write a macro that includes actions, state and reducers from a list of given submodule names? @sphinxc0re, @AshantiMutinta? :)


core/src/agent/actions/commit.rs
```diff
@@ -0,0 +1,60 @@
...
+use hash_table::entry::Entry;
+use instance::dispatch_action;
+use std::sync::{mpsc::Sender, Arc};
+
+/// Commit Action Creator
+/// This is the high-level commit function that wraps the whole commit process and is what should
+/// be called from zome api functions and other contexts that don't care about implementation details.
+///
+/// Returns a future that resolves to an ActionResponse.
+pub fn commit_entry(
```
ddd-mtl: Maybe prefix the fn name with `dispatch_` or some other convention to make clear its not actually doing the work directly.
We already use `reduce_` and `invoke_` for the other Commit related functions.

lucksus: hm, but when used with future composers it would read strange:
```
block_on(
  dispatch_validate_entry()
    .and_then(|_|{ 
      dispatch_commit_entry() 
    }
  );
```
because it does not continue after the dispatch, but after the whole task was done.

ddd-mtl: How about `async_` or `await_` or `promise_` or `future_` or ...

I think my preference is `promise_`

lucksus:
@ddd-mtl, yes, you could put it that way. Or: there is no supervisor, the commit process is defined within `invoke_commit()`.

Let's how far we get with this and if we need to change that at some point. I think this is definitely an improvement and might be enough.

core/src/agent/actions/commit.rs
thedavidmeister: @ddd-mtl the thing is that `async` `await` `promise` `future` are all Coming Soon in rust core :/

thedavidmeister: what if this was more like `CommitFuture::from_entry()`?

ddd-mtl: @thedavidmeister I know, that's exactly why I proposed that those prefixs. But I definitely prefer your idea. I'll add that its also with Context, so maybe more like `CommitFuture::from_entry_with_context()`

lucksus: Well.. I don't like focusing on the Future aspect so much that we represent that in the name of these functions.

After our Redux conversations, I actually think we should go the whole way to pure reducers which makes total sense since we now have these functions as *Action Creators*. Any asynchronous running of zome functions etc. should actually happen in here. So `validate_entry()` dispatches a `ValidationStarted` action **and** kicks-off the worker thread for the validation (which right now is in the reducer and makes it be not a pure function). That worker thread also dispatches a `ValidationResult` action when it's done. Reducers only take those action parameters and update the state.

I would leave the name for the action creators here as they are now, since they are scoped in action already, i.e. `nucleus::actions::validate_entry(entry.clone()).and_then(....`.

PR #323, opened 2018-09-14
Hashstring newtype

ddd-mtl:
Rust's `type` does not enforce types statically so I changed `HashString` into a newtype.
And since HashString is a real type of its own, I changed the functions in `hash\mod.rs`  to methods of HashString.

### DIFF
--- a/core/src/action.rs
+++ b/core/src/action.rs
@@ -1,6 +1,7 @@
 use agent::state::AgentState;
 use context::Context;
-use hash_table::{entry::Entry, HashString};
+use hash_table::{entry::Entry};
+use hash::HashString;
 use holochain_dna::Dna;
 use instance::Observer;
 use nucleus::{state::NucleusState, EntrySubmission, ZomeFnCall, ZomeFnResult};
@@ -61,7 +62,7 @@ impl Hash for ActionWrapper {
     }
 }
 
-#[derive(Clone, PartialEq, Hash, Debug)]
+#[derive(Clone, PartialEq, Debug)]
 pub enum Action {
     /// entry to Commit
     /// MUST already have passed all callback checks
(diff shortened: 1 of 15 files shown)


Connoropolous:
I like this a lot... having it super clear when a HashString is being passed around in the code is super handy. Over just a "string"

ddd-mtl:
@thedavidmeister I like your idea of `EntryHash` and `HeaderHash`. Maybe we will be able to do this by changing `HashString` into a trait? I can take a look after this gets merged.
I also don't like `struct` that much but that's how [Rust wants you to do it](https://doc.rust-lang.org/book/second-edition/ch19-03-advanced-traits.html), and swears its zero cost, so I'll tolerate since the gain outweighs this itch.
I tried doing `impl` on `String` but that fails because [you are not aloud to do `impl` on a type defined in another crate](https://doc.rust-lang.org/stable/error-index.html#E0116).

thedavidmeister:
@ddd-mtl dang, i guess it has to be a struct then

a trait makes sense to me as that would fit nicely into the CAS i think

zippy:
The reason it was and then wasn't in the go version, is that I made Hash it it's own type assuming there would be different implementations of hashes even at the DHT level.  But that just because super cumbersome because the libp2p library assumed just  a string for the hash type and I had to be converting things back and forth all the time when talking about a hash that's being interpreted as a PeerID vs one being interpreted as just a regular DHT hash.  So the deeper issue is that we want to make this compatible and easy to use when it comes time to be interfacing with the networking/DHT side.  I'm guessing the Trait approach will work well.

ddd-mtl:
@zippy Noted, thanks. I'm also counting on the network abstraction layer to scope this problem and make it less cumbersome.

Issue #188, opened 2018-08-02
linking


zippy:
parent of #95
parent of #92
parent of #186

ddd-mtl:
https://github.com/holochain/holochain-rust/pull/269

PR #325, opened 2018-09-17
link actions

thedavidmeister:
delivers the action and entry structions from #188


core/src/hash_table/links_entry.rs
```diff
@@ -0,0 +1,204 @@
...
+
+//-------------------------------------------------------------------------------------------------
+// Link
+//-------------------------------------------------------------------------------------------------
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, Hash)]
+pub struct Link {
+    base: HashString,
+    target: HashString,
+    tag: HashString,
```
ddd-mtl: I don't see how tag can be a HashString. Should be normal string, no?

core/src/hash_table/links_entry.rs
```diff
@@ -0,0 +1,204 @@
...
+}
+
+impl Link {
+    pub fn new(base: &str, target: &str, tag: &str) -> Self {
```
ddd-mtl: Can't new() have HashStrings as arguments?

core/src/nucleus/ribosome/api/get_links.rs
```diff
@@ -0,0 +1,75 @@
...
+    // Create GetLinks Action
+    let action_wrapper = ActionWrapper::new(Action::GetLinks(input));
+    // Send Action and block for result
+    let (sender, receiver) = channel();
+    ::instance::dispatch_action_with_observer(
+        &runtime.action_channel,
+        &runtime.observer_channel,
+        action_wrapper.clone(),
+        move |state: &::state::State| {
+            let mut actions_copy = state.agent().actions();
```
lucksus: Why is this dealing with the agent-state?

core/src/agent/state.rs
```diff
@@ -59,6 +60,8 @@ impl AgentState {
 pub enum ActionResponse {
     Commit(Result<Pair, HolochainError>),
     GetEntry(Option<Pair>),
+    GetLinks(Result<Vec<HashString>, HolochainError>),
+    LinkEntries(Result<Entry, HolochainError>),
```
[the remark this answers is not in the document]
ddd-mtl: A link binds two entries, hence `hdk::link_entries()` in the hdk API.
Alternate could be `create_link()`

core/src/hash_table/links_entry.rs
```diff
@@ -0,0 +1,204 @@
...
+    }
+    /// Committing a DnaEntry to source chain should work
+    #[test]
+    fn can_commit_multilink() {
+        // Create Context, Agent, Dna, and Commit AgentIdEntry Action
+        let context = test_context("alex");
+        let link1 = Link::new("12", "34", "fake");
+        let link2 = Link::new("56", "78", "faux");
+        let link3 = Link::new("90", "ab", "fake");
+        let link_entry = LinkListEntry::new(&[link1, link2, link3]);
```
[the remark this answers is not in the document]
ddd-mtl: The LinkEntry in the source-chain is just for one link, but the metadata on the base entry is stored per tag in the hash table since the API gets Links per tag and not per target entry, So the metadata points to a LinkListEntry storing all the links for that tag.

core/src/nucleus/ribosome/api/get_links.rs
zippy: Because there is no other place right now!  This should actually work exactly like how `get()` ends up working with a lookup in local chain/cache before dispatching an action.

core/src/hash_table/links_entry.rs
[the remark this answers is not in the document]
zippy: Just as a record here for people reading the PR.  @ddd-mtl and I talked about how part of the confusion is that we are using the same HT implementation for chain and dht, and "entry" could mean anything you put in an HT.  But, actually we need to distinguish those things and types that are chain-entries because they are what's in the DNA and they are the only things you can link on.  So, though we are adding this PR for now, the LinkListEntry is NOT something that you can commit to the chain, it's just data you get back from a DHT query.

PR #328, opened 2018-09-17
MDBOOK: organize articles better

Connoropolous:
Move things are terms that people will see and question into a glossary section.

Move Pairs article under "Extending Holochain" since its mainly a thing people looking at the source code would care about


Connoropolous:
Noted!

PR #332, opened 2018-09-18
MDBOOK: intro to building apps and intro to DNA

Connoropolous:
This is the beginning of a proposed "Building Apps" structure roughly of the form seen in this workflowy
https://workflowy.com/s/EzcM.WxqP4uJqte

### DIFF
--- a/doc/holochain_101/src/SUMMARY.md
+++ b/doc/holochain_101/src/SUMMARY.md
@@ -5,6 +5,7 @@
   - [How to contribute](./how_to_contribute.md)
 - [Planning a dApp](./planning_a_dapp.md)
 - [Building Holochain Apps](./building_apps.md)
+  - [Introduction to DNA](./intro_to_dna.md)
   - [Using Zome API global variables](./zome/api_globals.md)
   - [Using Zome API functions](./zome/api_functions.md)
   - [Using callback functions](./zome/callback_functions.md)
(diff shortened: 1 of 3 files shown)


Connoropolous:
Hi folks. Per my discussions with Emaline and David Meister, we are going to work on increasing the pace of documentation throughput, with a focus on content development now and QA reviews and detailed passes over the whole book later in the process. 
So highest priority for reviews is for factual correctness. According to that, I'm going to merge, but please feel free to make other suggested edits as review comments on this PR, or just make the updates to the merged version yourself and submit.

thedavidmeister:
@Connoropolous could we use a tag like "rough cut" or "draft" or something to indicate which content should be pushed through?

or is the idea that all content now should be pushed through?

Connoropolous:
I think that still operating from a case-by-case basis is better...

Within this section, there will be a LOT of articles I think, and they are ideally each fairly small, not too complex, and cover something very specific. I had an idea to post them up as PRs through small logical chunks, but all in all I am wanting to do this "section" very quickly as a draft. Get something up, get something in there... 

There will be enough work that is guaranteed for us to have to do in this section once the decisions which are in discussion right now are made and implemented. So I want to try to cover what I assess we've actually landed on, and leave out stuff that's still being determined

thedavidmeister:
@Connoropolous what you're facing isn't that different from the code situation tbh

we're slowly finding our way re: Messy Improvement vs. Perfect Deliverable

PR #334, opened 2018-09-18
ADR: Container API

lucksus:
This is an attempt to specify the Container and what we can do with it.
Other discussions around WebSocket interfaces happening right now in parallel actually should be contextualized with this.

### DIFF
--- a/doc/architecture/decisions/0016-container-api.md
+++ b/doc/architecture/decisions/0016-container-api.md
@@ -1,0 +1,103 @@
+# 16. Container API
+
+Date: 2018-09-17
+
+## Status
+
+Proposed
+
+## Context
+
+* We have several different roles for using a Holochain app (through a *container*):
+  * local, QML based UI components
+  * as a special case of the above: administration UI
+  * externally located (living in a browser) web UIs connected through some form of IPC
+  * *bridged* apps, i.e. Holochain app as the user of another Holochain app
+  * A Holo light-client connecting to the Holochain instance running on a HoloPort full-node
+* With HoloSqape and the Holochain-nodejs, we already have two container implementations
+
+A Holochain app cannot be used directly. Holochain is built as a library that exposes an interface for
+creating Holochain instances and calling their zome functions (see [core_api](/core_api/src/lib.rs)).
+This *core_api* requires its client to provide a context, which holds the representation of the agent
+(name and keys), a logger and persistence and will also include a networking proxy in the future.
+
+While it is possible to use this library directly in a monolithic app where the UI is tightly
+coupled to the app and everything linked to the Holochain library into the same executable, we regard this
+as a special case, since this does not allow for app composability (which we think is a crucial
+concept for the Holochain app ecosystem). 
+
+Instead, in the context of the Rust iteration of Holochain, we have been following the notion of
+providing a *Container* as a relevant concept (and implemention) for the deployment of Holochain apps:
+we provide a Container (i.e. [HoloSqape](https://github.com/holochain/holosqape)) for each supported platform (Linux, MacOS, Windows, iOS, Android)
+that gets installed on a host machine. Holochain apps get installed into the container.
+The Container:  
+ * manages installed hApps (Holochain Apps),
+ * manages agent identities (keys) ,
+ * should also enable hApps to call functions of other hApps - what we call *bridging*,
+ * has to implement access permissions to installed hApps.
+
+So far, the interface our Container implementation provides was growing organically
+in [container.h](https://github.com/holochain/holosqape/blob/master/bindings/container.h).
+
+With upcoming alternative container implementations (based on [Holochain-nodejs](https://github.com/holochain/holochain-nodejs)
+or a custom one for HoloPorts) we should drive the process of building this Container API
+consciously and with more coherence and visibility amongst our separate dev teams.
+
+We need a protocol for communication with a Holochain container and a specification of what upcoming
+containers have to implement, so that apps and UIs can be build against a standardised
+
+## Decision
+
+We establish the **Container as a fundamental module of the Holochain architecture/stack**
+by specifying its **API**, that can be assumed by UI components, Holochain apps (i.e. zome
+functions in the case of bridging)
+and remote processes alike to be implemented by the context a Holochain app is executed in.
+
+
+Fundamental to this API is user roles and permissions.
+Users will be able to use different subsets of the Container's API depending on their specific permissions. 
+
+### Generic API
+
+Every user (of the Container API, i.e. QML root widgets, admin UI in the case of HoloSqape, 
+and network services built on top of the Container) 
+will have these functions available, though functions can return with a permission denied
+error in case an app or capability was given that the current user is not allowed to use.
+
+* `call(app_hash, capability_name, function_name, parameters) -> {results}`
+* `connect(app_hash, capability_name, signal_name, callback_function)`
+* `installed_apps() -> [hash]`
+* `capabilities_of(app_hash) -> [String]`
+* `functions_of(app_hash, capability_name) -> [String]`
+* `request_access(app_hash, capability)`
+* `can_access?(app_hash, capability) -> bool`
+
+
+### Admin
+
+Only the admin user can manage apps and permissions:
+
+* `install_app(Dna) -> hash`
+* `uninstall_app(hash)`
+* `start_app(hash)`
+* `stop_app(hash)`
+* `promote_to_admin(user_id)`
+* `retract_admin(user_id)`
+* `grant_capability(user_id, app_hash, capability)`
+* `deny_capability(user_id, app_hash, capability)`
+
+### Extensible
+More API capabilities might be added in the future.
+
+## Consequences
+
+* We can build several separate pieces independently against the specified API
+* We can build different container implementations for different contexts but have components that
+  rely on a the existence of a container be decoupled from those specific implementations.
+  Components that rely on and could use this API:
+  * any kind of GUI
+  * app composition code like bridging
+  * external web UI
+  * Holo apps, orchastrating hosted apps
+  * Holo admin UI
+  * Holo light-client bridge


ddd-mtl:
Can this instead be a spec like the dna-spec, please?

lucksus:
Why, @ddd-mtl?

lucksus:
I agree that the functions itself should be moved out into their own spec file with much more comments and description, but the decision to actually promote the concept of a Container in this way I perceive to be an architecture decision.

doc/architecture/decisions/0015-container-api.md
```diff
@@ -0,0 +1,103 @@
...
+
+Fundamental to this API is user roles and permissions.
+Users will be able to use different subsets of the Container's API depending on their specific permissions. 
+
+### Generic API
+
+Every user (of the Container API, i.e. QML root widgets, admin UI in the case of HoloSqape, 
+and network services built on top of the Container) 
+will have these functions available, though functions can return with a permission denied
+error in case an app or capability was given that the current user is not allowed to use.
```
neonphog: Just for clarity, are capabilities issued to dpki umbrellas, and "`user`" in this case is one authorized device key under that umbrella?

doc/architecture/decisions/0015-container-api.md
```diff
@@ -0,0 +1,103 @@
...
+
+So far, the interface our Container implementation provides was growing organically
+in [container.h](https://github.com/holochain/holosqape/blob/master/bindings/container.h).
+
+With upcoming alternative container implementations (based on [Holochain-nodejs](https://github.com/holochain/holochain-nodejs)
+or a custom one for HoloPorts) we should drive the process of building this Container API
+consciously and with more coherence and visibility amongst our separate dev teams.
+
+We need a protocol for communication with a Holochain container and a specification of what upcoming
+containers have to implement, so that apps and UIs can be build against a standardised
```
neonphog: sentence trails off

ddd-mtl:
Your first comment literally says 'specify the Container', so its a spec. That and PR name also implies spec and not architecture decision.
I don't understand what the architecture decision is at hand here? Use containers? Not sure how we could do otherwise...

doc/architecture/decisions/0015-container-api.md
```diff
@@ -0,0 +1,103 @@
...
+
+Proposed
+
+## Context
+
+* We have several different roles for using a Holochain app (through a *container*):
+  * local, QML based UI components
+  * as a special case of the above: administration UI
+  * externally located (living in a browser) web UIs connected through some form of IPC
+  * *bridged* apps, i.e. Holochain app as the user of another Holochain app
```
neonphog: Is this suggesting that bridging communications should flow over a websocket channel? Feels like this should be direct function calls. Sorry, I may be conflating the other PR, this may indeed be just proposing a programming interface.

doc/architecture/decisions/0015-container-api.md
```diff
@@ -0,0 +1,103 @@
...
+
+Date: 2018-09-17
+
+## Status
+
+Proposed
+
+## Context
+
+* We have several different roles for using a Holochain app (through a *container*):
```
neonphog: Has the department of names department come up with a better name than "container" that isn't so overloaded with VM implications yet? Suggest we sort that out before merging this.

doc/architecture/decisions/0015-container-api.md
```diff
@@ -0,0 +1,103 @@
...
+as a special case, since this does not allow for app composability (which we think is a crucial
+concept for the Holochain app ecosystem). 
+
+Instead, in the context of the Rust iteration of Holochain, we have been following the notion of
+providing a *Container* as a relevant concept (and implemention) for the deployment of Holochain apps:
+we provide a Container (i.e. [HoloSqape](https://github.com/holochain/holosqape)) for each supported platform (Linux, MacOS, Windows, iOS, Android)
+that gets installed on a host machine. Holochain apps get installed into the container.
+The Container:  
+ * manages installed hApps (Holochain Apps),
+ * manages agent identities (keys) ,
```
neonphog: ... always includes the core Dpki app and through it provides key management / assists with persisting encrypted agent identities?

doc/architecture/decisions/0015-container-api.md
lucksus: Hm, in this case, `user` means user of the container API, which might be:
* some form of local UI (QML root UI in the case of HoloSqape, but could be the Holo hosting app for example)
* another app
* something that is calling in via a network connection, if the contianer is embedded in a service like application

doc/architecture/decisions/0015-container-api.md
lucksus: No, this API is meant as the functions any network service project could use in order to implement a websocket channel, but that would be *setting up a user that is bound to that websocket session* and not in the scope of a container.

The container just provides the means to spawn Holochain instances and make their capabilities available. Bridging would be a use-case of that where that happens within the container, and definitely within the process the container runs in.

it's just that from the perspective of the container (and Holochain apps fwiw) we need the same semantics for all these (quite different) cases.

doc/architecture/decisions/0015-container-api.md
lucksus: Hm, I guess that is right, that would make a lot of sense..

lucksus:
@neonphog, "unless the hosting app has access to the container api and is invoking functions on behalf of the light client user??", yes :) that is what I imagine. And I even see several ways of doing this:
* hosting app could be its own stand-alone app (like HoloSqape compiling to an executable) which links in a Container implementation
* hosting app is (just) a Holochain app. HoloPorts run HoloSqape (or whatever our main Holochain Container/Browser combo is at that point). Holo hosting app is granted admin privileges on that container so that it can install, start, stop apps and authorize websocket sessions (=API `users`) to access certain app's capabilities.

Sure, there is more to it to have web browser find the right HoloPort to connect to in the first place and so on.. But that is something we should be able to put in front of this. Only looking at what happens on the HoloPort that serves an app here..

doc/architecture/decisions/0015-container-api.md
samrose: @lucksus can you explain on a practical level what a "container" is?

samrose: @lucksus is a "container" a chrooted user? is it somehow tied to linux machine users? etc trying to understand how the operating system will know about the "boundaries" if this container

doc/architecture/decisions/0015-container-api.md
```diff
@@ -0,0 +1,103 @@
...
+
+* `install_app(Dna) -> hash`
+* `uninstall_app(hash)`
+* `start_app(hash)`
+* `stop_app(hash)`
+* `promote_to_admin(user_id)`
+* `retract_admin(user_id)`
+* `grant_capability(user_id, app_hash, capability)`
+* `deny_capability(user_id, app_hash, capability)`
+
```
lucksus: + clone(), restore(), backup()

doc/architecture/decisions/0015-container-api.md
samrose: @lucksus answered in the call that "container" == a library

doc/architecture/decisions/0015-container-api.md
```diff
@@ -0,0 +1,103 @@
...
+### Generic API
+
+Every user (of the Container API, i.e. QML root widgets, admin UI in the case of HoloSqape, 
+and network services built on top of the Container) 
+will have these functions available, though functions can return with a permission denied
+error in case an app or capability was given that the current user is not allowed to use.
+
+* `call(app_hash, capability_name, function_name, parameters) -> {results}`
+* `connect(app_hash, capability_name, signal_name, callback_function)`
```
Connoropolous: connect handles event streams then?

doc/architecture/decisions/0015-container-api.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Admin
+
+Only the admin user can manage apps and permissions:
+
+* `install_app(Dna) -> hash`
+* `uninstall_app(hash)`
+* `start_app(hash)`
+* `stop_app(hash)`
+* `promote_to_admin(user_id)`
```
Connoropolous: What is a "user_id" in this context?

doc/architecture/decisions/0015-container-api.md
samrose: I don't think there is an API method yet for event streams?

zippy: Correct, that hasn't been implemented but it has been referred to implicitly here: https://github.com/holochain/holochain-rust/blob/develop/doc/specs/dna.md   search for "signals".  These declared signals then become available for connecting to by the `connect` method.

doc/architecture/decisions/0015-container-api.md
zippy: The idea is that a "user" is who ever is using the container, i.e. not a human, but rather a particular UI, or a particular connection over the websocket connector.  This name will probably have to change as it's confusing.

Connoropolous: Right... I'd recommend "client" then, seems fitting

Connoropolous: I am also thinking out though... where exactly does this ID come from? As in... how does a client get assigned an ID, and where is that ID "registered". This seems an important question to clarify just a little bit more an important aspect of this

doc/architecture/decisions/0015-container-api.md
zippy: I actually like your distinctions from [your HDK doc](https://docs.google.com/document/d/1lFRQxrJqGI97A1S6WGsoL84qlVkP7HVjVF4cVA7zUF8/edit#) of Holochain System (for Container) and Holochain Service (for process that uses the System.

I mean system and service are kind of general words, but they seem to work.

doc/architecture/decisions/0015-container-api.md
lucksus: We actually have an ADR for that, which I've also linked in the context session.
Here again: https://github.com/holochain/holochain-rust/blob/develop/doc/architecture/decisions/0013-signals-listeners-model-and-api.md

doc/architecture/decisions/0015-container-api.md
```diff
@@ -0,0 +1,103 @@
...
+
+A Holochain app cannot be used directly. Holochain is built as a library that exposes an interface for
+creating Holochain instances and calling their zome functions (see [core_api](/core_api/src/lib.rs)).
+This *core_api* requires its client to provide a context, which holds the representation of the agent
+(name and keys), a logger and persistence and will also include a networking proxy in the future.
+
+While it is possible to use this library directly in a monolithic app where the UI is tightly
+coupled to the app and everything linked to the Holochain library into the same executable, we regard this
+as a special case, since this does not allow for app composability (which we think is a crucial
+concept for the Holochain app ecosystem).
```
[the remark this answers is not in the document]
lucksus: This sentence is not talking about UI composability, @ddd-mtl.
If you bundle the UI and Holochain System in an executable as in "use-case 1" (most left column in https://realtimeboard.com/app/board/o9J_kzZtCZI=/), you can't even have app composability because the app is hidden from the user and only usable through one UI. For app composability we need a (standardized) way to call zome functions from other apps (i.e. bridging) and a way for the user to set these up. This is what the Container is about. No reference to UI composability.

But maybe @zippy and @artbrock should give their approval for the importance of composability?

lucksus:
@ddd-mtl, hm, I'm a bit surprised by the amount of misunderstandings ;)

Regarding you saying "this is a spec": I agree that the part where we list functions does have spec quality. As I said earlier, I think we should have a spec where this is fleshed out. I would still keep some functions in in order to disambiguate and communicate more context - but flag it as exactly that: example for context and not binding.

The question of assuming a standardized container API as a thing that always needs to be there around Holochain apps is definitely an architectural decision. Following your definition: it has far reaching consequences in that we might build many other pieces (hApps themselves, services like HoloSqape or the Holo hosting service, ..) differently. And: if we wouldn't accidentally align all those different pieces without an agreed upon API, it would be hard to change this decision afterwards.

doc/architecture/decisions/0015-container-api.md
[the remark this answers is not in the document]
ddd-mtl: The paragraph does mention UI and would not be necessary if you think of app as something without UI.
I guess there is a misunderstanding around two words: 'app', and 'composability'.
I understand app as final end-user facing app with UI, so app composability encompass UI composability.
I understand 'composability' as the capacity to let developers compose an app out of pre-existing holochain-apps (where holochain-app is just DNA, no UI). Scenario 1 does allows that.
I see how there can be a different definition of 'composability ' as the property of letting end-users compose their own app out of pre-existing 'app building blocks'. Scenario 1 indeed does not allow that.

ddd-mtl:
Yes, I generally agree with decision at hand.
For me to approve this, I require that:
 - 'container' is clearly flagged as a temporary word.
 - There is a clear definition of what 'container' is, and that this definition is used properly throughout the ADR.
 - The decision clearly states that this 'container' layer is mandatory for using Holochain in an app.

doc/architecture/decisions/0015-container-api.md
```diff
@@ -0,0 +1,125 @@
...
+  mediating between any kind of user and all installed Holochain apps (here: app = Holochain instance =
+  embodied DNA = *Phenotype*).  The Container is necessarily the only architectural piece that has unique
+  and direct access to the [core_api](core_api/src/lib.rs) and thus the capabilites and zome functions
+  of Holochain apps.
+  
+  **Definition**: Any software module that uses Holochain's [core_api](core_api/src/lib.rs) to instantiate
+  a Holochain instance is called *Container*. It is the necessary layer that *contains* an instance,
+  as [core_api](core_api/src/lib.rs) being a library it (and instances created by it)
+  can not exist as a process themselves.
+
```
lucksus: Do see this being sufficient, @ddd-mtl?

doc/architecture/decisions/0015-container-api.md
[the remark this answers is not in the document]
zippy: So for me this depends on the definition of "crucial"  I do think it's very important for long term power and viability, I'm not so sure it's next step on critical path for adoption.  @artbrock and I keep talking about "UI hints" as something to have inside the Genome that provide semantic intent for different types of UI.  That seems very important to land early(ish) because it signals the intent to composibility.  I agree that the clarity on the term app would help because if you look at it as the very end user experience at the end of all development, then yes it's not so much about composibility.  But if app refers to one of the microservices that may have a UI component also, well that would be pretty crucial to compose with, otherwise how are they microservices?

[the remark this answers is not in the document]
lucksus: Yeah, I was using the latter semantic for composability and I have (so far always) used "app" to mean "Holochain micro-service = instance of one Genome = no UI.

doc/architecture/decisions/0015-container-api.md
ddd-mtl: almost!
By 'software module' you mean a library and/or an executable?
Typo: `capabilites`

doc/architecture/decisions/0015-container-api.md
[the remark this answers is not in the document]
ddd-mtl: So if I heard you right @zippy, you believe the important thing is to signal early that we intent holochain-apps to be UI-composable by developers making end-user facing apps?
@lucksus I understand you used 'app' for meaning 'holochain-app'. This is revealing to me that using terms like 'holochain-xxx' is prone to confusion, because "laziness" will push people to just say 'xxx' and ambiguity will start to form.

doc/architecture/decisions/0015-container-api.md
lucksus: Typo fixed, clarification added in parenthesis, @ddd-mtl.

PR #340, opened 2018-09-19
WIP: Move Pair from HashTable to Chain + renames

ddd-mtl:


core/src/chain/pair.rs
```diff
@@ -115,7 +108,7 @@ pub mod tests {
 
     /// dummy pair
     pub fn test_pair() -> Pair {
-        Pair::new(&test_chain(), &test_entry())
+        test_chain().create_next_pair(&test_entry())
```
zippy: Given that you are renaming most uses of "pair" to "entry" wouldn't it be good to similarly change the names in the tests?  Or is there a reason for keeping it pair?

core/src/chain/mod.rs
```diff
@@ -153,57 +224,83 @@ impl SourceChain for Chain {
...
 
-    fn top_pair_type(&self, t: &str) -> Option<Pair> {
+    fn top_pair_of_type(&self, t: &str) -> Option<Pair> {
         self.iter().find(|p| p.header().entry_type() == t)
     }
 
+    /// Whole process of authoring an entry.
+    /// 1. `validation` of the new entry using the ribosome and validation WASM code
+    /// 2. `pushing` the new entry onto the source chain, if valid
+    /// 3. `putting` the entry into the (distributed) hash table, if defined as public
```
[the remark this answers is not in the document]
ddd-mtl: fixed.
Somehow the comments made it through lol

core/src/hash_table/actor.rs
```diff
@@ -23,51 +23,63 @@ impl HashTable for ActorRef<Protocol> {
         unwrap_to!(response => Protocol::TeardownResult).clone()
     }
 
-    fn put_pair(&mut self, pair: &Pair) -> Result<(), HolochainError> {
-        let response = self.block_on_ask(Protocol::PutPair(pair.clone()));
-        unwrap_to!(response => Protocol::PutPairResult).clone()
+    fn put_entry(&mut self, entry: &Entry) -> Result<(), HolochainError> {
```
zippy: This makes it look like what goes in the hash table is entries, and is part of the confusion that leads there to be non-chain "entry" types and has led to the follow-up ticket on ["dht system entry types"](https://github.com/holochain/holochain-rust/issues/339)  I'm ok with approving this PR because we need to move all of the parts forward, but I'm just calling this out here for us to remember that the word "Entry" I think should be reserved for what can be added to the chain.  In the current code there are "DHT only" entry types for holding groups of links for example.   This will cause confusion and make us think we can commit such items to the chain, which we shouldn't be able to.

ddd-mtl: Ok. I'll add a todo here as a reminder.

Issue #243, opened 2018-08-25
optional Header in Pair


thedavidmeister:
when working with the network we want to get/set by entry hash rather than header because the header is related to the local chain only

for the persistence layer to work for both chain and network we want:

- header should be associated with chain module rather than hash map
- pair should be reworked/removed so that header is optional at hash table level
- clearer apis for when we are get/setting pairs vs. entries

the last point already seems to be happening as an opaque `get` is a bad method name. the trend i see across several PRs is that it is being organically factored out. Rust has conventions/opinions on this https://doc.rust-lang.org/1.0.0/style/style/naming/README.html#getter/setter-methods-[rfc-344]

ddd-mtl:
The convention doc you point to is about accessors, but here it seems more about inserting & retrieving data from a data structure which is different. The convention is more something like `insert/at`, or the `[]` operator.

thedavidmeister:
there's this too https://rust-lang-nursery.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

so maybe we just do `entry()` and `header()` with no `get_`, it is already that way in some places

whatever we choose, i'd like to be consistent if possible, across hash table, network, zomes, chains, actions, etc.

ddd-mtl:
I'm okay with `entry()`

I'm thinking `entry_at()` might be better, because I find it more intuitive that you may not get anything in return. Also we might want a slightly different naming to make it discernible that we are querying a hash table?

But again, I'm okay with `entry()` 👍

thedavidmeister:
if/when we get actors in it might even be fine to merge pair and header... header could potentially store the new chain struct, then it could lookup its own entry

thedavidmeister:
@lucksus do you mean redundant hash table copies on the network or locally?

there's no redundant hash table copies locally. after #135 only actor references are ever copied for a chain. i expect something similar for the network, there would only be a table actor reference to work with entries.

superfluous data across the network is a good point. this issue is about making headers optional for storage. could that address your concerns?

the methods exposed on the `HashTable` trait are flexible. we can expose `header()` and/or `pair()` to facilitate chain iteration and `entry()` to support everything else

lucksus:
I mean copies on the network, @thedavidmeister. The local hash table is a shard of the dht and each entry is stored reduntantly, n-times. I just wanted to point out that it does have negative side-effects if we would put include headers in the dht always. Not how it was meant to be and also different to the Go prototype.

[...]

PR #346, opened 2018-09-20
skeleton for CAS trait

thedavidmeister:
child of #243 

bb071367cc2b1dcc4af2b357acd9f97d

## EAV examples

The EAV trait proposed here uses simple constraints to determine what is returned.

A standard `std::collections::HashSet` of EAVs is always returned. The set may be empty.

### check a meta value exists for some entry

pass the entry and value addresses, as well as the attribute string to `fetch_eav`

```rust
let entry = ...
let attribute = ...
let value = ...

store.fetch_eav(Some(entry.address()), Some(attribute), Some(value.address()));
```

### iterate over all local relationships

leave all the constraints open to get a `HashSet` of all local EAVs.

`HashSet` support `iter()` so any further iteration can be handled from there.

```
store.fetch_eav(None, None, None).iter()
```

### implement linking between entries

use entry as a V, like EAE

```rust
// one to many example, a references both b and c
let entry_a = ...
let attribute = ...
let entry_b = ...
let entry_c = ...

// adds the EAE relationships to the store
store.add(EntityAttributeValue::new(&entry_a.address(), &attribute, &entry_b.address()));
store.add(EntityAttributeValue::new(&entry_a.address(), &attribute, &entry_c.address()));

// retrieve both relationships with EA*
store.fetch(Some(entry_a.address()), Some(attribute), None);
```


core/src/cas/mod.rs
```diff
@@ -0,0 +1,32 @@
...
+/// @see https://en.wikipedia.org/wiki/Content-addressable_storage
+pub trait AddressableContent {
+    fn address(&self) -> Address;
+    fn content(&self) -> Content;
+}
+
+/// content addressable store
+/// implements storage in memory or persistently
+/// anything implementing AddressableContent can be stored and retrieved
+pub trait Store {
```
lucksus: Hm, if we call this Store, I won't rename State to Store too, right?
@ddd-mtl, @thedavidmeister

core/src/cas/mod.rs
```diff
@@ -0,0 +1,32 @@
...
+    /// stores AddressableContent in the Store by its Address as Content
+    fn store(&self, content: &AddressableContent) -> Result<(), HolochainError>;
+    /// true if the Address is in the Store, false otherwise.
+    /// may be more efficient than retrieve depending on the implementation.
+    fn contains(&self, address: &Address) -> Result<bool, HolochainError>;
+    /// returns some Content String if it is in the Store
+    /// note: the original struct/type is NOT restored/deserialized
+    fn retrieve(&self, address: &Address) -> Result<Option<Content>, HolochainError>;
```
ddd-mtl: Is the AddressableContent removed from the store when calling this method?
If not, is the cas immutable?

ddd-mtl:
I like how we could define the DHT as:
```Rust
impl ContentAddressableStorage as HolographicStorage {
...
}
```

core/src/cas/storage.rs
```diff
@@ -0,0 +1,58 @@
...
+            self.storage.insert(content.address(), content.content());
+            Ok(())
+        }
+
+        fn contains(&self, address: &Address) -> Result<bool, HolochainError> {
+            Ok(self.storage.contains_key(address))
+        }
+
+        fn fetch<C: AddressableContent>(&self, address: &Address) -> Result<Option<C>, HolochainError> {
+            Ok(self.storage.get(address).and_then(|c| Some(C::from_content(c))))
```
ddd-mtl: Why not do `get_key_value()` instead of `get()` in `fetch()`

core/src/cas/content.rs
```diff
@@ -0,0 +1,81 @@
...
+            }
+        }
+    }
+
+    #[derive(Debug, PartialEq, Clone)]
+    /// another struct that can be content addressed
+    /// used to show ExampleCas storing multiple types
+    pub struct OtherExampleAddressableContent {
+        content: Content,
+    }
```
ddd-mtl: How about a Content that stores its address as a field and does the hashing in `from_content()`. I looks like a nice thing to try for testing performance.

core/src/cas/eav.rs
```diff
@@ -0,0 +1,173 @@
...
+// type Source ...
+
+#[derive(PartialEq, Eq, Hash, Clone, Debug)]
+pub struct EntityAttributeValue {
+    entity: Entity,
+    attribute: Attribute,
+    value: Value,
+    // index: Index,
+    // source: Source,
+}
```
ddd-mtl: I get and like the idea behind this struct. What I find questionnable is the name and how it is presented. Since value is constrained to an Address and the whole thing is for `AddressableContent`, I think it would clearer to have that reflected in the name. The current name `EntityAttributeValue` implies a general purpose struct but it is not, since its specific.

I would suggest something like `AddressableContentLink`, and have the same field names as `Link` (base, target, tag).
Maybe use `Relation` or `Edge` instead of `Link` to not confuse the two?

core/src/cas/eav.rs
```diff
@@ -0,0 +1,173 @@
...
+                    })
+                    .filter(|eav| {
+                        match value {
+                            Some(ref v) => &eav.value() == v,
+                            None => true,
+                        }
+                    })
+                    .collect::<HashSet<EntityAttributeValue>>();
+            Ok(filtered)
+        }
```
ddd-mtl: Cool, I like this! We might still add some more helper public methods with less arguments that internally calls this. 

For example:
 ```Rust
fn fetch_relations_from(&self, entity: &Entity) -> Result<HashSet<EntityAttributeValue>, HolochainError> {
  self.fetch_eav(Some(entity), None, None)
}
```

Comment (thedavidmeister) [reply core/src/cas/eav.rs]: @ddd-mtl oh yeah for sure, i just thought i'd get the low level constraint logic in place for this PR

IIRC combined CAS + EAV storage affords entire query languages based on constraints

e.g. https://en.wikipedia.org/wiki/Datalog

we don't need anything near that for the developer preview though ;)

File: core/src/cas/eav.rs (hunk 5 above)
Comment (ddd-mtl) [reply core/src/cas/eav.rs]: I like `AddressableContentRelationship`.  You can still leave all the EAV references in the comments.

Comment (thedavidmeister) [reply core/src/cas/eav.rs]: @ddd-mtl looks like EAV is also being used in the network proposal hackmd from @lucksus

Comment (lucksus) [reply core/src/cas/eav.rs]: Hm, I would actually prefer sticking to known vocabulary. And yes, with these traits we are describing the general case, which we then implement for `Links` or `Relations`.

```rust
impl EntityAttributValue for DHT::EntryLink {
...
}
```
no?

core/src/cas/eav.rs
```diff
@@ -0,0 +1,173 @@
...
+    /// eavs are retrieved through constraint based lookups
+    /// @see fetch_eav
+    fn add_eav(&mut self, eav: &EntityAttributeValue) -> Result<(), HolochainError>;
+    /// fetches the set of EntityAttributeValues that match constraints
+    /// None = no constraint
+    /// Some(Entity) = requires the given entity (e.g. all a/v pairs for the entity)
+    /// Some(Attribute) = requires the given attribute (e.g. all links)
+    /// Some(Value) = requires the given value (e.g. all entities referencing an Address)
+    fn fetch_eav(&self, entity: Option<Entity>, attribute: Option<Attribute>, value: Option<Value>) -> Result<HashSet<EntityAttributeValue>, HolochainError>;
+}
```
lucksus: You asked me before, @thedavidmeister, if we would need iterators. Yes we do. @neonphog as added callbacks to our p2p interface draft that enable the network to query complete lists of entries and metas we hold. See https://hackmd.io/_OIEfnTOSzWtW_7sUw6lrg?view.

core/src/cas/eav.rs
ddd-mtl: `EntityAttributValue` is a struct not a trait so your example would't compile ^^'''
I'm not sure if you are mistaken because you might be tired or because the name is indeed misleading and you just confirmed it.

Anyhow if you want to be descriptive to known vocabulary, this is not EAV since V **must** be an address like E, so its more like EAE. In a graph database thats a directed and named Edge / Relationship.

thedavidmeister: @lucksus yes, it's the general case that we can implement for `Links` or `Relations` _and_ `Address` is general enough to accept any value that can implement `AddressableContent`.

but, `EntityAttributeValue` is a struct not a trait!

`attribute` defines the "type". It is currently a `String` (e.g. `"link"`) to keep attributes open, but we could use an enum if we want the allowed values to be closed (e.g. `Attribute::Link`).

here's a contrived example implementing a relationship between `"1"` and `"2"` to show that 1 comes before 2.

```rust
impl AddressableContent for String {
  fn content (&self) -> Content { self.clone() }
  fn from_content (content: &Content) -> Self { content.clone() }
}

let store = ExampleEntityAttributeValueStorage::new();
let 1_before_2 = EntityAttributeValue::new("1".to_string().address(), "comes_before".to_string(), "2".to_string().address());
store.add(1_before_2);
```

thedavidmeister:
@ddd-mtl could you re-review this?

CAS merged so it is now just MVP EAV implementation + examples in tests

core/src/cas/eav.rs
ddd-mtl: Thanks for explanation. I finally get it 👍

ddd-mtl:
Maybe remove WIP prefix if the PR is review worthy.

thedavidmeister:
@ddd-mtl soz, i forgot about the WIP

PR #347, opened 2018-09-20
MDBOOK: remove pairs scratch content

Connoropolous:
pairs is having refactoring done on it, and this article is scratch anyways. 

Trying to clean up the book a bit so viewers of it get a good impression, even if it's incomplete


Connoropolous:
@thedavidmeister easy approve?

PR #348, opened 2018-09-20
MDBOOK: added glossary terms

chrisminnick:
Added the terms from https://developer.holochain.net/Glossary. I wrote an XSL script to convert it from XML (which is used in the Moodle devcamp) to markdown. I used an html definition list in the markdown, but let me know if it would be better to use a different pattern and I'll update the xsl and re-export.


doc/holochain_101/src/glossary.md
```diff
@@ -1 +1,142 @@
...
+   <dd><p>An entity which operates with agency -- meaning a human or a program, which makes
+                    choices outside of whatever code is written into holochain.</p></dd>
+   <dt>Application Engine</dt>
+   <dd><p>(a.k.a. Application Space) (See Nucleus & Ribosome)</p></dd>
+   <dt>Author</dt>
+   <dd><p>The agent who originally created the information on their source chain / author chain.</p></dd>
+   <dt>Base</dt>
+   <dd><p>Provide a hash for the primary resource that you want to define a relation to
+Tag: Provide a string that defines the relationship, and which defines how it should be looked up in the future
+Link: Provide a hash for the secondary resource that you want to associate with the Base
```
thedavidmeister: this isn't a glossary definition, it's a technical manual 😆

doc/holochain_101/src/glossary.md
```diff
@@ -1 +1,142 @@
...
+   <dt>Blacklist</dt>
+   <dd><p>A list of nodes who get blocked for attempting to propagate invalid information.</p></dd>
+   <dt>Blockchain</dt>
+   <dd><p>An architecture used by Bitcoin, Ethereum and many cryptocurrencies to decentralize maintenance of a global ledger for token accounting.</p></dd>
+   <dt>CALM</dt>
+   <dd><p>Consistency as Logical Monotonicity. (see Monotonicity)</p></dd>
+   <dt>CAP Theorem</dt>
+   <dd><p>Consistency-Availability-PartitionTolerance. Often depicted as a triangle, where you can only achieve two of three options. Unfortunately in the real-world, you always have to include 'P' because networks are not 100% reliable. So the question often comes down to are you optimzing for data Consistency or data Availablily (making sure new data is not visible to anyone until its visible to everyone). Holochains are eventually consistent, but our notion of Consistency is not that there is one set of absolutely true data, but rather you know exactly what was asserted by whom and when.</p></dd>
```
thedavidmeister: also see discussion at https://en.wikipedia.org/wiki/CAP_theorem and https://en.wikipedia.org/wiki/PACELC_theorem

P is described as an `if` rather than a constant

doc/holochain_101/src/glossary.md
```diff
@@ -1 +1,142 @@
...
+   <dt>Ceptr</dt>
+   <dd><p>Short for Receptor. The distributed application and communications framework that Holochains are one small part of. See Ceptr.org for more information.</p></dd>
+   <dt>Chain Entry</dt>
+   <dd><p>A new transaction added to your source chain.</p></dd>
+   <dt>Consensus</dt>
+   <dd><p>In the context of decentralized systems, it is the agreement among the nodes about the state of the system.</p></dd>
```
thedavidmeister: also see `Stability` and `Liveness` definitions as per the cardano WP for further discussion

doc/holochain_101/src/glossary.md
```diff
@@ -1 +1,142 @@
...
+   <dt>Countersigning</dt>
+   <dd><p>When all the agents involved with a transaction sign that transaction to each other's source chains.</p></dd>
+   <dt>Cryptographic Proof</dt>
+   <dd><p>Used to determine that data is untampered: Hashes are equal and Signatures match.</p></dd>
```
thedavidmeister: awkward phrasing

hashes/signatures are both examples of proofs but in different contexts

doc/holochain_101/src/glossary.md
```diff
@@ -1 +1,142 @@
...
+   <dt>DHT</dt>
+   <dd><p>Distributed Hash Table: A structure that enables data to be shared across many machines: easily retrieved, untampered proof by hash.</p></dd>
```
thedavidmeister: awkward phrasing here

doc/holochain_101/src/glossary.md
```diff
@@ -1 +1,142 @@
...
+   <dt>Data Positivism</dt>
+   <dd><p>The notion that data has "truth" in and of itself. Holochains are Data Relativistic, recognizing no independent truth, only who said what and when.</p></dd>
+   <dt>Decentralized</dt>
+   <dd><p>The elmination of the need for a central server or authority to maintain system integrity.</p></dd>
+   <dt>Distributed</dt>
+   <dd><p>A system whose integrity is maintained between ALL nodes of the network (see P2P).</p></dd>
+   <dt>End-to-End Encryption</dt>
+   <dd><p>The cryptographic keys used to encrypt and decrypt are stored only at the end points.</p></dd>
+   <dt>Entry</dt>
+   <dd><p>A record in a source chain recorded by the agent who controls that chain.</p></dd>
```
thedavidmeister: and shared with the DHT

doc/holochain_101/src/glossary.md
```diff
@@ -1 +1,142 @@
...
+   <dt>Eventual Consistency</dt>
+   <dd><p>A term in distributed computing which refers to the fact that data may be temporarily out of sync in parts of the network, but will eventually become consistent.</p></dd>
```
thedavidmeister: "out of sync" feels a bit fuzzy

thedavidmeister:
it's not 100% right in the rust world.

some definitions are tied to the go version.

some definitions are still being iterated internally.

overall i'd say it's worth pulling in so we've got a baseline to move forward with.

what do you think @Connoropolous ?

Connoropolous:
closing since I'm not a big fan of this content that we had in the glossary... can maybe draw on it as a resource, when looking again to implement this

PR #353, opened 2018-09-24
One WASM per Zome

ddd-mtl:
*Nico:*

# Change DNA structure so that we have only one WASM module per zome

I'm a bit sad that this change had to result in so many touched lines..
In particular, I had to refactor the callback mechanism because that was too tightly coupled with:
* Action::ExecuteZomeFunction which is meant for something else (= explicit zome function calls from the UI, not entry validations etc.)
* Capabilities, which don't apply to callbacks either

Now callbacks **do not** dispatch an action and observe state to wait for the result but cut to the chase and instantiate a ribosome locally and synchronously. Since callbacks are (currently) called from threads that are spawned from reducers in which their return value ends up in a more high-level action that is being dispatched as the state change (=validation result, initialization result, etc..) this makes perfect sense.

*(Could help to go through commits in sequence)*


lucksus:
In https://github.com/holochain/holochain-rust/pull/353/commits/8477b9581f1ec0e58e3d1d0d2a23c7654b242607 @ddd-mtl changed

screen shot 2018-09-26 at 11 56 25

into 

screen shot 2018-09-26 at 12 24 54

I actually like the pyramid of `match`es (that I've used to replace the old code in https://github.com/holochain/holochain-rust/pull/353/commits/6b784742769b64c9d155c768bbd48bb1ff9a8432) much more. Less code, more DRY, better readable and maintainable.

Think we had this discussion at some point before and my understanding is that `match`es are the idiomatic thing to do in Rust, and with this commit I thought I've really grokked this ;) Thoughts @AshantiMutinta, @thedavidmeister?

ddd-mtl:
@lucksus This is not a fair way to do it like this. But since you want to publicly "fight" about this I'll give a detailed explanation on how infinitely nesting matchs is unmaintainable and unreadable.

core/src/nucleus/ribosome/callback/mod.rs
```diff
@@ -168,6 +201,56 @@ pub fn call(
...
+                    action_channel,
+                    observer_channel,
+                    wasm,
+                    dna.name.clone(),
+                )
+            }
+        }
+    }
+
+    /*
```
thedavidmeister: get rid of this commented out code block

lucksus:
@ddd-mtl, there is no intention to fight at all. I think this difference in style is an interesting question and I would like to learn and improve my writing of idiomatic Rust code. I thought with this block I actually did that. You changed that to something else in the same PR so have a different opinion. I don't see any reason to fight, just a lot of potential to learn more.

thedavidmeister:
@lucksus @ddd-mtl i don't like nested matches because it looks like cyclomatic complexity to me

also i find the logic hard to follow

no idea what The Rust Way is

ddd-mtl:
@lucksus This is more than idiomatic Rust or not. If you look carefully, my refactor has the same number of matches. The 'philosophy' behind this refactor is language-agnostic. I'll write a detailed explanation later, and possibly in another space/format more suitable for the 'learning' intent.

PR #354, opened 2018-09-25
pull cas trait from 243-cas-trait

thedavidmeister:
delivers CAS trait only from #346 

followups:
- mdbook docs for CAS trait #355 

mdbook is listed as a followup because i figured we should wait until the current round of architectural changes lands before trying to document it all

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,6 +54,11 @@ jobs:
         - rm -rf $TRAVIS_BUILD_DIR/target/debug/.fingerprint/test_utils*/
         - rm -rf $TRAVIS_BUILD_DIR/target/debug/incremental/holochain*/
         - rm -rf $TRAVIS_BUILD_DIR/target/debug/incremental/test_utils*/
+        - rm -rf $TRAVIS_BUILD_DIR/target/debug/deps/holochain*
+        - rm -rf $TRAVIS_BUILD_DIR/target/debug/deps/libholochain*
+        - rm -rf $TRAVIS_BUILD_DIR/target/debug/deps/libtest_utils*
+        - rm -rf $TRAVIS_BUILD_DIR/target/debug/libtest_utils*
+        - rm -rf $TRAVIS_BUILD_DIR/target/debug/test_utils*
         - rm -rf $TRAVIS_BUILD_DIR/target/.rustc_info.json
         - rm -rf $HOME/.cargo/registry/index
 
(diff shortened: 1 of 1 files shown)


core/src/cas/content.rs
```diff
@@ -0,0 +1,86 @@
...
+
+/// can be stored as serialized content
+/// the content is the address, there is no "location" like a file system or URL
+/// @see https://en.wikipedia.org/wiki/Content-addressable_storage
+pub trait AddressableContent {
+    /// the Address the Content would be available at once stored in a ContentAddressableStorage
+    /// default implementation is provided as hashing Content with sha256
+    /// the default implementation should cover most use-cases
+    /// it is critical that there are no hash collisions across all stored AddressableContent
+    /// it is recommended to implement an "address space" prefix for address algorithms that don't
```
thedavidmeister: @ddd-mtl does this satisfy your concerns?

i feel `address()` should satisfy the expectation that it be universally unique, or as close to as is practical

there's two ways to achieve this:

- use a hash
- use a unique prefix

core/src/cas/content.rs
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        fn content(&self) -> Content {
+            self.content.clone()
+        }
+
+        fn from_content(content: &Content) -> Self {
+            OtherExampleAddressableContent {
+                content: content.clone(),
+                address: HashString::encode_from_str(&content, Hash::SHA2256)
```
thedavidmeister: @ddd-mtl here we go, this calculates address up front as per https://github.com/holochain/holochain-rust/pull/346/files#r219675042

core/src/cas/content.rs
ddd-mtl: yes

core/src/cas/content.rs
ddd-mtl: Perfect.
Maybe add the todo on not hardcoding the hashing algorithm?

ddd-mtl:
@lucksus I agree with you on that Nico, but I don't remember us stating such a rule and the whole repo needs some re-org anyway, so indeed we could do that later.

lucksus:
I expressed this as the idea I had in the back of my head for the repository structure when we were discussing architecture the other week and also what had me misunderstand the hash_table module as a high-level component when it was not.

PR #359, opened 2018-09-26
CrudStatus as AddressableContent

thedavidmeister:
lands `AddressableContent` implementation for `CrudStatus` from #346 

followup:

- #363 enum vs. bit flags

### DIFF
--- a/core/src/hash_table/status/mod.rs
+++ b/core/src/hash_table/status/mod.rs
@@ -29,6 +29,7 @@ impl ToString for CrudStatus {
             CrudStatus::REJECTED => "2",
             CrudStatus::DELETED => "4",
             CrudStatus::MODIFIED => "8",
+            CrudStatus::LOCKED => "16",
             _ => unreachable!(),
         }.to_string()
     }
@@ -41,6 +42,7 @@ impl<'a> From<&'a String> for CrudStatus {
             "2" => CrudStatus::REJECTED,
             "4" => CrudStatus::DELETED,
             "8" => CrudStatus::MODIFIED,
+            "16" => CrudStatus::LOCKED,
             _ => unreachable!(),
         }
     }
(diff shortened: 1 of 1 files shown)


core/src/hash_table/status/mod.rs
```diff
@@ -17,9 +19,53 @@ bitflags! {
...
+    fn to_string(&self) -> String {
+        // don't do self.bits().to_string() because that spits out values for default() and all()
+        // only explicit statuses are safe as strings
+        // the expectation is that strings will be stored, referenced and parsed
+        match self.to_owned() {
+            CrudStatus::LIVE => "1",
+            CrudStatus::REJECTED => "2",
+            CrudStatus::DELETED => "4",
+            CrudStatus::MODIFIED => "8",
+            CrudStatus::LOCKED => "16",
```
lucksus: Hm, so the address of a CrudStatus will be the hash of those strings: `"1"`, ...
Could potential collisions be a problem? Would it be better to have those strings be more specific?
`"HcCrudStatus:1"`

And an implicit assumption behind this match is that those states are disjunct. Why use powers of 2 then?

thedavidmeister: @lucksus collisions are harmless because the type system knows what to do on the way out

```rustlet foo: FooAddressableContent = store.fetch("xyz");

thedavidmeister: @lucksus the states might just end up in an enum as we discussed the other night, i'll add a followup for that

lucksus: I was rather thinking about managing (validating) the change of CrudStatus for entries in the DHT where a DHT node has to check if a new meta that is linking to a hash which represents a specific status is valid or not...

thedavidmeister: @lucksus as discussed on chat, this is what the A is for, to differentiate between relationships used in different contexts, `"1"` could be an entry or it could be a status but that's just abstractions we put on top of the string

i _think_ the combination of the type system + attribute names should give us enough context to use the right abstractions at the right time

PR #366, opened 2018-09-26
rename push_entry and push_pair to commit_entry and commit_pair

thedavidmeister:
lands renaming `push_entry` and `push_pair` to `commit_entry` and `commit_pair` from #269


ddd-mtl:
David extracted this is from 188-linking. I'm going to revert that in the 188-linking directly.

PR #370, opened 2018-09-26
361 memory CAS implementation

AshantiMutinta:
This is implementing the CAS memory storage with a hashmap

### DIFF
--- a/core/src/cas/memory/mod.rs
+++ b/core/src/cas/memory/mod.rs
@@ -49,7 +49,7 @@ pub mod tests {
 
     #[test]
     fn memory_round_trip() {
-        let mut cas = MemoryAddressableStorage::new();
+        let mut cas = MemoryStorage::new();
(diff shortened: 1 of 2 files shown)


core/src/cas/memory/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+        assert_eq!(
+            Ok(None),
+            cas.fetch::<ExampleAddressableContent>(&content.address())
+        );
+        assert_eq!(Ok(false), cas.contains(&other_content.address()));
+        assert_eq!(
+            Ok(None),
+            cas.fetch::<OtherExampleAddressableContent>(&other_content.address())
+        );
+        // round trip some AddressableContent through the FilesystemStorage
```
ddd-mtl: Change comment to `MemoryStorage`

AshantiMutinta: Woops! Thank you, didn't catch that. Thank you!

PR #371, opened 2018-09-26
implements addressable_content for HashString

thedavidmeister:
lands `AddressableContent` for `HashString` from #346

### DIFF
--- a/core/src/cas/memory/mod.rs
+++ b/core/src/cas/memory/mod.rs
@@ -10,7 +10,7 @@ pub struct MemoryStorage {
 }
 
 impl MemoryStorage {
-    fn new() -> MemoryStorage {
+    pub fn new() -> MemoryStorage {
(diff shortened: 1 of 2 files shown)


thedavidmeister:
@lucksus to reference an `Address` as a value

lucksus:
You mean as value in an EAV? So that we can link to any other address..? But wouldn’t we want the address itself as value then? I understand this code to take the HashString content and use the default implementation in AddressableContent to calculate the hash of this string holding a hash. So the value would be the hash of the hash.. right? That is not what we want, or do we?

So I think misses an overwriting implementation for ˋaddress()ˋ which returns the hash directly.

thedavidmeister:
@lucksus this might be premature, i can pull it out of the upstream branch

pretty easy to drop back in if we ever need it

PR #382, opened 2018-09-27
ADR 11 - Build HCadmin app with Qt/C++

lucksus:
This was treated as if accepted. https://github.com/holochain/holosqape is that.

### DIFF
--- a/doc/architecture/decisions/0011-build-hcadmin-app-with-qt-c.md
+++ b/doc/architecture/decisions/0011-build-hcadmin-app-with-qt-c.md
@@ -4,7 +4,7 @@ Date: 2018-05-16
 
 ## Status
 
-proposed
+Accepted
 
 ## Context
 


doc/architecture/decisions/0011-build-hcadmin-app-with-qt-c.md
```diff
@@ -4,7 +4,7 @@ Date: 2018-05-16
 
 ## Status
 
-proposed
```
zippy: I would say we should actually mark this as "obsoleted" by the Container ADR, which is a generalization of this ADR from an "architecture" point of view.  It doesn't mean we don't build holosqape, it just means that architecturally we now recognize that there are many ways to "admin" i.e. deploy your holochain instances which depend on the container type and the use-case type.

zippy:
closing this one in favor of continuing discussion on #389

PR #384, opened 2018-09-27
ADR 13 - Signal/Listener model and API

lucksus:


zippy:
This is an easy approve because so much is left in the consequences, and we know we want to do this.

PR #385, opened 2018-09-27
ADR 14 - P2P - IPC Abstraction

lucksus:


doc/architecture/decisions/0014-p2p-ipc-abstraction.md
```diff
@@ -4,7 +4,7 @@ Date: 2018-07-19
 
 ## Status
 
-Proposed
+Accepted
```
thedavidmeister: @lucksus @neonphog did we come to a conclusion on what lives where?

lucksus: Yes, in broad strokes still but we agree on the important pieces. I‘ve posted two links to hackmds on this (old) issue: 
https://github.com/holochain/holochain-rust/issues/99

But I would say that this discussion (and the agreed interface) goes beyond this ADR anyway.

thedavidmeister: @lucksus this ADR 

- proposes moving logic out of rust into a totally different programming language
- flags separation of concerns as a key benefit
- flags work overhead and complexity a a key drawback

isn't it a dependency that we first agree what is being separated before we can agree that it must be separated?

Issue #350, opened 2018-09-24
Refactor: (almost) pure Reducers


lucksus:

lucksus:
child of #349

Redux needs reducers to be pure functions.
We started without action creators.
Now we introduced action creators with the Future refactoring, and should also move side-effects out of the reducers into those action creators.

"almost" because some state was extracted into actors (chain) which the reducers still need to mutate.

PR #390, opened 2018-09-27
[REFACTOR] Channels: now sync and in context

lucksus:
preparation for #350 (pure reducer refactoring)

Sync channels (https://doc.rust-lang.org/std/sync/mpsc/fn.sync_channel.html) can easily be passed around between threads (which I need for #350).
Since I had to touch every line that mentions channels (or their `Senders`) to change to `SyncSender` and since we have the `context` object holding the state and being passed to everything, I've put those channels into the context.

### DIFF
--- a/core/src/context.rs
+++ b/core/src/context.rs
@@ -1,11 +1,14 @@
 use action::ActionWrapper;
 use error::HolochainError;
 use holochain_agent::Agent;
-use logger::Logger;
 use instance::Observer;
+use logger::Logger;
 use persister::Persister;
 use state::State;
-use std::sync::{Arc, mpsc::{sync_channel, SyncSender}, Mutex, RwLock, RwLockReadGuard};
+use std::sync::{
+    mpsc::{sync_channel, SyncSender},
+    Arc, Mutex, RwLock, RwLockReadGuard,
+};
 
 /// Context holds the components that parts of a Holochain instance need in order to operate.
 /// This includes components that are injected from the outside like logger and persister
@@ -18,7 +21,7 @@ pub struct Context {
     pub persister: Arc<Mutex<Persister>>,
     state: Option<Arc<RwLock<State>>>,
     pub action_channel: SyncSender<ActionWrapper>,
-    pub observer_channel: SyncSender<Observer>
+    pub observer_channel: SyncSender<Observer>,
 }
 
 impl Context {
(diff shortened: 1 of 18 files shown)


core/src/context.rs
```diff
@@ -23,11 +30,32 @@ impl Context {
         logger: Arc<Mutex<Logger>>,
         persister: Arc<Mutex<Persister>>,
     ) -> Context {
+        let (tx_action, _) = sync_channel(100);
+        let (tx_observer, _) = sync_channel(100);
```
ddd-mtl: Magic number: Please use a global/struct const instead and possibly add a comment explaining the rationale behind that number.

core/src/instance.rs
```diff
@@ -79,8 +79,8 @@ impl Instance {
 
     /// Returns recievers for actions and observers that get added to this instance
     fn initialize_channels(&mut self) -> (Receiver<ActionWrapper>, Receiver<Observer>) {
-        let (tx_action, rx_action) = channel::<ActionWrapper>();
-        let (tx_observer, rx_observer) = channel::<Observer>();
+        let (tx_action, rx_action) = sync_channel::<ActionWrapper>(100);
+        let (tx_observer, rx_observer) = sync_channel::<Observer>(100);
```
ddd-mtl: magic number

ddd-mtl:
@zippy Yeah I guess `Context` changed from "context of the Holochain instance" to "context of code execution", so the purpose changed but the name still works.

thedavidmeister:
@ddd-mtl @lucksus i'll see if i can clean up the magic number before my plane leaves...

lucksus:
Thanks for fixing the magic numbers, @thedavidmeister!

core/src/context.rs
```diff
@@ -15,19 +20,46 @@ pub struct Context {
...
     pub persister: Arc<Mutex<Persister>>,
     state: Option<Arc<RwLock<State>>>,
+    pub action_channel: SyncSender<ActionWrapper>,
+    pub observer_channel: SyncSender<Observer>,
 }
 
 impl Context {
+    pub fn default_channel_buffer_size() -> usize {
+        100
+    }
```
sphinxc0re: Please make this a constant

thedavidmeister: @sphinxc0re doesn't the rust compiler optimise this kind of thing away under the hood?

ddd-mtl: Yes, I believe so, but I agree with @sphinxc0re that a `const` would be clearer, more scan-able.
I don't see the advantage of this form. I would be happy to learn.

PR #393, opened 2018-09-28
251-DHT Mock

ddd-mtl:
New DHT state-slice with the upcoming refactors in mind.
Network is placeholder stuff for now.
Linking code will be added in a follow-up PR.

This state-slice is expected to reduce the following actions:
- Action::Commit(_, _) 
-  Action::GetEntry(_) 
-  Action::AddLink(_)
- Action::GetLinks(_)


core/src/dht/dht_reducers.rs
```diff
@@ -0,0 +1,144 @@
...
+}
+
+//
+pub(crate) fn reduce_add_link<CAS, EAVS>(
+    _context: Arc<Context>,
+    _old_store: &DhtStore<CAS, EAVS>,
+    _action_wrapper: &ActionWrapper,
+) -> Option<DhtStore<CAS, EAVS>>
+where
+    CAS: ContentAddressableStorage + Sized + Clone + PartialEq,
```
thedavidmeister: seems like a good candidate for a type alias

core/src/dht/dht_reducers.rs
```diff
@@ -0,0 +1,144 @@
...
+    if res.is_err() {
+        return None;
+    }
+    // ...and publish to the network
+    new_store.network_mut().publish(entry);
+    Some(new_store)
+}
+
+//
+pub(crate) fn reduce_get_entry<CAS, EAVS>(
```
zippy: I thought that GetEntry only needed to be in the reducer if we know that we don't have it already, i.e. that we do the request for getting an entry directly from the state, and then it's only an action if we didn't find it locally because that action of getting it from the network is precisely the state that needs to be managed.  i.e. that there's more like a RequestEntryFromNetwork action which the reducer uses then to store the value so retrieved.

core/src/dht/dht_reducers.rs
ddd-mtl: I tried:
```Rust
pub type ConcreteCas = ContentAddressableStorage + Sized + Clone + PartialEq;
```
but it gives me `type aliases cannot be used for traits`

And I tried this:
```Rust
pub trait ConcreteCas: ContentAddressableStorage + Sized + Clone + PartialEq {}
impl ConcreteCas for ContentAddressableStorage {}
impl ConcreteCas for Sized {}
impl ConcreteCas for Clone {}
impl ConcreteCas for PartialEq {}
```
But it also gives me lots of errors.
So can we leave that as a todo for later?

core/src/dht/dht_reducers.rs
ddd-mtl: Yes. So I guess we just need to rename the `GetEntry` Action to `GetEntryFromNetwork`, right @lucksus ?

core/src/dht/dht_reducers.rs
```diff
@@ -0,0 +1,144 @@
...
+    // Otherwise add it local storage...
+    let mut new_store = (*old_store).clone();
+    let res = new_store.content_storage_mut().add(entry);
+    if res.is_err() {
+        return None;
+    }
+    // ...and publish to the network
+    new_store.network_mut().publish(entry);
+    Some(new_store)
+}
```
lucksus: What is missing here is checking if the entry's type is a public one. If it is private it should not be published.

core/src/dht/dht_reducers.rs
lucksus: I would agree to rename it to something like that, or maybe LookupEntryInDht.
I think the reducer should be named this way already. You could also rename the action.

What is out of the scope of this PR is changing the agent/chain reducer for GetEntry and instead write an action creator..

core/src/dht/dht_reducers.rs
lucksus: `context.state().unwrap().nucleus().dna()`
and https://github.com/holochain/holochain-rust/blob/246a83026dacc406e06465cb1fe7a389d5e88cb0/dna/src/lib.rs#L238
might help

ddd-mtl: There is actually a lot of code needed for that to work.
I will add a todo and do a follow-up PR, okay?

PR #397, opened 2018-09-29
spike: use CAS in HashTable

thedavidmeister:
followups:

- #398 combine address and key
- #399 reuse path safety logic from `FileTable`
- #400 use the filesystem for instances


ddd-mtl:
Wondering if we still need FileTable and MemoryTable? Can't we just do:HashTable<CAS

thedavidmeister:
@ddd-mtl yeah man, that makes sense to me, actually i'm more worried about the serialization/deserialization of entries atm

will make a branch to isolate my concerns re: type handling

ddd-mtl:
@thedavidmeister yeah ok, makes sense. 👍

thedavidmeister:
hash table is gone

Issue #396, opened 2018-09-29
use CAS for hash_table storage


thedavidmeister:
this is the start of factoring functionality away from HashTable

PR #401, opened 2018-09-30
entry as addressable content

thedavidmeister:
lands entry addressable content from #396 

obviously needed to put entries in CAS

## changes

- rename entry.content to entry.value
  - avoid name conflicts with `content()` getter and `content()` from `AddressableContent`
- __entry type is used in entry.address__
  - this is to prevent address collisions for different entry types but same value in CAS round trips
  - without this multiple calls to `.add()` will make subsequent calls to `.fetch()` ambiguous with data loss
  - @zippy @lucksus i forgot why entry type is not included in the CAS?


ddd-mtl:
There is a typo in your PR description: You put `entry.address` instead of `entry.value` I believe.
My understanding is that entry_type is already stored in the CAS with the Header, so having it also in the Entry is redundant. When getting from the DHT, you should have contextual knowledge of what type that enrty is.

thedavidmeister:
@ddd-mtl i fixed the typo thanks

yes it is in the header but:

- if it is not on `Entry` struct then things like `entry.is_sys()` won't work
- if it is on the struct but not output by `content()` then CAS round trips will fail
- if it is on the struct and output by content but ignored by `address()` (this is how `key()` works atm) then CAS `.add()` calls result in data loss and/or ambiguous `.fetch()` calls

ddd-mtl:
Hmm yeah I see the problem.
The tricky thing is that we don't put Entries in the DHT. Entry is for the source chain only. The DHT only stores `entry.value`. So when getting a value from the DHT, the caller should be able to reconstruct the `Entry` with contextual knowledge.
I guess it's the same thing for the source chain, we don't want to store `entry_type` twice.
I see `Entry` as just a wrapper around `entry.value` with extra info to make the processing of `entry.value` easier around the codebase. Maybe rename `Entry` to `EntryWrapper` ?
If you look at the [source chain diagram](https://holochain.org/assets/images/holochain/private_source_chain_diagram.png), an entry does not store its type.
Definietly need input from @lucksus and @zippy on this.

ddd-mtl:
After this discussion, it seems to me now that `Entry` should not have `AddressableContent` trait, but `entry.value` should be the `AddressableContent`. That would also mean we don't need to rename `entry.content` to `entry.value`

lucksus:
Hm, I was about to agree with @ddd-mtl's last statement, but then Entry's AddressableContent implementation returns only the content/value as content, which results in only that being used to calculate the hash - which I understand to be correct. I'd suggest to talk about this tomorrow and basically have those trait implementations be driven by the DHT implementation..

thedavidmeister:
here's a WIP glossary:

- "chain link" or `ChainLink` = legacy `Header` = holds `Address` values and some meta like entry type
- "entry" or `Entry` = primary workhorse string that we build hApps around
- "entry link" = EAE references

core/src/hash_table/entry.rs
```diff
@@ -14,7 +15,7 @@ use std::{
 /// data is stored as a JSON string
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Entry {
```
lucksus: We just agreed on using the newtype pattern here and change this to
```pub struct Entry(String);```
https://doc.rust-lang.org/1.0.0/style/features/types/newtype.html

**Comment by thedavidmeister** [comment]
@ddd-mtl @lucksus so here's where i'm at

yes, we want to change `Entry` to a newtype but in my spike at https://github.com/holochain/holochain-rust/pull/425/files i discovered the need for a backing CAS to finish that refactor

getting errors like:

```
   --> core/src/chain/mod.rs:181:36
    |
181 |         self.table_actor.put_entry(&chain_header)?;
    |                                    ^^^^^^^^^^^^^ expected struct `hash_table::entry::Entry`, found struct `chain::header::ChainHeader`
    |
    = note: expected type `&hash_table::entry::Entry`
               found type `&chain::header::ChainHeader`
```

can we merge this branch knowing that `Entry` will be simplified in the near future?

Issue #387, opened 2018-09-27
wabt-sys build error


murrayjbrown:
Build fails compiling _wabt-sys:_

> error: failed to run custom build command for `wabt-sys v0.1.4`

The _wabt_ crate dependency in _test_utils/Cargo.toml_ appears to be an out-dated version (0.2.0), as compared with _core/Cargo.toml_ (version 0.4).

maackle:
fixed by #404, right?

PR #405, opened 2018-10-01
52 chain pair entry content boxing

murrayjbrown:
Resubmitted as a separate pull-request (was previously mistakenly bundled with fix for #387).
FYI @thedavidmeister


murrayjbrown:
@thedavidmeister Thanks for the heads-up. I've been working on DNA API test cases for hdk-rust.

PR #416, opened 2018-10-01
Initialize Application workflow with pure reducer

lucksus:
This is part of the #350 pure reducer refactoring

[x] Add action creator `initialize_application()` that returns future
[x] move side-effects from reducer over there
[x] make tests use new action creator with `block_on()`
[x] make core_api use action creator with `block_on()`
[x] Use Result to return initialization errors
[ ] implement timeout for future to make one last test in core_api green


lucksus:
I have problems trying to use something like https://github.com/alexcrichton/futures-timer to easily put a timeout on my InitializationFuture.

The tests are failing because there is one test that checks if core_api times out if initialization does not work. I was tempted to remove that test.

I believe `futures-timer` does not work because InitializationFuture is returning a Box<dyn Future<..>>, which I need because `initialize_application()` can also return a leaf-future.

Would love to get some advice from @AshantiMutinta :)

core/src/nucleus/actions/initialize.rs
```diff
@@ -0,0 +1,113 @@
...
+///
+/// Returns a future that resolves to an Ok(NucleusStatus) or an Err(String) which carries either
+/// the Dna error or errors from the genesis callback.
+///
+/// Use futures::executor::block_on to wait for an initialized instance.
+pub fn initialize_application(
+    dna: Dna,
+    context: Arc<Context>,
+) -> Box<dyn Future<Item = NucleusStatus, Error = String>> {
+    if context.state().unwrap().nucleus().status != NucleusStatus::New {
```
ddd-mtl: `NucleusStatus::InitializationFailed` should also work here.
Maybe we could add a method `can_initialize()` to the NucleusStatus enum.

AshantiMutinta: My regular question in these reviews hehe. But will this panic? is it meant to panic?

core/src/nucleus/actions/initialize.rs
```diff
@@ -0,0 +1,113 @@
...
+    thread::spawn(move || {
+        let action_wrapper = ActionWrapper::new(Action::InitApplication(dna.clone()));
+        dispatch_action_and_wait(
+            &context_clone.action_channel,
+            &context_clone.observer_channel,
+            action_wrapper.clone(),
+        );
+
+        // Create Commit Action for Genesis Entry
+        commit_entry(
```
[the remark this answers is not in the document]
lucksus: Do we need to block_on that future? I replaced the code in the reducer that was also just dispatching an action and not waiting for it to be processed..

core/src/nucleus/actions/initialize.rs
lucksus: Yep

lucksus: If this function was called with an uninitialized context, i.e. one that does not have a state, I'd say yes that should panic since it's not a user error but something that should bother a dev to fix it.

AshantiMutinta: That works for me and makes a lot of sense. Nice to see panic being unsed in the context library design.

core/src/nucleus/actions/initialize.rs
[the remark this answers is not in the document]
lucksus: Hm, I think the old code was too permissive. We should actually be prepared here to handle invalid DNA and let the validation fail. Will change that here and add a test for that..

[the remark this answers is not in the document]
lucksus: Hmm, ToEntry for DNA always creates an entry from a Dna struct. So with our current code, I can't create a Dna object that would not be able to be commited. So that is why this was ok to not wait for the commit.. but I will push the code that does wait, but without a test that tests it..

PR #417, opened 2018-10-02
MDBOOK: show status in chapter outline

Connoropolous:
I want people to know right away about the status of the book, when they come upon it

I also want them to know, at a glance, which articles to look into and which to ignore


emalinus:
Just "E" or complete? Might we also use an "IP" for in progress too?

Connoropolous:
@emalinus yeah, could be good to have another status option too... 

how might we qualify 'in progress'? and I'd recommend just (I) since space in the table of contents is precious

emalinus:
We could just say "'I' indicates that an article is in progress." I was gonna re-word some of the welcome stuff anyway, so I can make the change

Connoropolous:
The only hesitation I have about an (I) for in progress is that even the
"complete" articles are very much a work in progress so I would worry it'd
be meaningless, or make ones that didn't have an I claim to be "complete".

emalinus:
Ah, good point. Let's just leave it as you suggested, then, and trust that folks will be understanding of our progress :)

PR #419, opened 2018-10-02
wasm_utils split

ddd-mtl:
Just moved code around:
 - Moved some code from `wasm_utils/lib.rs` to `wasm_utils/memory_allocation.rs`
 - Moved some code from `wasm_utils/lib.rs` to `wasm_utils/memory_serialization.rs`
 - Moved SinglePageManager from Nucleus to Ribosome, because thats the only place where it is used and should be used.

Only code change is in `memory_allocation.rs` where I use:
`const U16_MAX: u32 = <u16>::max_value() as u32;`
instead of `std::u16::MAX as u32`

Also added a comment for wasm_utils crate.


Connoropolous:
@ddd-mtl not sure if this was in your awareness while making these changes, but these changes affect crates which import wasm_utils, such as hdk-rust, so its tests are failing, along with libs that depend on it.

It'd be handy to consider cross referencing a breaking change like this in the places it affects, if we can help it

ddd-mtl:
@Connoropolous Yes, once we have a first released version of hdk-rust, we need to up our game on stability. What do you mean by "cross referencing a breaking change like this in the places it affects", how does this take form?

Connoropolous:
Ah, I mean mentioning to people, and or in places (like repos, or channels) that an event has taken place

ddd-mtl:
Okay. I can't really know who is depending on this. Can you follow a folder in github? Maybe we could have some kind of tag or mention we could add in the PR that notifies people downstream? For now, how about I mention you directly everytime I submit a breaking PR?

Connoropolous:
Please yes to @ mentioning me about breaking PRs... 

We can't really know with external deps, but I thought that amongst us we (should have) a pretty good sense of how things depend on other things. Like with wasm_utils for example, since it's been well discussed

PR #420, opened 2018-10-02
Round Trip Test Added To Test Suite

AshantiMutinta:
Added round trip functionality to crudstatus. This is nice because it can be reused for a number of future tests.


core/src/cas/content.rs
```diff
@@ -128,6 +131,25 @@ pub mod tests {
                 other_addressable_content.address()
             );
         }
+
+        pub fn addressalbe_content_round_trip<T, K>(contents: Vec<T>, mut cas: K)
```
thedavidmeister: typo!

Issue #343, opened 2018-09-19
MDBOOK QA checklist


Connoropolous:
- [ ] add links in and between everything possible
- [ ] ensure all links are valid
- [ ] add licensing information
- [ ] add links to pages to feedback forms and how to contribute

PR #422, opened 2018-10-02
MDBook: Update welcome.md

emalinus:
Changed some language, like emphasizing "user-participants" over "users", and hApps over dApps (in this way also made a swift, hopefully non-distracting nod at blockchain)


doc/holochain_101/src/welcome.md
```diff
@@ -1,12 +1,24 @@
...
+Holochain is an open source software library that provides a way for businesses, communities,
+and other groups to build and run applications which are hosted and validated by the "users" themselves.
+Doing so provides a superior level of agency and autonomy over heavy reliance on the so-called "cloud" and other third parties.
+
+These applications are known more widely as peer-to-peer, decentralized applications, or dApps. To distinguish between dApps built on Blockchains and those built on Holochain, we preemptively call the latter "hApps". A more detailed comparison between Blockchain dApps and Holochain hApps is available [here](https://medium.com/holochain/beyond-blockchain-simple-scalable-cryptocurrencies-1eb7aebac6ae).
```
lucksus: @ddd-mtl has started this document that we have worked on together in the core team: https://docs.google.com/document/d/1lFRQxrJqGI97A1S6WGsoL84qlVkP7HVjVF4cVA7zUF8/

We started to define hApps and dApps there as both something you can do with Holochain. Please have a look. We can still discuss these semantics. I think it makes sense to understand dApps not materially defined by blockchain but by decentralised. It would make sense to introduce a new term, hApps, for something that enables users to blend and combine UI components with Genomes (Holochain networks and micro-services..)

Just putting this here as FYI and a request for us all to get coherent on these terms.

emalinus: @lucksus big thanks! Please share any other docs like this, as it'll make it waaay easier for folks like @pdaoust and I to be effective here!

Connoropolous: @emalinus do you think that this merits making any changes to the text? I have a feeling best thing to do right now is merge, and set an intention and 'follow-ups' to refine the language/terms in the book

emalinus: Sure - there are other additions. Please do merge and of course we'll follow up and refine. The doc Nico sent will provide a lot of direction for later passes.

Connoropolous: added this to #343  mdbook QA checklist

PR #425, opened 2018-10-02
spike: 421 entry no type

thedavidmeister:


core/src/agent/state.rs
```diff
@@ -96,14 +98,18 @@ fn reduce_commit_entry(
...
+    let (entry_type, entry) = match action {
+        Action::Commit(entry_type, entry) => (entry_type, entry),
+        _ => unreachable!(),
+    };
 
     // @TODO validation dispatch should go here rather than upstream in invoke_commit
     // @see https://github.com/holochain/holochain-rust/issues/256
 
-    let res = state.chain.push_entry(&entry);
+    // @TODO assumes entry already added to CAS
```
ddd-mtl: Shouldn't you also push entry in CAS here?
When/where would you expect the entry to be pushed in CAS otherwise?

core/src/nucleus/mod.rs
```diff
@@ -403,24 +404,24 @@ fn reduce_execute_zome_function(
     );
 }
 
-/// Reduce ValidateEntry Action
-/// Validate an Entry by calling its validation function
+/// Reduce ValidatePair Action
+/// Validate a Pair by calling its validation function
```
ddd-mtl: Since we are removing pair, maybe rename this `ValidateEntryHeader`? (and remove all remaining references to pair)

ddd-mtl:
I'm a bit lost here. Its a spike but its a request to merge into develop...
So are we suppose to not review it as "normal" code?
Is it some kind of github version of a bundle, were we validate all the PRs at the end or something? :)

thedavidmeister:
@ddd-mtl a spike is a branch on github that can be used to guide/prototype development in some way but doesn't need to land :)

the comment you left is basically where i stopped prototyping because i had what i needed in this case

other branches are going up that are polished and reviewer-friendly

ddd-mtl:
Okay, thanks for clarifying. So I guess there is an agreement that for a spike we should put "SPIKE:" at the beginning of PR name.

PR #430, opened 2018-10-03
428 memory storage eav

AshantiMutinta:
Implemented Memory EAV. Only note is that you can't really make generic lambdas in rust so I was forced to create a bounded function for the filter but it turned out okay. Also moved a eav into a module because I thought it made sense.


core/src/eav/memory/mod.rs
```diff
@@ -0,0 +1,212 @@
+use eav::{Attribute, Entity, EntityAttributeValue, EntityAttributeValueStorage, Value};
+use error::HolochainError;
+use std::collections::HashSet;
+
+pub struct MemoryStorageEav {
```
ddd-mtl: How about renaming this to `EavMemoryStorage` ?
that way we can rename the current `MemoryStorage` to `ContentMemoryStorage` and have consistent naming.

core/src/eav/memory/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+    fn fetch_eav(
+        &self,
+        entity: Option<Entity>,
+        attribute: Option<Attribute>,
+        value: Option<Value>,
+    ) -> Result<HashSet<EntityAttributeValue>, HolochainError> {
+        Ok(self
+            .eavs
+            .iter()
+            .cloned()
```
lucksus: Is this needed?

core/src/eav/memory/mod.rs
AshantiMutinta: That would be perfect! I will make those changes asap

core/src/eav/memory/mod.rs
lucksus: Right, but doesn't the collect at the end create a new HashSet since it's not into_iter()?

AshantiMutinta: Yes absolutely but not of type EntityAttributeValue they would be of &EntityAttributeValue :(

PR #441, opened 2018-10-05
FilesystemStorageActor a.k.a. thread safety for file CAS

thedavidmeister:
i believe this _should_ make the file backed CAS safe to use across threads

there is an actor under the hood
the actor reference re-uses existing actor reference paths in riker if the same dir is called
the actor reference is the new internal state of `FilesystemStorage` rather than the path

### DIFF
--- a/core/src/cas/file/actor.rs
+++ b/core/src/cas/file/actor.rs
@@ -7,6 +7,12 @@ use std::{
     path::{Path, MAIN_SEPARATOR},
 };
 
+const ACTOR_ID_ROOT: &'static str = "filesystem_storage_actor";
+
+fn path_to_actor_id(dir_path: &str) -> String {
+    format!("{}{}{}", ACTOR_ID_ROOT, MAIN_SEPARATOR, dir_path)
+}
+
 pub struct FilesystemStorageActor {
     /// path to the directory where content will be saved to disk
     dir_path: String,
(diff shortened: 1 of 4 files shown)


core/src/cas/file/actor.rs
```diff
@@ -0,0 +1,131 @@
...
+        Ok(write(self.address_to_path(address), content)?)
+    }
+
+    /// filesystem CAS contains. NOT thread safe.
+    fn unsafe_contains(&self, address: &Address) -> Result<bool, HolochainError> {
+        Ok(Path::new(&self.address_to_path(address)).is_file())
+    }
+
+    /// filesystem CAS fetch. NOT thread safe.
+    fn unsafe_fetch(&self, address: &Address) -> Result<Option<Content>, HolochainError> {
```
AshantiMutinta: I was afraid we were using unsafe rust and my eyes popped out haha. I think the naming gives the impression that we are using a subsection of rust that behaves in an unsafe context https://doc.rust-lang.org/book/second-edition/ch19-01-unsafe-rust.html

thedavidmeister: @AshantiMutinta that's fair. would you accept `unthreadable_fetch`?

AshantiMutinta: unthreadable is good but then again we have to ask what properties does this have which make them so unthreadable. I guess it is the lack of the ARC for the context? In that case I think that is a good name. Just highlighting the question of, if I had to put this in a thread::spawn, what would stop me? Is it unthreadable or unthreaded?

ddd-mtl: Random suggestion `single_threaded_fetch()` or  `st_fetch()`

PR #442, opened 2018-10-05
Validation of App Entries

lucksus:
This is part of the validation code needed for https://github.com/holochain/app-spec-rust/pull/19.
Dependency for https://github.com/holochain/hdk-rust/pull/15.

* new callback`validate_entry()` calls into WASM `validate_<entry-type-name>` instead of `validate_commit` and passes the entry and `ValidationData` as parameters (both as JSON in on Value)
* got rid of callback `validate_commit`
* added copies of some types to wasm_utils/src/validation.rs so I can access and deserialize those [in the HDK](https://github.com/holochain/hdk-rust/pull/15/files#diff-e1576943a5907b15388ab726037ebdb3R95)
* ValidationData is build in `invoke_commit`, but not yet filled with data because that needs a refactor first

This is not complete but I want to keep the PR clean and start a separate refactor branch and continue after that was merge in


wasm_utils/src/validation.rs
```diff
@@ -0,0 +1,97 @@
...
+    Commit,
+    Delete,
+}
+
+//#[derive(Serialize, Deserialize)]
+//#[serde(remote = "Either")]
+//pub enum EitherDef<L, R> {
+//    Left(L),
+//    Right(R),
+//}
```
ddd-mtl: dead code

PR #443, opened 2018-10-05
440 dht follow-up: Publish only public entries

ddd-mtl:
In `reduce_commit_entry()`, check if entry is "publishable" before publishing on the network.
system entries are handled differently, since, at least for now, we do not publish any system entry.

Added `Sharing::can_publish()` and `Dna::get_entry_type_def()`.
Added methods to EntryType, and took the occasion to move it in `holochain_dna` crate.


core/src/dht/dht_reducers.rs
```diff
@@ -52,9 +53,68 @@ where
...
+        .expect("Must have DNA to commit an app entry.");
+    let maybe_def = dna.get_entry_type_def(&entry_type.to_string());
+    if maybe_def.is_none() {
+        return None;
+    }
+    let entry_type_def = maybe_def.unwrap();
+
+    // Add it to local storage...
+    let mut new_store = (*old_store).clone();
+    let res = new_store.content_storage_mut().add(entry);
```
lucksus: Hm, I'm not sure if we should the entry to the local DHT shard at all if it is not public..
@zippy, what to do you say?

dna/src/entry_type.rs
```diff
@@ -0,0 +1,135 @@
+use std::{
```
lucksus: Yeah, really good to move this out of sys_entry. I did that in a local spike yesterday when starting to entangle our core types into a new crate for the HDK.

But why did you put EntryType into the DNA crate?
I don't think it belongs in here any more as it belongs into hash_table/sys_entry..

I would suggest moving it into a new crate. I created a `core_types` crate.

core/src/dht/dht_reducers.rs
zippy: That's what I implemented in the go version, i.e. that if it's private that the app developer knows they need to get from the chain, and there's a flag in the `get_entry` call for that case, that looks for the hash in the chain only.   I think this probably matters because if it gets into the DHT shard in the Holo case it may mean that the data actually travels outside of a security boundary that it shouldn't.

dna/src/entry_type.rs
ddd-mtl: I moved it in holochain_dna, because I want to use `EntryType::has_valid_app_name()` in that crate and did want to create a dependency to holochain_core. 
A new `core_types` crate might make sense but it feels really out of scope for this PR.
I also would like us to have clarity on the purpose and dependency of a new crate before making one.

core/src/dht/dht_reducers.rs
ddd-mtl: I'm a bit lost here. What needs changing for this to get approved and merged in?

lucksus: If the entry is not public it should not even be stored in the DHT shard. So these two lines should go down into the block guarded by the `if entry_type_def.sharing.clone().can_publish()`.

ddd-mtl: Okay. And can system entries be published ?

lucksus: Yes. All system entries get published except for the Dna, @ddd-mtl.

ddd-mtl: Thanks! This is the sort of info that I think belongs to a spec.

ddd-mtl: #spec

PR #449, opened 2018-10-06
ChainStore and removal of legacy chain

thedavidmeister:
standardising on the pattern established by `DhtStore`

## changes

- removed chain module
- moved `ChainHeader` to `agent::chain_header`
- refactored `ChainHeader` to be a standalone struct (no dependency on chain state)
- added `ChainStore` modelled off `DhtStore`
- implemented `.iter()` and `ChainStoreIterator`
  - `iter()` takes a `ChainHeader` and returns an iterator of all chain headers from that point
- implemented `iter_type()` and `ChainStoreTypeIterator`
  - `iter_type()` takes a `ChainHeader` and `EntryType` and returns an iterator of all chain headers of the passed type from that point
- both implemented iterators are deterministic based on the passed `ChainHeader`, assuming the CAS is populated
- updated agent reducer to use CAS instead of hash_table methods
- remove actor protocol variants that are no longer used
- did NOT reimplement methods like `.entry()` from chain
  - these methods did very little on top of the `.iter()`
  - e.g. `.entry()` is just `chain_store.iter(&Some(chain_header)).find(|chain_header| chain_header.entry_address() == entry.address())`
  - these methods aren't even being used atm


core/src/cas/memory/actor.rs
```diff
@@ -0,0 +1,90 @@
+use actor::{Protocol, SYS};
```
zippy: clean up looks like branch conflict

core/src/cas/memory/mod.rs
```diff
@@ -0,0 +1,75 @@
+use cas::{
```
zippy: same

PR #450, opened 2018-10-07
fixed bug where wasm serialization does not return error value

zippy:
I was unable to successfully write a test for this case, but the problem was that when an error occurred in the deserializing, it was putting the value of the memory, not the serde error into the returned Err()


zippy:
added a link to a ticket re the test

ddd-mtl:
This might fix a bug but it breaks the feature of retrieving error strings :/

PR #452, opened 2018-10-07
[REFACTOR] Extraction of core_types and cas_implementations

lucksus:
For validations I need several types in an extra crate to be able to use them in the HDK and thus in the Ribosome without creating a dependency to all the Holochain code.

I've started pulling the thread and ended up with this big spring-cleaning.
I did not change any code, only moved it.
I've commented out tests of ChainActor and ChainHeader, because @thedavidmeister is refactoring those anyway.

new crates:
## `core_types`
* does not depend on any other holochain crate
* contains:
  * Entry, EntryType, EntryMeta
  * error, ToJson, ToEntry,...
  * abstract CAS and EAV traits

## `cas_implementations`
I didn't want to put the CAS/EAV implementations in the `core_types` crate, so I've create this one.

### DIFF
--- a/cas_implementations/src/lib.rs
+++ b/cas_implementations/src/lib.rs
@@ -1,9 +1,9 @@
-/// CAS Implementations
-///
-/// (CAS == Content Addressable Storage)
-///
-/// This crate contains implementations for the CAS and EAV traits
-/// which are defined but not implemented in the core_types crate.
+//! CAS Implementations
+//!
+//! (CAS == Content Addressable Storage)
+//!
+//! This crate contains implementations for the CAS and EAV traits
+//! which are defined but not implemented in the core_types crate.
 
 extern crate futures;
 extern crate holochain_core_types;
--- a/core_types/src/lib.rs
+++ b/core_types/src/lib.rs
@@ -1,12 +1,12 @@
-/// Holochain Core Types
-///
-/// The reason for having this crate is to have a minimal but
-/// complete set of types that are used in most other Holochain
-/// crates, but that don't include Holochain itself.
-///
-/// Note: This is already quite big. Maybe break the CAS and EAV traits
-/// out into their separate crate as well since those are generic and not
-/// necessarily bound to Holochain.
+//! Holochain Core Types
+//!
+//! The reason for having this crate is to have a minimal but
+//! complete set of types that are used in most other Holochain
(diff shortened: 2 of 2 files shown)


cas_implementations/src/lib.rs
```diff
@@ -0,0 +1,14 @@
+extern crate futures;
```
ddd-mtl: Missing top comment explaining crate's purpose and expected use &dependencies.

core_types/src/lib.rs
```diff
@@ -0,0 +1,25 @@
+extern crate futures;
```
ddd-mtl: Same here

zippy:
Hmmm I actually kind of like @ddd-mtl suggestions, but this speaks to the bigger issue I have with hdk-rust being a separate repo right now and having a whole separate `wasm-test` directory with yet another test application in there for it's own integration tests.   I'd like to add this issue of crate & repo setup as a team discussion for this next week.

lucksus:
@ddd-mtl 

`core_types` is not the same as `wasm_utils`. Most of the stuff in there does not have anything to do with wasm. But it is a minimal complete set of types that we can extract from core without missing anything. It is included in (almost?) every other crate.

`cas_implementations` was inside the cas/eav directory in core. I did not want to move that together with the cas traits into core_types, since core_types was getting much bigger than I hoped. Since I had to disentangle it anyway I could provide the benefit of that work to future-us by putting it in another crate. That enables exactly what you're saying: creating another crate that uses that without including holochain. The only reason to not do that I would say is the 3h of mechnic stupid work it takes. But I did that last night.

lucksus:
core_types is much more than what a non-rust HDK would have to implement.
Which means, we could actually divide those types even more but that would involve more semantic changes. With this I was trying to only move but not change code.

lucksus:
I've added the crate descriptions, @ddd-mtl. Please go ahead and merge if you approve.

ddd-mtl:
I had to change comment formats so rustfmt doesn't whine.

@lucksus Thanks for explanation, its getting clearer for me, especially `cas_implementations`. 
For `core_types`, I think this needs more clarifying and thinking through as @zippy suggests, but I'm fine with merging this for now, since I'm counting on more code movements latter this week.

PR #460, opened 2018-10-10
Implemented file validation

AshantiMutinta:


core_types/src/file_validation.rs
```diff
@@ -0,0 +1,15 @@
+use error::HolochainError;
+use std::path::Path;
+
+pub fn validate_canoncical_path(dir_path: &str) -> Result<String, HolochainError> {
```
Connoropolous: Spelling error, `canoncical` -> `canonical` :)

PR #462, opened 2018-10-10
456 eav test suite

AshantiMutinta:
Implemented test suite for EAV


cas_implementations/src/eav/file/mod.rs
```diff
@@ -159,171 +159,28 @@ pub mod tests {
...
-            // open
-            (None, None, None),
-        ] {
-            println!("fetch");
-            assert_eq!(
-                expected,
-                eav_storage.fetch_eav(e, a, v).expect("could not fetch eav"),
-            );
-        }
+        EAVTestSuite::test_round_trip_test(
```
ddd-mtl: Pascal case acronyms please `EavTestSuite`

AshantiMutinta: Changing asap.

core_types/src/cas/storage.rs
```diff
@@ -231,6 +235,180 @@ where
     }
 }
 
+pub struct EAVTestSuite;
+
+impl EAVTestSuite {
+    pub fn test_round_trip_test(
```
lucksus: `test_rount_trip_test` sounds like we are testing the round_trip_test.. ;)

thedavidmeister: @lucksus yeah so i've been informally following a convention where `test_*` means "generate a test example thing" and `*_test` "a test for a thing"

PR #465, opened 2018-10-11
add entry type to entries

thedavidmeister:
## reference

puts entry type back in entries as discussed in https://hackmd.io/D65QC7fOQbe286K8FQuz_A

## changes

- ensure `agent` crate uses same serde as `core`
  - needed to derive `Deserialize`
  - needed for consistency in output JSON and therefore address hashes with core
- simplified `Agent` and `Identity` to use newtype pattern
  - nested structs were adding round trip complexity that isn't justified at this point
  - we can always add structs in if we find this is needed later
- replace ad-hoc `from_string()` fn with core `ToString` and `From<String>` for `Agent` and `Identity`
- put `EntryType` in `Entry` as `entry_type`
- put `Entry` content data in `Entry` as `value`
- disambiguate entry content and value everywhere
- entry content is the `AddressableContent` content as `Content`
  - i.e. JSON including the `entry_type` and `value`
- added tests for `Agent` and `Identity`
- everywhere that took an `(EntryType, Entry)` tuple now takes `Entry`
  - includes action enums
- cleans up `ChainHeader` a bit
  - new type for `Iso8601`
  - new type for `Signature`
  - fn sig should make more sense
- DhtStore `get` returns an `Option<Content>` now
  - previously was only `Content`
  - meant that empty string was returned, which cannot work as JSON
- reinstated/fixed/added tests for links


thedavidmeister:
@artbrock this is where the new data structures spec should go ;)

please include:

- what structs do we need? `Entry`, `ChainHeader`, ... ?
- what data is on each struct?
  - `Entry` => `value`, `entry_type`
  - `ChainHeader` => ...
  - ...
- what address generation logic overrides might we need?
  - `entry.address()` returns itself if `EntryType` is `AgentId`?
- anything else that seems relevant in the moment :)

thedavidmeister:
@lucksus FYI ^^

core_types/src/links_entry.rs
```diff
@@ -99,16 +121,142 @@ impl LinkListEntry {
...
     }
+
+    pub fn links(&self) -> &Vec<Link> {
+        &self.links
+    }
+}
+
+impl ToString for LinkListEntry {
+    fn to_string(&self) -> String {
+        serde_json::to_string(self).expect("LinkEntry failed to serialize")
```
ddd-mtl: Error message should be "LinkListEntry"

core_types/src/links_entry.rs
```diff
@@ -99,16 +121,142 @@ impl LinkListEntry {
...
+    }
+}
+
+impl From<String> for LinkListEntry {
+    fn from(s: String) -> LinkListEntry {
+        serde_json::from_str(&s).expect("LinkEntry failed to deserialize")
```
ddd-mtl: same here

core/src/nucleus/ribosome/api/commit.rs
```diff
@@ -15,7 +15,7 @@ use wasmi::{RuntimeArgs, RuntimeValue, Trap};
 #[derive(Deserialize, Default, Debug, Serialize)]
 struct CommitAppEntryArgs {
     entry_type_name: String,
-    entry_content: String,
+    entry_value: String,
```
lucksus: This is one of those changes that breaks the HDK if the copy of this struct on the HDK side is not changed in the same way!

core_types/src/entry.rs
```diff
@@ -9,7 +12,25 @@ use std::ops::Deref;
 /// Structure holding actual data in a source chain "Item"
 /// data is stored as a JSON string
 #[derive(Clone, Debug, Serialize, Deserialize)]
-pub struct Entry(String);
+pub struct Entry {
+    value: Content,
```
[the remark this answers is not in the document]
ddd-mtl: I think its because it will implement AddressableContent trait which has `content()`, so we can't have a private field named content otherwise you would have to break convention for getter name.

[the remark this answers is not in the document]
thedavidmeister: @lucksus @ddd-mtl yes exactly.

core_types/src/entry.rs
```diff
@@ -19,29 +40,33 @@ impl PartialEq for Entry {
...
 impl AddressableContent for Entry {
     fn content(&self) -> Content {
-        String::from(self.to_owned())
+        self.to_json()
+            .expect("could not convert Entry to Json Content")
     }
 
     fn from_content(content: &Content) -> Self {
-        Entry::from(content.to_string())
+        Entry::from_json(&content.to_string()).expect("could not convert Json Content to Entry")
```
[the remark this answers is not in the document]
thedavidmeister: @lucksus yes they must be because there is only one entry, it is `Entry`. everything "entry like" is actually just stuffed into value.

in the future `Entry` will be a trait rather than a struct, at which point this would be more meaningful.

core_types/src/links_entry.rs
thedavidmeister: @ddd-mtl please make edits like this directly when you see them

PR #466, opened 2018-10-11
MDBOOK: add built with holochain page

Connoropolous:
this is for the sake of dev camp

we are going to have people link to their projects, and in doing so, make (and learn about making) a contribution to the MDBOOK


thedavidmeister:
cool idea

would like to see how we can make this more active/visible in the future :)

PR #476, opened 2018-10-14
core_types that compile to WASM

lucksus:
The recent problems with the HDK basically boiled down to inconsistent Rust structs used for serializing/deserializing between native/WASM. We need to make sure that (at least for HDK-rust) we use the exact same types. Since validation the HDK needs also Entry, ChainHeader and all the dependencies that come with it.

For that reason I separated out those types (and their dependencies) into `core_types`.

Yesterday @zippy discovered (by trying to use it that way) that `core_types` does not compile in the HDK. It did not compile for target WASM.

That is what this PR fixes.

The culprits were `Walkdir` and `riker`. I've moved both to `cas_implementations`.
That solves this issue for now in the context of the HDK.

But we will have to solve that problem with riker as it will prevent us from compiling Holochain core to WASM which we need for Holo as discussed last week. Riker does not compile to WASM because it depends on config -> serde_hjson -> regex -> memchr, and the old version in memchr kills the build. There is a new version since January, but the latest serde_hsjon version is 2 years old. There is a PR up to fix this old dependency and I've commented on it as well: https://github.com/hjson/hjson-rust/pull/7. Hope they merge that soon.


cas_implementations/src/error.rs
```diff
@@ -0,0 +1,14 @@
...
+    fn from(error: WalkdirError) -> Self {
+        // adapted from https://docs.rs/walkdir/2.2.5/walkdir/struct.Error.html#example
+        let path = error.path().unwrap_or(Path::new("")).display();
+        let reason = match error.io_error() {
+            Some(inner) => reason_for_io_error(inner),
+            None => String::new(),
+        };
+        HolochainError::IoError(format!("error at path: {}, reason: {}", path, reason))
+    }
+}
```
Connoropolous: new line

PR #478, opened 2018-10-14
spike: Entry enum

thedavidmeister:


core_types/src/entry/entry_type.rs
```diff
@@ -14,17 +14,18 @@ macro_rules! sys_prefix {
...
+
     ChainHeader,
-    Key,
-    Link,
-    Migration,
-    /// TODO #339 - This is different kind of SystemEntry for the DHT only.
-    /// Should be moved into a different enum for DHT entry types.
-    LinkList,
+    ChainMigrate,
 }
```
ddd-mtl: How about separating this into two enums:

```Rust
enum EntryType {
 App(String),
 System(SystemEntryType),
}

enum SystemEntryType {
 AgentId,
 Dna,
 Link,
...
}
```

Because I believe there will be many places where we handle a SystemEntryType and don't want to make sure its not an AppEntry everytime.

thedavidmeister: @ddd-mtl yes we're doing an enum based approach :)

core_types/src/entry/agent.rs
```diff
@@ -0,0 +1,13 @@
+use keys::Key;
+
+pub struct Agent {
```
ddd-mtl: I believe this should be renamed `AgentId`, as to not be confused with the module Agent which is the redux state-slice.

thedavidmeister:
@ddd-mtl sort of? we're trying some idea out atm that @lucksus suggested.

watch this space.

PR #479, opened 2018-10-14
Extract Zome API serialization structs -> wasm_utils

lucksus:
Moving all structs that are used for serializing/deserializing arguments of API functions into `wasm_utils` so that HDK-rust can use those instead of having copies of those types which breaks things if only one side gets changed.

## api_serialization:
[x] CommitEntryArgs
[x] CommitEntryResult
[x] GetEntryArgs
[x] GetEntryResult

## validation
[x] ValidationData

I had to do quite some yak shaving to get here. To have the result code in `invoke_commit_entry` not confuse myself I refactored the futures to return almost the right things already.

I also switched the wasm-test projects to release to make use of the already configured compiler optimizations. Reduces binaries from ~9MB to ~500kB and initialization delays from ~20s to <1s for some tests. (Note: we have no advantage from building our wasm zome code with debug info since we have nothing in our ribosome that would use that)

This branch is needed to make https://github.com/holochain/hdk-rust/pull/32 work.


wasm_utils/src/api_serialization/mod.rs
```diff
@@ -0,0 +1,13 @@
...
+///
+/// When these types get changed their counter parts in all HDKs
+/// have to change with them! Otherwise we get deserialization
+/// errors in the ribosome.
+///
+/// For the case of HDK-rust we can use the exact same types by
+/// importing this module.
+
+pub mod commit;
+pub mod validation;
```
Connoropolous: newline

wasm_utils/src/api_serialization/validation.rs
```diff
@@ -0,0 +1,37 @@
...
+    Commit,
+    Modify,
+    Delete,
+}
+
+#[derive(Clone, Serialize, Deserialize)]
+pub enum LinkAction {
+    Commit,
+    Delete,
+}
```
Connoropolous: newline

core/src/agent/actions/commit.rs
```diff
@@ -33,13 +33,13 @@ pub struct CommitFuture {
 }
 
 impl Future for CommitFuture {
-    type Item = ActionResponse;
-    type Error = String;
+    type Item = Address;
+    type Error = HolochainError;
```
lucksus: Added `HolochainError::ValidationFailed` below. Much clearer interface for `CommitFuture`

core/src/nucleus/ribosome/api/commit.rs
```diff
@@ -27,20 +23,20 @@ fn build_validation_data_commit(
     // TODO: populate validation data with with chain content
     // I have left this out because filling the valiation data with
     // chain headers and entries does not work as long as ValidationData
-    // is defined with the type copies i've put in wasm_utils/src/validation.rs.
+    // is defined with the type copies i've put in wasm_utils/src/api_serialization/validation.rs.
```
Connoropolous: Avoid the word "i" and "I've" since the reader may not have any context as to who

core/src/nucleus/ribosome/api/commit.rs
```diff
@@ -27,20 +23,20 @@ fn build_validation_data_commit(
...
     // Doing this right requires a refactoring in which I extract all these types
```
Connoropolous: Again "I"

core_api/wasm-test/commit/Cargo.toml
```diff
@@ -9,11 +9,13 @@ crate-type = ["cdylib"]
...
 [workspace]
 members = []
 
 [dependencies]
 serde="1"
 serde_derive="1"
-holochain_wasm_utils = { path = "../../../wasm_utils"}
\ No newline at end of file
+holochain_wasm_utils = { path = "../../../wasm_utils"}
+holochain_core_types = { path = "../../../core_types"}
```
Connoropolous: newline

wasm_utils/src/api_serialization/commit.rs
```diff
@@ -0,0 +1,29 @@
+use holochain_core_types::{cas::content::Address, hash::HashString};
+
+/// Struct for input data received when Commit API function is invoked
+#[derive(Deserialize, Default, Debug, Serialize)]
+pub struct CommitEntryArgs {
+    pub entry_type_name: String,
+    pub entry_value: String,
+}
+#[derive(Deserialize, Serialize, Default, Debug)]
```
Connoropolous: Let's add a /// comment for this one too

wasm_utils/src/api_serialization/get_entry.rs
```diff
@@ -0,0 +1,34 @@
+use holochain_core_types::cas::content::Address;
+
+#[derive(Deserialize, Default, Debug, Serialize)]
```
Connoropolous: Let's add /// comments for these structs

wasm_utils/src/api_serialization/get_entry.rs
```diff
@@ -0,0 +1,34 @@
...
+        }
+    }
+
+    pub fn not_found() -> GetEntryResult {
+        GetEntryResult {
+            status: GetResultStatus::NotFound,
+            entry: String::from(""),
+        }
+    }
+}
```
Connoropolous: This is overall looovely

wasm_utils/src/api_serialization/validation.rs
```diff
@@ -0,0 +1,34 @@
+use holochain_core_types::{chain_header::ChainHeader, hash::HashString};
+
```
Connoropolous: Maybe there will be another LinkValidationData for links? Since ValidationData doesn't seem to include LinkAction?

wasm_utils/src/validation.rs
```diff
@@ -1,97 +0,0 @@
...
-//
-// These are copies of the same types else where.
-// We need to make these types available to the HDK
-// without having the HDK depend on the whole Holochain
-// library.
-// The same is true for many more types going forward.
-// I wanted to keep the PR small and have us merge
-// the basic validation code first.
-// I will post another PR soon that moves all needed types
-// into a new crate and will like remove this file again.
```
Connoropolous: Good to see this go

wasm_utils/src/api_serialization/mod.rs
```diff
@@ -0,0 +1,13 @@
+/// This module holds structs for all arguments and return types
+/// that get serialized and deserialized between core native and
```
Connoropolous: I guess it's true that this encapsulates stuff that gets serialized and deserialized, but we do also have the RibosomeErrorCode stuff which is another really important definition for the interface across wasm.

core/src/nucleus/ribosome/api/commit.rs
```diff
@@ -83,29 +79,26 @@ pub fn invoke_commit_app_entry(
     );
 
     let maybe_json = match task_result {
-        Ok(action_response) => match action_response {
-            ActionResponse::Commit(_) => action_response.to_json(),
-            _ => return ribosome_error_code!(ReceivedWrongActionResult),
-        },
+        Ok(address) => serde_json::to_string(&CommitEntryResult::success(address)),
+        Err(HolochainError::ValidationFailed(fail_string)) => {
+            serde_json::to_string(&CommitEntryResult::failure(fail_string))
```
Connoropolous: So... We didn't have a way before to pass back the validation failure reason?

wasm_utils/src/api_serialization/commit.rs
```diff
@@ -0,0 +1,29 @@
...
+pub struct CommitEntryResult {
```
ddd-mtl: Name is a bit misleading as its not a rust Result.

wasm_utils/src/api_serialization/commit.rs
```diff
@@ -0,0 +1,29 @@
...
+        }
+    }
+
+    pub fn failure(validation_failure: String) -> CommitEntryResult {
+        CommitEntryResult {
+            address: HashString::from(""),
+            validation_failure,
+        }
+    }
+}
```
ddd-mtl: I believe there is a better way to do this:
 - On success: Return the CommitEntryOutput struct. 
 - On failure: Return a ErrorOutput struct.

 On wasm side, if deserializing of CommitEntryOutput fails, then deserialize an ErrorOutput.

wasm_utils/src/api_serialization/get_entry.rs
ddd-mtl: Seems overkill for a boolean

wasm_utils/src/api_serialization/commit.rs
Connoropolous: I prefer that too

core/src/nucleus/ribosome/api/commit.rs
lucksus: `git blame` gives that context :)

wasm_utils/src/api_serialization/validation.rs
lucksus: Yep, there might be another one for links..

core/src/nucleus/ribosome/api/commit.rs
lucksus: Yes, we did, but it was handled exactly like an error below. The tests in HDK and app-spec were tailored for that. This is cleaner and more robust.

wasm_utils/src/api_serialization/commit.rs
lucksus: I thought about that but decided not to since I would either have to roll my own `try_deserialize!()` in `hdk::commit_entry()` or put commit/validation concern into the general `try_deserialize!()`. A fallback deserialization retry feels adequate for error handling, but a validation failure is not an error and specific to commit.

I struggled a bit with the implicit complexity we have here already. Given that we expect other HDK implementations to be added in the future, I'd rather keep this bridge simple and explicit rather then complex and implicit.

core/src/nucleus/ribosome/api/commit.rs
Connoropolous: I think you might feel differently if someone else started scattering the word "I" around the code comments. It's just an extra step, and not everyone knows how to use git blame, and those people could still be reading the source code. I really differ here

wasm_utils/src/api_serialization/get_entry.rs
lucksus: Well, the reason why I'm hesitant to put Rust specific Options or Results in here is because this is the go-to place for future HDK implementors to understand what JSON structures they have to parse and create.. and they might not know how and Option gets de-/serialized by serde..

core/src/nucleus/ribosome/api/commit.rs
lucksus: Using "I" in code comments creates an informal tone that IMO conveys that this meant by a developer for a developer. It is not meant as a code comment to stay there or to be used in docs. Hence the // instead of /// and all of it framed under TODO. I deliberately used this style here in that way.

Also, it is not changed in this PR. But I will take this comment block out within the coming few days when I'll work on the todo.

wasm_utils/src/api_serialization/get_entry.rs
zippy: I take this all back.  Lets leave the status as is for now (as an enum), because we may use it for added info about CRUD status.  This is fine for MVP.

PR #480, opened 2018-10-14
Update README.md

Connoropolous:
The README should be clearer that to install holochain-rust, you DO NOT follow any instructions from this repo itself, to avoid confusion.

the makefile doesn't work for mac... it does an apt-get without a check. This should be fixed, but until then...
created an issue #481 

Also, make changes back away from 'Genome' language, as per https://github.com/holochain/holochain-rust/pull/474/files


README.md
```diff
@@ -105,9 +107,6 @@ If your changes do not break the current environment, you can submit a separate
 
 Otherwise, you will need to speak to an admin who can force merge your full changes after testing locally.
 
-### Building for Android
-Note there is an article written for how to build Holochain for Android, read it [here](doc/holochain_101/src/holochain_across_platforms.md).
-
```
lucksus: Why remove this?

Connoropolous: The reason, at least for me, is because it feels a little arbitrary to put this in the readme, but not other info, that's in the book. We already link to the book, and 'building for android' is in the chapter structure, so I don't know exactly why we'd include it here

Connoropolous:
K! I will switch back to having that part in!

Connoropolous:
@lucksus updated to readd the android thing ,please change approve

Issue #232, opened 2018-08-20
Implement Dna Hash


ddd-mtl:
Hash of dna should be done and made available in core lib.

PR #483, opened 2018-10-15
Implement DNA hash Global

Connoropolous:
screen shot 2018-10-14 at 11 46 49 pm

#232


ddd-mtl:
Once the entry refactor is done. Dna should implement the Entry trait which is AddressableContent, so hash should be retrieved by just callind address()

Connoropolous:
Yes, nice. We can leave it wip till then... But check out the update just now too

PR #485, opened 2018-10-16
470 file storage passed in context.

AshantiMutinta:
Passing the File Storage in EAV and using it around the core.


core/src/dht/dht_store.rs
```diff
@@ -63,8 +63,9 @@ where
     // Linking
     // =======
     pub fn add_link(&mut self, _link: &Link) -> Result<(), HolochainError> {
-        // FIXME
-        Err(HolochainError::NotImplemented)
+        let link_eav = EntityAttributeValue::new(&_link.base(), &_link.tag(), &_link.target());
+        Ok(self.meta_storage.add_eav(&link_eav)?)
+        // Err(HolochainError::NotImplemented)
```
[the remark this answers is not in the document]
AshantiMutinta: Awesome, Will revert

Connoropolous:
So excited this is happening! 

Is it still WIP @AshantiMutinta or ready for another review?

AshantiMutinta:
> 
> 
> So excited this is happening!
> 
> Is it still WIP @AshantiMutinta or ready for another review?

yay!! Just a few more things and it should be fine! so still WIP for now 💃

thedavidmeister:
@AshantiMutinta not sure if this is already on your mind, but can we put the actual path in and make this all live?

something like `$HOME/.hc/storage/...` would work for me

core/src/agent/state.rs
```diff
@@ -134,20 +134,22 @@ fn reduce_commit_entry(
...
     ) -> Result<Address, HolochainError> {
-        state.chain.content_storage().add(entry)?;
-        state.chain.content_storage().add(chain_header)?;
+        let mut res_context = context.clone();
+        res_context.file_storage.add(entry)?;
+        res_context.file_storage.add(chain_header)?;
         Ok(entry.address())
     }
-    let res = response(state, &entry, &chain_header);
+    let result = response(&*context, state, &entry, &chain_header);
```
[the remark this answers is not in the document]
AshantiMutinta: Reference of the arc dereference :)

core/src/context.rs
```diff
@@ -10,6 +10,13 @@ use std::sync::{
     Arc, Mutex, RwLock, RwLockReadGuard,
 };
 
+use holochain_cas_implementations::cas::file::FilesystemStorage;
+
+#[cfg(target_os = "linux")]
```
Connoropolous: @AshantiMutinta does this target mac as well?

core_api_c_binding/src/lib.rs
```diff
@@ -27,17 +27,21 @@ impl Logger for NullLogger {
 pub unsafe extern "C" fn holochain_new(ptr: *mut Dna) -> *mut Holochain {
     let agent = Agent::from("c_bob".to_string());
 
-    let context = Arc::new(Context::new(
+    let context = Context::new(
         agent,
         Arc::new(Mutex::new(NullLogger {})),
         Arc::new(Mutex::new(SimplePersister::new())),
-    ));
+        "holostorage",
```
Connoropolous: What is this "holostorage" string? What does it do? 

I think it might not be a good choice of name, since it once again sets up a holo <> holochain confusion

AshantiMutinta: Changes still incoming :) have a push that uses the static storage variable instead.

core/src/context.rs
AshantiMutinta: Oh yes, will be having a push in a few minutes

core_api_c_binding/src/lib.rs
Connoropolous: Nice, sorry :p

AshantiMutinta: hehe no problem!! For this to work such questions need to be asked hehe

core/src/context.rs
sphinxc0re: There is already a crate for this that manages home directories

sphinxc0re: Specifically: https://crates.io/crates/dirs

sphinxc0re: Or: https://crates.io/crates/directories

AshantiMutinta: Nice, will make those updates and push! Thank you

core_api_c_binding/src/lib.rs
```diff
@@ -27,17 +27,21 @@ impl Logger for NullLogger {
 pub unsafe extern "C" fn holochain_new(ptr: *mut Dna) -> *mut Holochain {
     let agent = Agent::from("c_bob".to_string());
 
-    let context = Arc::new(Context::new(
+    let context = Context::new(
```
ddd-mtl: I would rename to `maybe_context` or some other name that makes it clear its a result.

core/src/agent/state.rs
```diff
@@ -134,20 +134,22 @@ fn reduce_commit_entry(
...
+        context: &Context,
         state: &mut AgentState,
         entry: &Entry,
         chain_header: &ChainHeader,
     ) -> Result<Address, HolochainError> {
-        state.chain.content_storage().add(entry)?;
-        state.chain.content_storage().add(chain_header)?;
+        let mut res_context = context.clone();
+        res_context.file_storage.add(entry)?;
+        res_context.file_storage.add(chain_header)?;
```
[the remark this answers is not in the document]
AshantiMutinta: Could it be refactored with the dht storage? I could revert it back if that is the case. I only added this because it was only an in memory store for persistence.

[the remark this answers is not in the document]
AshantiMutinta: Oh yeah Chainstore takes in a CAS so I will make that change definitely. Will probably cause an explosion of the state type all around but no problem

core/src/context.rs
```diff
@@ -22,6 +31,7 @@ pub struct Context {
     state: Option<Arc<RwLock<State>>>,
     pub action_channel: SyncSender<ActionWrapper>,
     pub observer_channel: SyncSender<Observer>,
+    pub file_storage: FilesystemStorage,
```
lucksus:
```suggestion
    pub file_storage: Box<dyn ContentAddressableStorage>,
```

core_api_c_binding/src/lib.rs
```diff
@@ -3,7 +3,7 @@ extern crate holochain_core;
 extern crate holochain_core_api;
 extern crate holochain_dna;
 
-use holochain_core::context::Context;
+use holochain_core::context::{Context, STORAGE_PATH};
```
lucksus: This is weird.

The storage path should not be in the module context, and not in the crate core.
Actually the storage path has to come in as a parameter to `holochain_new` below.

core_api_c_binding/src/lib.rs
```diff
@@ -27,17 +27,21 @@ impl Logger for NullLogger {
 pub unsafe extern "C" fn holochain_new(ptr: *mut Dna) -> *mut Holochain {
     let agent = Agent::from("c_bob".to_string());
 
-    let context = Arc::new(Context::new(
+    let context = Context::new(
         agent,
         Arc::new(Mutex::new(NullLogger {})),
         Arc::new(Mutex::new(SimplePersister::new())),
-    ));
+        STORAGE_PATH,
```
lucksus: This is where a `FileSystemStorage` should be created and put in a `Box`, so that core does not need to know anything about any CAS implementation.

wasm_utils/Cargo.toml
```diff
@@ -8,6 +8,7 @@ serde = "1.0"
 serde_derive = "1.0"
 serde_json = { version = "1.0", features = ["preserve_order"] }
 holochain_core_types = { path = "../core_types" }
+tempfile = "3"
```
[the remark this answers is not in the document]
lucksus: Yes

core/src/context.rs
lucksus: I don't think this belongs in here. The path should be set in the container. Ultimately, the CAS implementation has to be instantiated in the container and passed in to `holochain_new()` via the context. But we said for the DevCamp we always construct a `FileSystemStorage` in `core_api` with only the path set by the container.

In HoloSqape (and hcshell of that matter) I am using Qt's `QStandardPaths` to get the right path for the settings, which are actually very different for all three OSs (like $HOME/Library/Preferences/<App Name> for macOS and something else for Windows)

core/src/context.rs
AshantiMutinta: This needs to be done in another story because it means Context<T>. This causes an implosion of types which need to be changed across the system. This might also mean lifetimes but good suggestion!

hdk-rust/Cargo.toml
```diff
@@ -10,6 +10,7 @@ serde_json = "1.0"
 bitflags = "1.0"
 lazy_static = "1.1.0"
 holochain_wasm_utils = { path = "../wasm_utils" }
+tempfile="3"
```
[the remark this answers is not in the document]
AshantiMutinta: Just for creating the tempfiles in the tests :)

core/src/context.rs
thedavidmeister: @lucksus @AshantiMutinta we deliberately said that we would just hardcode for this PR to avoid the type implosion or lifetimes

i was actually kind of assuming a very lightweight actor to return the storage would avoid the type issues, but this is a future conversation

hdk-rust/Cargo.toml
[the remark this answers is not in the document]
ddd-mtl: Would you mind moving it to [dev-dependencies] then?

ddd-mtl:
still blocking for me:
 - tempfile crate added to normal dependencies and not dev-dependencies.

core/src/context.rs
lucksus: I don't understand what you mean with explosion of types and lifetimes.. 

I was saying hardcoding the path for the filestorage the gets set in the context when passing it to core, which means the context does depend on a specific CAS implementation.

The SoA says: "The storage implementation is passed into HC core via the Context."

AshantiMutinta: Passing a CAS into the context will mean that the struct has to be refactored to have a type definition e.g. struct context<T:CAS>{...}then you have your Box<T> in that. The reason being the type is valid(so it can be used in methods and return types but can't be constructed in a struct. Context touches so many things so you will likely have to touch every part that uses context that is if you don't pass it around an associated type wrapper

Hardcoding for now gets passed this and attempts and uses it as if it wasn't even a trait. Benefit I could see from doing it this right now is that it makes the refactor much easier only in the chainstore context cause not a lot of things will change. Context taking in a filesystemeav or filesystemcas could potentially be a huge refactor

core_api_c_binding/src/lib.rs
```diff
@@ -25,23 +32,42 @@ impl Logger for NullLogger {
...
-        Arc::new(Mutex::new(SimplePersister::new())),
-    ));
+    let context = get_context();
 
     assert!(!ptr.is_null());
     let dna = Box::from_raw(ptr);
 
-    match Holochain::new(*dna, context) {
-        Ok(hc) => Box::into_raw(Box::new(hc)),
+    match context {
```
sphinxc0re:
```suggestion
context.and_then(|con| Holochain::new(*dna, Arc::new(con))
    .and_then(|hc| Box::into_raw(Box::new(hc)))
).unwrap_or_else(|| std::ptr::null_mut())
```

sphinxc0re: I thought it might be bothersome to always sync up the last two lines (the ones with the `ptr::null_mut()`)

core/src/state.rs
```diff
@@ -20,17 +20,17 @@ pub struct State {
...
         let content_storage =
             MemoryStorage::new().expect("could not create new cas memory storage");
         let eav_storage = EavMemoryStorage::new().expect("could not create new eav memory storage");
-
+        let file_storage = &(*context).file_storage;
         State {
             nucleus: Arc::new(NucleusState::new()),
-            agent: Arc::new(AgentState::new(ChainStore::new(content_storage.clone()))),
+            agent: Arc::new(AgentState::new(ChainStore::new(file_storage.clone()))),
             dht: Arc::new(DhtStore::new(content_storage.clone(), eav_storage.clone())),
```
lucksus: Can we use the same storage for the DHT here as well? Is there a reason to leave DhtStore with the memory storage? @AshantiMutinta

AshantiMutinta: No you are right, just forgot the push  before I left. there in a few seconds.

Connoropolous:
If there is an argument for the container instance creating the folders, and passing them in, I am open to hearing that. THen we will build it into hcshell

lucksus:
But if the directory path exists, you could confirm that it works, @Connoropolous?

I think we actually should have the container manage that and thus pass in the right directory path. But  we will need a new function in core_api and c bindings for *loading* an instance instead of *creating* one. And we also need persistence of the Redux state for that to work anyway. I would actually vouch for having this PR go through as it is right now and add the right path handling together with the Redux state persistence and a `holochain_load()` function in `c_bindings` that HoloSqape and hcshell will use then..

Connoropolous:
@lucksus I can confirm that it DOES! woohoo, so that is the celebratory piece. 

Nico, I will retract my block, according to your added context :)

Connoropolous:
@lucksus could you please open up the issues for those things, since you've got the full context for them, and link to them from here as followups?

PR #488, opened 2018-10-17
hdk: Move API to api.rs

ddd-mtl:
and `global_fns.rs ` for the macro helpers.


Connoropolous:
Yeah, total agree with Nico there ^^ I will approve if we can make that change

Connoropolous:
This also bring to mind the renaming of this folder from `hdk-rust` to `hdk_rust` since it is oddly the only folder with a dash rather than an underscore. But that ALSO brings to mind Damiens rename suggestions, which don't seem to have consensus, about renaming these "zome frameworks". The one thing which is slowly coming clearer to me though is that actually, even a language for a zome is ideally like "a kit" because like in hdk-rust and hdk-assemblyscript, it implements these special syntaxes, decorators, macros, etc.

ddd-mtl:
Kind of depends on what's the difference between kit and framework for you.
How about Zome Development Kit - ZDK ?

For the imports, I guess it's up to the user to decide if they want to write hdk::api:: everytime or not.

Connoropolous:
ZDK doesn't have the same ring as HDK for sure, but I don't think I would have a problem with that. 
Whatever we do, if we rename it, it'll be a ton of work for me :p  so let's try and pick something that lasts. 

For the imports, you're right, one can just say in the code

`use hdk::api::*` and then they could even just write `debug(x)` and such right? :)

ddd-mtl:
Just `pub use api::*;` is a nice trick 👍

hdk-rust/wasm-test/src/lib.rs
```diff
@@ -132,7 +128,7 @@ struct TestEntryType {
 
 validations! {
     [ENTRY] validate_testEntryType {
-        [hdk::ValidationPackage::Entry]
+        [hdk::api::ValidationPackage::Entry]
```
lucksus: Missed this one. Does this compile?

PR #490, opened 2018-10-17
WIP: JsonString implementation

thedavidmeister:


core_types/src/chain_header.rs
```diff
@@ -109,25 +114,22 @@ impl ToEntry for ChainHeader {
...
-            .expect("entry is not a valid ChainHeader Entry");
+        ChainHeader::from(entry.value().to_owned())
     }
 }
 
 impl AddressableContent for ChainHeader {
     fn content(&self) -> Content {
-        self.to_json()
-            .expect("could not Jsonify ChainHeader as Content")
+        Content::from(self.to_owned())
```
ddd-mtl: I'd prefer `self.to_json()` or `self.to_content()`
This kind of hides the fact that we are serializing to a JsonString.

core_types/src/chain_header.rs
```diff
@@ -109,25 +114,22 @@ impl ToEntry for ChainHeader {
...
     }
 
     fn from_content(content: &Content) -> Self {
-        ChainHeader::from_json_str(content)
-            .expect("could not read Json as valid ChainHeader Content")
+        Self::from(content.to_owned())
```
ddd-mtl: Shouldn't `ChainHeader::from_json_str()` be removed from codebase?

thedavidmeister:
@ddd-mtl haha, this is still WIP!

i love that you love it though :D

the thing that I like about `Content` is exactly that we can swap out the underlying type and the compiler enforces sanity

thedavidmeister:
@ddd-mtl also i'm not opposed to implementing something like `.to_json()` but imo it should be sugar on top of `From`

that enables consistency across json handling, and with the native rust type conversion system

ddd-mtl:
Cool. For some reason I have a hard time seeing the WIP sign ^^''

core_types/src/chain_header.rs
thedavidmeister: @ddd-mtl does it? isn't the point of `From` that it defines type conversions?

cas_implementations/src/eav/file/actor.rs
```diff
@@ -167,7 +167,9 @@ impl EavFileStorageActor {
         Ok(entity_attribute_value_inter
             .into_iter()
             .filter(|e| e.is_ok())
-            .map(|eav_content| EntityAttributeValue::from_content(&JsonString::from(eav_content.unwrap())))
+            .map(|eav_content| {
+                EntityAttributeValue::from_content(&JsonString::from(eav_content.unwrap()))
```
sphinxc0re: You could just do `eav_content.unwrap().into()`

thedavidmeister: @sphinxc0re please provide code suggestions in github using the new functionality, or make minor changes directly and push them up

core_types/src/chain_header.rs
thedavidmeister: probably yes

core_api/src/lib.rs
```diff
@@ -49,7 +49,7 @@
 //! hc.stop().expect("couldn't stop the app");
 //!
 //!```
-
+#![feature(try_from)]
```
sphinxc0re: I feel like I shouldn't have to say this....... feature flags should not be used in production level software......

thedavidmeister: @sphinxc0re why not?

core_types/src/json.rs
```diff
@@ -57,17 +57,16 @@ impl<T: Serialize> From<Vec<T>> for JsonString {
 
 impl<T: Into<JsonString>, E: Into<JsonString>> From<Result<T, E>> for JsonString {
     fn from(result: Result<T, E>) -> JsonString {
-        JsonString::from(
-            serde_json::to_string(&match result {
+        JsonString::from(match result {
```
sphinxc0re: You could replace this encapsulating call to `JsonString::from` with a postfix `.into()`. Because every `From` implementation automatically generates an `Into` implementation

core/src/nucleus/ribosome/callback/validate_entry.rs
```diff
@@ -96,7 +96,7 @@ fn run_validation_callback(
     ) {
         Ok(call_result) => match call_result.is_null() {
             true => CallbackResult::Pass,
-            false => CallbackResult::Fail(call_result),
+            false => CallbackResult::Fail(call_result.to_string()),
```
sphinxc0re: FYI `bool` is not a monad. Could you please make this a regular if-expression? 🙏

thedavidmeister: @sphinxc0re i feel like i've touched enough LOC  already :P

was this way when i found it

core_types/src/json.rs
```diff
@@ -1,16 +1,277 @@
...
+
+pub type JsonResult = Result<JsonString, HolochainError>;
+
+/// standard boilerplate:
+// impl TryFrom<T> for JsonString {
+//     type Error = HolochainError;
+//     fn try_from(v: T) -> JsonResult {
+//         default_try_to_json(v)
+//     }
+// }
```
lucksus: // -> ///

core_types/src/json.rs
```diff
@@ -1,16 +1,277 @@
...
+    default_try_to_json(s).unwrap()
+}
+
+// standard boilerplate:
+// impl TryFrom<JsonString> for T {
+//     type Error = HolochainError;
+//     fn try_from(j: JsonString) -> HcResult<Self> {
+//         default_try_from_json(j)
+//     }
+// }
```
lucksus: // -> ///

hdk-rust/src/api.rs
```diff
@@ -326,31 +316,32 @@ pub fn link_entries<S: Into<String>>(
...
 
     let encoded_allocation_of_result: u32 =
         unsafe { hc_link_entries(allocation_of_input.encode() as u32) };
 
     // Deserialize complex result stored in memory and check for ERROR in encoding
-    let result: LinkEntriesResult = load_json(encoded_allocation_of_result as u32)
-        .map_err(|err_str| ZomeApiError::Internal(err_str))?;
+    let result: ZomeApiInternalResult = load_json(encoded_allocation_of_result as u32)?;
+
+    println!("link_entries result {:?}", result);
```
lucksus: Hm, I'd be surprised if this would work here. Inside the zome you need to use `hdk::debug()` to get an output..

hdk-rust/src/error.rs
```diff
@@ -1,43 +0,0 @@
...
-
-impl fmt::Display for ZomeApiError {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        // @TODO seems weird to use debug for display
-        // replacing {:?} with {} gives a stack overflow on to_string() (there's a test for this)
-        // what is the right way to do this?
-        // @see https://github.com/holochain/holochain-rust/issues/223
-        write!(f, "{:?}", self)
-    }
-}
```
lucksus: Why did you move this to core_types?

core_types/src/ribosome/callback.rs
```diff
@@ -0,0 +1,75 @@
...
+impl From<RibosomeReturnCode> for CallbackResult {
+    fn from(ribosome_return_code: RibosomeReturnCode) -> CallbackResult {
+        match ribosome_return_code {
+            RibosomeReturnCode::Failure(ribosome_error_code) => {
+                CallbackResult::Fail(ribosome_error_code.to_string())
+            }
+            RibosomeReturnCode::Success => CallbackResult::Pass,
+        }
+    }
+}
```
lucksus: Why is this here?

lucksus:
ah, I know see all those conflicting files. I think I had an old state in this tab ...

core_types/src/chain_header.rs
thedavidmeister: i'll try `into()` as that seems pretty standard

hdk-rust/src/error.rs
thedavidmeister: i don't remember, maybe a bad merge?

core/src/agent/state.rs
```diff
@@ -98,62 +96,31 @@ impl AgentStateSnapshot {
...
-                Ok(hash_list) => Ok(json!(hash_list).to_string()),
-                Err(err) => Ok((*err).to_json()?),
-            },
-            ActionResponse::LinkEntries(result) => match result {
-                Ok(entry) => Ok(format!("{{\"address\":\"{}\"}}", entry.address())),
-                Err(err) => Ok((*err).to_json()?),
-            },
-        }
-    }
+    LinkEntries(Result<SerializedEntry, HolochainError>),
```
lucksus: Why are you using SerializedEntry in those actions instead of just Entry, @thedavidmeister?

core_types/src/error/error.rs
```diff
@@ -157,6 +164,33 @@ impl From<FutureCanceled> for HolochainError {
     }
 }
 
+#[derive(Serialize, Deserialize, Default, Debug, DefaultJson)]
+pub struct ZomeApiInternalResult {
+    pub ok: bool,
+    pub value: String,
+    pub error: String,
+}
```
lucksus: Since this is the structure of the JSON that is passed through memory, it is something other HDKs need to implement as well. So far, we had all of those structs in wasm-utils/api_serialization.

I'm not sure if I want to request to move this over there for the sake of HDK developers finding it easily.

I don't want to spawn long discussions in this already big PR that has lots of awesome that I want to have merged in asap. So just marking it here for future discussion/changes, @thedavidmeister, @ddd-mtl, @Connoropolous, @willemolding.

core/src/nucleus/ribosome/runtime.rs
```diff
@@ -54,34 +55,33 @@ impl Runtime {
...
-        match maybe_json {
-            Err(_) => ribosome_error_code!(ResponseSerializationFailed),
-            Ok(json) => self.store_utf8(&json),
-        }
+    pub fn store_result<J: Into<JsonString>>(
+        &mut self,
+        result: Result<J, HolochainError>,
+    ) -> ZomeApiResult {
+        self.store_as_json_string(match result {
+            Ok(address) => ZomeApiInternalResult::success(address),
```
zippy: this makes it seem like the content of the the result OK(J) will always be an address, but from what I can tell it could be anything thats jsonifyable, not just an address.  Shouldn't we change the variable name to `x` or `content`?

core/src/agent/state.rs
zippy: I think this is explained in the PR documentation.  I think it's because of wanting to get at the entry type separate from the entry value and they are both inside the entry and thus get serialized twice.  Or something like that.  It's not totally clear to me.

lucksus:
@sphinxc0re, your approval is old. Can you re-review and and then merge please if you're ok with this?

wasm_utils/src/api_serialization/query.rs
```diff
@@ -1,12 +1,9 @@
...
 pub struct QueryArgs {
     pub entry_type_name: String,
     pub limit: u32,
 }
 
-#[derive(Deserialize, Default, Debug, Serialize)]
-pub struct QueryResult {
-    pub addresses: Vec<Address>,
-}
+pub type QueryResult = Vec<Address>;
```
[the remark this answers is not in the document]
zippy: I think it's fine to leave it as a type for right now because that should seamlessly convert back to struct once we add contents no?

core/src/nucleus/ribosome/api/get_links.rs
```diff
@@ -22,17 +28,14 @@ pub fn invoke_get_links(runtime: &mut Runtime, args: &RuntimeArgs) -> ZomeApiRes
...
-                .collect::<Vec<Address>>();
-            runtime.store_as_json(GetLinksResult { addresses })
-        }
-    }
+
+    runtime.store_result(match maybe_links {
+        Ok(links) => Ok(links
+            .iter()
+            .map(|eav| eav.value())
+            .collect::<Vec<Address>>()),
```
zippy: Whoops, yes we do need this back.  GetLinksResult needs to stay for when it returns more than just the addresses, i.e. the meta data

thedavidmeister:
added #590 @ddd-mtl

core_types/src/error/error.rs
lucksus: @ddd-mtl, oh I like this ZomeApiInternalResult and how it's used a lot. I was just talking about where it is. I think having this as the standard way makes understanding the zome API much easier. And especially implementing other HDKs. But therefore it should be somewhere where it's understood as this standard right away..

Why would you need to revert this?

core/src/nucleus/wasm-test/src/lib.rs
```diff
@@ -15,9 +22,9 @@ extern {
 
 /// Call HC API DEBUG function with proper input struct: a string
 /// return error code
-fn hdk_debug(mem_stack: &mut SinglePageStack, s: &str) {
+fn hdk_debug(mem_stack: &mut SinglePageStack, json_string: &JsonString) {
```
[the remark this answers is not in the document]
zippy: The problem here is that this is a "fake" test of the hdk, the real hdk::debug is `pub fn debug<J: TryInto<JsonString>>(msg: J) -> ZomeApiResult<()>`  And we really should replace this test with something more real.

hdk-rust/src/init_globals.rs
```diff
@@ -2,21 +2,26 @@
...
+use std::convert::TryInto;
 
 extern "C" {
     fn hc_init_globals(encoded_allocation_of_input: u32) -> u32;
 }
 
 // HC INIT GLOBALS - Secret Api Function
 // Retrieve all the public global values from the ribosome
-pub(crate) fn init_globals() -> ZomeApiGlobals {
+pub(crate) fn init_globals() -> ZomeApiResult<ZomeApiGlobals> {
```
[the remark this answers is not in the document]
thedavidmeister: making this consistent sets us up for a macro later

also i think that it should be a result for the call, but an unwrap/expect for deciding what to do about it

sure we don't want to recover, but that doesn't mean we want to hardcode the handling of it downstream

hdk-rust/src/error.rs
```diff
@@ -41,3 +77,5 @@ impl fmt::Display for ZomeApiError {
         write!(f, "{:?}", self)
     }
 }
+
+pub type ZomeApiResult<T> = Result<T, ZomeApiError>;
```
thedavidmeister: @ddd-mtl it is down here

core/src/nucleus/wasm-test/src/lib.rs
```diff
@@ -51,27 +58,27 @@ pub extern "C" fn debug_hello(encoded_allocation_of_input: usize) -> i32 {
...
 pub extern "C" fn debug_multiple(encoded_allocation_of_input: usize) -> i32 {
   let mut mem_stack = SinglePageStack::from_encoded_allocation(encoded_allocation_of_input as u32).unwrap();
-  hdk_debug(&mut mem_stack, "Hello");
-  hdk_debug(&mut mem_stack, "world");
-  hdk_debug(&mut mem_stack, "!");
-  return 0;
+  hdk_debug(&mut mem_stack, &JsonString::from(RawString::from("Hello")));
+  hdk_debug(&mut mem_stack, &JsonString::from(RawString::from("world")));
+  hdk_debug(&mut mem_stack, &JsonString::from(RawString::from("!")));
+  i32::from(RibosomeReturnCode::Success)
```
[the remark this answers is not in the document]
thedavidmeister: this is just because the fake hdk has no tooling :(

followup is to remove the fake hdks that only exist for testing and do more low level tests and also more tests in the real hdk

core/src/nucleus/ribosome/api/link_entries.rs
```diff
@@ -1,31 +1,30 @@
...
         Ok(entry_input) => entry_input,
         // Exit on error
-        Err(_) => return ribosome_error_code!(ArgumentDeserializationFailed),
+        Err(_) => {
+            println!(
+                "invoke_link_entries failed to deserialize LinkEntriesArgs: {:?}",
+                args_str
+            );
+            return ribosome_error_code!(ArgumentDeserializationFailed);
+        }
```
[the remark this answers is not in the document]
thedavidmeister: this is intentional because there is no other way to know what went wrong

the error code by itself doesn't contain enough information to actually debug anything

debugging this downstream involves rebuilding two or three repos per change

i agree that the println! is not ideal, but there does need to be a way to get debug information out, so let's discuss in a followup

core/src/nucleus/ribosome/api/init_globals.rs
```diff
@@ -44,28 +45,35 @@ pub fn invoke_init_globals(runtime: &mut Runtime, _args: &RuntimeArgs) -> ZomeAp
...
     #[test]
     /// test that bytes passed to debug end up in the log
     fn test_init_globals() {
         let input: Vec<u8> = vec![];
-        let (mut call_result, _) =
-            test_zome_api_function(ZomeApiFunction::InitGlobals.as_str(), input);
-        call_result.pop(); // Remove trailing character
-        let globals: ZomeApiGlobals = serde_json::from_str(&call_result).unwrap();
+        let (call_result, _) = test_zome_api_function(ZomeApiFunction::InitGlobals.as_str(), input);
+        println!("{:?}", call_result);
```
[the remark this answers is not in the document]
thedavidmeister: this is just info output from the test

core/src/nucleus/ribosome/api/get_links.rs
thedavidmeister: kk, i'll put it all back in with a followup, my bad

core/src/nucleus/ribosome/api/get_links.rs
```diff
@@ -10,10 +10,16 @@ use wasmi::{RuntimeArgs, RuntimeValue};
...
+    let input = match GetLinksArgs::try_from(args_str.clone()) {
         Ok(input) => input,
-        Err(_) => return ribosome_error_code!(ArgumentDeserializationFailed),
+        Err(_) => {
+            println!(
+                "invoke_get_links failed to deserialize GetLinksArgs: {:?}",
+                args_str
+            );
+            return ribosome_error_code!(ArgumentDeserializationFailed);
+        }
```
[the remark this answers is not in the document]
thedavidmeister: intentional, see below

core/src/nucleus/ribosome/api/debug.rs
```diff
@@ -6,7 +6,7 @@ use wasmi::{RuntimeArgs, RuntimeValue};
 /// Expecting a string as complex input argument
 /// Returns an HcApiReturnCode as I32
 pub fn invoke_debug(runtime: &mut Runtime, args: &RuntimeArgs) -> ZomeApiResult {
-    let payload = runtime.load_utf8_from_args(args);
+    let payload = runtime.load_json_string_from_args(args);
```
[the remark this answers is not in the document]
thedavidmeister: it can be json now, allows debugging entries, etc.

core/src/nucleus/ribosome/api/call.rs
```diff
@@ -80,15 +81,11 @@ pub fn invoke_call(runtime: &mut Runtime, args: &RuntimeArgs) -> ZomeApiResult {
...
-    // action_result should be a json str of the result of the zome function called
-    match action_result {
-        Err(hc_err) => runtime.store_as_json(core_error!(hc_err)),
-        Ok(json_str) => runtime.store_utf8(&json_str),
-    }
+    runtime.store_result(
+        receiver
+            .recv_timeout(RECV_DEFAULT_TIMEOUT_MS)
+            .expect("observer dropped before done"),
+    )
```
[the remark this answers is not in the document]
thedavidmeister: this is just me being "lispy"

happy to change it back

core/src/instance.rs
```diff
@@ -657,10 +658,12 @@ pub mod tests {
...
-        let instance = test_instance(dna);
-        assert!(instance.is_ok());
-        let instance = instance.unwrap();
-        assert!(instance.state().nucleus().has_initialized());
+        let maybe_instance = test_instance(dna);
+
+        match maybe_instance {
+            Ok(instance) => assert!(instance.state().nucleus().has_initialized()),
+            Err(e) => panic!(e),
+        }
```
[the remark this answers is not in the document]
thedavidmeister: i think this was a bad merge, i'll fix it

hdk-rust/src/init_globals.rs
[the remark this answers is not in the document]
ddd-mtl: `init_globals()` is not a normal api function as only we, hc core devs, use it to initiliaze globals at dna startup. So it is a really scoped use case which does need to comply with standard "macro'd" way of doing zome api functions.

PR #491, opened 2018-10-17
Waiting: Fix hdk functions

Connoropolous:
This is waiting on #488, it's branched off that branch

We can merge it after that, or into that if people approve of that


Connoropolous:
Yeah, I know. I did that so you could actually see a good diff indicator, but would only merge it into `develop` once that one is in to develop

lucksus:
👍

Connoropolous:
Oh, you are commenting on two different things, was confused for a sec. Yeah, I don't really have much of an idea about the bundle signature right now, but maybe we could create a separate issue to discuss that bit

PR #493, opened 2018-10-17
486 - Load and store strings in wasm memory

### LINKED ISSUE #486: wasm_utils: load and store string from wasm memory
+ tests

ddd-mtl:
closes #486 
Added `load_string()` and `store_string()` with tests.
Now works in Release with optimization level set to 'z'.

Cleaned-up `memory_serialization.rs`. Json functions have been moved around.

In the future, will probably rename the file to something like `memory_io.rs`.

Also since memory is a stack, might rename functions from store/load to stack/unstack/read.


wasm_utils/wasm-test/integration-test/src/lib.rs
```diff
@@ -24,24 +24,33 @@ pub extern "C" fn test_error_report(_: u32) -> u32 {
...
-//    let s = "some string";
-//    assert_eq!(0, stack.top());
-//    let res = store_string(&mut stack, s);
-//    //assert_eq!(obj.len(), stack.top() as usize);
-//    //res.unwrap().encode()
-//    0
-//}
+// General Note:
+// Can't do zome_assert!() while testing store_as_json() since
+// zome_assert!() internally uses store_as_json().
```
Connoropolous: Nice note... as I was reading the code I was like... hmmm

ddd-mtl:
I guess its a chicken and egg thing.
I can't use it in production code if its not there yet...
Planning on using it for at least error messages, yes.

wasm_utils/src/memory_serialization.rs
```diff
@@ -6,36 +6,72 @@ use serde::{Deserialize, Serialize};
...
+//-------------------------------------------------------------------------------------------------
+// String
+//-------------------------------------------------------------------------------------------------
+
+/// Write a string in wasm memory according to stack state.
+pub fn store_string(
+    stack: &mut SinglePageStack,
+    s: &str,
+) -> Result<SinglePageAllocation, RibosomeErrorCode> {
+    let bytes = s.to_string().into_bytes();
```
sphinxc0re: https://doc.rust-lang.org/1.4.0/std/primitive.str.html#method.as_bytes

Maybe that works better and doesn't copy

wasm_utils/src/memory_serialization.rs
```diff
@@ -6,36 +6,72 @@ use serde::{Deserialize, Serialize};
...
+    let len = bytes.len();
+    if len > <u16>::max_value() as usize {
+        return Err(RibosomeErrorCode::OutOfMemory);
+    }
+    return write_in_wasm_memory(stack, &bytes, len as u16);
```
sphinxc0re: Unnecessary return statement.

```rust
if len > u16::max_value() as usize {
    Err(RibosomeErrorCode::OutOfMemory);
} else {
    write_in_wasm_memory(stack, &bytes, len as u16)
}
```

wasm_utils/src/memory_serialization.rs
```diff
@@ -6,36 +6,72 @@ use serde::{Deserialize, Serialize};
...
+}
+
+// Sugar
+pub fn store_string_into_encoded_allocation(stack: &mut SinglePageStack, s: &str) -> i32 {
+    let allocation_of_output = store_string(stack, s).unwrap();
+    return allocation_of_output.encode() as i32;
```
sphinxc0re: Again, this return statement is unnecessary. Also, couldn't you just do:

```rust
store_string(stack, s).unwrap().encode() as i32
```
?

PR #499, opened 2018-10-18
Ribosome: Reworked call return value to be a HcResult<String>

ddd-mtl:


core/src/nucleus/ribosome/api/debug.rs
```diff
@@ -9,16 +9,17 @@ pub fn invoke_debug(
     runtime: &mut Runtime,
     args: &RuntimeArgs,
 ) -> Result<Option<RuntimeValue>, Trap> {
-    runtime.result = runtime.load_utf8_from_args(args);
-    println!("{}", runtime.result);
+    let payload = runtime.load_utf8_from_args(args);
+    println!("{}", payload);
+    // TODO log in logger as DEBUG
```
Connoropolous: could we get an issue open for this

core/src/nucleus/ribosome/api/init_globals.rs
```diff
@@ -22,7 +22,7 @@ pub fn invoke_init_globals(
     _args: &RuntimeArgs,
 ) -> Result<Option<RuntimeValue>, Trap> {
     let globals = InitGlobalsOutput {
-        app_name: runtime.app_name.to_string(),
+        app_name: runtime.dna_name.to_string(),
```
Connoropolous: I kinda agree with this, but then it's also confusing cause on the left hand "app_name" and on the right name "dna_name" and it makes you wonder about the difference, or how/why they're the same, and one is set to the other

core/src/nucleus/ribosome/api/mod.rs
```diff
@@ -349,33 +347,53 @@ pub fn call(
...
             let result = runtime.memory_manager.read(valid_allocation);
-            runtime.result = String::from_utf8(result).unwrap();
+            let maybe_zome_result = String::from_utf8(result);
+            match maybe_zome_result {
+                Err(err) => {
+                    return_text = err.to_string();
+                    return_result = Err(HolochainError::RibosomeFailed(err.to_string()));
+                }
+                Ok(json_str) => {
+                    return_text = json_str.clone();
```
Connoropolous: it seems like nothing at all actually happens with `return_text` am I wrong?

core/src/nucleus/ribosome/api/init_globals.rs
ddd-mtl: Yes. They are the same, and I'll rename all mention of 'app' in hdk-rust in a subsequent PR.
Seems out-of-scope of this one.
Expect: `APP_GLOBALS.app_name` => `ZOME_API_GLOBALS.dna_name`

Connoropolous: Cool. I did. It's a good change to make prior to the dev preview release

core/src/nucleus/ribosome/api/debug.rs
ddd-mtl: yes!

core/src/nucleus/ribosome/api/mod.rs
ddd-mtl: Its used right after in the log call.
I guess we can thank rusfmt for making that instruction unreadable.

Connoropolous: oh yeah I see. Yeah I totally missed it, and had a feeling somehow that I did

core/src/nucleus/ribosome/api/debug.rs
```diff
@@ -34,11 +35,11 @@ pub mod tests {
 
     #[test]
     /// test that bytes passed to debug end up in the log
-    fn test_debug() {
```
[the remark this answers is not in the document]
ddd-mtl: Yeah. Added debug to logger. It itches me to do this without log level but I guess its still better to have it in than not have it in.

core/src/nucleus/ribosome/api/mod.rs
```diff
@@ -349,33 +347,53 @@ pub fn call(
...
             .try_into()
             .unwrap();
     }
 
     // Handle result returned by invoked function
     let maybe_allocation = decode_encoded_allocation(returned_encoded_allocation);
+    let return_text: String;
+    let return_result: HcResult<String>;
     match maybe_allocation {
         // Nothing in memory, log return code
```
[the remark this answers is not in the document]
ddd-mtl: fixed

core_api/src/lib.rs
```diff
@@ -456,7 +456,7 @@ mod tests {
 
         // Call the exposed wasm function that calls the Commit API function
         let result = hc.call("test_zome", "test_cap", "debug_hello", r#"{}"#);
-        assert_eq!("\"Hello world!\"", result.unwrap());
+        assert!(result.unwrap().is_empty());
```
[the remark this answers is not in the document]
ddd-mtl: `debug_hello()` does not 'produce' anything, so we expect an empty result.
I did an update so you can now see the debug string in the logger.

ddd-mtl:
@zippy and others: I updated the integration tests in wasm_utils, and made them much simpler as it is not necessary to test logs anymore \o/

PR #500, opened 2018-10-18
Hdk unit tests

murrayjbrown:
Migrated hdk-rust API unit tests from holochain/hdk-rust repo (with considerable refactoring) per issues holochain/hdk-rust#13 &  holochain/hdk-rust#14.

Also, minor revision to _.gitignore_ to ignore _**/target/_ directories.


hdk-rust/src/api.rs
```diff
@@ -376,3 +406,680 @@ pub fn close_bundle(_action: BundleOnClose) -> Result<(), RibosomeError> {
...
+    fn test_make_hash_invalid() {
+        // check whether function implemented
+        let result = make_hash("", json!("test_data"));
+        assert_ne!(Some(RibosomeError::FunctionNotImplemented), result.err());
+
+        // test empty entry type parameter
+        // TODO: is this the right error?
+        let result = make_hash("", json!("test_data"));
+        assert!(result.is_err());
+        assert_eq!(Some(RibosomeError::HashNotFound), result.err());
```
lucksus: I don't understand how this test can pass.. (?!)
`make_hash()` is returning `Err(RibosomeError::FunctionNotImplemented)` above.

Are these test not run be the CI?

hdk-rust/src/api.rs
```diff
@@ -376,3 +406,680 @@ pub fn close_bundle(_action: BundleOnClose) -> Result<(), RibosomeError> {
     // FIXME
     Err(RibosomeError::FunctionNotImplemented)
 }
+//--------------------------------------------------------------------------------------------------
+// UNIT TESTS
+//--------------------------------------------------------------------------------------------------
+
+/// Unit tests
+#[cfg(test)]
+mod test {
```
lucksus: Ah! All those tests below don't run because it's `mod test` instead of `mod tests`.

murrayjbrown: @lucksus, I'd thought `mod test` was the Rust convention, although any function (whatever its name) annotated with `[#cfg(test)]` should be run by the cargo test harness. I'd successfully run these unit tests via `cargo test` and `make test` natively on Debian but I've since run them again via `. docker/docker-test` and they run fine (subject to the reported failures as noted above). Is there something special about the Travis-based(?) CI environment?

murrayjbrown:
@lucksus D'oh! I think I found why the tests didn't run. I hadn't noticed a compile error  (the call() function signature changed) after merging the latest _develop_ branch, so my local tests may have run agains old artifact. Sorry about that. I've since fixed it locally.

However, I encountered problems with the latest _develop_ branch. I'm getting test errors (using `. docker/run-test`) without any of my code merged:
```rust
clippy-preview-x86_64-unknown-linux-gnu (installed)
cargo  +nightly-2018-07-17 fmt -- --check
Diff in /holochain/core_api/src/lib.rs at line 68:
 
 use futures::executor::block_on;
 use holochain_core::{
-    context::Context, instance::Instance,
+    context::Context,
+    instance::Instance,
     nucleus::{actions::initialize::initialize_application, call_and_wait_for_result, ZomeFnCall},
     state::State,
 };
```
I understand  things are in flux so l can hold off until it settle down. Kindly ping me when the codebase is stable enough for me to merge my HDK API unit test changesets and update this PR.

Now, with regards to my previous comments (above) about the objective of these unit tests, I've given some thought to how we might accomodate both CI and unit developer needs. I propose the following: add unit tests to verify that functions are not (yet) implemented, which should pass for CI; and comment out the unit tests for validating intended results once the functions under test are eventually implemented. `// FIXME` comments will advise that the test case(s) verifying valid/invalid results should eventually be enabled (uncommented) and the _not-implemented_ variant(s) discarded.

I'd like some guidance, though, about the case where I've observed Wasm compiler errors (owing to a missing runtime dependency (e.g. `hc_commit_entry` function). My inclination is to comment them out entirely, leaving the issue to be debugged by whomever writes the function(s) under test, rather than crashing CI runs.

cc: @Connoropolous, @ddd-mtl

hdk-rust/src/api.rs
thedavidmeister: you can put `#[cfg(test)]` on anything to elide a code block whenever the tests for its crate are not currently being run

`#[test]` flags a function as a test to be run

.gitignore
```diff
@@ -1,5 +1,5 @@
 .qmake.stash
-/target/
+**/target/
```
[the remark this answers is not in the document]
murrayjbrown: I had asked @Connoropolous (DM on Mattermost) whether it should be changed to exclude new artifacts produced in: `wasm_utils/wasm-test/integration-test/target/`. Is it causing problems?

hdk-rust/src/api.rs
```diff
@@ -60,13 +61,42 @@ pub enum RibosomeError {
...
+}
+
+impl PartialEq<String> for RibosomeError {
+    fn eq(&self, failure_msg: &String) -> bool {
+        match self {
+            RibosomeFailed(msg) => {
+                if msg == failure_msg {
+                    return true;
+                }
+                false
```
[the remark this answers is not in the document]
murrayjbrown: :+1:

hdk-rust/src/api.rs
murrayjbrown: Thanks @thedavidmeister. Will revise to disable running unit tests pending implementation of functions under test; i.e.:
```rust
/// #[test] // FIXME when implemented
```

.gitignore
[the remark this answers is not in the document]
sphinxc0re: No, but I would rather have the Wasm tests exclude the target directories themselves. What would happen if, at some point, we were to introduce a new module called `target`. That would be excluded by this rule.

hdk-rust/src/api.rs
murrayjbrown: Unfortunately, this approach I described above results in unreferenced functions so I've reverted to my previous suggestion of only enabling those unit tests that pass and commenting out those that crash or are otherwise pending implementation of the function(s) under test.

.gitignore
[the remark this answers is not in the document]
murrayjbrown: I've reverted this change.

PR #505, opened 2018-10-18
ribosome::api::call()  => ribosome::run_dna()

ddd-mtl:
and moved it in its own file out of 'api' folder.
Move runtime to its own file as well.

There was too much confusion with call() of the hdk api.

### DIFF
--- a/core/src/nucleus/ribosome/run_dna.rs
+++ b/core/src/nucleus/ribosome/run_dna.rs
@@ -1,4 +1,3 @@
-
 use context::Context;
 use holochain_core_types::error::{HcResult, HolochainError};
 use holochain_wasm_utils::{
@@ -6,19 +5,13 @@ use holochain_wasm_utils::{
     memory_allocation::decode_encoded_allocation,
 };
 use nucleus::{
-    ribosome::{
-        api::{
-            ZomeApiFunction,
-        },
-        memory::SinglePageManager,
-        runtime::Runtime,
-    },
+    ribosome::{api::ZomeApiFunction, memory::SinglePageManager, Runtime},
     ZomeFnCall, ZomeFnResult,
 };
 use std::{str::FromStr, sync::Arc};
 use wasmi::{
-    self, Error as InterpreterError, FuncInstance, FuncRef, ImportsBuilder,
-    ModuleImportResolver, ModuleInstance, NopExternals, RuntimeValue, Signature, ValueType,
+    self, Error as InterpreterError, FuncInstance, FuncRef, ImportsBuilder, ModuleImportResolver,
+    ModuleInstance, NopExternals, RuntimeValue, Signature, ValueType,
 };
 
 /// Executes an exposed zome function in a wasm binary.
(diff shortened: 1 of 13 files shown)


core/src/nucleus/ribosome/run_dna.rs
```diff
@@ -0,0 +1,176 @@
...
+/// Executes an exposed zome function in a wasm binary.
+/// Multithreaded function
+/// panics if wasm binary isn't valid.
+pub fn run_dna(
+    dna_name: &str,
+    context: Arc<Context>,
+    wasm: Vec<u8>,
+    zome_call: &ZomeFnCall,
+    parameters: Option<Vec<u8>>,
+) -> ZomeFnResult {
```
Connoropolous: This function is such a monster lol

Connoropolous: Gradually getting better though

core/src/nucleus/ribosome/runtime.rs
```diff
@@ -0,0 +1,89 @@
...
+        let encoded_allocation = allocation_of_result.unwrap().encode();
+
+        // Return success in i32 format
+        Ok(Some(RuntimeValue::I32(encoded_allocation as i32)))
+    }
+}
+
+// Correlate the indexes of core API functions with a call to the actual function
+// by implementing the Externals trait from Wasmi.
+impl Externals for Runtime {
```
Connoropolous: Nice, then I like how this is out here now

Connoropolous: I was wondering whether it actually needed to be defined within the function

core/src/nucleus/ribosome/run_dna.rs
ddd-mtl: yep. There is no intent to make it smaller, thats why I'm giving it its own file.
I did shrunk it a bit my removing that Runtime impl.

core/src/nucleus/ribosome/runtime.rs
ddd-mtl: Not 'needed', it just helped scope things a bit.

core/src/nucleus/ribosome/run_dna.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    // invoke_index and resolve_func work together to enable callable host functions
+    // within WASM modules, which is how the core API functions
+    // read about the Externals trait for more detail
+
+    // Correlate the names of the core ZomeApiFunction's with their indexes
+    // and declare its function signature (which is always the same)
+    struct RuntimeModuleImportResolver;
+    impl ModuleImportResolver for RuntimeModuleImportResolver {
+        fn resolve_func(
```
Connoropolous: @ddd-mtl couldn't we possibly define `RuntimeModuleImportResolver` outside of this function too, along with its `impl ModuleImportResolver...`

ddd-mtl: probably yes. We just took this code from the sample in wasmi

ddd-mtl:
@Connoropolous I was planning on continuing working in this branch :/
But I guess it would be better if I made a branch with my name in it to make it clear.

Connoropolous:
Oops, sorry @ddd-mtl I didn't think folks did that, and was in a habit of deleting them for some reason

PR #507, opened 2018-10-19
implements get with options as per app-spec pr #32

zippy:
This PR implements passing in an Options argument to get_entry and it's returning the GetEntryResult itself instead of the Option<String> where the string is the json entry.

This PR does not implement any of the Options themselves, that'll be for a separate PR, it just sets up the structure.


Connoropolous:
Opened up new #511 for the follow up @zippy

PR #514, opened 2018-10-21
513 create storage dirs

AshantiMutinta:


AshantiMutinta:
Closed because i branched out from wrong place

PR #515, opened 2018-10-21
create storage path if it does not exist

### LINKED ISSUE #513: change core_api_c_bindings so `get_context` creates the file dirs if they don't exist
Here...

cas_path and storage_path need to be created if they don't exist, otherwise the `.unwrap()` throws a panic

https://github.com/holochain/holochain-rust/blob/ed5a99c89f7985831b9f6fecca95005b015c57f4/core_api_c_binding/src/lib.rs#L51-L70

AshantiMutinta:
closes #513

### DIFF
--- a/cas_implementations/src/path.rs
+++ b/cas_implementations/src/path.rs
@@ -5,7 +5,7 @@ pub fn storage_path(path: &Path, folder_name: &str) -> HcResult<String> {
     let full_path = path.join(".hc").join("storage").join(folder_name);
     let path_as_string = full_path.to_str().ok_or(HolochainError::IoError(
         "Could not find home directory".to_string(),
-    ))?;;
+    ))?;
     Ok(String::from(path_as_string))
 }
 
--- a/cas_implementations/src/path.rs
+++ b/cas_implementations/src/path.rs
@@ -1,10 +1,30 @@
-use holochain_core_types::error::HolochainError;
-use std::path::Path;
+use holochain_core_types::error::{HcResult, HolochainError};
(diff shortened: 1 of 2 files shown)


cas_implementations/src/path.rs
```diff
@@ -1,10 +1,30 @@
...
-use std::path::Path;
+use holochain_core_types::error::{HcResult, HolochainError};
+use std::{fs::DirBuilder, path::Path};
 
-pub fn storage_path(path: &Path, folder_name: &str) -> Result<String, HolochainError> {
+pub fn storage_path(path: &Path, folder_name: &str) -> HcResult<String> {
     let full_path = path.join(".hc").join("storage").join(folder_name);
     let path_as_string = full_path.to_str().ok_or(HolochainError::IoError(
         "Could not find home directory".to_string(),
-    ))?;
```
Connoropolous: Now there's two semicolons for some reason

lucksus:
Hm, I think we really should do this in the container.
But I'm fine with merging this for now, if it is less work for the devcamp.
But I'm also not quite sure if it actually is.. ;) because I believe we have to touch the c-binding interface anyway and add a `Holochain::load()` function. That should get a path from the container. HoloSqape is already maintaining a settings directory and has code for creating the directory if it does not exists. Also that directory is more adjusted to the specific platform patterns..

lucksus:
So, I'll refrain from approving or requesting changes. If you all want to do it this way, go ahead. But I don't think it actually spares us effort.

PR #519, opened 2018-10-22
Storage path from container and scaffold for Holochain::load(storage_path: String)

### LINKED ISSUE #518: we need a Holochain::load() function.
@lucksus said:
"I don't think we can get around passing in a path from the container"

HoloSqape and hcshell already manage a setting directory that also gets created if not existent.

lucksus:
closes #518

[x] Get storage path from container
[x] Add load function scaffold


core_api_c_binding/src/lib.rs
```diff
@@ -50,29 +49,35 @@ pub unsafe extern "C" fn holochain_new(ptr: *mut Dna) -> *mut Holochain {
...
+        },
+        Err(_) => std::ptr::null_mut(),
     }
 }
 
+fn get_context(path: &String) -> Result<Context, HolochainError> {
+    let agent = Agent::from("c_bob".to_string());
+    let cas_path = format!("{}/cas", path);
+    let eav_path = format!("{}/eav", path);
+    create_path_if_not_exists(&cas_path)?;
```
Connoropolous: Nice, I like the creation of the subdirs, while passing in the top level path

core_api_c_binding/src/lib.rs
```diff
@@ -50,29 +49,35 @@ pub unsafe extern "C" fn holochain_new(ptr: *mut Dna) -> *mut Holochain {
...
+    create_path_if_not_exists(&eav_path)?;
+    Context::new(
+        agent,
+        Arc::new(Mutex::new(NullLogger {})),
+        Arc::new(Mutex::new(SimplePersister::new())),
+        FilesystemStorage::new(&cas_path).unwrap(),
```
AshantiMutinta: How does unwrap work when interacting with c extern functions? Can we use question mark instead?

lucksus: I only changed the indentation of this line. The C string is converted in line 54.

lucksus: :+1:

PR #526, opened 2018-10-23
Renamed ZomeApiGlobals and moved to wasm_utils

ddd-mtl:
Renamed everything called 'App' expect entry_type.

### DIFF
--- a/core/src/nucleus/ribosome/api/init_globals.rs
+++ b/core/src/nucleus/ribosome/api/init_globals.rs
@@ -1,17 +1,8 @@
 use nucleus::ribosome::Runtime;
 use wasmi::{RuntimeArgs, RuntimeValue, Trap};
-
 use serde_json;
-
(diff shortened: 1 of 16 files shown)


core/src/nucleus/ribosome/api/init_globals.rs
```diff
@@ -21,22 +12,18 @@ pub fn invoke_init_globals(
     runtime: &mut Runtime,
     _args: &RuntimeArgs,
 ) -> Result<Option<RuntimeValue>, Trap> {
-    let globals = InitGlobalsOutput {
-        app_name: runtime.dna_name.to_string(),
-
+    #[cfg_attr(rustfmt, rustfmt_skip)]
```
sphinxc0re: That seems unnecessary. Please don't align the values with spaces or at all. It makes this much harder to maintain

PR #527, opened 2018-10-23
HDK Rust Docs housekeeping, improvements Part 1

Connoropolous:
unify the AppGlobals struct, fixup hdk api comments


hdk-rust/src/api.rs
```diff
@@ -170,37 +153,22 @@ pub fn debug(msg: &str) -> Result<(), ZomeApiError> {
...
-    _data: S,
-    _pub_key: S,
-) -> ZomeApiResult<bool> {
-    // FIXME
-    Err(ZomeApiError::FunctionNotImplemented)
-}
-
-/// FIXME DOC
+/// Attempts to commit an entry to your local source chain. The entry
+/// will have to pass the defined validation rules for that entry type. Returns either an
```
lucksus:
```suggestion
/// will have to pass the defined validation rules for that entry type. If the entry type is defined as public, will also publish the entry to the DHT. Returns either an
```

Connoropolous:
thanks Damien! Just merged with develop again and will merge right once travis is done

Connoropolous:
merging because I tested locally, and there's one test failing on travis which isn't related to this PR. We are looking into it and going to fix

PR #529, opened 2018-10-23
hdk::get_links() MVP

lucksus:
[x] invoke_get_links
[x] tests
[x] hdk::get_links
[x] integration tests in HDK

# Links round-trip test:
Zome function:
```rust
    links_roundtrip: | | {
        let entry1_hash = hdk::commit_entry("testEntryType", json!({
            "stuff": "entry1"
        })).unwrap();
        let entry2_hash = hdk::commit_entry("testEntryType", json!({
            "stuff": "entry2"
        })).unwrap();
        let entry3_hash = hdk::commit_entry("testEntryType", json!({
            "stuff": "entry3"
        })).unwrap();


        hdk::link_entries(&entry1_hash, &entry2_hash, "test-tag").expect("Can't link?!");
        hdk::link_entries(&entry1_hash, &entry3_hash, "test-tag").expect("Can't link?!");

        match hdk::get_links(&entry1_hash, "test-tag") {
            Ok(links) => json!({"links": links}),
            Err(error) => json!({"error": error}),
        }
    }
```

Test in HDK-rust:
```rust
#[test]
fn can_roundtrip_links() {
    let (mut hc, _) = start_holochain_instance();
    let result = hc.call("test_zome", "test_cap", "links_roundtrip", r#"{}"#);
    assert!(result.is_ok(), "\t result = {:?}", result);
    let result_string = result.unwrap();
    let ordering1: bool = result_string == r#"{"links":["QmStYP5FYC61PfKKMYZpqBSMRJCAUeuSS8Vuz4EQL5uvK2","QmW6vfGv7fWMPQsgwd63HJhtoZmHTrf9MSNXCkG6LZxyog"]}"#;
    let ordering2: bool = result_string == r#"{"links":["QmW6vfGv7fWMPQsgwd63HJhtoZmHTrf9MSNXCkG6LZxyog","QmStYP5FYC61PfKKMYZpqBSMRJCAUeuSS8Vuz4EQL5uvK2"]}"#;
    assert!(ordering1 || ordering2);
}
```


hdk-rust/src/api.rs
```diff
@@ -340,8 +341,35 @@ pub fn remove_entry<S: Into<String>>(_entry: HashString, _message: S) -> ZomeApi
...
+    let result: GetLinksResult = load_json(encoded_allocation_of_result as u32)
+        .map_err(|err_str| ZomeApiError::Internal(err_str))?;
+
+    // Free result & input allocations and all allocations made inside commit()
+    mem_stack
+        .deallocate(allocation_of_input)
+        .expect("deallocate failed");
+
+    if result.ok {
+        Ok(result.links)
```
zippy: I think this should be just like get_entry() in that it should return the result (i.e. the LinkEntriesResult object) not the links out of the result because the result will end up having the meta data (crud status,sources) and so-on, so we should have that be the the same starting from now.  It's just a tiny change.

lucksus: Ok, will change this now.

core/src/nucleus/ribosome/api/get_links.rs
```diff
@@ -16,47 +15,132 @@ pub fn invoke_get_links(
...
+        links: get_links_result
+            .clone()
+            .unwrap_or(HashSet::new())
+            .iter()
+            .map(|eav| eav.value())
+            .collect::<Vec<Address>>(),
+        error: get_links_result
+            .map_err(|holochain_error| holochain_error.to_string())
+            .err()
+            .unwrap_or(String::from("")),
```
ddd-mtl: Note: With upcoming error system overhaul, we won't have to do this anymore and will have a standard way to send back an error, instead of putting it in the normal result struct.

hdk-rust/tests/integration_test.rs
```diff
@@ -226,10 +227,18 @@ fn has_populated_validation_data() {
...
     let result = hc.call("test_zome", "test_cap", "link_two_entries", r#"{}"#);
     assert!(result.is_ok(), "\t result = {:?}", result);
     assert_eq!(result.unwrap(), r#"{"ok":true}"#);
 }
+
+#[test]
+fn can_roundtrip_links() {
+    let (mut hc, _) = start_holochain_instance();
+    let result = hc.call("test_zome", "test_cap", "links_roundtrip", r#"{}"#);
+    assert!(result.is_ok(), "\t result = {:?}", result);
```
ddd-mtl: that `\t` doesn't seem necessary as the error message is wrapped in a panic format.

core/src/nucleus/ribosome/api/get_links.rs
lucksus: I prefer this explicit way actually. The results of these functions need to be understood by every HDK that will follow in the future. I would like to make it easily understandable by people not deep into Rust.

`ribosome_error_code!` above fooled me today as it does return an `Ok( <something> )`. Since everything that this function returns gets serialized into a string that then the HDK sees, returning special strings that are not listed in `wasm_utils::api_serialization` might confuse people. (It did confuse me already)

PR #531, opened 2018-10-23
520 persist agent state

AshantiMutinta:
Attempted to serialize the whole agentstate but with the time crunch and the many places that the 'Serialize Trait' touches it became a bigger job. This is a good compromise in the middle and the chain header can be loaded by just fetching eavs with the chain_header attribute.

Update  : Attemtping to implement a custom serializer for the agent state using the serde api. Using CAS instead of eav


lucksus:
I understand this code to add an EAV between every entry and its header on commit.
How does that persist (and restore?) the chain header?

AshantiMutinta:
> 
> 
> I understand this code to add an EAV between every entry and its header on commit.
> How does that persist (and restore?) the chain header?

A fetch on an "eav" of chain-header type would return every chain header on the hashset. From there you just navigate and get the latest from the data structure.

thedavidmeister:
i don't understand the direction this is going in

IIRC we'll run into circular dependencies as soon as we try to serialize/persist the EAV, which seems like it would be the next logical thing to persist after the CAS storage?

how often are we planning to persisting the store? is every reduction going to result in some data in the EAV?

why do we need custom serialization? what's bad about default JSON handling for structs/values containing the minimum data that we want persisted between save/load operations?

based on recent experience, i'm pretty wary of relying directly on serde to indiscriminately serialize whatever is thrown at it. it's far too easy to end up with `String` soup everywhere, never knowing which strings should be input for serde and which were output from serde.

could we get away with 1 standard struct like: 

```rs
AgentStoreSnapshot{
  keys: JsonString,
  top_pair: JsonString,
  cas_path: JsonString,
  eav_path: JsonString,
}
```

and just bounce this out to the file-system via JSON whenever appropriate?

AshantiMutinta:
> cies as soon as we try to serialize/persist the EAV, which seems like it would be the next logical thing to persist after the CAS storage?
> 
> how often are we planning to be persisting the store? is every reduction going to result in some data in the EAV?
> 
> why do we need custom serialization? what's bad about default JSON handling for structs/values containing the minimum data that we want persisted between save/load operations?

I reworked this during the night. Will push but basically I wrote a custom serializer for the agentstate that only takes in the variables we need. I am going to be using a CAS with a 'redux' address.

thedavidmeister:
> Will push but basically I wrote a custom serializer for the agentstate that only takes in the variables we need

implementing `From<AgentState> for AgentStateSnapshot` then using default serialization for the snapshot makes more sense to me, you can still cherry pick the minimum data to persist but we retain the type safety that we lose through serialization into a generic `String`

> I am going to be using a CAS with a 'redux' address.

interesting! looking forward to seeing what that looks like

AshantiMutinta:
> nting `From<Age`

Awesome, it should be noted that the there is some certain type of safety with the custom visitors and serializers, only comes during runtimes and 'OtherTypes' I am thinking for now, I can do it this way and then put in an issue to fix this asap? Only because of the dev camp crunch

AshantiMutinta:
> 
> 
> > nting `From<Age`
> 
> Awesome, it should be noted that the there is some certain type of safety with the custom visitors and serializers, only comes during runtimes and 'OtherTypes' I am thinking for now, I can do it this way and then put in an issue to fix this asap? Only because of the dev camp crunch

Actually let me try this first.

sphinxc0re:
But implementing from would imply a conversion. I don't think this is the case here. We can create a snapshot from an `AgentState` and it can continue being mutated

thedavidmeister:
ok, so is there a native trait that we should implement that covers this use case? @sphinxc0re @AshantiMutinta 

the main thing i want to avoid is "flattening" the type system down into `String` any earlier than absolutely necessary as it makes everything downstream much harder to work with

sphinxc0re:
There is no native trait for that AKAIK. I would just create a method on `AgentState`

core/src/agent/state.rs
```diff
@@ -30,6 +39,66 @@ pub struct AgentState {
...
 
+impl Serialize for AgentState {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        // 3 is the number of fields in the struct.
+        let mut state = serializer.serialize_struct("AgentState", 3)?;
+        state.serialize_field("keys", &self.chain)?;
+        state.serialize_field("chain_store", &self.chain)?;
```
lucksus: I don't think we need to serialize that one because that is just a wrapper around the CAS. Would mean we could avoid the custom serializer for the filesystemstorage.

core/src/agent/state.rs
```diff
@@ -59,9 +128,45 @@ impl AgentState {
...
         self.top_chain_header.clone()
     }
+
+    pub fn from_json_str(header_str: &str) -> serde_json::Result<Self> {
+        serde_json::from_str(header_str)
+    }
 }
 
-#[derive(Clone, Debug, PartialEq)]
+impl ToEntry for AgentState {
```
lucksus: This feels strange. The redux state is not an entry and we dont' want to commit redux states to source chains or share them in a DHT, right? Why do you want to treat the AgentState as an Entry?

lucksus:
Yeah, I resonate with @thedavidmeister's comment above about a snapshot. I would basically just write two functions that translate holochain_core::state::State (taking the whole thing, but just cherry picking agent().top_chain_header for now) into a string, and back. Using something like ```rust
#[derive(Serialize, Deserialize)]
struct HolochainStateSnapshot {
  pub top_chain_header: String
}
```
and:
```rust
fn serialize_state(state: State) -> String {
  serde_json::to_string(HolochainStateSnapshot{
    top_chain_header: state.agent().top_chain_header(),
  }).expect("Can't serialize HolochainStateSnapshot?!")
}
```

no?

core/src/agent/state.rs
AshantiMutinta: You can't initiate the agentstate without a CAS. Unless we make the CAS part of it an optional

core/src/agent/state.rs
AshantiMutinta: Just a copy and paste from addressable content. Will push this out.

core/src/agent/state.rs
```diff
@@ -133,15 +194,19 @@ fn reduce_commit_entry(
...
     fn response(
+        context: Arc<Context>,
         state: &mut AgentState,
         entry: &Entry,
         chain_header: &ChainHeader,
     ) -> Result<Address, HolochainError> {
         state.chain.content_storage().add(entry)?;
         state.chain.content_storage().add(chain_header)?;
+        let agent = AgentStateSnapshot::new(chain_header.clone());
+        state.chain.content_storage().add(&agent)?;
```
lucksus: Why adding the agent snapshot to the CAS?
And how could we get it out again, if that is meant as the persistence of the agent state?

core/src/agent/state.rs
```diff
@@ -61,7 +93,36 @@ impl AgentState {
...
+    }
+}
+
+impl ToJson for AgentStateSnapshot {
+    fn to_json(&self) -> Result<String, HolochainError> {
+        Ok(serde_json::to_string(self)?)
+    }
+}
+
+impl AddressableContent for AgentStateSnapshot {
```
thedavidmeister:
```suggestion
impl AddressableContent for AgentStateSnapshot {
    fn address(&self) -> Address {
        Address::from("AgentState")
    }
    
```

core/src/agent/state.rs
thedavidmeister: at first i was surprised/concerned by this

now i think it's actually important to do it this way, any given `AgentStateSnapshot` only makes sense for a specific CAS - at the hodl podl we said that we wanted to keep the "many CAS" option cleanly available

we can hardcode the `.address()` in the `AddressableContent` implementation for `AgentStateSnapshot` to something predictable and that should sidestep a whole bunch of potential "dangling snapshot" issues where a snapshot points to a missing/out of sync CAS.

core/src/agent/state.rs
```diff
@@ -30,6 +39,17 @@ pub struct AgentState {
     top_chain_header: Option<ChainHeader>,
 }
 
+#[derive(Debug, Deserialize, Serialize)]
+pub struct AgentStateSnapshot {
+    top_chain_header: ChainHeader,
+}
+
+impl AgentStateSnapshot {
+    pub fn top_chain_header(self) -> ChainHeader {
```
sphinxc0re: This method can only be called once 😁 
```suggestion
    pub fn top_chain_header(&self) -> ChainHeader {
```

core/src/agent/state.rs
lucksus: Hm, a deterministic address for the snapshot was the missing puzzle piece, at least for my perspective. Yeah, that sounds doable and interesting :)

core_api/src/lib.rs
```diff
@@ -112,8 +113,22 @@ impl Holochain {
         }
     }
 
-    pub fn load(_path: String, _context: Arc<Context>) -> Result<Self, HolochainError> {
-        Err(HolochainError::NotImplemented)
+    pub fn load(_path: String, context: Arc<Context>) -> Result<Self, HolochainError> {
```
lucksus: _path -> path

core/src/agent/state.rs
lucksus: +1

core/src/agent/state.rs
lucksus: +1

core/src/agent/state.rs
```diff
@@ -30,6 +39,17 @@ pub struct AgentState {
...
 }
 
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct AgentStateSnapshot {
+    top_chain_header: ChainHeader,
+}
+
+impl AgentStateSnapshot {
+    pub fn top_chain_header(&self) -> ChainHeader {
+        self.top_chain_header.clone()
```
sphinxc0re: I would return a reference to the `top_chain_header` and leave it to the callee to clone it 🤔

PR #532, opened 2018-10-23
Rename AGENT_KEY_HASH to AGENT_ADDRESS

lucksus:
*WAS: Fill `AGENT_KEY_HASH` with hash of agent entry.*

As we discussed today in our heartbeat, `AGENT_KEY_HASH` is the thing to hang an agent's entry off of. But, the name is misleading as we are also discussing to have a special treatment of public keys as to not hash them but have the key be its address.

These details are not relevant to the usage of this anchor from the perspective of the app developer. We just need something that is in the address space of all entries and that represents the agent. Be it the hash of a key or the key itself or something different in the future.. In all cases, it is the address used to talk about an agent. Hence `AGENT_ADDRESS`.

### DIFF
--- a/core/src/nucleus/ribosome/api/init_globals.rs
+++ b/core/src/nucleus/ribosome/api/init_globals.rs
@@ -19,7 +19,7 @@ pub fn invoke_init_globals(
         dna_hash: HashString::from(""),
         agent_id_str: runtime.context.agent.to_string(),
         // TODO #233 - Implement agent pub key hash
-        agent_key_hash: HashString::encode_from_str("FIXME-agent_key_hash", Multihash::SHA2256),
+        agent_address: HashString::encode_from_str("FIXME-agent_address", Multihash::SHA2256),
         agent_initial_hash: HashString::from(""),
         agent_latest_hash: HashString::from(""),
     };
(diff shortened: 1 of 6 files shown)


thedavidmeister:
- as per a squirl in the hodl podl
  - https://hackmd.io/D65QC7fOQbe286K8FQuz_A
  - agent entry address will be a combination of public signing, encryption and dpki keys
- personally i have no problems linking to addresses that don't hold (known) entries
  - i have not thought much about possible ramifications of this!
  - linking to/from hardcoded hashes like `000..000` and `000...001` etc. might be a useful hack in some situations, or a really bad anti-pattern 🤔 i dunno?
- it would be pretty easy (after i land the new entry definition setup) to have an entry type that uses the identity function for entry addresses, rather than sha256.
  - this would allow things like public keys to be themselves in the DHT
  - we wouldn't need the hack ^^ because it would be really easy to drop real entries into any address we want at that point

lucksus:
@ddd-mtl, @thedavidmeister please note the changed title and description and please re-review :)

ddd-mtl:
I don't understand if `AGENT_ADDRESS` is supposed to be different from `AGENT_INITIAL_HASH` or not? Changing the name does't change the fact that having 2 global const with the same value doesn't make any sense. If its an API thing then it could be done in hdk-rust only. No need to pass this around through wasm memory.

PR #536, opened 2018-10-24
hdk::query()

ddd-mtl:
Added basic query() to ChainStore.

interface:
```Rust
pub struct QueryArgs {
    pub entry_type_name: String,
    pub limit: u32,
}

pub struct QueryResult {
    pub hashes: Vec<HashString>,
}
```

### DIFF
--- a/core/src/agent/chain_store.rs
+++ b/core/src/agent/chain_store.rs
@@ -1,5 +1,6 @@
 use holochain_core_types::{
     cas::storage::ContentAddressableStorage, chain_header::ChainHeader, entry_type::EntryType,
+    cas::content::Address,
 };
 
 #[derive(Debug, PartialEq, Clone)]
@@ -37,6 +38,22 @@ where
             self.iter(start_chain_header)
                 .find(|chain_header| chain_header.entry_type() == entry_type),
         )
+    }
+
+    pub fn query(
+        &self,
+        start_chain_header: &Option<ChainHeader>,
+        entry_type: EntryType,
+        limit: u32)
+        -> Vec<Address> {
+        let mut result: Vec<Address> = Vec::new();
+        for header in self.iter_type(start_chain_header, &entry_type) {
+            result.push(header.entry_address().clone());
+            if limit != 0 && result.len() as u32 >= limit {
+                break;
+            }
+        }
+        result
     }
 }
 
@@ -139,8 +156,8 @@ pub mod tests {
     use holochain_core_types::{
         cas::{content::AddressableContent, storage::ContentAddressableStorage},
         chain_header::{test_chain_header, ChainHeader},
-        entry::{test_entry, test_entry_b},
-        signature::{test_signature, test_signature_b},
+        entry::{test_entry, test_entry_b, test_entry_c},
+        signature::{test_signature, test_signature_b, test_signature_c},
         time::test_iso_8601,
     };
 
@@ -261,4 +278,54 @@ pub mod tests {
         }
         assert_eq!(expected, found);
     }
+
+    #[test]
+    /// show query() implementation
+    fn query_test() {
+        let chain_store = test_chain_store();
+
+        let chain_header_a = test_chain_header();
+        let entry = test_entry_b();
+        let chain_header_b = ChainHeader::new(
+            &entry.entry_type(),
+            &entry.address(),
+            &test_signature_b(),
+            &Some(chain_header_a.address()),
+            &None,
+            &test_iso_8601(),
+        );
+        let entry = test_entry_c();
+        let chain_header_c = ChainHeader::new(
+            &entry.entry_type(),
+            &entry.address(),
+            &test_signature_c(),
+            &Some(chain_header_b.address()),
+            &Some(chain_header_b.address()),
+            &test_iso_8601(),
+        );
+
+        chain_store
+            .content_storage()
+            .add(&chain_header_a)
+            .expect("could not add header to cas");
+        chain_store
+            .content_storage()
+            .add(&chain_header_b)
+            .expect("could not add header to cas");
+        chain_store
+            .content_storage()
+            .add(&chain_header_c)
+            .expect("could not add header to cas");
+
+        let found = chain_store.query(&Some(chain_header_c.clone()), entry.entry_type().to_owned(), 0);
+        let expected = vec![
+            chain_header_c.entry_address().clone(),
+            chain_header_b.entry_address().clone()];
+        assert_eq!(expected, found);
+
+        let found = chain_store.query(&Some(chain_header_c.clone()), entry.entry_type().to_owned(), 1);
+        let expected = vec![chain_header_c.entry_address().clone()];
+        assert_eq!(expected, found);
+    }
+
 }
--- a/core/src/nucleus/ribosome/api/query.rs
+++ b/core/src/nucleus/ribosome/api/query.rs
@@ -1,0 +1,38 @@
+use holochain_core_types::{
+    entry_type::EntryType,
+};
+use holochain_wasm_utils::api_serialization::{QueryArgs, QueryResult};
+use nucleus::ribosome::Runtime;
+use serde_json;
+use std::str::FromStr;
+use wasmi::{RuntimeArgs, RuntimeValue, Trap};
+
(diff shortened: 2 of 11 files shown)


core/src/nucleus/ribosome/api/query.rs
```diff
@@ -0,0 +1,38 @@
...
+    let input: QueryArgs = match serde_json::from_str(&args_str) {
+        Ok(input) => input,
+        Err(_) => return ribosome_error_code!(ArgumentDeserializationFailed),
+    };
+    // Get entry_type
+    let maybe_entry_type = EntryType::from_str(&input.entry_type_name);
+    if maybe_entry_type.is_err() {
+        return ribosome_error_code!(UnknownEntryType);
+    }
+    let entry_type = maybe_entry_type.unwrap();
```
sphinxc0re: Better do

```rust
let entry_type = EntryType::from_str(&input.entry_type_name)
    .map_err(|_| ribosome_error_code!(UnknownEntryType))?;
```
It's much cleaner

hdk-rust/wasm-test/src/lib.rs
```diff
@@ -167,6 +168,47 @@ zome_functions! {
     }
 }
 
+zome_functions! {
+    check_query: | | {
+
+        // Query DNA entry
+        let result = hdk::query(&EntryType::Dna.to_string(), 0);
+        assert!(result.is_ok());
+        assert!(result.unwrap().len() == 1);
```
sphinxc0re: This way, we don't know why the test failed. Better do something like:

```rust
let query_result = hdk::query(&EntryType::Dna.to_string(), 0).expect("blah blah blah");
assert!(query_result.len() == 1);
```

Does this even work with wasm?

hdk-rust/tests/integration_test.rs
```diff
@@ -242,3 +243,20 @@ fn can_roundtrip_links() {
...
+        "test_cap",
+        "check_query",
+        r#"{ "entry_type_name": "testEntryType", "limit": "0" }"#,
+    );
+    assert!(result.is_ok(), "result = {:?}", result);
+    assert_eq!(
+        result.unwrap(),
+        r#"["QmStYP5FYC61PfKKMYZpqBSMRJCAUeuSS8Vuz4EQL5uvK2"]"#,
+    );
+}
```
sphinxc0re: Small one: Missing newline

core/src/nucleus/ribosome/api/query.rs
lucksus: Sadly, that doesn’t work. ˋribosome_error_code!ˋ does not return an Err but an Ok..

sphinxc0re: Okay, then that's wrong. An error is an error. An ok value can't be an error

thedavidmeister: well yes it can be

ok means "the zome ran" not "the result of the zome had no errors"

thedavidmeister: would also be out of scope to change that

sphinxc0re: I agree, but we should discuss this. This is logically and semantically wrong.

hdk-rust/src/api.rs
```diff
@@ -373,8 +374,27 @@ pub fn get_links<S: Into<String>>(base: &HashString, tag: S) -> ZomeApiResult<Ge
 }
 
 /// Not Yet Available
-pub fn query() -> ZomeApiResult<Vec<String>> {
-    Err(ZomeApiError::FunctionNotImplemented)
+/// entry_type: Specify type of entry to retrieve
```
thedavidmeister:
```suggestion
/// entry_type_name: Specify type of entry to retrieve
```

hdk-rust/tests/integration_test.rs
thedavidmeister:
```suggestion
}

```

core_types/src/signature.rs
```diff
@@ -14,3 +14,7 @@ pub fn test_signature() -> Signature {
 pub fn test_signature_b() -> Signature {
     Signature::from("another-fake-signature")
 }
+
+pub fn test_signature_c() -> Signature {
+    Signature::from("sig-c")
+}
```
thedavidmeister:
```suggestion
}

```

core/src/agent/chain_store.rs
```diff
@@ -38,6 +39,22 @@ where
                 .find(|chain_header| chain_header.entry_type() == entry_type),
         )
     }
+
+    pub fn query(
+        &self,
+        start_chain_header: &Option<ChainHeader>,
```
thedavidmeister: it seems weird that this is in the query here but not exposed in `QueryArgs`

extra weird that `limit` is exposed in `QueryArgs`

so we almost have pagination but not quite.

it makes sense to me that `limit` and `start` would almost always be used together to page through results?

core/src/nucleus/ribosome/api/query.rs
lucksus: It did confuse me already.

hdk-rust/src/api.rs
```diff
@@ -373,8 +374,27 @@ pub fn get_links<S: Into<String>>(base: &HashString, tag: S) -> ZomeApiResult<Ge
...
+/// limit: max number of entries to retrieve
```
Connoropolous:
```suggestion
/// limit: Max number of entries to retrieve
```

hdk-rust/src/api.rs
```diff
@@ -373,8 +374,27 @@ pub fn get_links<S: Into<String>>(base: &HashString, tag: S) -> ZomeApiResult<Ge
 }
 
 /// Not Yet Available
```
Connoropolous:
```suggestion
/// Returns a list of entries from your local source chain, that match a given type.
```

core/src/nucleus/ribosome/api/query.rs
ddd-mtl: I designed I new error system: https://drive.google.com/drive/u/1/folders/12VqxGg_kar2mW_HESP9kKAQnJOuT-qBy

How's that for a discussion? :)

hdk-rust/wasm-test/src/lib.rs
ddd-mtl: Not really. You get a "Trap: Unreachable" in any case.
I made a zome_assert! macro but it doesn't work well inside the zome_functions! macro for now.

core/src/agent/chain_store.rs
ddd-mtl: 'start' and 'end' would make sense yes, but 'limit' is not 'end'.
'end' would be a chainHeader address where you would have to stop iterating, which is different from the limit of number of results the query/find gives back.

ddd-mtl:
Can someone do rustfmt please?

lucksus:
rustfmt done, @ddd-mtl

wasm_utils/src/api_serialization/query.rs
```diff
@@ -0,0 +1,12 @@
...
+
+#[derive(Deserialize, Default, Debug, Serialize)]
+pub struct QueryArgs {
+    pub entry_type_name: String,
+    pub limit: u32,
+}
+
+#[derive(Deserialize, Default, Debug, Serialize)]
+pub struct QueryResult {
+    pub hashes: Vec<HashString>,
```
Connoropolous: Could we make this `addresses`? please, for consistency?

ddd-mtl: FYI, this is internal and not public facing.

Connoropolous: well... it get passed through the wasm interface, so its sorta public-y

ddd-mtl: then I guess we should also change the globals from hash to address like:
DNA_HASH -> DNA_ADDRESS

Connoropolous: yeah... its currently a mix, but Id love to get to address in all the right places. Just merge this as is if you feel its best, i will unblock

PR #557, opened 2018-10-25
Persistence fix: Load DNA from source chain

lucksus:
For holochain_load() to really work we need to construct a state with everything a running Instance needs. And that means: the DNA.

* retrieving the DNA from the source chain on load
* putting it into the state and making sure that state is used in the newly created instance

### DIFF
--- a/dna/src/lib.rs
+++ b/dna/src/lib.rs
@@ -264,7 +264,7 @@ impl ToEntry for Dna {
     }
 
     fn from_entry(entry: &Entry) -> Self {
-        return Dna::from_json_str(&entry.value()).expect("entry is not a valid Dna Entry");
+        Dna::from_json_str(&entry.value()).expect("entry is not a valid Dna Entry")
     }
 }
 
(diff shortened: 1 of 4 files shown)


dna/src/lib.rs
```diff
@@ -264,7 +264,7 @@ impl ToEntry for Dna {
     }
 
     fn from_entry(entry: &Entry) -> Self {
-        return Dna::from_json_str(&entry.content()).expect("entry is not a valid Dna Entry");
+        return Dna::from_json_str(&entry.value()).expect("entry is not a valid Dna Entry");
```
lucksus: THIS was a mean bug!

sphinxc0re: Please remove the `return`

lucksus: jeez

lucksus:
Pushed removal of return, @sphinxc0re. Please merge at will.
(and just go ahead push such little changes in the future generally)

PR #558, opened 2018-10-25
Implemented hdk::call()

ddd-mtl:


hdk-rust/src/api.rs
```diff
@@ -159,12 +159,48 @@ pub fn debug(msg: &str) -> ZomeApiResult<()> {
 
 /// Not Yet Available
```
lucksus: Rust doc? @Connoropolous?

zippy: Just a simple thing like `call() allows you to call a function in another zome.`  should suffice

PR #562, opened 2018-10-27
clarify how to add a holochain back end to existing project

jryanhaber:
Suggesting we clarify this, running into it myself.


doc/holochain_101/src/new_project.md
```diff
@@ -2,7 +2,7 @@
 
 The command line tools discussed in the last article can be used to easily create a new folder on your computer, that contains all the initial folders and files needed for a Holochain application. 
 
-You will typically want to start a new Holochain application this way, since creating them all by hand would waste your time.
+You will typically want to start a new Holochain application this way, since creating them all by hand would waste your time.  If you are adding a holochain back end to an existing project, then you will typically follow these same steps.
```
Connoropolous: What if we just rephrased the original sentence to "You will typically want to create a new project folder for a Holochain application this way..."

since the addition of "project folder" can mean that it doesn't need to be an app that you haven't started yet, it just means a new project folder, for whatever you're working on

jryanhaber: My sense is that we could aim towards addressing the questions of 
"what if its for a new holo app, how do I proceed" 
and 
"what if its for an existing project, adding holochain, how do I proceed?"
in one go ... 

Both are questions I found myself asking when I got to this part of the guide... 

But your suggestion does actually feel like an improvement so I'll edit the pr.

PR #566, opened 2018-10-29
[ERROR] step4: CoreError instead of *Result structs

ddd-mtl:
Changed `RibosomeErrorReport` to `CoreError`. 
and related macros: `ribosome_error_report!()` -> `core_error!()`
Changed `memory_serialization::load_*()` to return a HolochainError.
Removed `GetLinksResult`, `LinkEntriesResult` & `QueryResult` and use CoreError instead.
Created type:
`pub type ZomeApiResult = Result<Option<RuntimeValue>, Trap>`
Created helper function `Runtime::store_as_json()`
Added `Ribosome(RibosomeErrorCode)` variant to HolochainError so that the Zome has a way to send RibosomeErrorCode back to the Holochain Instance.


hdk-rust/src/api.rs
```diff
@@ -425,38 +404,31 @@ pub fn remove_entry<S: Into<String>>(_entry: HashString, _message: S) -> ZomeApi
...
-/// Consumes two values, the first of which is the address of an entry, `base`, and the second of which is a string, `tag`,
-/// used to describe the relationship between the `base` and other entries you wish to lookup. Returns a list of addresses of other
-/// entries which matched as being linked by the given `tag`. Links are created in the first place using the Zome API function `link_entries`.
-pub fn get_links<S: Into<String>>(base: &HashString, tag: S) -> ZomeApiResult<GetLinksResult> {
+/// Consumes two values, the first of which is the address of an entry, `base`,
+/// and the second of which is a string, `tag`, used to describe the relationship between the `base`
+/// and other entries you wish to lookup. Returns a list of addresses of other entries which matched
+/// as being linked by the given `tag`.
+/// Links are created in the first place using the Zome API function `link_entries`.
+pub fn get_links<S: Into<String>>(base: &HashString, tag: S) -> ZomeApiResult<Vec<HashString>> {
```
lucksus: Do we really want to get rid of a GetLinksResult? I suppose there will be more information that should be available to the zome around links, esp. crud status and DHT saturation etc... I thought these result structs where partly placeholder for where we can add stuff in the (not too distant) future without breaking our API signature. @zippy?

lucksus:
Nice work! I love how this simplifies the return-value creation in our `invoke_*` functions. As I mentioned, I'm not sure if we really want to get rid of return-value structs for the sake of stable API function signatures.. Would love get a comment on that from others as well and abstain from approving for now. But will fix the fmt errors.

hdk-rust/src/api.rs
zippy: Yes, that's what I was expecting, that it would work exactly like get_entry with an options so you can say if you want sources etc.

ddd-mtl: ...

ddd-mtl:
@lucksus Thanks for develop merge and fmt.
I put back QueryResult and GetLinksResult. I didn't put back LinkEntriesResult as there is no intention to add more there in the future.

PR #572, opened 2018-10-29
small change - move riker ActorSystem & Protocol code into core_types

neonphog:
Moves ActorSystem and Protocol into core_types to centralize access from other crates (e.g. p2p networking).

This PR is no code change, just the relocation.


zippy:
Hmm, I think you aren't going to get buy-in on this one, because I think the intent of core-types was to not include things that shouldn't be available on the wasm side.  @lucksus?

neonphog:
@lucksus @zippy - spoke with @thedavidmeister - we're going to just have two ActorSystems / Protocols for the moment - closing this PR

lucksus:
Is there a way to not use Riker for `net`? We will want to compile Holochain to WASM for Holo soon. Putting Riker into `net` which we need in `core` makes that impossible, currently. @neonphog

neonphog:
@lucksus @zippy - since this PR is closed, moved discussion over to an issue: https://github.com/holochain/holochain-rust/issues/574

PR #576, opened 2018-10-30
Changed Link EAV attribute name from 'link:{}' to 'link__{}'

ddd-mtl:
because ':' is not allowed for filesystem paths on windows and currently eav file storage is using the attribute name as a folder name.


thedavidmeister:
i've added to the acorn board a definition/validation of what characters _are_ allowed in EAV attribute names

otherwise this will come up again

PR #578, opened 2018-10-31
[WIP]Container toml parsing

AshantiMutinta:
Taking the structure from https://hackmd.io/OcT2cI1ETfu4DHyvn4QZ5A?both#

This is creating the initial steps into loading a toml and create a container out of it. For the sake of compile time validation, configurations are organized as structs(in the future items with types will be enums which will allow for dynamic creation of traits or structs).

The only problem is that for every configuration and field that is put in the toml, we would have to reflect it in the binary which can get annoying but I would rather have compile time checking as that is where the compiler shines at.

There are areas of using an "Any" type but it would make it a mess to work with.
https://doc.rust-lang.org/std/any/trait.Any.html

There is also the question of do we need these configuration types or should we just use the existing structs. Is that a redesign on our TOML or the application itself?


core/src/container/config.rs
```diff
@@ -0,0 +1,97 @@
...
+#[derive(Deserialize)]
+pub struct InstanceConfiguration {
+    id: String,
+    #[serde(rename = "DNA")]
+    dna: String,
+}
+
+#[derive(Deserialize)]
+pub struct Configuration {
+    #[serde(rename = "Agents")]
```
sphinxc0re: I would actually call that one `agent`. Because in toml, you'll be able declare an agent configuration by saying:

```toml
[[agent]]
property = "value"
...

[[agent]]
property = "value"
...

[[agent]]
property = "value"
...


```

It's the same with the `Cargo.toml`. They use singular for lists to make items more intuitive as in: "I'm describing an agent.... now another agent"

sphinxc0re: Also if this is optional anyway, you might wanna add `#[serde(default)]` to the field. This way you can drop the `Option<T>`

PR #580, opened 2018-11-02
Point wabt to my fork which fixes the build.rs for windows msvc

ddd-mtl:


ddd-mtl:
My hack is based on the currently open issue: https://github.com/pepyakin/wabt-rs/issues/20
If it gets fixed we will be able to point back to it.
My fix is hack-ish  so I don't intend to submit a PR fix. It is okay for us because we only use it for test.

Connoropolous:
Okay cool.

pepyakin:
Fwiw, the problem mentioned here was fixed. Updated version was released on crates.io under 0.7.2 version!

PR #593, opened 2018-11-06
CI Travis windows

ddd-mtl:
Compiles but does not launch tests as wabt requires python which we can't install for windows when there are secrets in the travis repo settings.


sphinxc0re:
Okay, it seems like you can't install these via `choco`. But you can install them via the official installers (http://zeromq.org/distro:microsoft-windows)

sphinxc0re:
There is one repo that lets you install zmq via `choco`: https://github.com/keenbrowne/chocolatey-zeromq

ddd-mtl:
@sphinxc0re How do I tell travis to install that chocolatey package?

ddd-mtl:
It won't work because its v4.0.4. We need at least v4.1

sphinxc0re:
@ddd-mtl `choco` is installed automatically. You just have to use it. `choco install ...`

zippy:
Doing an admin merge to get this working, despite the code-coverage and binaries issue.  The former we need to address, but is not specific to this pull request as it happend after JsonString.  The latter we will handle as a follow-up when we have more staffing to handle DevOps.

PR #594, opened 2018-11-06
504 cas eav refactor

AshantiMutinta:
refactoring EAV and CAS to use trait objects instead of concrete. Some of the extra implemented trait objects were done because Arc mutexes cannot derive them and for a trait to be a trait object it cannot come from a supertrait of clone or partialord as they both content self references.


core/src/agent/state.rs
```diff
@@ -194,13 +192,20 @@ fn reduce_get_entry(
...
     let address = unwrap_to!(action => Action::GetEntry);
+    let storage = &state.chain().content_storage().clone();
+    let json = storage
+        .read()
+        .unwrap()
+        .fetch(&address)
+        .expect("could not fetch from CAS");
+    let result: Option<SerializedEntry> = match json {
+        Some(js) => match js.try_into() {
+            Ok(info) => Some(info),
```
lucksus: Not really important, but why `info`? This is the SerializedEntry, no?

core/src/context.rs
```diff
@@ -23,8 +24,8 @@ pub struct Context {
     state: Option<Arc<RwLock<State>>>,
     pub action_channel: SyncSender<ActionWrapper>,
     pub observer_channel: SyncSender<Observer>,
-    pub file_storage: FilesystemStorage,
-    pub eav_storage: EavFileStorage,
```
lucksus: Yay! :)

core/src/agent/state.rs
AshantiMutinta: oh that is a typo, i think I meant to type "into" which is not a really good name in itself haha. I will change that!

sphinxc0re: Also, you can replace this `match` statement with a call to `.ok()`

sphinxc0re: it converts a `Result<T, E>` to a `Option<T>`

AshantiMutinta: Oh yes! That is a good idea! Will make those changes! @sphinxc0re

sphinxc0re:
```rust
let result: Option<SerializedEntry> = json.and_then(|js| js.try_into().ok())
```

PR #597, opened 2018-11-07
net_connection p2p abstraction - STEP 1

neonphog:
Don't be scared by the line count! It's mostly unit tests and `From` trait helpers.

The code submitted here replaces most of the functionality in the `net` and `net_ipc` crates. To see a more complete integration / usage, you can see my [net-connector](https://github.com/holochain/holochain-rust/compare/net-connection-step1...net-connector) branch + the [test_bin](https://github.com/holochain/holochain-rust/blob/net-connector/test_bin/src/test_bin_ipc.rs#L45).

We'll still need those crates, they'll just be much simpler because the message protocol for the abstraction is taken care of here.

Main Points:

- `Protocol` defines a simple low-level protocol for communicating with an abstract p2p module
- `ProtocolWrapper` defines a higher-level developer friendly wrapper around protocol
- `NetConnection / NetWorker` define a light-weight messaging strategy that can easily be integrated with Riker should we decide to in the future. For now, we just get `Protocol` messages sent to the closure passed in when the system is initialized.

Next Steps:

In an effort to have smaller pull requests, I broke this down into separate PRs, if this is merged, expect a quick follow on:

- `STEP 2` - will delete all the unused parts of the `net` and `net_ipc` crates.
- `STEP 3` - will add back in the ipc zmq socket management in the `net_ipc` crate, but using the `net_connection` abstraction
- `STEP 4` - will update the `net` crate with a `P2pNetwork` class that uses the `net_connection` abstraction, as well as a `test_bin` binary that exercises it.

### DIFF
--- a/net_connection/Cargo.toml
+++ b/net_connection/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "holochain_net_connection"
+version = "0.1.0"
+authors = ["neonphog <neonphog@gmail.com>"]
+
+[dependencies]
+byteorder = "1"
+failure = "0.1.3"
+holochain_core_types = { path = "../core_types" }
+rmp = "0.8"
+rmp-serde = "0.13.7"
+serde = "1"
+serde_bytes = "0.10.4"
+serde_derive = "1"
+serde_json = "1"
--- a/net_connection/src/net_connection_thread.rs
+++ b/net_connection/src/net_connection_thread.rs
@@ -1,0 +1,183 @@
+use super::NetResult;
+
+use super::{
+    net_connection::{NetConnection, NetHandler, NetWorkerFactory},
+    protocol::Protocol,
+};
+
+use std::{thread, time};
+
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    mpsc, Arc,
+};
+
+/// a NetConnection instance that is managed on another thread
+#[derive(Debug)]
+pub struct NetConnectionThread {
+    keep_running: Arc<AtomicBool>,
+    send_channel: mpsc::Sender<Protocol>,
+    thread: thread::JoinHandle<()>,
+}
+
+impl NetConnection for NetConnectionThread {
+    /// send a message to the worker within this NetConnectionThread instance
+    fn send(&mut self, data: Protocol) -> NetResult<()> {
+        self.send_channel.send(data)?;
+        Ok(())
+    }
+}
+
+impl NetConnectionThread {
+    /// stop (join) the worker thread
+    pub fn stop(self) -> NetResult<()> {
+        self.keep_running.store(false, Ordering::Relaxed);
+        match self.thread.join() {
+            Ok(_) => Ok(()),
+            Err(_) => {
+                bail!("NetConnectionThread failed to join on stop() call");
+            }
+        }
+    }
+
+    /// create a new NetConnectionThread instance with given handler / worker
+    pub fn new(handler: NetHandler, mut worker_factory: NetWorkerFactory) -> NetResult<Self> {
+        let keep_running = Arc::new(AtomicBool::new(true));
+        let keep_running2 = keep_running.clone();
+
+        let (sender, receiver) = mpsc::channel();
+        Ok(NetConnectionThread {
+            keep_running,
+            send_channel: sender,
+            thread: thread::spawn(move || {
+                let mut us = 100_u64;
+                let mut worker = match worker_factory(handler) {
+                    Ok(w) => w,
+                    Err(e) => panic!("{:?}", e),
+                };
+
+                while keep_running2.load(Ordering::Relaxed) {
+                    let mut did_something = false;
+
+                    match receiver.try_recv() {
+                        Ok(data) => {
+                            did_something = true;
+                            match worker.receive(data) {
+                                Ok(_) => (),
+                                Err(e) => panic!("{:?}", e),
+                            };
+                        }
+                        Err(_) => (),
+                    };
+
+                    match worker.tick() {
+                        Ok(b) => {
+                            if b {
+                                did_something = true;
+                            }
+                        }
+                        Err(e) => panic!("{:?}", e),
+                    };
+
+                    if did_something {
+                        us = 100_u64;
+                    } else {
+                        us *= 2_u64;
+                        if us > 10_000_u64 {
+                            us = 10_000_u64;
+                        }
+                    }
+
+                    thread::sleep(time::Duration::from_micros(us));
+                }
+            }),
+        })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use super::super::net_connection::NetWorker;
+
+    struct DefWorker;
+
+    impl NetWorker for DefWorker {}
+
+    #[test]
+    fn it_can_defaults() {
+        let mut con = NetConnectionThread::new(
+            Box::new(move |_r| Ok(())),
+            Box::new(|_h| Ok(Box::new(DefWorker))),
+        ).unwrap();
+
+        con.send("test".into()).unwrap();
+        con.stop().unwrap();
+    }
+
+    struct Worker {
+        handler: NetHandler,
+    }
+
+    impl NetWorker for Worker {
+        fn tick(&mut self) -> NetResult<bool> {
+            (self.handler)(Ok("tick".into()))?;
+            Ok(true)
+        }
+
+        fn receive(&mut self, data: Protocol) -> NetResult<()> {
+            (self.handler)(Ok(data))
+        }
+    }
+
+    #[test]
+    fn it_invokes_connection_thread() {
+        let (sender, receiver) = mpsc::channel();
+
+        let mut con = NetConnectionThread::new(
+            Box::new(move |r| {
+                sender.send(r?)?;
+                Ok(())
+            }),
+            Box::new(|h| Ok(Box::new(Worker { handler: h }))),
+        ).unwrap();
+
+        con.send("test".into()).unwrap();
+
+        let res;
+        loop {
+            let tmp = receiver.recv().unwrap();
+
+            if &(tmp.as_json_string()) == "tick" {
+                continue;
+            } else {
+                res = tmp;
+                break;
+            }
+        }
+
+        assert_eq!("test".to_string(), res.as_json_string());
+
+        con.stop().unwrap();
+    }
+
+    #[test]
+    fn it_can_tick() {
+        let (sender, receiver) = mpsc::channel();
+
+        let con = NetConnectionThread::new(
+            Box::new(move |r| {
+                sender.send(r?)?;
+                Ok(())
+            }),
+            Box::new(|h| Ok(Box::new(Worker { handler: h }))),
+        ).unwrap();
+
+        let res = receiver.recv().unwrap();
+
+        assert_eq!("tick".to_string(), res.as_json_string());
+
+        con.stop().unwrap();
+    }
+}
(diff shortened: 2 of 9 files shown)


net_connection/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+[dependencies]
+byteorder = "1"
+failure = "0.1.3"
+holochain_core_types = { path = "../core_types" }
+rmp = "0.8"
+rmp-serde = "0.13.7"
+serde_bytes = "0.10.4"
+serde = "1.0"
+serde_derive = "1.0"
+serde_json = { version = "1.0", features = ["preserve_order"] }
```
thedavidmeister: i just set this to preserve order

net_connection/src/protocol.rs
```diff
@@ -0,0 +1,276 @@
...
+impl<'a> From<&'a Protocol> for NamedBinaryData {
+    fn from(p: &'a Protocol) -> Self {
+        match p {
+            Protocol::NamedBinary(nb) => NamedBinaryData {
+                name: b"namedBinary".to_vec(),
+                data: rmp_serde::to_vec_named(nb).unwrap(),
+            },
+            Protocol::Json(j) => NamedBinaryData {
+                name: b"json".to_vec(),
+                data: String::from(j).as_bytes().to_vec(),
```
thedavidmeister: `JsonString` offers `into_bytes()` that gives a `Vec<u8>` of the inner string, should be what you want?

net_connection/src/protocol.rs
```diff
@@ -0,0 +1,276 @@
...
+impl Protocol {
+    simple_access! {
+        is_named_binary as_named_binary NamedBinary NamedBinaryData
+        is_json as_json Json JsonString
+        is_ping as_ping Ping PingData
+        is_pong as_pong Pong PongData
+    }
+
+    /// get a json string straight out of the Protocol enum
+    pub fn as_json_string(&self) -> String {
```
thedavidmeister: should return `JsonString` no?

there are good reasons not to put known-json data into a raw `String` primitive, documented https://github.com/holochain/holochain-rust/pull/490 and https://hackmd.io/EmRHEEpuRRihr_K7MThVZg#centralize-serialization-logic

if there is a use-case for having raw `String` values containing json floating around, i'd like to know so i can maybe extend what i've done to cover it?

net_connection/src/protocol_wrapper.rs
```diff
@@ -0,0 +1,433 @@
...
+
+    /// [send] send our response to a previous `HandleSend`
+    HandleSendResult(SendResultData),
+}
+
+/// upgrade a Protocol reference to a ProtocolWrapper instance
+impl<'a> From<&'a Protocol> for ProtocolWrapper {
+    fn from(p: &'a Protocol) -> Self {
+        if let Protocol::Json(json) = p {
+            let json: serde_json::Value = serde_json::from_str(json.into()).unwrap();
```
thedavidmeister: hmm, this doesn't feel right

firstly an unwrap is dangerous when deserializing JSON as it quite commonly does not deserialize, especially as our data structures are unstable and tend to drift from month to month atm

secondly, could this be represented with an enum that we do `MyEnum::try_from(json)?` against
the matching can then be done against the enum variants rather than raw serde traversals

net_connection/src/protocol_wrapper.rs
```diff
@@ -0,0 +1,433 @@
...
+
+/// downgrade a ProtocolWrapper instance back into a Protocol message
+impl<'a> From<&'a ProtocolWrapper> for Protocol {
+    // david.b (neonphog) - tarpaulin doesn't cover macros... skipping this fn
+    #[cfg_attr(tarpaulin, skip)]
+    fn from(w: &'a ProtocolWrapper) -> Self {
+        match w {
+            ProtocolWrapper::RawProtocol(p) => p.clone(),
+            ProtocolWrapper::RequestState => Protocol::Json(
+                json!({
```
thedavidmeister: this approach is fragile as the json structure must be manually maintained across everywhere that handles it

a more robust technique is to define a struct/enum and derive `DefaultJson`

something like this:

```rust
#[derive(Debug, Serialize, Deserialize, DefaultJson)]
pub struct RequestStateData{ method: Method }

Protocol::Json(RequestStateData{ method: Method::RequestState }.into())
```

then later:

```rust
RequestStateData::try_from(json)?
```

net_connection/src/protocol.rs
```diff
@@ -0,0 +1,276 @@
...
+        if let Protocol::Json(data) = self {
+            String::from(data)
+        } else {
+            panic!("as_json_string called with bad type");
```
thedavidmeister: this means the compiler is broken >.<

should be `unreachable()!` really

net_connection/src/protocol_wrapper.rs
```diff
@@ -0,0 +1,433 @@
...
+    fn from(p: &'a Protocol) -> Self {
+        if let Protocol::Json(json) = p {
+            let json: serde_json::Value = serde_json::from_str(json.into()).unwrap();
+            let method = &json["method"];
+
+            if method == "requestState" {
+                return ProtocolWrapper::RequestState;
+            } else if method == "state" {
+                return ProtocolWrapper::State(StateData {
+                    state: match json["state"].as_str() {
```
sphinxc0re: Since there is no early return in any of those `match` expressions, there is no need for one. You can just use the methods supplied by `Option<T>` to the code more simple.

net_connection/src/protocol.rs
sphinxc0re: But this code isn't really unreachable, is it?

net_connection/src/protocol_wrapper.rs
sphinxc0re: For example:
```rust
json["state"].as_str().unwrap_or("undefined").to_string()
```

net_connection/src/protocol.rs
lucksus: It reached ff this function is called on a Protocol::NamedBinary for instance, no?

net_connection/src/protocol.rs
neonphog: :+1: done

net_connection/src/protocol.rs
neonphog: :+1: updated to JsonString

net_connection/src/protocol_wrapper.rs
neonphog: :+1: updated to try_from

net_connection/src/protocol_wrapper.rs
neonphog: :+1: updated to try_from

net_connection/src/protocol.rs
neonphog: correct, this is for calling as_* on a type that is not that type

net_connection/src/protocol_wrapper.rs
```diff
@@ -0,0 +1,433 @@
...
+}
+
+/// upgrade a Protocol reference to a ProtocolWrapper instance
+impl<'a> From<&'a Protocol> for ProtocolWrapper {
+    fn from(p: &'a Protocol) -> Self {
+        if let Protocol::Json(json) = p {
+            let json: serde_json::Value = serde_json::from_str(json.into()).unwrap();
+            let method = &json["method"];
+
+            if method == "requestState" {
```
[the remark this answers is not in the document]
neonphog: :+1: updated to use match, but then ended up deleting this code

net_connection/src/protocol_wrapper.rs
neonphog: :+1: updated, but then ended up deleting this code...

neonphog:
@thedavidmeister @sphinxc0re - updated per comments, could you please re-review?

@thedavidmeister - can you take an extra close look at the core_types_derive/src/lib.rs changes for allowing converts from references?

net_connection/Cargo.toml
neonphog: :+1: - awesome, yep, definitely needed : )

net_connection/src/net_connection_thread.rs
```diff
@@ -0,0 +1,183 @@
...
+        let keep_running = Arc::new(AtomicBool::new(true));
+        let keep_running2 = keep_running.clone();
+
+        let (sender, receiver) = mpsc::channel();
+        Ok(NetConnectionThread {
+            keep_running,
+            send_channel: sender,
+            thread: thread::spawn(move || {
+                let mut us = 100_u64;
+                let mut worker = match worker_factory(handler) {
```
sphinxc0re: Better do:

```rust
 let mut worker = worker_factory(handler).unwrap_or_else(|e| panic!("{:?}", e));
```

net_connection/src/net_connection_thread.rs
```diff
@@ -0,0 +1,183 @@
...
+                    match receiver.try_recv() {
+                        Ok(data) => {
+                            did_something = true;
+                            match worker.receive(data) {
+                                Ok(_) => (),
+                                Err(e) => panic!("{:?}", e),
+                            };
+                        }
+                        Err(_) => (),
+                    };
```
sphinxc0re:
```rust
receiver.try_recv().and_then(|data| {
    did_something = true;
    worker.receive(data)
        .unwrap_or_else(|e| panic!("{:?}", e));
}).unwrap_or(());
```

net_connection/src/net_connection_thread.rs
```diff
@@ -0,0 +1,183 @@
...
+
+                    match worker.tick() {
+                        Ok(b) => {
+                            if b {
+                                did_something = true;
+                            }
+                        }
+                        Err(e) => panic!("{:?}", e),
+                    };
```
sphinxc0re:
```rust
worker.tick().and_then(|b| if b { did_something = true; }).unwrap_or_else(|e| panic!("{:?}", e));
```

net_connection/src/net_connection_thread.rs
neonphog: @sphinxc0re - happy to make these changes, just curious about the rational. Is it mainly stylistic? From my perspective, I always feel like direct code is preferable to closures even if the compiler will probably optimize them to the same thing...

net_connection/src/net_connection_thread.rs
neonphog: updated

net_connection/src/net_connection_thread.rs
neonphog: updated

net_connection/src/net_connection_thread.rs
neonphog: updated

net_connection/src/net_connection_thread.rs
sphinxc0re: Because it makes the control flow more clear. If you are using a function that takes a closure, you can't `return` from the outer function and thus it's easier to grasp the control flow directly.

net_connection/src/net_connection_thread.rs
```diff
@@ -0,0 +1,178 @@
...
+        self.send_channel.send(data)?;
+        Ok(())
+    }
+}
+
+impl NetConnectionThread {
+    /// stop (join) the worker thread
+    pub fn stop(self) -> NetResult<()> {
+        self.keep_running.store(false, Ordering::Relaxed);
+        match self.thread.join() {
```
sphinxc0re: You can make this a simple `if` using:

```rust
if self.thread.join().is_err() {
    bail!("NetConnectionThread failed to join on stop() call");
}

Ok(())
```

But that's really personal preference

PR #598, opened 2018-11-07
Container config structs

lucksus:
# Beginnings of Holochain Rust Container
This is the first PR in row and laying ground work for a container api and crate.
Converted the projected config files described [here](https://hackmd.io/OcT2cI1ETfu4DHyvn4QZ5A?both) into Rust structs.

* renamed `core_api` to `container_api` and moved the config structs there
* created `container` crate to be filled with a thin wrapper around `container_api` that creates an executable (added an example config the serves the goal to implement)


README.md
```diff
@@ -26,20 +26,25 @@ This is the home of the Holochain Rust libraries, being rewritten from [Go](http
 This `holochain-rust` repository implements three distinct yet overlapping aspects of the Holochain framework.
 
 1. The core Holochain functionality that executes applications
-2. A wrapper used to instantiate, manage, and run applications
-3. A library and syntax for use in Rust based development of Zomes within applications
+2. A wrapper used to instantiate, manage, and run applications, called *container*
```
zippy: s/application/DNA/

container_api/src/config.rs
```diff
@@ -0,0 +1,223 @@
...
+pub struct DNAConfiguration {
+    id: String,
+    file: String,
+    hash: String,
+}
+
+impl Into<HcResult<Dna>> for DNAConfiguration {
+    fn into(self) -> HcResult<Dna> {
+        let mut f = File::open(self.file)
+            .map_err(|_| HolochainError::IoError(String::from("Could read from file")))?;
```
zippy: Shouldn't that be "could _NOT_ read from file"?

container_api/src/config.rs
```diff
@@ -0,0 +1,223 @@
...
+        let mut contents = String::new();
+        f.read_to_string(&mut contents)
+            .map_err(|_| HolochainError::IoError(String::from("Could read from file")))?;
```
zippy: ditto

README.md
```diff
@@ -26,20 +26,25 @@ This is the home of the Holochain Rust libraries, being rewritten from [Go](http
...
 
 A container is a Holochain utility or service that manages and runs Holochain applications. The first such containers are the GUI driven [holosqape](https://github.com/holochain/holosqape) and the CLI driven [hcshell](https://github.com/holochain/holosqape#hcshell) container for test running.
 
-To implement a Rust based container, one could use the Rust crate exposed in [core_api](./core_api). To implement a container in a C based language, the [core_api_c_binding](./core_api_c_binding) code could be used, such as HoloSqape does.
+In order to develop a Rust based container toolset we added this crate [container_api](container_api).
+There is another crate [container](container) which is a wrapper around [container_api](container_api) that builds to an executable which is intended to become the main, highly configurable and GUI less container implementation that can be run as a background system service.
```
zippy: I thought we talked about not having an actual binary in the `holochain-rust` repo, i.e. that it's just a library, does this change that?

lucksus: Hm, there is no change to core, that stays a library. But I don't think it is a bad idea to have the container crate live in the same repository for now. Since it's a separate crate, we could move it out any time. But it also would make a good example of how to use container_api..

lucksus: Do you want to have it be in a separate repo?

lucksus:
implemented, @zippy

container/example-config/basic.toml
```diff
@@ -0,0 +1,27 @@
...
+name = "Holo Tester"
+key_file = "holo_tester.key"
+
+[[dna]]
+id = "app spec rust"
+file = "app-spec-rust.hcpkg"
+hash = "Qm328wyq38924y"
+
+[[instance]]
+id = "app spec instance"
```
maackle: We had also talked about just using (agent_key, dna_hash) as the unique id for instances. If this separate notion of `id` is not implemented, we could take it out...

container/example-config/basic.toml
```diff
@@ -0,0 +1,27 @@
...
+[instance.storage]
+type = "file"
+path = "app_spec_storage"
+
+
+[[interface]]
+type = "websocket"
+port = 8888
+[[interface.instance]]
+id = "app spec instance"
```
maackle: Why is interface not included with the instance config? Seems like a one-to-one relationship. Then we don't even need to refer to an id.

container_api/src/config.rs
```diff
@@ -0,0 +1,223 @@
...
+    #[serde(rename = "agent")]
+    agents: Option<Vec<AgentConfiguration>>,
+    #[serde(rename = "dna")]
+    dnas: Option<Vec<DNAConfiguration>>,
+    #[serde(rename = "instance")]
+    instances: Option<Vec<InstanceConfiguration>>,
+    #[serde(rename = "interface")]
+    interfaces: Option<Vec<InterfaceConfiguration>>,
+    #[serde(rename = "bridge")]
+    bridges: Option<Vec<Bridge>>,
```
maackle: Might be more idiomatic to just use the plural names in the toml config too, i.e. `[[bridges]]`. The [toml docs](https://github.com/toml-lang/toml#array-of-tables) suggest this.

container/example-config/basic.toml
lucksus: (agent_key, dna_hash) could be used to set the id when creating instance configurations through code. That is actually what I'm intending to do. But I think its cleaner and easier to understand (especially if somebody writing a config manually) if there is an id field in the config file.

container/example-config/basic.toml
lucksus: I don't think it is a one-to-one relationship. I think it will be a rather regular case to have:
* an admin interface that can install/manage apps and also access every app
* an interface without admin rights that access all instances of the local user
* a few special interfaces (for command line scripts for instance) that have access to a distinct subset of instances

container_api/src/config.rs
lucksus: Had it that way first. @sphinxc0re convinced me to do this way to have a more natural language compatible read of config files. I don't have a strong opinion but lean towards singular. Anybody with a strong opinion and good argument for it?

container_api/src/config.rs
```diff
@@ -0,0 +1,223 @@
...
+}
+
+#[derive(Deserialize)]
+pub struct DNAConfiguration {
+    id: String,
+    file: String,
+    hash: String,
+}
+
+impl Into<HcResult<Dna>> for DNAConfiguration {
```
maackle: Feels a bit heavy to include file IO in an `into()` conversion but I'm not versed enough to know how people do it.

lucksus: Inherited that impl from @AshantiMutinta.
I do think this is adequate for an Into impl. What do you say, Ashanti?

On a related note, this could be a TryInto, no?

container/example-config/basic.toml
maackle: We had also talked about just using (agent_key, dna_hash) as the unique id for instances. If this separate notion of id is not implemented, we could take it out...

container/example-config/basic.toml
maackle: Why is interface not included with the instance config? Seems like a one-to-one relationship. Then we don't even need to refer to an id.

container_api/src/config.rs
maackle: Might be more idiomatic to just use the plural names in the toml config too, e.g. `[[bridges]]`. The [toml docs](https://github.com/toml-lang/toml#array-of-tables) suggest this.

maackle:
(Those "deleted comments" were from my first pass through when I simply made comments, then I deleted them to include them in a proper review. Strange that they say they were from lucksus though...)

lucksus:
@maackle, I had answered your comments already.

lucksus:
Ok, let me try again ;D

container/example-config/basic.toml
lucksus: I intend to fill the id this way when the config gets created from code (`install_app()`), but I like having an explicit id in the config file especially if a human writes it manually.

container/example-config/basic.toml
lucksus: It's not one-to-one, we can have several different interfaces with access to different subsets of instances.

container_api/src/config.rs
lucksus: Had it that way before, @sphinxc0re convinced me to use singular in the files, which feels like more natural read of the config. Don't have a strong opinion but like singular more since we will have several blocks of these and each block is just one instance..

lucksus: I'd prefer to keep it. Also, I haven't used toml much. Does it feel strange to have singular array instances? To me it feels strange to have them be plural.
Will change if you feel strongly it should be plural.

maackle: I agree that it *looks* cleaner singular, but note that this equivalent TOML is also valid (an array with inline tables):

```
agent = [ 
    {id = "test agent 1", name = "Holo Tester 1", key_file = "holo_tester1.key"},
    {id = "test agent 2", name = "Holo Tester 2", key_file = "holo_tester2.key"}
]
```

in which case it clearly should be plural. I think it's more that TOML syntax is awkward for arrays. So I do feel it should be plural, but I don't feel strongly enough about it to block this PR.

maackle:
Ack, sorry about that @lucksus, I should have refreshed the page. Anyway, I made one more case for pluralizing the config keys, but regardless of how that turns out, I still approve.

container_api/src/config.rs
lucksus: Ok, I've switched to plurals.

container_api/src/config.rs
```diff
@@ -159,31 +153,31 @@ hash = "Qm328wyq38924y"
...
+[instances.storage]
 type = "file"
 path = "app_spec_storage"
 
-[[interface]]
+[[interfaces]]
 type = "websocket"
 port = 8888
-[[interface.instance]]
+[[interfaces.instances]]
```
maackle: This is one case where plurals makes it clearer -- originally I thought it was one-to-one because it was `interface.instance`

lucksus: gotcha 👍

README.md
zippy: that's fine.  Just a question about your thinking.

container_api/src/config.rs
AshantiMutinta: Definitely could be a tryinto, think it is okay to have the IO in there as the struct is a configuration struct but open to other ideas definitely @lucksus @maackle

container_api/src/config.rs
```diff
@@ -0,0 +1,217 @@
...
+    bridges: Option<Vec<Bridge>>,
+}
+
+#[derive(Deserialize, Clone)]
+pub struct AgentConfiguration {
+    id: String,
+    key_file: Option<String>,
+}
+
+impl Into<Agent> for AgentConfiguration {
```
sphinxc0re: `Into<T>` is implicitly generated when you implement `From<T>` for a type. It shouldn't be necessary to implement `Into<T>`

lucksus: Done

PR #601, opened 2018-11-09
Added start to query function

sphinxc0re:


hdk-rust/src/api.rs
```diff
@@ -53,37 +50,6 @@ lazy_static! {
...
-    fn from(agent_initial_hash: AGENT_INITIAL_HASH) -> JsonString {
-        JsonString::from(HashString::from(agent_initial_hash.to_string()))
-    }
-}
-impl From<AGENT_LATEST_HASH> for JsonString {
-    fn from(agent_latest_hash: AGENT_LATEST_HASH) -> JsonString {
-        JsonString::from(HashString::from(agent_latest_hash.to_string()))
-    }
-}
-
```
lucksus: Why did you remove these From implementations? This makes possible to write

```rust
hdk::link_entries(AGENT_ADDRESS, "post", post)
```

instead of

```rust
hdk::link_entries(JsonString::from(HashString::from(AGENT_ADDRESS.to_string())), "post", post)
```

sphinxc0re: They looked pretty useless to me. Will update

wasm_utils/src/macros.rs
```diff
@@ -26,7 +26,7 @@ macro_rules! ribosome_success {
 macro_rules! ribosome_error_code {
     ($s:ident) => {
         Ok(Some(RuntimeValue::I32(
-            ::holochain_wasm_utils::holochain_core_types::error::RibosomeErrorCode::$s as i32,
+            $crate::holochain_core_types::error::RibosomeErrorCode::$s as i32,
```
lucksus: Also, are you sure this works in all contexts?

sphinxc0re: sure, why wouldn't it?

thedavidmeister:
FYI for contextual info on why this is good https://developer.twitter.com/en/docs/tweets/timelines/guides/working-with-timelines.html

PR #602, opened 2018-11-09
Hash -> Address; hash_entry -> entry_address

sphinxc0re:


zippy:
@Connoropolous I could only find one reference in the mdbook, and that was to make_hash, and even older name (which I updated).  Is there some place missing?

thedavidmeister:
imma clean this up

PR #609, opened 2018-11-10
587 cas try from 2

thedavidmeister:
### try_from_content

`AddressableContent` now implements `try_from_content()` instead of `from_content()`

I wanted to use native `TryFrom<Content>` and `Into<Content>` but this causes issues elsewhere in the codebase. These traits use `Self` in the return type making it impossible to use `AddressableContent` as a trait object. We need `AddressableContent` as a trait object to support configurable `ContentAddressableStorage`.

### codecov

- deleted unused impl for walkdir error (the module isn't even included by lib.rs)
- added test for `invoke_hash_entry`

## followup

i accidentally pushed some compiled wasm that somehow was left over from `core_api` being removed and also dropped from `.gitignore` 😳 

might want to run BFG over the repo at some point to clean up the history

- run BFG over the repo #610


core_types/src/cas/content.rs
```diff
@@ -33,7 +34,7 @@ pub trait AddressableContent {
 
     /// restore/deserialize the original struct/type from serialized Content
     /// the default implementation covers anything that implements From<JsonString> for Foo
-    fn from_content(content: &Content) -> Self
+    fn try_from_content(content: &Content) -> Result<Self, HolochainError>
```
lucksus: You say that we can't use `TryFrom<T>` because `try_from()` returns `Self` and thus would prevent AddressableContent from being made into a trait object. This function has `Self` wrapped in a Result. Doesn't that prevent trait objects as well? Or does `Result` store its content on the heap? @thedavidmeister

thedavidmeister: @lucksus hmmmm 🤔 i think i need @AshantiMutinta to weigh in on this one!

this definitely works now while `TryFrom` definitely broke when i tried it...

AshantiMutinta: The self in the result definitely stops it from being a trait object sadly. I think it is fine because there is no AddressableContent trait object being used for now and if we ever need one we could refactor it in.(The issue would be if we would want that AddressableContent in a trait that needs AC to be a trait object). This seems fine to me!

lucksus: So what made the TryFrom solution break? If it was not this, maybe we could use TryFrom? But I don't want to halt merging this. Feel free to either investigate what's going on here, or merge, @thedavidmeister

PR #611, opened 2018-11-10
2018 11 10 fix travis cache

thedavidmeister:
adds items to the travis cache blacklist


core/src/nucleus/ribosome/api/hash_entry.rs
```diff
@@ -15,6 +15,7 @@ pub fn get_entry_type(dna: &Dna, entry_type_name: &str) -> Result<EntryType, Opt
             holochain_core_types::error::RibosomeErrorCode::UnknownEntryType as i32,
         ))
     })?;
+
```
zippy: Umm this stuff seems not relevant to this PR, but that's ok I guess...  :-P

thedavidmeister: @zippy haha this is me triggering travis builds so i can inspect whether the cache changes worked... because you cannot tell until the next run 😅

zippy: ah grr, the pain of travis...

PR #613, opened 2018-11-10
move dna into core types entry

thedavidmeister:
## changes

`Dna` implements `ToEntry` so it is moving under `core_types::entry`

## followup

- Dna using serde_json::Value where it could possibly use JsonString #612


zippy:
worried about approving and this before fixing CI on app-spec-rust and holochain-cmd because I think this will break both those repos.

thedavidmeister:
also as @lucksus and i discussed on the call, Dna already has a dep on core_types, so actually it needs to be in there if the enum references it in the future or we will have circular deps

so the outcome, as he said, is making this a sibling rather than a child of entry

thedavidmeister:
@lucksus i'm force merging this to get things moving, it incorporates all your feedback from our meeting

PR #614, opened 2018-11-10
move entry types under entry

thedavidmeister:
## changes

entry type definitions sit under/alongside entry data definitions


thedavidmeister:
should be fixed now, was an issue in the wasm build not main test suite

PR #617, opened 2018-11-11
move chain header under entry

thedavidmeister:
## changes

moves chain header alongside/under entry


thedavidmeister:
actually this should NOT happen, as per discussion at #3565

PR #618, opened 2018-11-11
move links under entry

thedavidmeister:
## changes

moves links alongside entry


thedavidmeister:
@lucksus @zippy as per #3565 this moves links _next to_ entry in core types

PR #619, opened 2018-11-12
spike: using cargo make

thedavidmeister:


thedavidmeister:
@lucksus on the whole i would say this spike was pretty unsatisfying

- not clear how to chain conditions, so i can do `if` but not `if, else if, else if` or `switch`
- not possible to set toolchain version
- very quickly devolves into the same kind of low level scripting required in standard `Makefile` to ensure deps and cross platform compatibility

there _are_ some aesthetic wins in using cargo make vs. Makefile, but there are some functional losses too, and the big issues we are facing like "works/doesn't work on my computer" aren't resolved

PR #620, opened 2018-11-12
588 define zome jsonstring

thedavidmeister:
## changes

- allows anything that implements `Into<JsonString>` to be returned by zome api functions using rust hdk
- sweeps rust hdk tests/examples to use real types and not just JsonString everywhere
- returns `ZomeApiResult<T>` for a lot of the existing zomes as this matches what the hdk returns
- uses `?` and directly return hdk fns as much as possible in zomes to be neat
- updates scenario tests
- extends `JsonString` a bit, e.g. `u32` and `empty_object`
- new trait `JsonError` for any type that can be the `Err` value in a `Result` that should be `JsonString`able
  - currently available for `ZomeApiError` and `HolochainError`
- implements `String` as an `Err` value option for `Result` that can be `JsonString`able

## breaking changes

- zome functions should no longer return `JsonString`
  - this probably doesn't work at all now
  - if you feel the need to return a `JsonString` (e.g. the result of `hdk::call`) then consider whether you really want to return `ZomeApiResult<JsonString>`, `RawString` or `String`
  - try to match the return signatures of relevant `hdk` calls as much as possible, use `?` as much as possible
    - e.g. `hdk::debug("foo")?; // -> ZomeApiResult<T>`
 - `ZomeApiError` no longer serializes as `{"error": ...}` and should be used with `Result` or `ZomeApiResult` which will serialize as `{"Err": ...}`

## followups

- there is still a bit of awkward serialization juggling in scenario tests for things that work with `call` inside the zome function
  - this is because we have a nested call that serializes the inner zome result
  - there is a followup in the SOA tree to make this nicer to work with
  - for now we get things like `JSON.parse(result.Err.Internal)` in tape
- entries are a bit awkward to work with in zomes
  - in general `Entry::App()` is just one example of an app type, but from the perspective of zomes this is the main/only type
  - want some sugar
  - leads to things like `JSON.parse(result.Ok.App[1])` in tape
- Simplify zome macro: inputs and outputs #725


thedavidmeister:
@Connoropolous this question is exactly what i was going back and forward on haha

app_spec/zomes/blog/code/src/blog.rs
```diff
@@ -55,13 +40,10 @@ pub fn handle_hash_post(content: String) -> JsonString {
...
 
-    match hdk::entry_address(&post_entry) {
-        Ok(address) => AddressResponse{address}.into(),
-        Err(hdk_error) => hdk_error.into(),
-    }
+    hdk::entry_address(&post_entry)
 }
 
-pub fn handle_create_post(content: String, in_reply_to: HashString) -> JsonString {
+pub fn handle_create_post(content: String, in_reply_to: Option<Address>) -> ZomeApiResult<Address> {
```
maackle: Good, I was hoping someone would wrap this in `Option`

app_spec/zomes/blog/code/src/lib.rs
```diff
@@ -28,43 +32,43 @@ define_zome! {
...
+                outputs: |post: ZomeApiResult<JsonString>|,
                 handler: blog::handle_check_sum
             }
 
-            hash_post: {
+            post_address: {
                 inputs: |content: String|,
-                outputs: |result: JsonString|,
-                handler: blog::handle_hash_post
+                outputs: |result: ZomeApiResult<Address>|,
```
maackle: While we're changing the macro behavior, can we consider removing the requirement of naming inputs and outputs? Maybe as a separate PR? The names shouldn't matter, especially for the outputs, it's just noise. (right?)

core_types/src/json.rs
```diff
@@ -81,9 +98,44 @@ impl<T: Serialize> From<Vec<T>> for JsonString {
...
-        JsonString::from(serde_json::to_string(&result).expect("could not Jsonify result"))
+        let is_ok = result.is_ok();
+        let inner_json: JsonString = match result {
+            Ok(inner) => inner.into(),
+            Err(inner) => inner.into(),
+        };
+        let inner_string = String::from(inner_json);
+        format!(
+            "{{\"{}\":{}}}",
+            if is_ok { "Ok" } else { "Err" },
```
maackle: This seems like the right approach, always putting successes and errors behind different keys, but I want to bikeshed a little and propose that the names change, or at least get lowercased. Having a title-cased member name feels awkward to me in any language except Visual Basic, and doesn't even feel Rusty. Maybe the error one could be simply named `"error"`. I'd lean towards `"value"` for the success case.

```javascript
t.ok(result.value)  // assert not an error
t.ok(result.error)  // assert there was an error
```

thedavidmeister: @maackle this is exactly what serde would have done :man_shrugging: 

_every_ enum that we serialize is title case and `Result` is just an enum under the hood, so we get `Ok` and `Err`

while i agree in principle that this gives a somewhat awkward interface, i don't think that the solution is to inconsistently override casing on an ad-hoc basis

can we make this a followup because implementing it properly would need some thought

app_spec/zomes/blog/code/src/lib.rs
thedavidmeister: @maackle yes this bugs me too

inputs might be tricky because the named parameters map to a struct inside the macro, and it is pretty standard rust fare to have `name: Type`

outputs is always a single value so the name seems unneccessary

however, i think this is out of scope for this branch, we should put it somewhere in the SOA tree

app_spec/zomes/blog/code/src/blog.rs
thedavidmeister: reminds me that this is a breaking change, `""` will error now. it must be `null` or some value.

thedavidmeister: oh, except that it isn't really breaking because it's just an example haha

core_types/src/json.rs
thedavidmeister: @maackle thinking about this a little more, maybe this logic should go in the node container?

that way the container can decide what is idiomatic for node and lowercase everything across the board. other containers could camel case, kebab case, etc.

for an example of why this goes way beyond `Ok` and `Err` in `Result`, we also have things like this (which i think is much worse) in app spec:

```js
const error = JSON.parse(result.Err.Internal)
```

note `Internal` is also title case and represents a `ZomeApiError::Internal` enum. Rust only serializes the variant name because the compiler always forces you to BYO type, but we don't have that safety/context in JS. If we had another type `FooError::Internal` with the same variant name but different enum, there would be no way of knowing what we meant just from looking at the string.

also note the need to `JSON.parse` that error because it is serialized into a string value due to the way that `hdk::call` works. nested zome calls serialize their internal results each step of the way. In this way `call` (and i assume bridge?) behaves differently to other zome fns that deserialize the results for you on the way out.

and we also have:

```js
const entry_value = JSON.parse(result.Ok.App[1])
```

which (from a JS perspective) is a gross way to work with entries for many reasons.

from the Rust perspective it is great though, it preserves our type safety all the way through JSON, which is the only reason we can get core types at all in zomes! (without heroic effort)

the convo we had the other day about wrapping _everything_ in a giant enum would only make this much worse...

unfortunately all of this seems to just come with the territory of using `serde`. If we want compiler-driven serialization logic then our serialization naturally ends up looking a lot like our compiler types :/

yes we can implement custom serialization logic and serde attributes, but that also seems like a lot of boilerplate/overhead too :(

i already added some items to the SOA around cleaning up the awkwardness in general. we could be putting sugar in the JS side of things that uses contextual knowledge (e.g. that rust/hc is not dynamically typed so returns predictable structures) to do things like `parse_call` and `parse_entry`.

the reason i believe all of this is out of scope for this branch (aside from general PR sizing concerns) is that nothing ^^ actually changed in this branch, all i've done is exposed what was already there :sweat_smile:

app_spec/zomes/blog/code/src/lib.rs
maackle: Started a discussion about it here: https://github.com/holochain/holochain-rust/issues/725

core_types/src/json.rs
maackle: @thedavidmeister I see, yeah I wasn't thinking about all the possible enums that can serialized and sent as return values. If you think about the entire structure of a zome function return value though, I think it is worth distinguishing the serialization of the root-level `Result` that everything is wrapped in vs. the rest of the enums and structs inside that. Because, if every zome function returns a `Result`, the user has no control over its serialization (unless we do provide config for this on the container side, which does sound like a nice idea), but if they want they can map the other values into more serialize-friendly types before jamming them into the Result on the way out (by using serde `rename` and `tag` attributes on user-defined types).

So I'm not proposing making *all* serialization more idiomatic for other languages, but actually wondering if it makes sense to give special treatment to this top-level Result (which I think it does), since it is indeed being treated specially here even though it's doing what serde would have done.

If others think this should wait for another PR I will shut up about it 😛

thedavidmeister: @maackle it's not true that every zome function returns a `Result` - it is _hdk_ functions that return a `Result`.

zome functions that internally call `hdk::call` actually have serialized internal results - could be an arbitrary level of nesting for this

zome functions that don't use the hdk won't necessarily have a `Result` at all - check out `handle_sum` for example

so the `Result` is not "root level that everything is wrapped in at all" :thinking:

thedavidmeister: another example of when this might be relevant...

we've been getting feedback from the community that they want to do their own serialization that is not JSON at all... so that would look like returning `RawString` from a zome function not a `Result`

PR #623, opened 2018-11-14
AgentId core_type

neonphog:
Provides a type for agent identifiers per https://docs.google.com/document/d/1qoAk2NpML_WTZtNE44ZZ33WNIuHnrrPd-ojj93SZMck/edit#heading=h.3a4cbeqn4a2 including the reed-solomon parity bytes in the base64url string representation.

Note that this doesn't provide any utilities for generating the IDs, that'll be part of the dpki effort... we can just fill them with random data for testing until then.


core_types/src/agent_id.rs
```diff
@@ -0,0 +1,156 @@
+//! Agent Ids are 64 bytes long.
+//! - The first 32 bytes 0-32 are the public side of a libsodium signing keypair
```
sphinxc0re: Shouldn't it be `0-31`?

core_types/src/agent_id.rs
```diff
@@ -0,0 +1,156 @@
...
+//! - The next 32 bytes 32-64 are the public side of a libsodium kx keypair
```
sphinxc0re: Shouldn't it be `32-63`?

core_types/src/agent_id.rs
```diff
@@ -0,0 +1,156 @@
...
+        array_ref![self.0, 0, 32]
+    }
+
+    /// get the encryption (kx) public key portion of this identity
+    pub fn get_enc(&self) -> &[u8; 32] {
+        array_ref![self.0, 32, 32]
+    }
+}
+
+impl std::fmt::Display for AgentIdPub {
```
sphinxc0re: Why are you implementing this by hand. You could just `#[derive(Debug)]` on the struct

core_types/src/agent_id.rs
```diff
@@ -0,0 +1,156 @@
...
+        )
+    }
+}
+
+impl<'a> From<&'a AgentIdPub> for String {
+    /// stringify this agent public identity
+    fn from(id: &AgentIdPub) -> Self {
+        let enc = Encoder::new(PARITY_LEN);
+        let enc = *enc.encode(&id.0[0..64]);
+        base64::encode(&enc[..]).replace("+", "-").replace("/", "_")
```
sphinxc0re: What is this replacement for?

core_types/src/agent_id.rs
neonphog: You'd think so, right? Apparently rust can only derive Debug for u8 arrays up to 32 bytes... If you know a way of doing this, I'm all ears.

core_types/src/agent_id.rs
neonphog: The [base64url variant](https://tools.ietf.org/html/rfc4648#section-5) makes the ids easier to use in filenames and urls, (no forward slash).

core_types/src/agent_id.rs
neonphog: Sorry, I tend to assume the top-end is exclusive, but I can never remember if it's a paren or bracket that denotes that. I can update to `[0-32)`.

core_types/src/agent_id.rs
neonphog: Can update to `[32-64)`

core_types/src/agent_id.rs
sphinxc0re: 🙈 Oh boy... this again. Well... I guess we can't do any better than this

neonphog:
@zippy @lucksus @sphinxc0re @thedavidmeister

After looking at `entry/agent.rs` I decided it would be more efficient to store the user string in memory... the expensive part is generating the reed-solomon parity bits. we can always just base64 decode the string ignoring the parity if we need the keys from the identity.

I made the alterations, but it turns out usage of the `Agent` entry is quite far and wide. Will need a re-review from you guys, so I dismissed your approvals.

I have pretty much zero knowledge of the HDK, so I'm sure I broke something in there. Especially: https://github.com/holochain/holochain-rust/pull/623/files#diff-705bf18c1b860bbe8c60fc825cfcc880R82

There seems to be some hard-coded assumptions of sha256 hashes that will need to be updated to support the raw agent id addresses.

core_types/src/entry/agent.rs
```diff
@@ -1,43 +1,104 @@
...
+    pub fn generate_fake(s: &str) -> Self {
+        let mut s = s.to_string();
+        while s.len() < 85 {
+            s.push_str("+");
+        }
+        s.push_str("A==");
+        let s = base64::decode(&s)
+            .expect("could not decode the generated fake base64 string - use the base64 alphabet");
+        let s = IdentityBuffer(array_ref![s, 0, 64].clone());
+        Agent::render(&s)
```
lucksus: What exactly is happening here? Seems like you are creating something that is a syntactilly correct (public?) key from a string? I understood those names in the tests ("Terence", "Florence") to be the nick associated with a key.

core_types/src/entry/agent.rs
```diff
@@ -1,43 +1,104 @@
...
+    }
+
+    pub fn render(b: &IdentityBuffer) -> Self {
+        let enc = Encoder::new(PARITY_LEN);
+        let enc = *enc.encode(&b.0);
+        Agent(base64::encode(&enc[..]).replace("+", "-").replace("/", "_"))
```
lucksus: Why do we need this? This is the reciprocal functions, turning a byte vector into a string representation? (Comments for context would be good here I think)

neonphog:
@lucksus 
> I'm a bit confused by that implementation. Seems like you are using the Identity string to store the key in there? With merging I meant having both in the end:
> 
> - an Identity string that is a (user input) name, like "neonphog"
> - a byte vector that holds the key

So, I determined that it would be more cpu efficient to store the user copy-able base64url string representation of the key in memory rather than the key itself, so we didn't have to recalculate the parity bits every time it is string-ified. The additional important bit from my perspective is that this string is also the `Address` (i.e. NOT the sha265 hash of it, but the actual string itself).

If we want to store additional metadata along with this, I can add that back in. I'm not sure I exactly understand the utility of the nicknames though... they will not be namespaced / unique... This would make the entry content look something like:

Address: `MTIzNDU2Nzg5MDEyMzQ1Njc4OTAxMjM0NTY3ODkwMTIzNDU2Nzg5MDEyMzQ1Njc4OTAxMjM0NTY3ODkwMTIzNBkd`
Content:
```json
{
  "value": "{\"nick\":\"bob\",\"key\":\"MTIzNDU2Nzg5MDEyMzQ1Njc4OTAxMjM0NTY3ODkwMTIzNDU2Nzg5MDEyMzQ1Njc4OTAxMjM0NTY3ODkwMTIzNBkd\"}",
  "entry_type": "%agent_id"
}
```

core_types/src/entry/agent.rs
neonphog: Yep, we need users to be able to copy / paste these ids around, and we add some parity bits that can act as a checksum as well as doing up to one byte of reed-solomon error correction. Added some comments.

neonphog:
> I think I would prefer "identifier" over "nick" as the name for this attribute

I don't want to call it anything that can be shortened to `ID`... i think of the address as the agent id. If someone comes up with something that doesn't conflate with `id` or `username`, i'm happy to change it. Maybe `hint`?

lucksus:
"identifier" is what we had in proto, I believe. No strong opinion. Also fine with nick or nickname.

PR #639, opened 2018-11-17
spike: lets see what happens when travis runs nix

thedavidmeister:


thedavidmeister:
this is old nix and stale results, ther'es another newer branch up #713

PR #640, opened 2018-11-18
move all entry content to siblings of entry

thedavidmeister:
easy approve, moves all the entry types to siblings of entry


zippy:
CI failing because we dropped to 85% exactly.  I added a couple skips for code that's apparently being tested elsewhere, I hope that was a reasonable move.

PR #642, opened 2018-11-18
Link validation

lucksus:
done:
* added links to entry! macro 
* completed `link_entries()` flow with a LinkAddEntry that gets validated and committed before setting the DHT's EAV
* getting the validation package definition for link entries
* building the validation package for links from that definition
* actually calling the validation callbacks for link entry validation
* added new callbacks `__hdk_validate_entry` and `__hdk_get_validation_package_for_link`
* more tests for `link_entries`
* Rustdoc for `link!`, `to!` and `from!`

# Example
`entry!` now has a links section that takes an *array* of links that can be defined with the macros `from!` and `to!`. Every link definition has a `validation_package` and `validation` closure just like an entry:
```rust
            links: [
                to!(
                    "testEntryType",
                    tag: "test-tag",
                    validation_package: || {
                        hdk::ValidationPackageDefinition::ChainFull
                    },
                    validation: |source: Address, target: Address, ctx: hdk::ValidationData | {
                        Ok(())
                    }
                )
            ]
```

# Impl
Works very similar to entry validation but a bit more complex. To find out which validation closure (and package) needs to be called for a given link, we need to first get the base and target of the link and search in the DNA's entry types for according `to` or `from` link definitions that would match the given link. This happens in https://github.com/holochain/holochain-rust/blob/link-validation/core/src/nucleus/ribosome/callback/links_utils.rs.


core/src/context.rs
```diff
@@ -97,6 +104,43 @@ impl Context {
             Some(ref s) => Some(s.read().unwrap()),
         }
     }
+
```
lucksus: These functions are just moved for easy reuse.

core/src/nucleus/ribosome/api/link_entries.rs
```diff
@@ -22,8 +33,44 @@ pub fn invoke_link_entries(runtime: &mut Runtime, args: &RuntimeArgs) -> ZomeApi
...
+    let entry = link_add_entry.to_entry();
+
+    // Wait for future to be resolved
+    let result: Result<(), HolochainError> = block_on(
+        // 1. Build the context needed for validation of the entry
+        build_validation_package(&link_add_entry.to_entry(), &runtime.context)
+            .and_then(|validation_package| {
+                Ok(ValidationData {
+                    package: validation_package,
+                    sources: vec![HashString::from("<insert your agent key here>")],
```
zippy: Should this be a HashString, or an Address?

core/src/nucleus/ribosome/callback/links_utils.rs
```diff
@@ -0,0 +1,82 @@
...
+    base_type: &EntryType,
+    tag: &String,
+    target_type: &EntryType,
+    context: &Arc<Context>,
+) -> Option<LinkDefinitionPath> {
+    let dna = context.get_dna().expect("No DNA found?!");
+    match base_type {
+        EntryType::App(app_entry_type) => dna
+            .get_entry_type_def(&app_entry_type)
+            .expect("Found entry type that is not defined in DNA?!")
```
zippy: I don't think these can be expect statements, because doesn' that mean that someone could craft a validation request that would cause your machine to panic?  I think this function has to be able to return a HolochainError that can be passed back to it's caller (`validate_link_entry()`) no?

core/src/nucleus/ribosome/callback/validate_entry.rs
```diff
@@ -1,53 +1,97 @@
...
+) -> Result<CallbackResult, HolochainError> {
+    let link_add_entry = LinkAddEntry::from_entry(&entry);
+    let link = link_add_entry.link().clone();
+    let (base, target) = links_utils::get_link_entries(&link, &context)?;
+    let link_definition_path = links_utils::find_link_definition_in_dna(
+        &base.entry_type(),
+        link.tag(),
+        &target.entry_type(),
+        &context,
+    ).ok_or(HolochainError::NotImplemented)?;
```
zippy: What I see here implied is that this becomes part of the "system level" validation of links which makes it impossible for an app developer to create ad-hoc links manually (as we have done in holochain-proto) that aren't defined `entry!`  

I'm actually fine with this as MVP and getting off the ground because its conceptually simple, but I want to note that it decreases the expressivity from the initial design.  I think we might want have to do something that allows a "manual/custom" override for defining a link type with validation that's not inferred from the `entry!` definition.  But we can add that later.

hdk-rust/tests/integration_test.rs
```diff
@@ -57,10 +57,25 @@ fn start_holochain_instance() -> (Holochain, Arc<Mutex<TestLogger>>) {
     ]);
     let mut dna = create_test_dna_with_cap("test_zome", "test_cap", &capabability, &wasm);
 
-    dna.zomes.get_mut("test_zome").unwrap().entry_types.insert(
-        String::from("validation_package_tester"),
-        EntryTypeDef::new(),
-    );
+    // TODO: construct test DNA using the auto-generated JSON feature
```
zippy: Can you please make sure this gets documented in a ticket or a leaf on the tree?

core/src/nucleus/ribosome/callback/links_utils.rs
lucksus: I started writing why I thought this is not the case but I now think you are very right.
Will change :)

hdk-rust/tests/integration_test.rs
lucksus: https://realtimeboard.com/app/board/o9J_kyiXmFs=/?moveToWidget=3074457346314650824

lucksus:
@Connoropolous, I did add rustdoc comments! Was about to ping you for approving that I did well.. ;) What did I miss?

lucksus:
Are you talking about code outside the HDK?

lucksus:
Added comments for link_utils.rs. Need more, @Connoropolous?

core/src/nucleus/ribosome/callback/validate_entry.rs
```diff
@@ -1,53 +1,97 @@
...
         )?),
         EntryType::Dna => Ok(CallbackResult::Pass),
+        EntryType::LinkAdd => Ok(validate_link_entry(
+            entry.clone(),
+            validation_data,
+            context,
+        )?),
         _ => Ok(CallbackResult::NotImplemented),
     }
 }
```
Connoropolous: Could you add one here?

core/src/nucleus/ribosome/callback/validate_entry.rs
```diff
@@ -1,53 +1,97 @@
...
+};
 use nucleus::{
-    ribosome::{
-        self,
-        callback::{get_dna, get_wasm, CallbackResult},
-    },
+    ribosome::{self, callback::CallbackResult},
     ZomeFnCall,
 };
 use std::sync::Arc;
```
Connoropolous: I know you didn't add this function just now but here?

hdk-rust/src/entry_definition.rs
```diff
@@ -1,26 +1,34 @@
...
 
 pub struct ValidatingEntryType {
     pub name: String,
     pub entry_type_definition: EntryTypeDef,
     pub package_creator: PackageCreator,
     pub validator: Validator,
-    pub link_validators: HashMap<String, LinkValidator>,
+    pub links: Vec<ValidatingLinkDefinition>,
+}
+
```
Connoropolous: here

zippy:
@lucksus one thought about the naming of `from!` and `to!`  how about `as_base!` and `as_target!`  very clear and explicit?

lucksus:
I wouldn't know which term you would want to use for which direction. Is the surrounding entry type the base when you say `as_base!` or the one that follows in the macro?

PR #643, opened 2018-11-19
WIP: Adds two config getters

maackle:
I suspect Holo is going to need to look up Instances based on Agent and DNA hashes rather than by ID. This adds some additional getters to find Instances by these parameters. If we find it's not necessary we can close this.


maackle:
No longer needed, got clarity after discussion with Nico.

PR #648, opened 2018-11-19
Proposal to use an enum to handle Interface types

maackle:
* Removes the need for stringly-typed `InterfaceConfiguration` and `StorageConfiguration`, using enums instead
* Uses the `Default` impl for some struct fields instead of using `Option` to specify optional TOML fields


container_api/src/config.rs
```diff
@@ -30,26 +30,31 @@ use toml;
...
+    #[serde(default)]
+    pub instances: Vec<InstanceConfiguration>,
     /// List of interfaces any UI can use to access zome functions
-    pub interfaces: Option<Vec<InterfaceConfiguration>>,
+    #[serde(default)]
+    pub interfaces: Vec<InterfaceConfiguration>,
     /// List of bridges between instances
-    pub bridges: Option<Vec<Bridge>>,
+    #[serde(default)]
+    pub bridges: Vec<Bridge>,
```
sphinxc0re: All these defaults can't be right. Is an empty configuration a valid one?

container_api/src/config.rs
```diff
@@ -177,13 +177,14 @@ pub struct LoggerConfiguration {
...
-    pub password: Option<String>,
-    pub url: Option<String>,
-    pub path: Option<String>,
+#[serde(tag="type")]
+pub enum StorageConfiguration {
+    #[serde(rename = "memory")]
+    Memory,
+    #[serde(rename = "file")]
+    File {
+        path: String,
```
sphinxc0re: Please make this a `PathBuf`

container_api/src/config.rs
maackle: @sphinxc0re sure, it's OK to just not specify bridges in the config if you don't need them right?

It makes it so absence of a list corresponds to the empty list

maackle: @sphinxc0re which would you recommend should be required? `agents` and `dna`?

sphinxc0re: I'm not sure, that's why I'm asking. What exactly does the agent need to get running?

container_api/src/config.rs
maackle: @sphinxc0re I started doing this but if you follow the trail, this gets fed into several other functions including `create_context` which also take a `String` and should probably take a `Path`. I'd like to defer it to a separate small PR.

container_api/src/config.rs
maackle: @sphinxc0re you're right, according to `check_consistency`, agents, dnas, and instances should be required, so I changed it.

lucksus: 👍

zippy:
I'd like to be able to merge this but it looks like some tests are failing...

container_api/src/config.rs
sphinxc0re: 😱 When a path is needed `PathBuf` should be used. This sounds like a followup.

container_api/src/config.rs
```diff
@@ -307,46 +312,63 @@ pub mod tests {
...
+            panic!("Expected `file` type StorageConfiguration!")
+        }
+
+        let interfaces = config.interfaces;
+        let interface_config_0 = interfaces.get(0).unwrap();
+        let interface_config_1 = interfaces.get(1).unwrap();
+        if let InterfaceProtocol::Websocket { port } = interface_config_0.protocol {
+            assert_eq!(port, 8888);
+        } else {
+            panic!();
```
sphinxc0re: Please specify a proper error message here.

container_api/src/config.rs
```diff
@@ -307,46 +312,63 @@ pub mod tests {
...
+        }
+        if let InterfaceProtocol::DomainSocket { ref file } = interface_config_1.protocol {
+            assert_eq!(file, "/tmp/holochain.sock");
+        } else {
+            panic!();
```
sphinxc0re: See above

PR #649, opened 2018-11-20
2018 11 20 entry enum

thedavidmeister:
# blockers

https://github.com/holochain/holochain-rust/pull/668

# changes

- renamed `agent::Agent` to `agent::AgentId`
- merged `Entry` and `SerializedEntry`
- `Entry` is an enum
- application vs. system entries now differentiated by `Entry::App(AppEntryType, AppEntryValue)` and everything else e.g. `Entry::Dna(Dna)`
- `AppEntryType` exists as a new type wrapper around string names e.g. `AppEntryType::from("myType")`
- `AppEntryValue` is an alias to `JsonString`
- custom serialization/deserialization logic for the `Entry::App(..)` variant to handle `JsonString` and treat as a tuple
- serialized entry data was: `{\"value\":\"\\\"test entry value\\\"\",\"entry_type\":\"testEntryType\"}` is now `{\"App\":[\"testEntryType\",\"\\\"test entry value\\\"\"]}`
- entries can now be serialized/deserialized in a single step
- `LinkListEntry` renamed to `LinkList` and used as `Entry::LinkList(LinkList)`
- `ToEntry` trait removed in favour of `Entry::Foo(Foo)` pattern, where type is implicit as the variant or explicit using `Entry::Foo(Foo).entry_type()` method
- Validation logic takes `Entry` instead of `String`
- renamed `get_zome_name_for_entry_type` to `get_zome_name_for_app_entry_type`
- removed `Entry::new(foo)` pattern in favour of `Entry::Foo(foo)`
- removed `Deref` for `Entry` as it is not used AFAICS
- renamed `LinkAddEntry` to `LinkAdd`, use as `Entry::LinkAdd(LinkAdd)`
- renamed `LinkListEntry` to `LinkList`, use as `Entry::LinkList(LinkList)`
- cleaned out some cruft unused code
- `handle_commit_entry` in HDK WASM tests now takes an `Entry` argument

# blockers

merge https://github.com/holochain/holochain-rust/pull/642 first if it is ready as it will have merge conflicts with this - touches the same validation logic

# followup

- merge EntryType and Entry in validation #662


thedavidmeister:
this has reverted to WIP status due to bugs in app_spec

thedavidmeister:
i'm getting totally opague errors out of app_spec/packager that are very time consuming to debug for someone not familiar with that code

looks like something i did causes the wasm not to end up in the generated JSON files in the packager

i can see the wasm and the JSON files separately, but when i view it in the output JSON it is just an empty string for code

new strategy is to clean up panics throughout the codebase in other branches until a real error apears in this branch, rather than have this end up being another thousand LOC

app_spec/zomes/blog/code/src/blog.rs
```diff
@@ -45,100 +45,100 @@ pub fn handle_check_sum(num1: u32, num2: u32) -> JsonString {
...
+    );
 
-    // match hdk::entry_address(&post_entry) {
-    //     Ok(address) => AddressResponse{address}.into(),
-    //     Err(hdk_error) => hdk_error.into(),
-    // }
+    match hdk::entry_address(&post_entry) {
+        Ok(address) => AddressResponse{address}.into(),
+        Err(hdk_error) => hdk_error.into(),
+    }
```
sphinxc0re: How about some declarative programming here:
```rust
hdk::entry_address(&post_entry)
    .map(|address| AddressResponse { address }.into())
    .unwrap_or_else(|hdk_err| hdk_err.into())
```

app_spec/zomes/blog/code/src/blog.rs
```diff
@@ -45,100 +45,100 @@ pub fn handle_check_sum(num1: u32, num2: u32) -> JsonString {
...
+                "authored_posts"
+            );
+
+            if link_result.is_err() {
+                return link_result.into()
+            }
+
+            let in_reply_to = in_reply_to;
+            if !in_reply_to.to_string().is_empty() {
+                if let Ok(_) = hdk::get_entry_result(in_reply_to.clone(), GetEntryOptions{}) {
```
sphinxc0re: please use `is_ok()` in such a case.

app_spec/zomes/blog/code/src/blog.rs
thedavidmeister: pattern matching doesn't fall under declarative programming?

thedavidmeister: http://wiki.c2.com/?PatternMatching

> A feature of FunctionalProgramming and LogicProgramming languages ... which is fundamental to making these programming paradigms declarative

sphinxc0re: Okay, it's more using the tools, the standard library supplies instead of reinventing the wheel over and over again. It makes it easier to reason about the code

core/src/nucleus/ribosome/run_dna.rs
```diff
@@ -27,7 +27,10 @@ pub fn run_dna(
     parameters: Option<Vec<u8>>,
 ) -> ZomeFnResult {
     // Create wasm module from wasm binary
-    let module = wasmi::Module::from_buffer(wasm).expect("wasm binary should be valid");
+    let module = match wasmi::Module::from_buffer(wasm) {
+        Ok(module) => module,
+        Err(e) => return Err(HolochainError::ErrorGeneric(e.into())),
+    };
```
sphinxc0re: You can use the `?` operator here

app_spec/zomes/blog/code/src/blog.rs
zippy: In that particular case the code looks almost identical and just as easy to reason about!

core/src/nucleus/actions/validate.rs
```diff
@@ -43,7 +43,17 @@ pub fn validate_entry<'a>(
...
+                EntryType::App(app_entry_type) => app_entry_type,
+                _ => {
+                    return FutureObj::new(Box::new(future::err(HolochainError::ValidationFailed(
+                        format!(
+                            "Attempted to validate system entry type {:?}",
+                            entry.entry_type()
+                        ),
+                    ))));
+                }
+            })
```
lucksus: Hm, these lines would be the reason imo to change `fn validate_entry` to be `async` which would mean we could replace this `return FutureObj...` with a `?`..

lucksus: Would also be nice to have a function:
```rust
fn to_app_entry_type() -> Result<String, HolochainError>
```
to change this to
```
.get_zome_name_for_entry_type(&entry.entry_type().to_app_entry_type()?)
```

core/src/nucleus/ribosome/callback/validate_entry.rs
```diff
@@ -61,8 +63,15 @@ fn validate_link_entry(
     validation_data: ValidationData,
     context: Arc<Context>,
 ) -> Result<CallbackResult, HolochainError> {
-    let link_add_entry = LinkAddEntry::from_entry(&entry);
-    let link = link_add_entry.link().clone();
+    let link_add = match entry {
+        Entry::LinkAdd(link_add) => link_add,
+        _ => {
```
lucksus: Also would be nice to have
```rust
impl Entry {
  pub fn to_link_add() -> Result<LinkAddEntry, HolochainError> {..},
  pub fn to_app_entry() -> Result<JsonString, HolochainError> {..},
  ...
}
```

core/src/nucleus/ribosome/run_dna.rs
lucksus: Please don't stall merges with change requests like this. We agreed that our etiquette suggests for pushing a commit to fix those kind of issues ourselves.

lucksus: done

lucksus: please confirm and hit merge, @sphinxc0re

core_types/src/dna/mod.rs
```diff
@@ -178,13 +178,19 @@ impl Dna {
...
+        let entry_type_name = String::from(app_entry_type.to_owned());
         // pre-condition: must be a valid app entry_type name
-        assert!(EntryType::has_valid_app_name(entry_type_name));
+        assert!(EntryType::has_valid_app_name(&entry_type_name));
         // Browse through the zomes
         for (zome_name, zome) in &self.zomes {
             for (zome_entry_type_name, _) in &zome.entry_types {
-                if *zome_entry_type_name == entry_type_name {
+                if *zome_entry_type_name
+                    == EntryType::App(AppEntryType::from(entry_type_name.to_string()))
```
zippy: yech, this seems so much clunkier that what it was. Sorry, just griping about rust, not really a useful comment, though I do see that his double nested for loop could be refactored into a closure and use by both `get_zome_name_for_app_entry_type` and `get_entry_type_def`  maybe a folowup?

core_types/src/dna/mod.rs
```diff
@@ -364,9 +357,12 @@ pub mod tests {
         };
         let mut zome = zome::Zome::default();
         zome.entry_types
-            .insert("".to_string(), zome::entry_types::EntryTypeDef::new());
+            .insert("".into(), zome::entry_types::EntryTypeDef::new());
         dna.zomes.insert("".to_string(), zome);
 
+        let expected = JsonString::from(dna.clone());
+        println!("{:?}", expected);
```
zippy: forgot to remove println from test?

core_types/src/dna/mod.rs
thedavidmeister: @zippy `EntryType::from(entry_type_name)` probably works too

core/src/nucleus/actions/build_validation_package.rs
```diff
@@ -36,7 +36,19 @@ pub fn build_validation_package(entry: &Entry, context: &Arc<Context>) -> Valida
             .nucleus()
             .dna()
             .unwrap()
-            .get_zome_name_for_entry_type(&entry.entry_type().to_string())
+            .get_zome_name_for_app_entry_type(&match entry.entry_type() {
```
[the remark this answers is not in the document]
sphinxc0re: I wouldn't do that either. If the match statement returns, the whole chain of function calls before is useless.

core/src/link_tests.rs
```diff
@@ -66,25 +66,25 @@ pub mod tests {
...
+                    assert_eq!(entry.entry_type(), EntryType::LinkList,);
+                    assert_eq!(entry.content(), link_list_entry.content());
                     true
                 }
                 _ => false,
             });
     }
 
-    /// Committing a LinkListEntry to source chain should work
+    /// Committing a LinkList to source chain should work
```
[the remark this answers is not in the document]
ddd-mtl: yeah, LinkListEntry can be removed, its not useful anymore.

sphinxc0re:
@lucksus I'm on it

PR #650, opened 2018-11-20
Added nodejs as test runner

sphinxc0re:
There is still one test failing and I can't see why. I fixed this in my previous version so this is really odd.


Connoropolous:
@zippy @sphinxc0re I know why its failing its because our latest npm publish of holochain-nodejs didn't work out quite right. We used the wrong approach. Trying to fix that now

PR #652, opened 2018-11-21
RFC: on get entry result struct

thedavidmeister:
prelim notes/thoughts:

- as this involves `Entry` in the struct, it would be best to wait until after https://github.com/holochain/holochain-rust/pull/649 lands so we don't have to juggle `SerializedEntry`
- what is `Map<Address, Address>` representing in `crud_links`? should this be an `EAV`?
- would it sit better as `Vec<Results>` rather than `Results{x: Vec<>, y: Vec<>, z: Vec<>}`?
- this is replacing `Option<Entry>`, so what are the vecs? previous states? multiple entries? what does "not found" look like?
- how should `GetEntryOptions` be extended?


zippy:
This has now been completed

PR #653, opened 2018-11-21
WIP: add wasm to workspace

thedavidmeister:


Cargo.toml
```diff
@@ -14,3 +18,8 @@ members = [
   "test_bin",
 ]
 exclude = ["test_utils"]
+
+[profile.release]
```
sphinxc0re: This can't be good. This means that every panic in the whole project will abort instead of unwinding the stack. I don't think that's what we want

thedavidmeister: @sphinxc0re this is pulled from existing config

sphinxc0re: But this shouldn't be in the root `Cargo.toml`. Every wasm project should define it's own config

thedavidmeister:
@lucksus @sphinxc0re i'm going to have another crack at this in context of #784 

it will look different to what is here i think... the compilation options are being controlled a bit differently now

PR #657, opened 2018-11-21
Crud works locally

ddd-mtl:
See: https://hackmd.io/VH6oxwQWSKG1M9nqXNfNZg?both

# CRUD works locally
Implemented hc_update_entry() and hc_remove_entry()

## API changes
@Connoropolous 

HDK API now has get_entry_initial() and get_entry_history().
get_entry() gets latest entry.

## Internals

Added UpdateEntry and RemoveEntry actions and ActionCreators.
Everything is mostly done in the DHT, since metadata is stored in its eav store.
DHT now holds a generic actions hashMap that returns an address.

Added link_crud in ChainHeader which is filled in by the Action::Commit which now has an extra argument: Option<Address>. Normal commit action is None, while modify entry commit is Some(...).

Found decent work-around for missing update/remove_eav() api in eav store.

### get_entry
The Ribosome's hc_get_entry() now takes a GetEntryArgs as input which holds a GetEntryOptions.
It gives back a GetEntryResult which holds all the entries it walked.
The get_entry action creator now walks the crud-links in a recursive way. see get_entry_rec()
Tests got changed accordingly.

## Misc.
Updated build.bat script for local windows building.


build.bat
```diff
@@ -1,6 +1,4 @@
-cd core/src/nucleus/wasm-test && cargo +nightly-2018-10-12 build --release --target wasm32-unknown-unknown && cd ../../../../
-cd core_api/wasm-test/round_trip && cargo +nightly-2018-10-12 build --release --target wasm32-unknown-unknown && cd ../../../
-cd core_api/wasm-test/commit && cargo +nightly-2018-10-12 build --release --target wasm32-unknown-unknown && cd ../../../
+cd container_api/wasm-test && cargo +nightly-2018-10-12 build --release --target wasm32-unknown-unknown && cd ../../
```
Connoropolous: You can use the `--manifest-path` option to clean up these lines so they don't need the cd commands :)

ddd-mtl: I know. I'm doing this on purpose so when it fails, you are in the failed directory.

Connoropolous: ah

core/src/nucleus/ribosome/api/update_entry.rs
```diff
@@ -0,0 +1,81 @@
...
+    // Wait for future to be resolved
+    let task_result: Result<Address, HolochainError> = block_on(
+        // 1. Build the context needed for validation of the entry
+        build_validation_package(&entry, &runtime.context)
+            .and_then(|validation_package| {
+                Ok(ValidationData {
+                    package: validation_package,
+                    sources: vec![HashString::from("<insert your agent key here>")],
+                    lifecycle: EntryLifecycle::Chain,
+                    action: EntryAction::Commit,
```
lucksus:
```suggestion
                    action: EntryAction::Modify,
```

core/src/nucleus/ribosome/api/update_entry.rs
```diff
@@ -0,0 +1,81 @@
...
+                validate_entry(
+                    entry.entry_type().clone(),
+                    entry.clone(),
+                    validation_data,
+                    &runtime.context,
+                )
+            })
+            // 3. Commit the valid entry to chain and DHT
+            .and_then(|_| {
+                commit_entry(
```
lucksus: We either have to
* tell commit_entry somehow that this is a modification commit so that it creates a different chain header
or we need to
* introduce a modify_entry action that takes care of that.

core/src/nucleus/ribosome/api/remove_entry.rs
```diff
@@ -0,0 +1,31 @@
...
+    // Exit on error
+    if try_address.is_err() {
+        println!(
+            "invoke_remove_entry failed to deserialize Address: {:?}",
+            args_str
+        );
+        return ribosome_error_code!(ArgumentDeserializationFailed);
+    }
+    let address = try_address.unwrap();
+    let future = remove_entry(&runtime.context, &runtime.context.action_channel, address);
```
lucksus: remove_entry only triggers the DHT EAV handling for this entry. There is no validation of this action currently. We need to create, validate and commit a `EntryType::Delete` entry.

ddd-mtl:
@lucksus I did as you suggested and build is green.

build.bat
thedavidmeister: hmmm, this might need to revert to `--manifest-path` at some point, but i can't remember why...

core/src/nucleus/actions/get_entry.rs
```diff
@@ -48,15 +48,17 @@ pub(crate) fn get_entry_meta_from_dht(
...
     if has_deleted {
         crud_status = CrudStatus::DELETED;
     } else {
         let has_modified = status_eavs
             .iter()
             .filter(|e| CrudStatus::from(String::from(e.value())) == CrudStatus::MODIFIED)
             .collect::<HashSet<&EntityAttributeValue>>()
-            .len() > 0;
+            .len()
+            > 0;
```
sphinxc0re: Better use `!something.is_empty()`

core/src/nucleus/ribosome/callback/links_utils.rs
```diff
@@ -15,13 +15,29 @@ pub fn get_link_entries(
...
-    let target = block_on(get_entry(&context, target_address.clone()))?.ok_or(
-        HolochainError::ErrorGeneric(String::from("Target for link not found")),
-    )?;
-    Ok((base, target))
+    let entry_args = &GetEntryArgs {
+        address: base_address.clone(),
+        options: GetEntryOptions::default(),
+    };
+    let base_result = block_on(get_entry(&context, entry_args))?;
+    if base_result.entries.len() == 0 {
```
sphinxc0re: better use `is_empty()`

core/src/nucleus/ribosome/callback/links_utils.rs
```diff
@@ -15,13 +15,29 @@ pub fn get_link_entries(
...
+            "Base for link not found",
+        )));
+    }
+    let base = base_result.entries.iter().next().unwrap();
+    let entry_args = &GetEntryArgs {
+        address: target_address.clone(),
+        options: GetEntryOptions::default(),
+    };
+    let target_result = block_on(get_entry(&context, entry_args))?;
+    if target_result.entries.len() == 0 {
```
sphinxc0re: better use `is_empty()`

core/src/nucleus/ribosome/api/update_entry.rs
ddd-mtl: Done

ddd-mtl:
Back to WIP status now that DHT and Network got separated.

core/src/action.rs
```diff
@@ -73,10 +73,13 @@ impl Hash for ActionWrapper {
 pub enum Action {
     /// entry to Commit
     /// MUST already have passed all callback checks
-    Commit(Entry),
+    Commit((Entry, Option<Address>)),
     /// GetEntry by address
     GetEntry(Address),
-
+    ///
+    UpdateEntry((Address, Address)),
```
sphinxc0re: Why the extra `()`

core/src/dht/dht_reducers.rs
```diff
@@ -39,8 +46,11 @@ pub fn reduce(
 /// Maps incoming action to the correct reducer
 fn resolve_reducer(action_wrapper: &ActionWrapper) -> Option<DhtReducer> {
     match action_wrapper.action() {
+        Action::Commit(_) => Some(reduce_add_crud_meta),
```
lucksus: This should not happen on `Commit`, but on `Hold`. The DHT should not listen to `Commit`. That was short cut that doesn't make sense when we have networking.

core/src/dht/dht_reducers.rs
```diff
@@ -68,6 +78,32 @@ pub(crate) fn reduce_hold_entry(
...
+        // TODO #439 - Log the error. Once we have better logging.
+        println!(
+            "reduce_add_crud_meta: meta_storage write failed!: {:?}",
+            res.err().unwrap()
+        );
+        return None;
+    }
+    // Done
+    Some(new_store)
+}
```
lucksus: This function can and should be called from `reduce_hold_entry`, instead of bound to `Action::Commit`.

core/src/dht/dht_reducers.rs
ddd-mtl: I don't think that works because `Hold` can be called for anything: create, modify delete.
I need to do a specific thing only when its a 'create' and `Commit` seemed to do just that.
Generally I see a confusion between 'Commit' and 'Create' which needs to be cleared out.

core/src/action.rs
ddd-mtl: Its an enum variant, which is a tuple.
Don't know any other way to make the unwrap_to! macro work.

core/src/nucleus/ribosome/api/remove_entry.rs
```diff
@@ -0,0 +1,74 @@
...
+            // 3. Commit the valid entry to chain and DHT
+            .and_then(|_| commit_entry(deletion_entry.clone(), None, &runtime.context))
+            // 4. Remove the entry in DHT metadata
+            .and_then(|_| {
+                remove_entry(
+                    &runtime.context,
+                    &runtime.context.action_channel,
+                    deleted_entry_address,
+                )
+            }),
```
lucksus: This should go to `workflows/remove_entry.rs`

lucksus: Well, actually, this should just be the `author_entry` workflow with this deletion entry and then the publish reducer has to discern what "meta" message to send to the network. The DHT should only be triggered from the network handler.

core/src/nucleus/ribosome/api/update_entry.rs
```diff
@@ -0,0 +1,84 @@
...
+            // 3. Commit the valid entry to chain and DHT
+            .and_then(|_| commit_entry(entry.clone(), Some(chain_header_address), &runtime.context))
+            // 4. Update the entry in DHT metadata
+            .and_then(|new_address| {
+                update_entry(
+                    &runtime.context,
+                    &runtime.context.action_channel,
+                    entry_args.address.clone(),
+                    new_address,
+                )
```
lucksus: Same comment as above in  core/src/nucleus/ribosome/api/remove_entry.rs

core/src/nucleus/ribosome/callback/validate_entry.rs
```diff
@@ -52,6 +52,10 @@ pub fn validate_entry(
             context,
         )?),
 
+        // Deletion entries are not validated currently and always valid
+        // TODO: Specify how Deletion can be commited to chain.
+        EntryType::Delete => Ok(CallbackResult::Pass),
```
lucksus: Deletion should be validated with the same validation callback used for commit. You are already setting the action field in the validation data to `EntryAction::Delete` above. That gets passed into the callback anyway.

core_types/src/chain_header.rs
```diff
@@ -30,6 +30,8 @@ pub struct ChainHeader {
     link: Option<Address>,
     /// Key to the most recent header of the same type, None is valid only for the first of that type
     link_same_type: Option<Address>,
+    /// Key to the header of the previous version of this chain header's entry
+    link_crud: Option<Address>,
```
lucksus: Not sure if `link_crud` is the best name for this as it is only for modification and the existence of this address means that this header is actually a header of a modification event. I think the term `modification` should be more prominent somehow.

hdk-rust/src/api.rs
```diff
@@ -689,13 +720,36 @@ pub fn verify_signature<S: Into<String>>(
 }
 
 /// Not Yet Available
```
lucksus: Rustdoc neeeds updating

ddd-mtl: Done

core_types/src/chain_header.rs
ddd-mtl: it also links to the deletion entry, so its not just modification.

core/src/nucleus/ribosome/callback/validate_entry.rs
ddd-mtl: I don't understand how it can be same validation as commit/create we are not putting new data that needs to be validated.

core/src/dht/dht_reducers.rs
```diff
@@ -131,11 +167,225 @@ pub(crate) fn reduce_add_link(
...
+            Err(HolochainError::ErrorGeneric(String::from(
+                "new_entry is not present in DHT's CAS",
+            ))),
+        );
+        return Some(new_store);
+    }
+    // pre-condition: old_entry's latest version must have LIVE crud-status
+    // get latest entry
+    let mut entry_result = GetEntryResult::new();
+    let res = get_entry_rec(
```
lucksus: Hm, this won't work when we switch to network code. `get_entry_rec` will need to do recursive network look-ups. That can't happen inside a reducer since it has to involve actions that are reduced as well. This (DHT) reducer should only care for the last step, i.e. if it has the entry that needs it crud modify link to be set.

`get_entry_rec` should be called in the according action creator of `update_entry`.
And that action creator should be called in network/handler.rs when processing a DhtMetaData with attribute "modify".

hdk-rust/tests/integration_test.rs
```diff
@@ -340,10 +347,11 @@ fn can_link_entries() {
...
 }
 
-/* This test now fails because handle_links_roundtrip doesn't take into
- account how long it takes for the links to propigate on the network
-the correct test would be to wait for a propigation period
+// This test now fails because handle_links_roundtrip doesn't take into
+// account how long it takes for the links to propigate on the network
+//the correct test would be to wait for a propigation period
 #[test]
+#[cfg(not(windows))]
```
lucksus: This doesn't compile because of missing use statements.

core/src/nucleus/ribosome/callback/validate_entry.rs
lucksus: From a DNA developer perspective: the same (the one and only) validation callback is called, but with a `EntryAction::Delete`. The content should be the existing (to be deleted) entry.

core/src/action.rs
sphinxc0re: I don't think it's a good idea to make this unnecessary change just because `unwrap_to!` doesn't work with it.

sphinxc0re: nvm 😁 I saw the usage of this and really like it.

core/src/agent/state.rs
```diff
@@ -75,16 +79,20 @@ impl AgentState {
...
+            address: agent_entry_address,
+            options: GetEntryOptions::default(),
+        };
+        let agent_entry_result = await!(get_entry(context, &entry_args))?;
+        if agent_entry_result.entries.is_empty() {
+            return Err(HolochainError::ErrorGeneric(
+                "Agent entry not found".to_string(),
+            ));
+        }
+        let agent_entry = agent_entry_result.entries.iter().next().unwrap().clone();
```
sphinxc0re: What type is `agent_entry_result.entries` ? couldn't you just do:
```rust
let agent_entry = agent_entry_result.entries[0].clone();
```
?

PR #661, opened 2018-11-22
Websocket container interface MVP

maackle:
* Adds `Interface` trait which only needs a `jsonrpc_core::IoHandler` injected into it, paving the way for third-party interface authors (different drivers)
* Adds `WebsocketInterface` struct, the first implementation of `Interface`

## TODO/Unfinished:
Maybe we can merge this before doing these extras, or maybe should add some:

* [X] Tests for `ContainerApiDispatcher`
* [ ] Can only start an interface, currently cannot stop one (interfaces run in their own threads, so this needs to be implemented per interface)
* [ ] Finish HttpInterface because why not?
* [ ] Make Configuration extensible to allow custom config types for interfaces (otherwise the extensible Interface trait is kinda worthless)
    * one idea here is to make Configuration take a type parameter for InterfaceProtocol which is a struct of all the different kinds of interfaces, and also inject a function that takes an InterfaceConfiguration and creates a Box<Interface> or something


container_api/src/container.rs
```diff
@@ -8,49 +7,96 @@ use Holochain;
 
 use holochain_core::{logger::Logger, persister::SimplePersister};
 use holochain_core_types::agent::Agent;
+use holochain_net::p2p_network::P2pNetwork;
```
maackle: remove

container_api/src/container.rs
```diff
@@ -8,49 +7,96 @@ use Holochain;
...
+use interface::{self, DispatchRpc, InstanceMap, Interface, RpcDispatcher};
+use jsonrpc::JsonRpc;
 
 /// Main representation of the container.
 /// Holds a `HashMap` of Holochain instances referenced by ID.
-///
+
 /// A primary point in this struct is
-/// ```load_config(&mut self, config: &Configuration) -> Result<(), String>```
+/// `load_config(&mut self, config: &Configuration) -> Result<(), String>`
```
maackle: changed this because it was breaking my vim syntax highlighting 🤔

container_api/src/container.rs
```diff
@@ -8,49 +7,96 @@ use Holochain;
...
 /// While doing so it has to load DNA files referenced in the configuration.
 /// In order to not bind this code to the assumption that there is a filesystem
 /// and also enable easier testing,
 /// a DnaLoader has to be injected on creation.
 /// This is a closure that returns a Dna object for a given path string.
 pub struct Container {
-    pub instances: HashMap<String, Holochain>,
+    pub instances: InstanceMap,
+    config: Configuration,
+    // interface_threads: HashMap<String, thread::JoinHandle<Result<(), String>>>,
```
maackle: remove comment

container_api/src/container.rs
```diff
@@ -8,49 +7,96 @@ use Holochain;
...
 type DnaLoader = Arc<Box<FnMut(&String) -> Result<Dna, HolochainError> + Send>>;
 
 impl Container {
     /// Creates a new instance with the default DnaLoader that actually loads files.
-    pub fn new() -> Self {
+    pub fn with_config(config: Configuration) -> Self {
         Container {
             instances: HashMap::new(),
+            interface_threads: HashMap::new(),
+            config,
```
maackle: need to have a Configuration object handy to spin up instances from

container_api/src/container.rs
```diff
@@ -119,19 +163,58 @@ impl Container {
...
+            iface.run(Arc::new(dispatcher))
+        })
+    }
 }
 
 impl<'a> TryFrom<&'a Configuration> for Container {
     type Error = HolochainError;
     fn try_from(config: &'a Configuration) -> Result<Self, Self::Error> {
-        let mut container = Container::new();
+        let mut container = Container::with_config((*config).clone());
```
maackle: yes this is redundant, need to clean up the API more to prevent loading the config twice

container_api/src/jsonrpc.rs
```diff
@@ -0,0 +1,49 @@
...
+    pub method: String,
+    pub params: Value,
+    pub id: u32,
+}
+
+pub fn jsonrpc_success(id: u32, result: JsonString) -> JsonString {
+    json!({
+        "jsonrpc": "2.0",
+        // Question, is there a better way to safely go from JsonString into Value?
+        "result": serde_json::from_str::<Value>(&result.to_string()).unwrap(),
```
maackle: Question for ya @thedavidmeister, can we make an `impl Into<serde_json::Value> for JsonString`? Or is there one I missed?

container_api/src/container.rs
```diff
@@ -8,59 +7,93 @@ use Holochain;
...
+    pub dna_loader: DnaLoader,
 }
 
+type InterfaceThreadHandle = thread::JoinHandle<Result<(), String>>;
 type DnaLoader = Arc<Box<FnMut(&String) -> Result<Dna, HolochainError> + Send>>;
 
 impl Container {
     /// Creates a new instance with the default DnaLoader that actually loads files.
-    pub fn new() -> Self {
+    pub fn with_config(config: Configuration) -> Self {
```
lucksus:
```suggestion
    pub fn from_config(config: Configuration) -> Self {
```

container_api/src/container.rs
```diff
@@ -111,6 +145,16 @@ impl Container {
         }
     }
 
+    fn start_interface(&mut self, config: &InterfaceConfiguration) -> Result<(), String> {
+        if self.interface_threads.contains_key(&config.id) {
+            return Err(format!("Interface {} already started!", config.id));
+        }
+        let dispatcher = self.make_dispatcher(config);
```
lucksus: dispatcher is not used?!

container_api/src/container.rs
```diff
@@ -119,19 +163,58 @@ impl Container {
...
+            .map(|(id, val)| (id.clone(), val.clone()))
+            .collect();
+        RpcDispatcher::new(&self.config, instance_subset)
+    }
+
+    fn spawn_interface_thread(
+        &self,
+        interface_config: InterfaceConfiguration,
+    ) -> InterfaceThreadHandle {
+        let dispatcher = self.make_dispatcher(&interface_config);
```
lucksus: Ok, this is where the dispatcher gets created **and used**.

container_api/src/container.rs
lucksus:
```suggestion
```

container_api/src/interface.rs
```diff
@@ -0,0 +1,82 @@
...
+use config::{Configuration, InstanceConfiguration};
+
+pub type InterfaceError = String;
+pub type InstanceMap = HashMap<String, Arc<Mutex<Holochain>>>;
+
+pub trait DispatchRpc {
+    fn dispatch_rpc(&self, rpc: &JsonRpcRequest) -> JsonString;
+}
+
+pub struct RpcDispatcher {
```
lucksus: I think I would call this thing `ContainerAPI` and move it in its own file. The implementation of `dispatch_rpc` below will grow over time to be the central point of this whole crate, no?

container_api/src/container.rs
maackle: @lucksus don't we use the `with_` convention for constructors? Why `from_` here?

lucksus: Actually, it should be the `From<T>` trait: https://doc.rust-lang.org/std/convert/trait.From.html

But the config could be broken, which is why I then used TryFrom below. Why do we need a new() at all?

container_api/src/interface.rs
lucksus: I would call the set of available JSON-RPC functions Container API.

container/example-config/basic.toml
```diff
@@ -21,7 +21,17 @@ path = "example-config/tmp-storage"
...
 [[interfaces]]
+id = "websocket interface"
+[interfaces.driver]
 type = "websocket"
-port = 8888
+port = 3000
 [[interfaces.instances]]
 id = "app spec instance"
+
+# [[interfaces]]
```
thedavidmeister: why is this left in with comments?

container_api/src/container.rs
```diff
@@ -8,68 +7,95 @@ use Holochain;
...
+    pub fn start_all_interfaces(&mut self) {
+        self.interface_threads = self
+            .config
+            .interfaces
+            .iter()
+            .map(|ic| (ic.id.clone(), self.spawn_interface_thread(ic.clone())))
+            .collect()
+    }
+
+    // pub fn stop_all_interfaces(&mut self) {
```
thedavidmeister: remove commented out code

container_api/src/container.rs
```diff
@@ -8,68 +7,95 @@ use Holochain;
...
-    pub fn start_all(&mut self) {
-        let _ = self.instances.iter_mut().for_each(|(id, hc)| {
+    pub fn start_all_instances(&mut self) {
+        self.instances.iter_mut().for_each(|(id, hc)| {
             println!("Starting instance \"{}\"...", id);
-            match hc.start() {
-                Ok(()) => println!("ok"),
-                Err(err) => println!("Error: {}", err),
-            }
+            hc.write().unwrap().start().expect("Could not start server")
```
thedavidmeister: return a result?

container_api/src/container.rs
```diff
@@ -8,68 +7,95 @@ use Holochain;
...
-    pub fn stop_all(&mut self) {
-        let _ = self.instances.iter_mut().for_each(|(id, hc)| {
+    pub fn stop_all_instances(&mut self) {
+        self.instances.iter_mut().for_each(|(id, hc)| {
             println!("Stopping instance \"{}\"...", id);
-            match hc.stop() {
-                Ok(()) => println!("ok"),
-                Err(err) => println!("Error: {}", err),
-            }
+            hc.write().unwrap().stop().expect("Could not stop server")
```
thedavidmeister: return a result?

container_api/src/container.rs
```diff
@@ -119,19 +155,60 @@ impl Container {
...
         Ok(container)
     }
 }
 
+/// This can eventually be dependency injected for third party Interface definitions
+fn make_interface(
+    interface_config: &InterfaceConfiguration,
+) -> Box<Interface<ContainerApiDispatcher>> {
+    match interface_config.driver {
+        // InterfaceDriver::Http { port } => Box::new(interface_impls::http::HttpInterface::new(port)),
```
thedavidmeister: cruft

container_api/src/container.rs
```diff
@@ -257,16 +341,16 @@ pub mod tests {
     fn test_container_load_config() {
         let config = load_configuration::<Configuration>(test_toml()).unwrap();
 
-        let mut container = Container {
-            instances: HashMap::new(),
-            dna_loader: test_dna_loader(),
-        };
+        // TODO: redundant
```
thedavidmeister: add a followup issue?

container_api/src/container.rs
```diff
@@ -283,4 +367,23 @@ pub mod tests {
...
+        let mut container = Container::with_config(config.clone());
+        container.dna_loader = test_dna_loader();
+        assert!(container.load_config(&config).is_ok());
+
+        let instance_config = &config.interfaces[0];
+        let dispatcher = container.make_dispatcher(&instance_config);
+        let io = dispatcher.io;
+
+        let request = r#"{"jsonrpc": "2.0", "method": "info/instances", "params": null, "id": 1}"#;
```
thedavidmeister: JsonString?

container_api/src/interface.rs
```diff
@@ -0,0 +1,107 @@
...
+                                let zome_name = zome_name.clone();
+                                let cap_name = cap_name.clone();
+                                let method_name = format!(
+                                    "{}/{}/{}/{}",
+                                    instance_id, zome_name, cap_name, func_name
+                                );
+                                let hc_lock_inner = hc_lock.clone();
+                                self.io.add_method(&method_name, move |params| {
+                                    let mut hc = hc_lock_inner.write().unwrap();
+                                    let params_string = serde_json::to_string(&params).map_err(
```
thedavidmeister: JsonString?

container_api/src/interface_impls/http.rs
```diff
@@ -0,0 +1,34 @@
...
+            let method = request.url().to_string();
+            println!("{}", method);
+            let response = Response::from_string(method);
+            request.respond(response).unwrap();
+        }
+        unimplemented!();
+    }
+}
+
+// fn mk_err(msg: &str) -> JsonString {
```
thedavidmeister: cruft

container_api/src/container.rs
maackle: Not in this case, it's a test, the string is being created raw, and it's being used in a function that only takes String

container_api/src/interface.rs
maackle: Not until `hc.call` accepts JsonString for params

container_api/src/container.rs
maackle: done here https://github.com/holochain/holochain-rust/issues/674

container_api/src/container.rs
maackle: 👍

container_api/src/interface.rs
```diff
@@ -0,0 +1,108 @@
...
+        let mut this = Self { instances, io };
+        this.setup_info_api(instance_configs);
+        this.setup_zome_api();
+        this
+    }
+
+    fn setup_info_api(&mut self, instance_configs: HashMap<String, InstanceConfiguration>) {
+        self.io.add_method("info/instances", move |_| {
+            let configs = instance_configs.clone();
+            let config_string = serde_json::to_string(&configs)
```
thedavidmeister: JsonString

container_api/src/config.rs
```diff
@@ -100,6 +100,14 @@ impl Configuration {
             .and_then(|instance_config| Some(instance_config.clone()))
     }
 
+    /// Returns the interface configuration with the given ID if present
+    pub fn interface_by_id(&self, id: &String) -> Option<InterfaceConfiguration> {
+        self.interfaces
+            .iter()
+            .find(|ic| &ic.id == id)
+            .and_then(|interface_config| Some(interface_config.clone()))
```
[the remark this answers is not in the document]
lucksus: Please go ahead and fix small things like this yourself. Other than this, this branch can be merged and should enable you to build 'hc run'.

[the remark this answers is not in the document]
zippy: Yes, I agree!  Lets get this merged!

container/src/main.rs
```diff
@@ -51,7 +51,11 @@ fn main() {
...
                 );
                 println!("Starting all of them...");
-                container.start_all();
+                container
+                    .start_all_instances()
+                    .expect("Could not start instances!");
+                println!("Starting interfaces...");
+                container.start_all_interfaces();
                 println!("Done.");
                 loop {}
```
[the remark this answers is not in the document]
maackle: @sphinxc0re what, the `loop`? I didn't write that but it lets the thing keep running. Or do you mean `start_all_interfaces`? Well, I think the code explains it but it spawns a new thread for each configured interface, and currently those threads just run forever

maackle:
@lucksus yeah HTTP should be easy although it feels like overkill to use JSON-RPC for HTTP requests so might want to broaden the Interface abstraction. The IoHandler is strictly for JSON-RPC requests.

Also I should mention there can be some cleanup here, the `DispatchRpc` trait is mostly redundant, it existed when I had a custom dispatcher, before I was using `IoHandler` for dispatching. Now it's just a useless wrapper around an `IoHandler`, but that is maybe where we can put the more general handling  for requests to get us away from only JSON-RPC.

PR #663, opened 2018-11-22
Setup for holochain-nodejs deploys

Connoropolous:
What this does, and how to use it:

We need a way to upload binary releases of holochain-nodejs to this github repo.
So we use travis automation. These changes are required to make it work for macos, linux, and windows.
There are instructions in the nodejs_container README for how to perform a release. It involves tagging a commit on `develop` or `master` with a tag of the format `holochain-nodejsY.Y.Y` where Y.Y.Y matches the version number in `nodejs_container/package.json` (which should first have been modified using `npm version [type]`

This will cause travis to upload binary files for use by npm packages here
https://github.com/holochain/holochain-rust/releases

this goes with the remote package name in package.json
https://github.com/holochain/holochain-rust/blob/holochain-nodejs0.1.22/nodejs_container/package.json#L38


.travis.yml
```diff
@@ -4,6 +4,7 @@ branches:
   only:
     - master
     - develop
+    - /^holochain-nodejs\d+\.\d+\.\d+/
```
Connoropolous: This line will get removed before merging it in, just needed it for testing

.travis.yml
```diff
@@ -96,24 +97,21 @@ _nodejs_deploy_template: &_nodejs_deploy_template
...
+          secure: SfAyYHw8iC74CY8SrSnEsA0GPLxpCwguYziw19ZlBz6Nq+HxqNWoexw87TdrTIkxUKJpNxoLzqeKmigpAzr8Q4Y8j72tD5JyTSLwTnjChclM1BMRh7NR3dL8gTrBRuRZ4ALJr4EP1sxXI16W3lEz7hrS4tyws2e0W3xb80Os7p9fNQHpKCAncON622mDSznPbUXbUCNdosqZq+dZua6RZ9xmMaba1La3+vZn+hzvcNvI2mtiK/KOIDbo4uzZEUf7URtKmtkslilIc+3568iuCjPl5zV1s+/VN2/SEuxcxuVtxXUV6ZVguniF0z4L5YOCwnnadUJtlXYxxQznRu/YT7xvlvitSNulDm81jACiwKZ7LbbXssfzIb+gTBXiTdFyUDJUEHL8U992K7oDJN2nM6jPsqnmjuqalFcV/raVmpZA/gA3eCqyCw8is363NqEFpGgczBnL84h81x63RrXsXb6oZq2E5fQkIHP/4srG/s359achguHmaf/g69NiAnhCz9P2xF15zFqRPZnUA7q9ONpSAJrkOWVxBEntfIsLL6P07y3rCdXhs6A6Q5lh5YcWdOH2zEtsVXaI2/AbVgB4g0CWChts6tpqrNa5QzRIF82LLjhQTZOkpBUwMliunzmHMhIz0ojHMjay0JrCuYF2lbUfQ/f3S4Clnz19Z5DPo/A=
```
Connoropolous: When using travis-ci.org, you have to run
`travis setup releases --org`

Then what I do is copy out the resulting `secure` prop from the file, reset the branch, and then copy the new `secure` prop back in because I don't want to overwrite the rest of the configuration

docker/Makefile
```diff
@@ -62,7 +62,7 @@ install_system_libzmq:
 		if ! which apt-get ; then \
 			if which brew ; then \
 				echo "\033[0;93m## Attempting to install zmq using homebrew ##\033[0m"; \
-				brew install zmq \
+				brew install zmq; \
```
Connoropolous: apparently this wasn't tested, cause it didn't work without the semicolon

.travis.yml
```diff
@@ -73,7 +74,7 @@ _macosx_nodejs_template: &_macosx_nodejs_template
   rust: nightly
 
 _to_deploy_template: &_to_deploy_template
-  if: tag =~ /^v\d+\.\d+\.\d+/
+  if: tag =~ /^holochain-nodejs\d+\.\d+\.\d+/
```
maackle: Might be nice to have a dash if possible or even a dash-vee: 
`holochain-nodejs-1.2.3` or
`holochain-nodejs-v1.2.3`

nodejs_container/README.md
```diff
@@ -39,22 +52,23 @@ Commit this.
 
 Push it to github.
 
-Create a tag on github of the format `holochain-nodejsY.Y.Y` where `Y.Y.Y` is the version number of the tag. This is really important.
+Create a tag on github of the format `holochain-nodejsY.Y.Y` where `Y.Y.Y` is the version number of the tag. This is really important, as only a tag with this format will trigger release builds to happen. This is configured in the .travis.yml file.
```
maackle: see above, of dashes and vees

.travis.yml
Connoropolous: you know, I like that idea. Lets do it

cmd/src/cli/js-tests-scaffold/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "dependencies": {
-    "@holochain/holochain-nodejs": "^0.1.21",
+    "@holochain/holochain-nodejs": "^0.1.22",
```
zippy: This seems like a problem.  Won't it break some of our tests cmd regression tests when building out a new feature?

Connoropolous: Not sure what you mean. 
I can't think of anything that this would break

scripts/build_nodejs_container.sh
```diff
@@ -0,0 +1,3 @@
+cd nodejs_container
```
thedavidmeister: hmm, i'll have to chase this with nix, but that's ok

cmd/src/cli/js-tests-scaffold/package.json
zippy: If we have a test, that tests whether the code generated by `hc init` works, i.e. runs `hc test` on that code, then it could fail because it would use a fixed version, not the "current" version in development.  At one point I thought we had such a test but maybe we don't.

PR #665, opened 2018-11-23
Rust Edition 2018

lucksus:
The new Rust edition becomes stable in two weeks (December 6th).
https://rust-lang-nursery.github.io/edition-guide/rust-2018

`async`/`await` almost made it into that edition as a stable thing, but is a feature in 2018 that can be activated:
https://github.com/rust-lang/rfcs/blob/master/text/2394-async_await.md

Newer Future releases (0.3) are tailored for async/await:
https://rust-lang-nursery.github.io/futures-rs/blog/2018/07/19/futures-0.3.0-alpha.1.html

In the course of working on the network module, I ran into problems that @thedavidmeister had some months back with actors: with the current 0.2 Futures we can not nest `block_on`s. That really is a problem.

Since it's only two weeks until Rust 2018 is stable, I suggest to make the switch now so I can proceed with the networking code.

The only two things changed here are:
* modules used from within the same crate need a `crate::` prefix (https://rust-lang-nursery.github.io/edition-guide/rust-2018/module-system/path-clarity.html). That is changed in this PR in almost every file.
* `try` has become a reserved word. we had one variable named `try`


AshantiMutinta:
imo the closer we are to the stable channel the better. This is a plus for me.

maackle:
Looks good, also a suggestion for other readers, this PR reads way better one commit at a time rather than all changes at once.

zippy:
Looks good but CI failing with tarpaulin run.

PR #671, opened 2018-11-26
CAS/EAV without actor

lucksus:
This (hopefully) fixes the problems I was facing in the networking PR: nested `block_on`s.

This removes `block_on`s, Futures, Actors and Riker from CAS implementations and just uses `Arc<RwLock<_>>`s to make the storages thread-safe.

-500 LOC


cas_implementations/src/cas/file.rs
```diff
@@ -1,78 +1,77 @@
...
 use uuid::Uuid;
 
-#[derive(Clone, PartialEq, Debug)]
+#[derive(Clone, Debug)]
 pub struct FilesystemStorage {
-    actor: ActorRef<Protocol>,
+    /// path to the directory where content will be saved to disk
+    dir_path: String,
     id: Uuid,
+    lock: Arc<RwLock<()>>,
```
AshantiMutinta: This looks great so far more lightweight! I was wondering if we need the extra guard since we are using this behind the ARC and Mutex

lucksus: Hm, you're saying because we have the storage wrapped in `Arc<RwLock<>>` in the context already we don't even need this lock here? You're maybe right...

cas_implementations/src/cas/file.rs
```diff
@@ -100,7 +99,7 @@ pub mod tests {
 
     #[test]
     /// show that content of different types can round trip through the same storage
-    /// this is copied straight from the example with a file CAS
+    /// this is copied straight from the example with a aasdf CAS
```
zippy:
```suggestion
    /// this is copied straight from the example with a file CAS
```

cas_implementations/src/cas/file.rs
AshantiMutinta: Oh yeah! Cause the writes will be around that lock so you can't do a write and read without unwrapping which in order to do that frees it from it's lock :)

PR #682, opened 2018-11-27
MDBOOK: Update intro_to_command_line_tools.md

chrisminnick:


doc/holochain_101/src/intro_to_command_line_tools.md
```diff
@@ -1,12 +1,8 @@
 # Intro to Command Line Tools
 
-It will likely not come as much of a surprise that to develop a Holochain app requires the use of the command line, or terminal, of your computer. 
+There are a set of custom-designed tools for working with Holochain that can be installed as utilities to your command line to simplify and accelerate the process of Holochain app development. These command line tools are required if you wish to be able to attempt what you are reading about as you continue through the articles on building an app.
 
-Fortunately, there are a set of custom designed tools for working with Holochain that can be installed as utilities to your command line to simplify and accelerate the process of Holochain app development.
+To install the command line tools, go to the repository at [github.com/holochain/holochain-rust/tree/develop/cmd](https://github.com/holochain/holochain-rust/tree/develop/cmd) and then follow the instructions in the README file.
```
Connoropolous: I wonder whether here we should point instead to the new installation instructions... they are a bit more involved, but at the same time, they are a bit more useful

PR #684, opened 2018-11-27
Added basic hc-run

sphinxc0re:
Introduces a command like:

`hc run -p 3400 -b`

where -p is for the port to run the websocket server on... (--port)
-b is short for build, or --package

This will become much more configurable in the future, this is the bare minimum configuration. 
Default port is 8888

Shows CLI feedback something like:
screen shot 2018-11-27 at 10 59 00 pm


Connoropolous:
I wonder whether we would want to pull out `hc agent` during this PR too, since it seems it does nothing?

@lucksus @sphinxc0re @zippy @thedavidmeister

sphinxc0re:
@Connoropolous I thought of `hc agent` as what is currently `container`. A "power-user-node".

cmd/src/cli/run.rs
```diff
@@ -0,0 +1,47 @@
...
+        file: package::DEFAULT_BUNDLE_FILE_NAME.into(),
+        hash: "Qm328wyq38924ybogus".into(),
+    };
+
+    let base_config = Configuration {
+        agents: vec![agent_config],
+        dnas: vec![dna_config],
+        ..Default::default()
+    };
+
```
lucksus: We need an `InstanceConfiguration` and an `InterfaceConfiguration`, no?

lucksus: Does this work without??

cmd/src/cli/run.rs
```diff
@@ -0,0 +1,47 @@
+use cli::{self, package};
+use error::DefaultResult;
+use holochain_container_api::{
+    config::{AgentConfiguration, Configuration, DNAConfiguration},
+    container::Container,
+};
+
+pub fn run(package: bool) -> DefaultResult<()> {
```
Connoropolous: Can you give us some /// rustdoc comments up here about this?

cmd/src/cli/run.rs
```diff
@@ -0,0 +1,47 @@
...
+
+    let mut container = Container::with_config(base_config);
+
+    container.start_all_interfaces();
+    container.start_all_instances()?;
+
+    let mut rl = rustyline::Editor::<()>::new();
+
+    loop {
+        let readline = rl.readline(">> ")?;
```
Connoropolous: What does this part do?

sphinxc0re: To not have an infinite loop making interaction inpossible, I added a simple readline. It displays a simple shell that you can easily `ctrl-c` out of.

cmd/src/cli/run.rs
sphinxc0re: sure

cmd/src/cli/run.rs
sphinxc0re: 🤷‍♂️ 
Give me some pointers. It's not really easy to create these configurations

cmd/src/main.rs
```diff
@@ -96,21 +89,34 @@ enum Cli {
         )]
         language: String,
     },
+    #[structopt(
+        name = "run",
+        alias = "r",
+        about = "starts a development container with an open websocket"
```
Connoropolous:
```suggestion
        about = "Starts a development container with an open websocket"
```

cmd/src/cli/run.rs
```diff
@@ -0,0 +1,67 @@
...
+        instances: vec![instance_config],
+        interfaces: vec![interface_config],
+        ..Default::default()
+    };
+
+    let mut container = Container::with_config(base_config);
+
+    container.start_all_interfaces();
+    container.start_all_instances()?;
+
```
Connoropolous: @sphinxc0re we need just a tiny bit of CLI feedback here...
tell the user the service is running

Connoropolous: And also tell the user that they can type `exit` to exit

cmd/src/cli/run.rs
lucksus: https://github.com/holochain/holochain-rust/blob/develop/container/example-config/basic.toml

cmd/src/main.rs
```diff
@@ -96,21 +89,40 @@ enum Cli {
...
+    Run {
+        #[structopt(
+            long,
+            short,
+            help = "The port to run the websocket server at"
+        )]
+        port: Option<u16>,
+        #[structopt(
+            long,
+            short = "b",
```
sphinxc0re: why `b`?

cmd/src/cli/run.rs
```diff
@@ -0,0 +1,77 @@
...
+        file: package::DEFAULT_BUNDLE_FILE_NAME.into(),
+        hash: "Qm328wyq38924ybogus".into(),
+    };
+
+    let instance_config = InstanceConfiguration {
+        id: "test-instance".into(),
+        dna: "hc-run-dna".into(),
+        agent: "hc-run-agent".into(),
+        logger: Default::default(),
+        storage: StorageConfiguration::Memory,
```
lucksus: I think we said it should be persistent. Then this should be set to file storage.

Connoropolous: Should it? Dunno, wasn't present to that conversation

PR #685, opened 2018-11-28
container fixes

zippy:
fixed container tests to always use temp dir.

### DIFF
--- a/cas_implementations/src/path.rs
+++ b/cas_implementations/src/path.rs
@@ -12,8 +12,9 @@ pub fn storage_path(path: &Path, folder_name: &str) -> HcResult<String> {
 pub fn create_path_if_not_exists(path: &str) -> HcResult<()> {
     if !Path::new(path).exists() {
         return DirBuilder::new()
+            .recursive(true)
             .create(path)
-            .map_err(|_| HolochainError::IoError("Could not create directory".to_string()));
+            .map_err(|e| HolochainError::IoError(format!("Error while attempting to create directory {}: {}",path,e)));
     }
     Ok(())
 }
@@ -39,11 +40,9 @@ pub mod tests {
         let bad_path = storage_path(Path::new("/foo"), "bar").unwrap();
         let result = create_path_if_not_exists(&bad_path);
         match result {
-            Ok(()) => unreachable!(),
-            Err(err) => assert_eq!(
-                err,
-                HolochainError::IoError("Could not create directory".to_string())
-            ),
+            Ok(()) => panic!("expected error"),
+            Err(HolochainError::IoError(_)) => (),
+            Err(_) => panic!("expected IoError"),
         };
         let dir = tempdir().unwrap();
         let file_path = dir.path();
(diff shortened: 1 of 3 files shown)


cas_implementations/src/path.rs
```diff
@@ -11,9 +11,12 @@ pub fn storage_path(path: &Path, folder_name: &str) -> HcResult<String> {
 
 pub fn create_path_if_not_exists(path: &str) -> HcResult<()> {
     if !Path::new(path).exists() {
-        return DirBuilder::new()
-            .create(path)
-            .map_err(|_| HolochainError::IoError("Could not create directory".to_string()));
+        return DirBuilder::new().recursive(true).create(path).map_err(|e| {
```
thedavidmeister: can we use `?` instead of `map_err` here?

container/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+Our recommended pattern for the installation of the container is to download the binary for your platform from our [releases](https://github.com/holochain/holochain-rust/releases) page. Otherwise, you can proceed with the more complex instructions for building from source, below.
+
+### Building From Source
+
+These dependencies need to be installed in order to compile, and use `holochain_container`:
+
+- [Rust](https://www.rust-lang.org/en-US/install.html)
+  - needs to be the `nightly` build, so use the following commands, once you have first installed Rust
+  - `rustup toolchain install nightly`
```
thedavidmeister: is an unpinned nightly safe?

container_api/src/container.rs
```diff
@@ -345,7 +356,9 @@ pub mod tests {
 
     #[test]
     fn test_container_load_config() {
-        let config = load_configuration::<Configuration>(test_toml()).unwrap();
+        let dir = tempdir().unwrap();
+        let path = dir.path().to_str().unwrap().to_string();
```
thedavidmeister: `dir.path().to_string_lossy()` avoids the need for an unwrap and coercion

container_api/src/container.rs
```diff
@@ -361,7 +374,9 @@ pub mod tests {
 
     #[test]
     fn test_container_try_from_configuration() {
-        let config = load_configuration::<Configuration>(test_toml()).unwrap();
+        let dir = tempdir().unwrap();
+        let path = dir.path().to_str().unwrap().to_string();
```
thedavidmeister: `to_string_lossy`

container_api/src/container.rs
```diff
@@ -376,7 +391,9 @@ pub mod tests {
 
     #[test]
     fn test_rpc_info_instances() {
-        let config = load_configuration::<Configuration>(test_toml()).unwrap();
+        let dir = tempdir().unwrap();
+        let path = dir.path().to_str().unwrap().to_string();
```
thedavidmeister: `to_string_lossy`

cas_implementations/src/path.rs
zippy: no because I want to create a much more informative error.  The ? passes back something with no contextual info.

container/README.md
zippy: probably not for ever, but this is also what's in the cmd README so I'm just going to flag this for followup. and I think it should work for now...

PR #686, opened 2018-11-28
Check currently installed version of Rust, offer to un/re-install (OLD: See #782)

pjkundert:
o Properly export PATH (in case this is the first installation of Rust)
o Use target-specific variables to select TOOLS/CORE_RUST_VERSION

If the wrong version of Rust is installed, we can check and offer to replace it with
one compatible with holochain-rust.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,8 @@ lint: fmt_check clippy
 # Check if Rust version is correct, and offer to uninstall mismatching version.  Requires
 # RUST_VERSION to be set (defaults to CORE_RUST_VERSION; see install_rustup..., below).  We'll
 # export PATH to default Rust installation here in the Makefile, in case this is the first time
-# rustup has been run, and we don't have a rustup-modified .profile loaded yet.
+# rustup has been run, and we don't have a rustup-modified .profile loaded yet.  If not a terminal
+# (stin is not a tty), defaults to not replace.
 export PATH := $(HOME)/.cargo/bin:$(PATH)
 RUST_VERSION = $(CORE_RUST_VERSION)
 .PHONY: version_rustup
@@ -47,8 +48,8 @@ version_rustup:
 	    echo "\033[0;93m## Current Rust version installed: ##\033[0m"; \
 	    if ! rustup show | grep "$(RUST_VERSION)"; then \
 	        rustup show; \
-		echo "\033[0;93m## Replace current Rust version (using rustup uninstall) with '$(RUST_VERSION)' ##\033[0m"; \
-	        read -p "Continue? (y/N) " yes; \
+		echo "\033[0;93m## Replace current Rust version (after rustup self uninstall) with '$(RUST_VERSION)' ##\033[0m"; \
+	        [ -t 0 ] && read -p "Continue? (y/N) " yes; \
 	        if [[ "$${yes:0:1}" == "y" ]] || [[ "$${yes:0:1}" == "Y" ]]; then \
 	            echo "\033[0;93m## Uninstalling Rust... ##\033[0m"; \
 	            rustup self uninstall || true; \
--- a/Makefile
+++ b/Makefile
@@ -36,24 +36,25 @@ C_BINDING_CLEAN = $(foreach dir,$(C_BINDING_DIRS),$(dir)Makefile $(dir).qmake.st
 # apply formatting / style guidelines
 lint: fmt_check clippy
 
-# Check if Rust version is correct, and offer to uninstall mismatching version.  Requires
-# RUST_VERSION to be set (defaults to CORE_RUST_VERSION; see install_rustup..., below).  We'll
+# Check if Rust version is correct, and offer to change to the correct version.  Requires
+# RUST_VERSION to be set (defaults to CORE_RUST_VERSION; see install_rustup..., below).  We'll also
 # export PATH to default Rust installation here in the Makefile, in case this is the first time
-# rustup has been run, and we don't have a rustup-modified .profile loaded yet.  If connected to a terminal
-# (stdin is a tty) and not running under a Continuous Integration test (CI), defaults to not replace.
+# rustup has been installed/run, and we don't have a rustup-modified .profile loaded yet.  If
+# connected to a terminal (stdin is a tty) and not running under a Continuous Integration test (CI),
+# defaults to not replace.
 export PATH := $(HOME)/.cargo/bin:$(PATH)
 RUST_VERSION = $(CORE_RUST_VERSION)
 .PHONY: version_rustup
 version_rustup:
 	@if which rustup; then \
 	    echo "\033[0;93m## Current Rust version installed (need: '$(RUST_VERSION)'): ##\033[0m"; \
-	    if ! rustup show | grep "$(RUST_VERSION)"; then \
+	    if ! rustup show 2>/dev/null | grep -qe "$(RUST_VERSION).*(default)"; then \
 	        rustup show; \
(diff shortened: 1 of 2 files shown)


zippy:
Hmm, I really like where this is going, but it has a problem that it breaks [travis CI](https://travis-ci.org/holochain/holochain-rust/jobs/460570199) because travis can't answer an interactive question when it runs.    To make this work there would have to be some other check to see if we are in travis and not ask the question.

thedavidmeister:
hey, could i get some more info on the problem we're tackling here?

using make commands should already use the correct toolchain

pjkundert:
This will require a lot more work; there are many, many targets in the Makefile that select and/or depend on various versions of Rust, using various methods.  We'll need to try to unify this Rust version  management, but this pull request doesn't (yet) do it correctly.

Ignore this pull for now...

zippy:
Thanks for your work on this.  It's been a tricky problem to manage because occasionally we need a new feature for the main build, but we don't want to change how `fmt` and `clippy` behave so we have kept those at a different build.  Right now I think they are set to the same thing.

Also if you are using nix, @thedavidmeister has much easier solutions than this make.

pjkundert:
Further work on this; I think its getting close.  The current Makefile Rust versioning is pretty broken, and this makes it a little clearer, I hope.  Specifically, I think it will now install and change the default Rust version for most things automatically, *but* it will allow a knowlegable user to avoid changing the Rust version (ie. when they are testing a new Rust toolchain) -- by entering "no" to the query.

pjkundert:
Only the lint target (fmt_check and clippy) now fail; everything else passes the Travis-CI tests.  Cleaned up some more of the component install version specification issues, so these should pass now, hopefully.

Makefile
```diff
@@ -35,36 +36,66 @@ C_BINDING_CLEAN = $(foreach dir,$(C_BINDING_DIRS),$(dir)Makefile $(dir).qmake.st
...
+version_rustup:
+	@if which rustup; then \
+	    echo "\033[0;93m## Current Rust version installed (need: '$(RUST_VERSION)'): ##\033[0m"; \
+	    if ! rustup show 2>/dev/null | grep -qe "$(RUST_VERSION).*(default)"; then \
+	        rustup show; \
+		echo "\033[0;93m## Change current Rust version to '$(RUST_VERSION)' ##\033[0m"; \
+	        [ -t 1 ] && [[ "$(CI)" == "" ]] && read -p "Continue? (Y/n) " yes; \
+	        if [[ "$${yes:0:1}" != "n" ]] && [[ "$${yes:0:1}" != "N" ]]; then \
+	            echo "\033[0;93m## Selecting Rust version '$(RUST_VERSION)'... ##\033[0m"; \
+	            rustup default $(RUST_VERSION); \
```
sphinxc0re: `rustup` has a override feature that should be used here instead of `rustup default`

sphinxc0re: For example:
```shell
rustup override set $(RUST_VERSION)
```
will set the the rustc and cargo version only for the current directory

pjkundert: Ah; so there are *two* places that need to be checked, in order, to determine the version of toolchain selected; rustup override list (to determine if the correct toolchain is already selected), and then rustup show (to see if A) if the toolchain is installed, and B) if it is the default).

Actually, it looks like it'll automatically install the correct toolchain if it doesn't exist, so only `rustup override ...` is required to test/set toolchain selection.  OK, I'll make the change.

thedavidmeister:
the makefile should be improved especially if it is broken somehow but i have a few comments:

- i don't think the makefile should try to interact with the OS global environment state at all
  - prompting to re/uninstall a global rust as per the OP/title says to me "HC does not play nice with other rust development on your machine" so i don't like that approach
  - the makefile was written assuming a green fields OS (e.g. CI/travis) and `shell.nix` was added to start tackling the much more complex problem of an active dev environment
- if we're making changes to the makefile, please make sure that `shell.nix` is updated accordingly

pjkundert:
You could be right; I was just trying to make the Makefile work, more or less as it useful, just less ... busted.  I'll be changing it to only do "local" Rust environment changes via `rustup override ...` instead of `rustup default ...`, so it'll only affect the local holochain-rust directory, as recommended by @sphinxc0re.

As much as we might ideally want to convert to other build management tools, I think the Makefile is still the primary method that most new devs will interact holochain-rust, at least initially, so it has to work, and (for now) things are tightly constrained to specific Rust versions, which were not being detected/handled by the Makefile, leading to failures...  If you have any other suggestions, I'm all ears!  Specifically, I don't know what targets in that Makefile might be obsolete, so I've retained them all, as much as I can test them.

thedavidmeister:
@pjkundert yeah the "local override" approach sounds better! :+1: 

i'm not anti-makefile at all, if it sounded like i was saying we should change build tools or stop using makefiles as a build/test tool, my bad :sweat_smile: 

taking a step back, perhaps the makefile should even fail _more_ obviously and do _less_ to the active rust environment? better debugging might be preferable to better "magic" :thinking: 

what i'm trying to say is that i don't feel comfortable with the makefile as an OS/development environment/package management tool (even though i personally wrote a lot of the code that makes it look that way :P), especially if we're now talking about supporting BYO environments that can have arbitrary existing config/tooling/cruft

for environment management there are better tools like nix, ansible, salt, etc. that can sit inside docker

we currently maintain multiple docker boxes to docker hub for OS management and they are intended to support local dev and testing on BYO environments, so using docker + make is an option RN that doesn't require nix and has been supported from day 1

however, docker is slow and resource intensive with daemons and VMs, etc. so if you don't like that solution it might be worth investigating whether nix could work for you ;)

if we were clearer/stricter about the difference between building vs. environment management then `shell.nix` could simply use the `Makefile` for most of what it does - the only reason i don't do that in nix already is _because_ the Makefile currently attempts to install cargo/rust and other deps automatically as part of running build/test tasks in ways that would either fail/error or conflict with the Rust mozilla overlays provided for nixos

it's probably better to talk about nix as a docker alternative rather than a Makefile alternative

sphinxc0re:
We could use the toolchain file: https://github.com/rust-lang/rustup.rs#the-toolchain-file

sphinxc0re:
We could use the toolchain file: https://github.com/rust-lang/rustup.rs#the-toolchain-file

sphinxc0re:
We could use the toolchain file: https://github.com/rust-lang/rustup.rs#the-toolchain-file

sphinxc0re:
We could use the toolchain file: https://github.com/rust-lang/rustup.rs#the-toolchain-file

sphinxc0re:
We could use the toolchain file: https://github.com/rust-lang/rustup.rs#the-toolchain-file or override precedence: https://github.com/rust-lang/rustup.rs#override-precedence

thedavidmeister:
@sphinxc0re nice, i didn't know about those, thanks for the links

i expect we still want clean and reliable separation between environment management vs doing builds given that we're not a 100% pure rust project and have external cross-platform dependencies to satisfy

checkout the list of environment libs etc. that monero cli is maintaining (the GUI has even more) for local builds:

https://github.com/monero-project/monero#compiling-monero-from-source

we're heading in that direction! e.g. the recent zmq pain and older Qt issues

thedavidmeister:
ugh... installing rustup will still break nix shell but i don't think this is the branch to fix that

pjkundert:
I think this pull should be ready merge into develop.

zippy:
Hey @pjkundert pjkundert, can you please start working on branches in the main repo instead of your fork (I did set you up as team member)?  Then we could more easily contribute changes to it, for example I just wanted go and update your branch so that CI would pass, but I can't!

pjkundert:
Archived; see #782

PR #702, opened 2018-11-29
Reenable code coverage

zippy:
I have just re-enabled code coverage, time to get things back above 85%!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -152,13 +152,13 @@ _cmd_deploy_template: &_cmd_deploy_template
 
 jobs:
   include:
-    - name: "Holochain tests [code coverage temporarily deactivated]"
+    - name: "Holochain tests & code coverage"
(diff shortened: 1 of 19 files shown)


zippy:
You're kidding!  To get code coverage passing we have to agree to a different formatting!?!?

lucksus:
Updated the nightly version. Don't know why, but that usually results in a different rustfmt output.. that is why we pinned it in the first place, right?

lucksus:
But that update to nightly, rebuilding tarpaulin for version 0.6.9 is what fixed it.
Updating futures to 0.3.0-alpha10 and changing the fmt are collateral.

zippy:
I'm ok us bringing the tools version to the same version as the core version right now, knowing that going forward we can still bump the core version WITHOUT bumping the tools version.

PR #706, opened 2018-11-29
rust scaffold fixes sufficient for 0.0.3

Connoropolous:
- [ ] add fuller list of dependencies
- [x] add holochain_core_types_derive

Have added actual sample zome code and test that works with zome.  This is good to have until we expand `hc generate` to be able to add entries and functions, at which point it should be removed, but until then, it's better than not having it.  One followup needed:

- [ ] make adding a zome also add a test for that zome instead of having it hard coded into the test file.


cmd/src/cli/js-tests-scaffold/index.js
```diff
@@ -1,21 +1,45 @@
 // This test file uses the tape testing framework.
 // To learn more, go here: https://github.com/substack/tape
 const test = require('tape');
-const Container = require('@holochain/holochain-nodejs');
+const { ConfigBuilder, Container } = require('@holochain/holochain-nodejs');
 
-// instantiate an app from the DNA JSON bundle
-const app = Container.instanceFromNameAndDna("app", "dist/bundle.json")
+// IIFE to keep config-only stuff out of test scope
```
Connoropolous: not sure what IIFE is, better not to have a comment in a scaffold file with some insider language

cmd/src/cli/js-tests-scaffold/index.js
```diff
@@ -1,21 +1,45 @@
...
+  const instanceAlice = ConfigBuilder.instance(agentAlice, dna)
+
+  return ConfigBuilder.container(instanceAlice)
+})()
+
+// Initialize the Container
+const container = new Container(config)
+container.start()
+
+// This function is a bit of temporary boilerplate to construct a convenient object
```
Connoropolous: I don't believe that we should use this in the scaffold file, better to take it out and use an unwrapped version of 'call'  further down in the code

cmd/src/cli/scaffold/rust/lib.rs
```diff
@@ -1,15 +1,75 @@
...
+    entry_definition::ValidatingEntryType,
+    error::ZomeApiResult,
+};
+use hdk::holochain_core_types::{
+    cas::content::Address, entry::Entry, dna::entry_types::Sharing, error::HolochainError, json::JsonString,
+};
 
 // see https://developer.holochain.org/api/0.0.2/hdk/ for info on using the hdk library
 
+// This is a sample zome that defines an entry type "MyEntry" that can be committed to the
```
Connoropolous: I don't believe in all the changes from here down, I don't think it'd be better to complete the easy 'entry type' generator thing if we want to make it easier to get boilerplate for this stuff and/or just use the link in the comment to give people a pointer precisely to a place where they can see code very similar to what's here.

core/src/network/actions/custom_send.rs
```diff
@@ -37,11 +32,11 @@ pub async fn custom_send(
     let action_wrapper = ActionWrapper::new(Action::SendDirectMessage(direct_message_data));
     dispatch_action(context.action_channel(), action_wrapper);
 
-    async {
+    /* async {
```
Connoropolous: what happens here? how can this code just be commented out?

zippy: I mentioned this in the PR text.  There's a followup to fix that.  Don't know how to do it yet.

cmd/src/cli/scaffold/rust/lib.rs
zippy: My goal is having `hc init` and `hc generate` produce something that actually works when you do `hc test`.  I'd rather have this kind of boilerplate with comments and instructions included on how to change it, rather than point people elsewhere, and have them try and figure out where to put all the parts.  Much more error prone and less likelyhood to get started quickly.

cmd/src/cli/js-tests-scaffold/index.js
zippy: k.  I'll fix that.

pdaoust: MDN suggests 'self-executing anonymous function' as an alternative to 'IIFE'. Or could just call it a closure, and people will notice that it's an immediately invoked one.

zippy:
hey @Connoropolous could you check this over once more.  I think it's ready to merge for 0.0.3

cmd/src/cli/scaffold/rust.rs
```diff
@@ -36,8 +37,11 @@ fn generate_cargo_toml(name: &str, contents: &str) -> DefaultResult<String> {
     let edition = maybe_package
         .and_then(|p| p.get("edition"))
         .unwrap_or(&edition_default);
+    let branch = maybe_package
```
Connoropolous: @zippy why would a branch be set in the package? does package mean cargo.toml?

zippy: My intent is that you can specify which branch of holochain you are are building the package against in your cargo toml.  The `maybe_package` was already there so I just used it, I think it actually should by `maybe_cargo`

zippy: Actually it the "package" section in the cargo file, and there's no branch in there.  So that's bogus.  I'll fix that.

cmd/src/cli/js-tests-scaffold/index.js
```diff
@@ -1,21 +1,38 @@
...
+
+// closure to keep config-only stuff out of test scope
+const container = (() => {
+    const agentAlice = Config.agent("alice")
+
+    const dna = Config.dna(dnaPath)
+
+    const instanceAlice = Config.instance(agentAlice, dna)
+
+    const containerConfig = Config.container(instanceAlice)
```
maackle: #827 introduces one more small change, where `Config.container` accepts a single array rather than multiple arguments, because in later iterations we'll likely need to pass config as well. I actually updated this file over in that branch too, we should probably use the one here so whoever merges this feel free to toss out the other one. (It may be me who deals with that conflict, so note to self)

```suggestion
    const containerConfig = Config.container([instanceAlice])
```

cmd/src/cli/js-tests-scaffold/index.js
zippy: fixed

core/src/network/actions/custom_send.rs
zippy: this got fixed

cmd/src/cli/js-tests-scaffold/index.js
zippy: updated

PR #708, opened 2018-11-30
Code-Coverage Party #2

zippy:
Goal:  get above new target of 75%  when we get there changes from WIP and we merge.

added --ignore-tests which drops the coverage substantially.  May have to change the goal!


thedavidmeister:
@zippy i dropped the codecov target to 75% with the idea that we will be having many more codecov parties

container_api/src/interface.rs
```diff
@@ -106,3 +108,45 @@ impl DispatchRpc for ContainerApiDispatcher {
...
+        let mut instances = InstanceMap::new();
+        instances.insert("test_instance".into(), Arc::new(RwLock::new(holochain)));
+        (config, instances)
+    }
+
+    #[test]
+    fn test_new_dispatcher() {
+        let (config, instances) = example_config_and_instances();
+        let dispatcher = ContainerApiDispatcher::new(&config, instances.clone());
+        assert!(match dispatcher.instances.get("test_instance") {
```
sphinxc0re: Please use `is_some()` here. No need for reinventing the wheel.

```rust
assert!(dispatcher.instances.get("test_instance").is_some());
```

PR #713, opened 2018-12-02
2018 12 02 circle nix

thedavidmeister:
## changes

takes significant load off travis

this splits our tests into two sets:

- things that run nicely on nix to date
  - fmt (1 min on travis + sudo 2-3 mins = 3 mins)
  - codecov (23 min on travis + sudo 2-3 mins = 25 mins)
  - app spec tests (27 min on travis + sudo 2-3 mins = 30 mins)
- things that i haven't got working on nix to date

using most recent develop build on travis for reference: https://travis-ci.org/holochain/holochain-rust/builds/467005369

**60 mins total time is benchmark for these 3 tasks on travis**

image

**can see that we've shaved ~50 minutes off the total build time**

wall time is still 45 mins because windows, but this unlocks more parallel builds

image

runing in sequence on circle takes 20 minutes uncached

https://circleci.com/gh/holochain/holochain-rust/48#usage-queue/containers/0

**running in sequence on circle takes 9:51 minutes cached for app spec + codecov + fmt**

https://circleci.com/gh/holochain/holochain-rust/49

image

the former moves to circle because:

- we can run circle in parallel to travis if we want
  - circle is 4 free builds for FOSS
  - $50 p.m. for extra paid containers
  - mac and linux but i don't think windows

- circle has much better debugging tools
  - can ssh into a build to "look around" and try things out
    - don't have to wait for 1-2 hours for a round trip to try to run `ls` :roll_eyes: 

rebuilding a bad build with ssh mode on:

image

poking around to see the cruft breaking this build's checkout:

image

  - can use the circle cli tool to run circle builds locally (travis cli only allows queuing remote builds)
    - can keep working even if there is a build queue remotely

running circle _locally with a local docker build_:

image

  - travis kills builds with log output that is too long!
    - it got to the point where i was pastebinning console output to avoid blowing up logs :(
    - the logs i needed were the verbose compiler info that would tell me why fingerprints fail
    - i can't turn off the logs and back on because the fingerprints (cache) only update on successful travis builds, also that would literally take 3+ hours to round trip through the queue once
- circle is faster
  - setup is faster, 39s to pull the image down vs. the hidden 2-3 minutes sudo adds to travis + 25s
  - can actually configure circle to have more cores/ram on paid account 

configurable build resources, goes up to 8 cores:

image

https://circleci.com/docs/2.0/configuration-reference/#resource_class
  - caching is easier to configure
    - e.g. can invalidate cache based on a checksum of cargo.toml(s) rather than scanning file times
    - it actually works! compilation resumes correctly on circle
    - caches are _immutable_ per key so are either there or not, they'll never silently rebuild out from 

getting straight to the point without recompiling cached deps needlessly:

image

underneath you and add many minutes and weight to the start and end of builds
- nix works better on circle
  - there's a docker box that works nicely, it has been fixed since i tried it last nixorg/nix:circleci
  - the docker box is setup to encourage pinned nixpkgs versions for more reliable builds
  - nix itself works on travis, the issue is really that logs/caches/debug keep blowing up builds so they shut themselves down or enqueue for hours
- circle has better workflows
  - can do fan out and in, etc. not just parallel builds - https://circleci.com/docs/2.0/workflows/
  - has browser notifications, leave a circle tab open and get notified pass/fail when it finishes

the latter remains running on travis

also adds a docker box running alpine + nix that warms builds for our commands (alpine is much smaller than ubuntu!)

`holochain/holochain-rust:circle`

adds more targets to the .dockerignore file to shrink build context from 2GB+ down to 100MB-

renormalizes tools version to core version in make and run a fmt to bump inline with nix-shell

commented out some flaky tests - they _were_ skipped on windows but also seem to hang on nix :(

## followups

- get more things working in nix (notably cmd tests)
- should windows be on something else that better supports rust compilation cache? e.g. appveyor?
- the cache logic can be further refined, each target can checksum its own `Cargo.toml` giving very fine grained recompliation caching
- `neon info forcing rebuild for new build settings` whatever this is about, slowing down tests :rage: ???
- split the 3 steps into concurrent builds
- swap out our ubuntu docker for alpine everywhere?
- really want an up to date "clean" command to flush all these heavy targets that get compiled and slow build tooling down massively (multiple GBs atm)
- fix flaky tests


thedavidmeister:
i agree disabling tests sucks

my plan is not to stop here at all, i'm going to keep chugging along speeding, standardising and fixing

PR #718, opened 2018-12-03
Network enabled get_entry

lucksus:
## Local vs. Network
We now have two `get_entry` functions:
* `nucleus::actions::get_entry` which only tries to find the entry locally
* `network::actions::get_entry` which only tries to get the entry via a network request

I've added a workflow that uses both.

## Flow of `network::actions::get_entry`
1. `network::actions::get_entry::get_entry` -> `Action::GetEntry(entry)`
2. `network::reduce_get_entry` creates a `None` entry in `NetworkState::get_entry_results` and sends a `ProtocolWrapper::GetDht(GetDhtData)`
3. In `handler.rs`, the `ProtocolWrapper::GetDht(get_dht_data)` branch calls `nucleus::actions::get_entry` (the local version) and creates a `Action::RespondGet(GetDhtData, Option<Entry>)` with the result
4. `network::reduce_respond_get` takes that potential entry, serializes it and sends a `ProtocolWrapper::GetDhtResult(DhtData)`
5. Handler for `ProtocolWrapper::GetDhtResult` just creates a `Action::HandleGetResult(DhtData)`
6. `network::reduce_handle_get_result` writes the potential entry into `network_state.get_entry_results`
7. Future returned by `network::actions::get_entry::get_entry` resolves


zippy:
Are you expecting main branch to fail tests ?
```
test network::tests::get_non_existant_entry ... FAILED
test network::tests::get_when_alone ... FAILED
```

zippy:
Hmmm @lucksus, to start I just ran the tests locally on my machine, using `make test_holochain` and I get a lockup here: 
```
test can_get_entry ... test can_get_entry has been running for over 60 seconds
```

lucksus:
Ok, fixed that. This is happening now because many test instances that run simultaneously are in the same virtual network because they all use the same vanilla test DNA. I've set the test name now as UUID where relevant. Fixed a bunch failing tests and running full test suite local now..

zippy:
And now only the windows build fails!

core/src/network/actions/mod.rs
```diff
@@ -6,4 +7,5 @@ use holochain_core_types::{cas::content::Address, error::HcResult};
 #[derive(Clone, Debug)]
 pub enum ActionResponse {
     Publish(HcResult<Address>),
+    RespondeGet(HcResult<()>),
```
ddd-mtl: typo? Respond**e**Get

lucksus: *<italian accent>* "RespondGet"

core/src/network/reducers/respond_get.rs
```diff
@@ -0,0 +1,61 @@
...
+    (network_state.network.is_some()
+        && network_state.dna_hash.is_some() & network_state.agent_id.is_some())
+    .ok_or("Network not initialized".to_string())?;
+
+    let data = DhtData {
+        msg_id: get_dht_data.msg_id.clone(),
+        dna_hash: network_state.dna_hash.clone().unwrap(),
+        agent_id: get_dht_data.from_agent_id.clone(),
+        address: get_dht_data.address.clone(),
+        content: serde_json::from_str(&serde_json::to_string(&maybe_entry).unwrap()).unwrap(),
```
ddd-mtl: isn't serde_json deprecated? Shouldn't you use JsonString instead?
Also, I'm not sure I understand the usefulness of serializing an option that could be None.

lucksus: The `DhtData` struct is defined with serde, `content` is of type `serde_json::Value`. Yeah, @thedavidmeister brought this up before. This is the network-module API as @neonphog built it for us. We should change this to JsonString but I didn't want to put that into the scope of this PR. It is still just a few lines like this.

The option is for being able to respond with "I don't have the requested entry".

PR #721, opened 2018-12-04
Real sources in ChainHeader

lucksus:
# Context
* Implents this SoA: https://realtimeboard.com/app/board/o9J_kyiXmFs=/?moveToWidget=3074457346318844110
* Needed for validation in a network context (we need to know who to ask for the validation package)

# Changes
* added `sources: Vec<Address>` to `ChainHeader`
* changed `fn create_new_chain_header()` to get the real agent_id from the chain or context and put its address in the header
* as a consequence had to fix tests that manually call `instance.process_action` or reducers without initializing the context (they do now)


core/src/agent/state.rs
```diff
@@ -144,10 +147,18 @@ pub enum ActionResponse {
...
+        .state()
+        .expect("create_new_chain_header called without state")
+        .agent();
+    let agent_address = agent_state
+        .get_agent_address()
+        .unwrap_or(context.agent_id.address());
     ChainHeader::new(
         &entry.entry_type(),
         &entry.address(),
+        &vec![agent_address],
```
lucksus: This is the most important line in this PR ;)

lucksus: as in: everything else needed to happen as a consequence to get this line in :D

core_types/src/chain_header.rs
```diff
@@ -24,6 +25,9 @@ pub struct ChainHeader {
     entry_type: EntryType,
     /// Key to the entry of this header
     entry_address: Address,
+    /// Address(es) of the agent(s) that authored and signed this entry.
+    /// Backed by the entry_signature below.
+    sources: Vec<Address>,
     /// agent's cryptographic signature of the entry
```
zippy: So, just checking, the reason to put the source in the header itself is to not have to store it as meta-data when published to the DHT, and, in the future use cases of multiple signers of an entry, because in the local chain where there is no meta-data, you would have to do something else (like put the address of the signers in the entry itself).  But otherwise, we could be getting the address from the meta data (if on the dht), or by looking for the most recent agent entry on the chain.  Am I missing something?

core_types/src/chain_header.rs
```diff
@@ -24,6 +25,9 @@ pub struct ChainHeader {
...
     entry_signature: Signature,
```
zippy: if sources are going to be a Vec, then we should changes this to `entry_signatures` to match right now, otherwise this is too confusing.

core/src/agent/state.rs
thedavidmeister: image

core_types/src/chain_header.rs
thedavidmeister: right, so the idea is that the nth signature was produced by the nth address/key?

core/src/agent/state.rs
lucksus: yup ;)

lucksus:
What is so complicated with 
```rust
pub fn create_new_chain_header(entry: &Entry, context: Arc<Context>) -> ChainHeader
```
@sphinxc0re?

lucksus:
@zippy I've turned the signatures into a vector as well. If you approve, please go ahead and hit merge.

core_types/src/chain_header.rs
sphinxc0re: Are multiple Agents able to create a single entry together?

lucksus: Yes, that's what is supposed to happen in the Holo fuel transactor for example: the two parties exchange the transaction (pre-flight) entry with node-to-node messages and both sign it and both commit it to their source chain.

sphinxc0re:
@lucksus I'm talking about `ChainHeader::new`. It has so many parameters and is used in many tests.

lucksus:
Yeah, that's true. I didn't start questioning if all of that needs to happen that way, just left it at: I understand that tests need to use the internals for testing. As long as there this this high-level function that is used in non-test code, it's kinda fine. But yeah, would be great if we could get our tests less verbose as well.

sphinxc0re:
Follow-up it is then

core_types/src/chain_header.rs
pdaoust: @lucksus no problem. Happy to provide context. And even if HF isn't using it anymore, I figured it would be idiomatic enough that it still needed to show up in core.

PR #726, opened 2018-12-05
moved network config to InstanceConfiguration

zippy:
currently `container_api::instantiate_from_config()` hard codes the network configuration.  This PR moves the config into InstanceConfiguration.  This is necessary to connect a real network to instances, and also looks forward to containers that could connect different DNA to different networks.

### DIFF
--- a/container/example-config/basic.toml
+++ b/container/example-config/basic.toml
@@ -12,7 +12,7 @@ hash = "Qm328wyq38924y"
 id = "app spec instance"
 dna = "app spec rust"
 agent = "test agent"
-network = "{\"backend\": \"mock\"}"
+network = ""
 [instances.logger]
 type = "simple"
 file = "app_spec.log"
--- a/container/example-config/basic.toml
+++ b/container/example-config/basic.toml
@@ -12,7 +12,6 @@ hash = "Qm328wyq38924y"
 id = "app spec instance"
 dna = "app spec rust"
 agent = "test agent"
-network = ""
 [instances.logger]
 type = "simple"
 file = "app_spec.log"
--- a/container/example-config/basic.toml
+++ b/container/example-config/basic.toml
@@ -12,6 +12,7 @@ hash = "Qm328wyq38924y"
 id = "app spec instance"
 dna = "app spec rust"
 agent = "test agent"
+network = "{\"backend\": \"mock\"}"
 [instances.logger]
 type = "simple"
 file = "app_spec.log"
--- a/container_api/src/config.rs
+++ b/container_api/src/config.rs
@@ -139,7 +139,7 @@ impl TryFrom<DNAConfiguration> for Dna {
 }
 
 /// An instance combines a DNA with an agent.
-/// Each instance has its own storage and logger configuration.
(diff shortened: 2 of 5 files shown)


container_api/src/config.rs
```diff
@@ -139,14 +139,15 @@ impl TryFrom<DNAConfiguration> for Dna {
...
-/// Each instance has its own storage and logger configuration.
+/// Each instance has its own network, storage and logger configuration.
 #[derive(Deserialize, Serialize, Clone)]
 pub struct InstanceConfiguration {
     pub id: String,
     pub dna: String,
     pub agent: String,
     pub logger: LoggerConfiguration,
     pub storage: StorageConfiguration,
+    pub network: String,
```
lucksus: I would make it optional (`Option<String>`) because I don't think this should be in the config files as the container user shouldn't (have to) even know that we have separated out our network module from core (which is the main reason for having this config).

container/example-config/basic.toml
```diff
@@ -12,6 +12,7 @@ hash = "Qm328wyq38924y"
 id = "app spec instance"
 dna = "app spec rust"
 agent = "test agent"
+network = "{\"backend\": \"mock\"}"
```
lucksus: Yeah, this shouldn't be here.

container_api/src/config.rs
lucksus:
```suggestion
    pub network: Option<String>,
```

lucksus: This change would make several other changes redundant.

lucksus: With this we can have a config.toml be valid even without mentioning network at all. Not even with `network=""`, which is not what I meant.

container/example-config/basic.toml
```diff
@@ -12,6 +12,7 @@ hash = "Qm328wyq38924y"
 id = "app spec instance"
 dna = "app spec rust"
 agent = "test agent"
+network = ""
```
lucksus: I want to get rid of this line completely here.

container_api/src/container.rs
```diff
@@ -235,15 +237,22 @@ fn instantiate_from_config(
                 ))
             })?;
 
-            let network = json!({"backend": "mock"}).into();
+            let network_config = if instance_config.network == "" {
```
lucksus:
```suggestion
            let network_config = instance_config.network.unwrap_or(DEFAULT_NETWORK_CONFIG.to_string());
```

container_api/src/container.rs
```diff
@@ -219,6 +220,7 @@ fn instantiate_from_config(
     id: &String,
     config: &Configuration,
     dna_loader: &mut DnaLoader,
+    default_network_config: &String,
```
lucksus:
```suggestion
```

container_api/src/config.rs
zippy: just pushed Option<String> fix

container/example-config/basic.toml
zippy: fixed

PR #727, opened 2018-12-05
Get ValidationPackage over network

lucksus:
# Context
Implements this state: https://realtimeboard.com/app/board/o9J_kyiXmFs=/?moveToWidget=3074457346314392714

screenshot 2018-12-07 at 17 36 48

# Implementation steps in this PR
- [x] Protocol for node2node messages `DirectMessage` with variants `RequestValidationPackage` and `ValidationPackage`
- [x] `Action::GetValidationPackage` with reducer and action creator that lets the network send a `DirectMessage::RequestValidationPackage` to the entries source
- [x] handler for the request that triggers the new workflow:
- [x] `workflows::respond_validation_package_request` that checks if we have the entry in the source chain, builds the validation package and then triggers sending of the response via:
- [x] `Action::SendDirectMessage` that makes the network send any (initial or responding) node2node message
- [x] handler for the response of a validation package request `DirectMessage::ValidationPackage` that updates the state so the `GetValidationPackageFuture` can resolve

# Additional required changes in here
* Had to make sure that `Entry::address()` returns the right address also for agent_id entries which override the hashing of the content and instead have their public key be the address (https://github.com/holochain/holochain-rust/pull/727/commits/6f494af8967b5b55cbe32b4a581b28d7280e8d2b and https://github.com/holochain/holochain-rust/pull/727/commits/b66068a7129db1e4013d50302f88dd3c937407ac)
* `ProtocolWrapper::SendResult` and `ProtocolWrapper::HandleSendResult` were mixed up (right, @neonphog? Or did I get it wrong?) I've swapped them to have *handle* always mean that the network requests core to handle something. (https://github.com/holochain/holochain-rust/pull/727/commits/76fa8420a9ea43014891b284ea4ece3a823f27ad and https://github.com/holochain/holochain-rust/pull/727/commits/e812ae798223730d94c709fe83bdb8b1f59db04f)


Connoropolous:
You might already be thinking about/ working on this (like last time) , but I'd love to see /// comments on any structs and functions

net/src/mock_worker.rs
```diff
@@ -389,7 +389,7 @@ mod tests {
 
         if let ProtocolWrapper::HandleSend(msg) = res {
             cli2.receive(
-                ProtocolWrapper::HandleSendResult(MessageData {
+                ProtocolWrapper::SendResult(MessageData {
```
neonphog: @lucksus 
> ProtocolWrapper::SendResult and ProtocolWrapper::HandleSendResult were mixed up (right, @neonphog? Or did I get it wrong?) I've swapped them to have handle always mean that the network requests core to handle something

I was thinking of this the other way. When I wrote it, I was thinking the Sender would send the `SendMessage` and receive the `SendResult` and the remote node (the Handler) would receive the `HandleSend` and send the `HandleSendResult`.

Either way the names are a bit confusing... so if you have other suggestions, feel free to adjust.

core_types/src/entry/mod.rs
```diff
@@ -98,6 +99,13 @@ impl PartialEq for Entry {
 }
 
 impl AddressableContent for Entry {
+    fn address(&self) -> Address {
+        match &self {
+            Entry::AgentId(agent_id) => agent_id.address(),
+            _ => Address::encode_from_str(&String::from(self.content()), Hash::SHA2256),
+        }
+    }
```
neonphog: :+1:

core/src/network/handler/send.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    match message {
+        DirectMessage::Custom(_) => log(&context, "DirectMessage::Custom not implemented"),
+        DirectMessage::RequestValidationPackage(address) => {
+            // Async functions only get executed when they are polled.
+            // I don't want to wait for this workflow to finish here as it would block the
+            // network thread, so I use block_on to poll the async function but do that in
+            // another thread:
+            thread::spawn(move || {
+                block_on(respond_validation_package_request(
```
neonphog: Agree, we don't want to block the network thread, but we also want to be careful not to duplicate the pattern of spawning hard os threads too much. Isn't that the point of thread pool executors in futures?

core/src/action.rs
```diff
@@ -72,23 +72,74 @@ impl Hash for ActionWrapper {
...
-    ReturnZomeFunctionResult(ExecuteZomeFnResponse),
+    /// We got a response for our GET request which needs to be
+    /// added to the state.
+    /// Triggered from the network handler.
+    HandleGetResult(DhtData),
 
+    /// Sends a direct message object to the given address.
+    /// 3rd parameter is the message id
+    /// 4th parameter is true for a response to a previous message, false for a new interaction
+    SendDirectMessage((Address, DirectMessage, String, bool)),
```
neonphog: It may be time ditch the tuple in favor of a sub-struct so the parameters can be named.

```rust
pub struct DirecmMessageData {
    address: Address,
    message: DirectMessage,
    msg_id: String,
    continue: bool,
}

pub enum Action {
    SendDirectMessage(DirectMessageData)
```

net/src/mock_worker.rs
lucksus: Ah, ok. Sorry. Makes sense. I think I will switch back then.

core/src/network/handler/send.rs
lucksus: Well yeah, we could (and maybe should soon) implement our own executor for that. With the old futures I could have just called that function and not wait for the future result. But with async/await they decided for not executing the async code at all until somebody starts polling.. Would leave it like this for now and add a custom global future executor soon.

core/src/network/actions/get_validation_package.rs
```diff
@@ -0,0 +1,70 @@
...
+}
+
+impl Unpin for GetValidationPackageFuture {}
+
+impl Future for GetValidationPackageFuture {
+    type Output = HcResult<Option<ValidationPackage>>;
+
+    fn poll(self: Pin<&mut Self>, lw: &LocalWaker) -> Poll<Self::Output> {
+        let state = self.context.state().unwrap().network();
+        if state.network.is_none() || state.dna_hash.is_none() || state.agent_id.is_none() {
```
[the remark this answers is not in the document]
lucksus: There actually is one. Will use it here as well. ;)

core/src/network/handler/store.rs
```diff
@@ -0,0 +1,31 @@
...
+    match dht_meta_data.attribute.as_ref() {
+        "link" => {
+            let link_add = match entry_with_header.entry {
+                Entry::LinkAdd(link_add) => link_add,
+                _ => unreachable!(),
+            };
+            let link = link_add.link().clone();
+            let _ = block_on(add_link(&link, &context.clone()));
+        }
+        _ => {}
```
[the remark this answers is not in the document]
lucksus: There will be more attributes soon with CRUD. This is already a hint towards that actually..

core/src/network/reducers/get_validation_package.rs
```diff
@@ -0,0 +1,42 @@
...
+use holochain_core_types::{chain_header::ChainHeader, error::HolochainError};
+use std::sync::Arc;
+
+fn inner(network_state: &mut NetworkState, header: &ChainHeader) -> Result<(), HolochainError> {
+    initialized(network_state)?;
+
+    let source_address = header
+        .sources()
+        .first()
+        .expect("A header must have at least one source");
```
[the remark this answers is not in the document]
lucksus: 👍

core/src/network/handler/send.rs
neonphog: :+1:

core/src/network/direct_message.rs
```diff
@@ -0,0 +1,21 @@
+use holochain_core_types::{cas::content::Address, validation::ValidationPackage};
+
+/// These are the different kind of node-to-node messages
```
zippy: I think we should add "low-level system" so as not to confuse the App level node-to-node messaging.

core/src/network/reducers/publish.rs
```diff
@@ -23,25 +19,17 @@ fn publish_entry(
...
-        })
-        .expect("Network has to be Some because of check above")
+    send(
+        network_state,
+        ProtocolWrapper::PublishDht(DhtData {
+            msg_id: "?".to_string(),
+            dna_hash: network_state.dna_hash.clone().unwrap(),
+            agent_id: network_state.agent_id.clone().unwrap(),
+            address: entry.address().to_string(),
+            content: serde_json::from_str(&serde_json::to_string(&entry_with_header).unwrap())
```
zippy: can't we use JsonString here?

core/src/network/reducers/publish.rs
```diff
@@ -61,37 +49,26 @@ fn publish_link(
...
-        .expect("Network has to be Some because of check above")
+    send(
+        network_state,
+        ProtocolWrapper::PublishDhtMeta(DhtMetaData {
+            msg_id: "?".to_string(),
+            dna_hash: network_state.dna_hash.clone().unwrap(),
+            agent_id: network_state.agent_id.clone().unwrap(),
+            address: link.base().to_string(),
+            attribute: String::from("link"),
+            content: serde_json::from_str(&serde_json::to_string(&entry_with_header).unwrap())
```
zippy: ditto JsonString

core/src/network/reducers/respond_get.rs
```diff
@@ -1,44 +1,29 @@
...
-        })
-        .expect("Network has to be Some because of check above")
+    send(
+        network_state,
+        ProtocolWrapper::GetDhtResult(DhtData {
+            msg_id: get_dht_data.msg_id.clone(),
+            dna_hash: network_state.dna_hash.clone().unwrap(),
+            agent_id: get_dht_data.from_agent_id.clone(),
+            address: get_dht_data.address.clone(),
+            content: serde_json::from_str(&serde_json::to_string(&maybe_entry).unwrap()).unwrap(),
```
zippy: ditto JsonString

core/src/network/reducers/send_direct_message.rs
```diff
@@ -0,0 +1,44 @@
...
+
+pub fn reduce_send_direct_message(
+    _context: Arc<Context>,
+    network_state: &mut NetworkState,
+    action_wrapper: &ActionWrapper,
+) {
+    let action = action_wrapper.action();
+    let dm_data = unwrap_to!(action => crate::action::Action::SendDirectMessage);
+    if let Err(error) = inner(network_state, dm_data) {
+        println!("Error sending direct message: {:?}", error);
```
zippy: println!?  at least this should be log, no?

core/src/network/state.rs
```diff
@@ -8,8 +14,22 @@ use std::{
 };
 
 type Actions = HashMap<ActionWrapper, ActionResponse>;
+
+/// This represents the state of a get_entry network process:
+/// None: process started, but no response yet from the network
+/// Some(Err(_)): there was a problem at some point
+/// Some(Ok(None)): no problem but also no entry -> it does not exist
+/// Some(Ok(Some(entry))): we have it
 type GetEntryResult = Option<Result<Option<Entry>, HolochainError>>;
```
zippy: I like this enum for state!

core/src/workflows/respond_validation_package_request.rs
```diff
@@ -0,0 +1,47 @@
...
+
+pub async fn respond_validation_package_request(
+    to_agent_id: Address,
+    msg_id: String,
+    requested_entry_address: Address,
+    context: Arc<Context>,
+) {
+    let maybe_validation_package = match get_entry(&requested_entry_address, &context) {
+        Ok(entry) => await!(build_validation_package(&entry, &context)).ok(),
+        Err(_) => None,
```
zippy: Do we really send a None value back across the network if there was an error?  I have the feeling that we at least have to tell the requester that we had some kind of error...  followup?

core/src/action.rs
```diff
@@ -77,9 +78,38 @@ impl Hash for ActionWrapper {
...
+    /// Does not validate, assumes link is valid.
+    AddLink(Link),
+
+    // ----------------
+    // Network actions:
+    // ----------------
+    /// Create a network proxy instance from the given JSON config.
+    /// 2nd and 3rd parameter are the DNA hash and the agent id
+    /// which are needed to register with the network.
+    InitNetwork((JsonString, String, String)),
```
thedavidmeister: followup to use real types for all params?

core/src/action.rs
```diff
@@ -88,16 +118,42 @@ pub enum Action {
...
-    ExecuteZomeFunction(ZomeFnCall),
-    /// return the result of a zome WASM function call
-    ReturnZomeFunctionResult(ExecuteZomeFnResponse),
+    /// We got a response for our GET request which needs to be
+    /// added to the state.
+    /// Triggered from the network handler.
+    HandleGetResult(DhtData),
+
+    /// Sends a direct message object to the given address.
+    /// 3rd parameter is the message id
```
thedavidmeister: parameters?

core_types/src/entry/mod.rs
```diff
@@ -100,6 +101,13 @@ impl PartialEq for Entry {
 }
 
 impl AddressableContent for Entry {
+    fn address(&self) -> Address {
+        match &self {
+            Entry::AgentId(agent_id) => agent_id.address(),
+            _ => Address::encode_from_str(&String::from(self.content()), Hash::SHA2256),
```
thedavidmeister: hmm, this seems like it should always defer to the `.address()` method of the inner value?

lucksus:
@thedavidmeister raw serde handling because ˋProtocolWrapperˋ‘s data structs use ˋserde_json::Valueˋ. You pointed to that earlier and I agree that should be changed. But in this PR I didn’t want to broaden the scope to include those changes to the net crates and instead just use them.

core/src/network/reducers/publish.rs
lucksus: We can and we should. But I didn’t want to change the net crates in this PR and instead just use it. Follow up.

core_types/src/entry/mod.rs
lucksus: Are you sure? Wouldn't that mean that the entry type is not included in the hash? In Montreal we agreed that the "content" that gets hashed needs to include the type.

PR #728, opened 2018-12-06
2018 12 06 json string docs

thedavidmeister:
## changes

- mdbook docs for various `JsonString` tips, tricks, considerations and tradeoffs


doc/holochain_101/src/json_string.md
```diff
@@ -0,0 +1,823 @@
...
+
+- https://github.com/appcypher/awesome-wasm-langs
+- https://github.com/mbasso/awesome-wasm
+
+The second requirement means that we must be very mindful of choosing a
+serialization format that can round trip through as many languages as possible.
+
+In the end, this is the main reason we chose JSON for communication with core.
+
+Note that when we started on an AssemblyScript (ostentisbly JavaScript) HDK
```
Connoropolous: @thedavidmeister there still isn't ... we never got what we worked on merged upstream!

Connoropolous:
Overall, amazing to have this, epic write up David!

doc/holochain_101/src/json_string.md
```diff
@@ -0,0 +1,823 @@
...
+  - the `Serialize` implementation of `foo` decides this for us
+  - in the case of nested data e.g. hash maps, `Serialize` works recursively
+
+OR using the manual `json!` macro:
+
+```rust
+let foo_json = json!({"foo": foo.inner()});
+```
+
+Notes:
```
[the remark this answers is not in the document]
thedavidmeister: i just meant it is slightly less typing, i'll update

doc/holochain_101/src/json_string.md
```diff
@@ -0,0 +1,888 @@
...
+let restored_bar: MyBar = foo.bar()?;
+```
+
+This is how the `ContentAddressableStorage` trait used to work. It would
+"magically" restore the correct `Content` from storage based on an `Address`
+and type alone, provided the compiler had the type info available at compile
+time.
+
+We had to sacrifice this neat trick due to incompatible constraints put on the
+CAS by the reducer store etc. but it should work well in other scenarios :)
```
lucksus: „Reducer store?“
I think we had to choose to not use rust‘s type system for this so we can have dynamically linked CAS implementations.

thedavidmeister: @lucksus yeah, how would you say that?

PR #729, opened 2018-12-06
Switched to persistent storage

sphinxc0re:


lucksus:
I don't see how these changes change persistence. Is this still WIP or did you forget to push more commits?

sphinxc0re:
I changed the storage to file system. See: https://github.com/holochain/holochain-rust/pull/729/files#diff-baff0a5332cbd7e8bcb76c945c13fd4eR38

sphinxc0re:
Okay, then I'll think of something

sphinxc0re:
Okay, then I'll think of something

cmd/src/cli/run.rs
```diff
@@ -1,39 +1,58 @@
...
+const LOCAL_STORAGE_PATH: &str = ".hc";
+
+const AGENT_CONFIG_ID: &str = "hc-run-agent";
+const DNA_CONFIG_ID: &str = "hc-run-dna";
+const INSTANCE_CONFIG_ID: &str = "test-instance";
+const INTERFACE_CONFIG_ID: &str = "websocket-interface";
 
 /// Starts a small container with the current application running
-pub fn run(package: bool, port: u16) -> DefaultResult<()> {
+pub fn run(package: bool, port: u16, persist: bool) -> DefaultResult<()> {
```
thedavidmeister: persist is not a boolean though is it? in the not-so-distant-future we will have sqlite as an option too right?

zippy: Fur the purposes of dev and `hc run` I think on option is fine for now...

thedavidmeister: @zippy would it be OK to swap it with an enum that implements `From<String>` so we can extend it easily as we go? 

now is the easiest moment to make that refactor i think

sphinxc0re: Good idea but this seems like a follow-up though

cmd/README.md
```diff
@@ -187,6 +187,10 @@ where `test/test.js` is the path of the file.
 
 You have the flexibility to write tests in quite a variety of ways, open to you to explore.
 
+## Running your application
+For the purpose of *testing* APIs or prototyping user interfaces, you can run your DNA using `hc run`.
+This will start the application and open a WebSocket on port `8888` or any other port specified by the `-p`/`--port` option. The `-b`/`--package` flag will build your DNA before running it. By default, none of the data your application is putting on the SourceChain. To persist data onto the file system, use the `--persist` flag.
```
lucksus:
```suggestion
This will start the application and open a WebSocket on port `8888` or any other port specified by the `-p`/`--port` option. The `-b`/`--package` flag will build your DNA before running it. By default, none of the data your application is putting on the source chain gets persisted. To persist data onto the file system, use the `--persist` flag.
```

thedavidmeister:
followup it is then

persistence config should be an enum not true/false #747

PR #730, opened 2018-12-06
[WIP] Initialize Network on load

AshantiMutinta:
this pull requests initializes the network on load as well as isolates the extra work in it's own workflow module.

### DIFF
--- a/container_api/src/holochain.rs
+++ b/container_api/src/holochain.rs
@@ -65,10 +65,10 @@ use futures::{executor::block_on, TryFutureExt};
 use holochain_core::{
     context::Context,
     instance::Instance,
-    network::actions::initialize_network::initialize_network,
     nucleus::{actions::initialize::initialize_application, call_and_wait_for_result, ZomeFnCall},
     persister::{Persister, SimplePersister},
     state::State,
+    workflows::network
 };
 use holochain_core_types::{dna::Dna, error::HolochainError, json::JsonString};
 use std::sync::Arc;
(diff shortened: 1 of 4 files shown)


lucksus:
Hm, I don't see much benefit in moving the initialize_network code (on its own) into a separate workflow. What we should have imho is an over all initialize workflow that applies in all cases (load and new instance) which always initializes the network but checks the source chain if it is initialized and other wise runs `initialize_application` as well..

container_api/src/holochain.rs
```diff
@@ -109,15 +105,15 @@ impl Holochain {
     pub fn load(_path: String, context: Arc<Context>) -> Result<Self, HolochainError> {
         let persister = SimplePersister::new(context.file_storage.clone());
         let loaded_state = persister
-            .load(context.clone())
-            .unwrap_or(Some(State::new(context.clone())))
-            .unwrap();
+            .load(context.clone())?
+            .unwrap_or(State::new(context.clone()));
         // TODO get the network state initialized!!
```
zippy: I think this comment can be deleted now!

zippy:
@lucksus, what you say makes sense, but I think this PR is a step forward, so I'll add that to the SoA...

container_api/src/holochain.rs
```diff
@@ -87,17 +87,13 @@ impl Holochain {
         let mut instance = Instance::new(context.clone());
         let name = dna.name.clone();
         instance.start_action_loop(context.clone());
-        let context = instance.initialize_context(context.clone());
-        let context2 = context.clone();
-        let result = block_on(
-            initialize_application(dna, &context2).and_then(|_| initialize_network(&context)),
-        );
+        let result = block_on(network::initialize(&instance, Some(dna), context.clone()));
```
lucksus: Yeah, I was thinking this misses `initialize_application` but that is included in this workflow. But `network::initialize` is misleading then.

PR #731, opened 2018-12-06
fixup hdk examples

Connoropolous:


hdk-rust/src/api.rs
```diff
@@ -196,7 +196,7 @@ pub enum BundleOnClose {
 /// # fn main() {
 /// pub fn handle_some_function(content: String) -> JsonString {
 ///     // ...
-///     hdk::debug("write a message to the logs");
+///     let _ = hdk::debug("write a message to the logs")?;
```
maackle: Since debug returns `ZomeApiResult<()>` you don't need to bind the `()` output.

```suggestion
///     hdk::debug("write a message to the logs")?;
```

PR #732, opened 2018-12-06
Tracing (Internal) Signals

maackle:
- Wrap `Instance` and `Context` signal senders in `Option` rather than initializing with broken channels
- Add Signal struct with Internal variant
- Emit `Signal::Internal` for most Actions (all but InitApplication)
- Test for said emitting
- Adds utility for replacing history.len() checks with a signal receiver


core/src/instance.rs
```diff
@@ -570,7 +634,7 @@ pub mod tests {
     pub fn can_process_action() {
         let mut instance = Instance::new(test_context("jason"));
 
-        let context = instance.initialize_context(test_context("jane"));
```
lucksus: Pretty sure that removing this call (and the ones below) to `initialize_context` is what makes the tests fail..

core/src/instance.rs
```diff
@@ -765,7 +829,6 @@ pub mod tests {
         let instance = Instance::new(test_context("jason"));
         let state_observers: Vec<Observer> = Vec::new();
         let (_, rx_observer) = channel::<Observer>();
-        let context = instance.initialize_context(context);
```
lucksus: this line is needed

core/src/instance.rs
```diff
@@ -796,7 +859,6 @@ pub mod tests {
...
```
lucksus: this line is needed

core/src/instance.rs
```diff
@@ -174,24 +200,61 @@ impl Instance {
...
+            let fire_away = match action {
+                AddLink(_) => true,
+                InitApplication(_) => false,
+                Hold(ref entry) => match entry {
+                    Entry::App(_, _) => true,
+                    _ => false,
+                },
+                Commit((ref entry, _)) => match entry {
+                    Entry::App(_, _) => true,
+                    _ => false,
```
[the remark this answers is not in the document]
maackle: Yeah good point, I replaced it with a user-provided closure

core/src/nucleus/actions/build_validation_package.rs
```diff
@@ -133,6 +133,8 @@ pub fn build_validation_package(entry: &Entry, context: &Arc<Context>) -> Valida
 
             context
                 .action_channel
+                .as_ref()
+                .expect("action channel to be initialized in reducer")
```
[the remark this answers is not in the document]
maackle: Yes, thanks for the thorough read-through!

container_api/src/container.rs
```diff
@@ -141,6 +174,11 @@ impl Container {
             .collect();
 
         if errors.len() == 0 {
+            self.instances = instances;
+            // @NB: if the Instance constructor ever changes such that it can accept a SyncSender rather than
+            // spitting out a Receiver, we can just clone that Sender instead of gathering the receivers
+            let signal_rx = combine_receivers(signal_rxs);
+            self.spawn_signal_thread(signal_rx);
```
zippy: if spawn_signal_thread returns a Result then this could just have a `?`

core/src/context.rs
```diff
@@ -143,6 +143,29 @@ impl Context {
         dna.get_wasm_from_zome_name(zome)
             .and_then(|wasm| Some(wasm.clone()).filter(|_| !wasm.code.is_empty()))
     }
+
+    // @NB: these three getters smell bad because previously Instance and Context had SyncSenders
+    // rather than Option<SyncSenders>, but these would be initialized by default to broken channels
+    // which would panic if `send` was called upon them. These `expect`s just bring more visibility to
+    // that potential failure mode.
```
zippy: nice!

zippy:
I'm mildly in favor of just merging this before we figure out the windows problem and adding it to the list.  However  I did notice something weird.  Look at line 848 of https://travis-ci.com/holochain/holochain-rust/jobs/165253486

why are there those compiler warnings?  I don't see them when I run holochain-cmd tests locally.

PR #733, opened 2018-12-07
Introduce ability to run foreign networking code as a subprocess

neonphog:
> 310 additions and 145 deletions

Tested against https://github.com/holochain/n3h/commit/e467be91661c13e039e60319d8c0438d76cccdd4

With `cargo run -p holochain_test_bin --bin test_bin_ipc ~/projects/n3h`

Makes working with external ipc networking modules a lot easier if we don't have to manually execute the processes. Also crucial for integration tests, allowing us to spin up clusters of nodes and verify functionality.


Connoropolous:
Looks exciting @neonphog . Can you give me a little more detail in terms of 
a. usage
b. the slightly zoomed out picture of what this PR enables
?

neonphog:
@Connoropolous 

> Looks exciting @neonphog . Can you give me a little more detail in terms of
> a. usage
> b. the slightly zoomed out picture of what this PR enables

Before this PR, my workflow was:
- Edit networking code
- Launch one or more networking processes
- Copy the ipc uris from those processes
- Launch holochain-rust tests with ipc uris

After this PR:
- Edit networking code
- Launch holochain-rust tests

---

Since rust can now spawn the networking process and glean configuration info from the process output, there is no configuration dance.

Outside of a testing context, the normal user will probably not require any special networking configuration. This means they can just run the holochain process, and it will automagically manage the networking process in the background.

zippy:
When I run this I get:
```
$ cargo run -p holochain_test_bin --bin test_bin_ipc ../n3h
    Finished dev [unoptimized + debuginfo] target(s) in 0.23s
     Running `target/debug/test_bin_ipc ../n3h`
node1 path: /tmp/.tmpzZ5LYd
node2 path: /tmp/.tmpI6BBN5
thread '<unnamed>' panicked at 'Os { code: 2, kind: NotFound, message: "No such file or directory" }', net_connection/src/net_connection_thread.rs:59:77
note: Run with `RUST_BACKTRACE=1` for a backtrace.
thread '<unnamed>' panicked at 'Os { code: 2, kind: NotFound, message: "No such file or directory" }', net_connection/src/net_connection_thread.rs:59:77
```
is this what you expect?

zippy:
I did a git clone of  n3h, and an `npm install` from that directory and then when I run you test I get the error output below.  Probably missing a setup step?

```
$ cargo run -p holochain_test_bin --bin test_bin_ipc ~/code/metacurrency/holochain/n3h
    Finished dev [unoptimized + debuginfo] target(s) in 0.22s
     Running `target/debug/test_bin_ipc /home/eric/code/metacurrency/holochain/n3h`
node1 path: /tmp/.tmp9STpqq
node2 path: /tmp/.tmpAliWWW
module.js:487
    throw err;
    ^

Error: Cannot find module '@holochain/n3h-common'
    at Function.Module._resolveFilename (module.js:485:15)
    at Function.Module._load (module.js:437:25)
    at Module.require (module.js:513:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/home/eric/code/metacurrency/holochain/n3h/packages/n3h/lib/index.js:5:40)
    at Module._compile (module.js:569:30)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:503:32)
    at tryModuleLoad (module.js:466:12)
    at Function.Module._load (module.js:458:3)
module.js:487
    throw err;
    ^

Error: Cannot find module '@holochain/n3h-common'
    at Function.Module._resolveFilename (module.js:485:15)
    at Function.Module._load (module.js:437:25)
    at Module.require (module.js:513:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/home/eric/code/metacurrency/holochain/n3h/packages/n3h/lib/index.js:5:40)
    at Module._compile (module.js:569:30)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:503:32)
    at tryModuleLoad (module.js:466:12)
    at Function.Module._load (module.js:458:3)
```

neonphog:
@zippy 

> I did a git clone of n3h, and an npm install from that directory and then when I run your test I get the error output below. Probably missing a setup step?

Yeah, you need to install the monorepo sub-package dependencies with:

```shell
npm run bootstrap
```

zippy:
Yep, that did the trick.  Works!

neonphog:
@lucksus

> Please correct me if I'm misunderstanding something, but having spawn commands and args in the config of P2pNetwork carries the implicit assumption that we have only one instance (= one app) per node, no?

I don't believe so, no. I was imagining the container spawning P2pNetwork, and passing it into an app instance which could then call `trackApp` on it. While minimal, there is some overhead to the ipc connection, so we should share that where possible.

Even if we want separate ipc connections for apps, this spawn PR doesn't preclude that... if you don't pass in the spawn config, but instead pass in an ipc URI, you will get the old behavior.

neonphog:
@lucksus 

> We need one P2pNetwork per instance because
> a) the handler needs to connect the right instance

Yes, this will require some thought. Likely some kind of routing module at the container level that understands the protocol well enough to direct messages known to be dna / agent specific correctly, and broadcast any others to all app instances.

> and b) each of these network proxies registers with a specific DNA hash.

Yep, that is what the trackApp message does, but the networking code is happy to process / multiplex as many dna/agent pairs as you feel like registering.

lucksus:
@neonphog, well I was operating under the assumption that we have to have one `P2pNetwork` instance per Holochain instance. When I saw the handler passed in in the constructor I had to move construction of it into core: https://github.com/holochain/holochain-rust/blob/develop/core/src/network/reducers/init.rs. Also, since `P2pNetwork::send(&mut self, data: Protocol)` takes a mutable self-reference I have to put it somewhere that enables core to change it. I first tried a different way, but I then made network a first-class module with state and reducers. So the network is seen as part of the instance's state and mutating it (which includes sending (`send(&mut self, data: Protocol)`) something) happens through an action and a state reduction step. After building out the [code](https://github.com/holochain/holochain-rust/tree/develop/core/src/network) like this last week I'd say it feels very right.


> 
Yes, this will require some thought. Likely some kind of routing module at the container level that understands the protocol well enough to direct messages known to be dna / agent specific correctly, and broadcast any others to all app instances.

So that implies this is not happening right now? Which means that a `P2pNetwork` instance receives all network messages also those directed to other agent_ids (and also other dna_hashes?) than the one we provided in TrackApp? But `MockWorker` is not implemented that way. It does only send messages to the specific channel for each `agent_id` and `dna_hash`... 

So if we stick to one `P2pNetwork` per instance because of the reasons above, each handler would have to filter and ignore messages that are not meant for our dna_hash/agent_id combination? Shouldn't that go into the network module?

And: in order to spawn the other process, the container could just instantiate another `P2pNetwork` with spawn config, right?

net/src/ipc_net_worker.rs
```diff
@@ -103,14 +108,42 @@ impl IpcNetWorker {
         if config["socketType"] != "zmq" {
             bail!("unexpected socketType: {}", config["socketType"]);
         }
+        let mut block_connect = true;
+        if let Some(b) = config["blockConnect"].as_bool() {
+            block_connect = b;
+        }
```
sphinxc0re: Maybe do:
```rust
let block_connect = config["blockConnect"].as_bool().unwrap_or(true);
```

neonphog: Updated. I totally appreciate your suggestion, and the terseness of the change, and at the same time... if I'm not allowed to use if let or match statements anymore, I wish someone would add a cli flag to the compiler or something to disallow them in the first place to prevent the back-n-forth...

neonphog:
@sphinxc0re - PR updated, please take another look when you get a chance

net/src/ipc_net_worker.rs
```diff
@@ -108,10 +108,7 @@ impl IpcNetWorker {
         if config["socketType"] != "zmq" {
             bail!("unexpected socketType: {}", config["socketType"]);
         }
-        let mut block_connect = true;
-        if let Some(b) = config["blockConnect"].as_bool() {
-            block_connect = b;
-        }
+        let block_connect = config["blockConnect"].as_bool().unwrap_or(true);
         if let None = config["ipcUri"].as_str() {
```
sphinxc0re: Please use
```rust
if config["ipcUri"].as_str().is_none() {
```
here

net/src/ipc_net_worker.rs
```diff
@@ -103,14 +108,39 @@ impl IpcNetWorker {
         if config["socketType"] != "zmq" {
             bail!("unexpected socketType: {}", config["socketType"]);
         }
+        let block_connect = config["blockConnect"].as_bool().unwrap_or(true);
         if let None = config["ipcUri"].as_str() {
+            if config["spawn"].is_object() {
+                let s = config["spawn"].as_object().unwrap();
```
sphinxc0re: This is the perfect place to use an `if let`:
```rust
if let Some(object) = config["spawn"].as_object() {
    ...
```

net/src/ipc_net_worker.rs
```diff
@@ -103,14 +108,39 @@ impl IpcNetWorker {
...
+                if s["cmd"].is_string()
+                    && s["args"].is_array()
+                    && s["workDir"].is_string()
+                    && s["env"].is_object()
+                {
+                    let mut env = HashMap::new();
+                    for (k, v) in s["env"].as_object().unwrap().iter() {
+                        env.insert(k.to_string(), v.as_str().unwrap().to_string());
+                    }
```
sphinxc0re: You can use an iterator and `map` + `collect` here to make this more declarative and easier to understand

net/src/ipc_net_worker.rs
```diff
@@ -119,13 +149,91 @@ impl IpcNetWorker {
...
+    /// create a new IpcNetWorker instance pointing to a process that we spawn here
+    fn priv_spawn(
+        handler: NetHandler,
+        cmd: String,
+        args: Vec<String>,
+        work_dir: String,
+        env: HashMap<String, String>,
+        block_connect: bool,
+    ) -> NetResult<Self> {
+        let mut proc = std::process::Command::new(cmd);
```
sphinxc0re: isn't proc a reserved word in Rust? 😁

net/src/ipc_net_worker.rs
```diff
@@ -119,13 +149,91 @@ impl IpcNetWorker {
...
+        proc.stdout(std::process::Stdio::piped());
+        proc.args(&args);
+        proc.envs(&env);
+        proc.current_dir(work_dir);
```
sphinxc0re: `Command::new()` returns a struct that can be used in a "builder-like" manner. For example:
```rust
let proc_handler = Command::new(cmd)
    .stdout(std::process::Stdio::piped())
    .args(&args)
    .envs(&env)
    .current_dir(work_dir)
    .spawn()?;
```

net/src/ipc_net_worker.rs
```diff
@@ -119,13 +149,91 @@ impl IpcNetWorker {
...
+        let mut proc = proc.spawn()?;
+
+        let re_ready = regex::Regex::new("#IPC-READY#")?;
+        let re_ipc = regex::Regex::new("(?m)^#IPC-BINDING#:(.+)$")?;
+        let re_p2p = regex::Regex::new("(?m)^#P2P-BINDING#:(.+)$")?;
+
+        let mut ipc_binding = "".to_string();
```
sphinxc0re: If you need an empty string, always use `String::new()`, which gives you one

net/src/ipc_net_worker.rs
```diff
@@ -119,13 +149,91 @@ impl IpcNetWorker {
...
+        let mut p2p_bindings: Vec<String> = Vec::new();
+
+        if let Some(ref mut stdout) = proc.stdout {
```
sphinxc0re: So... I don't really get what the purpose of this whole codeblock is. It seems like you are reading the output of the subprocess. Could you elaborate?

net_connection/src/net_connection_thread.rs
```diff
@@ -32,16 +32,21 @@ impl NetConnectionThread {
     /// stop (join) the worker thread
     pub fn stop(self) -> NetResult<()> {
         self.keep_running.store(false, Ordering::Relaxed);
-        match self.thread.join() {
-            Ok(_) => Ok(()),
-            Err(_) => {
-                bail!("NetConnectionThread failed to join on stop() call");
-            }
+        if let Err(_) = self.thread.join() {
```
sphinxc0re: Again, please use the `is_err()` method

test_bin/src/test_bin_ipc.rs
```diff
@@ -56,140 +81,108 @@ fn exec() -> NetResult<()> {
...
 
-    // yay, everything worked
-    println!("test complete");
+    let connect_result_2 = node2.wait(Box::new(is_peer_connected))?;
+    println!("got connect result 2: {:?}", connect_result_2);
 
-    // shut down the P2pNetwork instance
-    con.stop()?;
+    node1.drop();
+    node2.drop();
```
sphinxc0re: Why are you dropping those by hand?

net/src/ipc_net_worker.rs
neonphog: Why?

This should compile into equivalently efficient optimized byte code, correct? The terseness is equivalent and the readability to my eyes only differs subjectively.

test_bin/src/test_bin_ipc.rs
neonphog: erm... sorry, bad name. changing to `stop` for now. `stop` on P2pNetwork right now takes `self` (moved, not a reference). I'd like to fix that, but as a separate effort not part of this.

net/src/ipc_net_worker.rs
neonphog: changed... although I still needed some stack space so I could print out the spawn command for debugging.

net/src/ipc_net_worker.rs
neonphog: updated to `child`

net/src/ipc_net_worker.rs
neonphog: done

net/src/ipc_net_worker.rs
neonphog: done

net/src/ipc_net_worker.rs
neonphog: good call. added some comments.

You are correct, we need to know the child process is up and running before we try to connect to it over ipc. It monitors stdout, gleans some binding info, and then proceeds when it gets the `#IPC-READY#` signal.

net_connection/src/net_connection_thread.rs
neonphog: I feel like this might be subjective style, no? If you use a `_` doesn't the compiler optimize out the unused stack variable?

Went ahead and changed this, incase the optimizer isn't that smart.

neonphog:
@sphinxc0re - PR updated, please take another look at your convenience.

To expand on https://github.com/holochain/holochain-rust/pull/733#discussion_r240565738 a little:

I really appreciate your feedback and attention to readability, and I hope you don't think I'm being too stubborn.

From my perspective rust has a lot of ways syntactically of doing the same things. Some of them are arguably better for readability, but some of them are less obvious, and, for me, anything that can't be boiled down to a code-style checker and verified pre-commit feels very arbitrary to me. Especially for a line of code that wasn't changed in this PR.

net/src/ipc_net_worker.rs
sphinxc0re: It's just a common way of writing this in Rust. Also clippy would argue about this 😉

PR #736, opened 2018-12-10
Validate on DHT hold

lucksus:
# Only hold valid entries in the DHT
New workflow 
```rust
pub async fn hold_entry_workflow<'a>(
    entry_with_header: &'a EntryWithHeader,
    context: &'a Arc<Context>,
) -> Result<Address, HolochainError>
``` 
that runs the full validation process on the entry before adding it to the local DHT shard.

Used in the network handler for `ProtocolWrapper::StoreDht`

# Pre-/Side work
* Extracted and reused `AgentState::get_header_for_entry`
* Renamed `EntryWithHeader::entry_body` back to just `EntryWithHeader::entry`

# Scenario test that simulates an attack
The test
```rust
fn test_reject_invalid_entry_on_hold_workflow()
```
creates an instance that has a different DNA with loose validation rules but that spoofes its DNA hash to be in the same mock network as the second instance on which we expect the entry validation to fail.

### DIFF
--- a/core/src/agent/state.rs
+++ b/core/src/agent/state.rs
@@ -91,6 +91,12 @@ impl AgentState {
             _ => unreachable!(),
         }
     }
+
+    pub fn get_header_for_entry(&self, entry: &Entry) -> Option<ChainHeader> {
+        self.chain()
+            .iter_type(&self.top_chain_header(), &entry.entry_type())
+            .find(|h| h.entry_address() == &entry.address())
+    }
 }
 
 #[derive(Clone, Debug, Deserialize, Serialize, DefaultJson)]
--- a/core/src/network/entry_with_header.rs
+++ b/core/src/network/entry_with_header.rs
@@ -6,14 +6,14 @@ use std::{convert::TryInto, sync::Arc};
 
 #[derive(Serialize, Deserialize)]
 pub struct EntryWithHeader {
-    pub entry_body: Entry,
+    pub entry: Entry,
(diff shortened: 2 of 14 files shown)


core/src/agent/state.rs
```diff
@@ -91,6 +91,12 @@ impl AgentState {
             _ => unreachable!(),
         }
     }
+
+    pub fn get_header_for_entry(&self, entry: &Entry) -> Option<ChainHeader> {
```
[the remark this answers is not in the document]
lucksus: Well, this function is returning `None` if the entry is not in the source chain, which is what I need here.

core/src/network/entry_with_header.rs
```diff
@@ -6,14 +6,14 @@ use std::{convert::TryInto, sync::Arc};
...
-    pub entry_body: Entry,
+    pub entry: Entry,
     pub header: ChainHeader,
 }
 
 impl EntryWithHeader {
     pub fn new(entry: Entry, header: ChainHeader) -> EntryWithHeader {
         EntryWithHeader {
-            entry_body: entry,
+            entry: entry,
```
sphinxc0re: `entry: entry` -> `entry`

lucksus: Hey @sphinxc0re, please don't block for small single changes like this but instead use Github's suggestions feature:
```suggestion
            entry,
```

PR #737, opened 2018-12-10
Validate links on DHT hold meta

lucksus:
Similar to https://github.com/holochain/holochain-rust/pull/736 on which this sits on top.

# Only hold valid links in the DHT
New workflow 
```rust
pub async fn hold_link_workflow<'a>(
    entry_with_header: &'a EntryWithHeader,
    context: &'a Arc<Context>,
) -> Result<Address, HolochainError>
``` 
that runs the full validation process on links before adding them to the local DHT shard.

Used in the network handler for `ProtocolWrapper::StoreDhtMeta`

# Pre-/Side work
* Since this broke the `can_roundtrip_links` test again, I've refactored it to also do polling for the result.
* To see if this fixes it for windows again I've removed the `#[cfg(not(windows))]` for that test.

# Scenario test that simulates an attack
The test
```rust
fn test_reject_invalid_link_on_hold_workflow()
```
creates an instance that has a different DNA with loose validation rules but that spoofes its DNA hash to be in the same mock network as the second instance on which we expect the entry validation to fail.

### DIFF
--- a/app_spec/test/test.js
+++ b/app_spec/test/test.js
@@ -89,7 +89,7 @@ test('posts_by_agent', (t) => {
   t.deepEqual(result.Ok, {"addresses":[]})
 })
 
-test('my_posts', (t) => {
+test('my_posts', async (t) => {
   t.plan(1)
 
   app.call("blog", "main", "create_post",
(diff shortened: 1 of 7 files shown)


thedavidmeister:
i'll get deeper into this after #736 gets in :)

PR #740, opened 2018-12-11
Add regex-based "glob" pattern matching to hdk::query (OLD: See #781)

pjkundert:
o Change the Holochain query API to accept a `Vec<String>` vs. `String`
o If multiple EntryType names (or one with "glob" wildcards) is used,
  then produce a GlobSetBuilder matcher to recognize the selected names.

There is no way to get more than a single EntryType via hdk::query.

The primary purpose of this change is to support querying multiple EntryType names in a single hdk::query request.  To do this efficiently, a single regex encompassing all of the desired EntryType names is produced.

The globset::GlobSetBuilder is used to do this, and also supports the full "glob" pattern matching on EntryType names.  This includes the concept of path-like namespaces for EntryType names.  For example, if you have 3 EntryTypes `"one"`, `"two"` and `"three"` related to the concept "tx" in your Holochain application, you can name them `"tx/one"`, `"tx/two"` and `"tx/three"`.  Then, when you want to query all of your tx-related entries from your local chain, you'd call

```
hdk::query( vec!("tx/*".to_string()))
```

Later, when you add a fourth "tx/four" EntryType, the above code would automatically get all the tx-related EntryTypes.

You can also do things like get all EntryTypes except the System EntryTypes: `"[!%]*"`.

NOTE: Changes the hdk::query API!

### DIFF
--- a/core/src/agent/chain_store.rs
+++ b/core/src/agent/chain_store.rs
@@ -1,3 +1,4 @@
+use globset::{ GlobBuilder, GlobSetBuilder };
 use holochain_core_types::{
     cas::{
         content::{Address, AddressableContent},
@@ -5,12 +6,12 @@ use holochain_core_types::{
     },
     chain_header::ChainHeader,
     entry::entry_type::EntryType,
-    error::RibosomeErrorCode,
-    error::RibosomeErrorCode::*,
+    error::RibosomeErrorCode::{self, *},
 };
-use std::sync::{Arc, RwLock};
-use globset::{ GlobBuilder, GlobSetBuilder };
-use std::str::FromStr;
+use std::{
+    str::FromStr,
+    sync::{Arc, RwLock}
+};
 
 #[derive(Debug, Clone)]
 pub struct ChainStore {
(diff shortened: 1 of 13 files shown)


hdk-rust/src/api.rs
```diff
@@ -849,16 +849,16 @@ pub fn get_links<S: Into<String>>(base: &Address, tag: S) -> ZomeApiResult<GetLi
 }
 
 /// Returns a list of entries from your local source chain, that match a given type.
-/// entry_type_name: Specify type of entry to retrieve
+/// entry_type_names: Specify type of entry(s) to retrieve, as a Vec<String> of 0 or more names.
 /// limit: Max number of entries to retrieve
-pub fn query(entry_type_name: &str, start: u32, limit: u32) -> ZomeApiResult<QueryResult> {
+pub fn query(entry_type_names: Vec<String>, start: u32, limit: u32) -> ZomeApiResult<QueryResult> {
```
sphinxc0re: I'm not so sure about the overall API here. The current `query` function seems a little bit unergonomic. Maybe we could provide a `QueryBuilder`?

core/src/agent/chain_store.rs
```diff
@@ -109,17 +115,19 @@ impl ChainStore {
                                  .literal_separator( true )
                                  .build() {
                                      Ok(pat) => pat,
-                                     Err(_) => return Err( UnknownEntryType ),
+                                     Err(_) => return Err(UnknownEntryType),
```
sphinxc0re: You can replace this `match` statement with:
```rust
GlobBuilder::new( name )
    .literal_separator(true)
    .literal_separator(true)
    .build().map_err(|_| Err(UnknownEntryType))?

File: hdk-rust/src/api.rs (hunk 1 above)
Comment (pjkundert) [reply hdk-rust/src/api.rs]: Perhaps, although I hope querying the local chain remains a pretty simple function.  This just restores the basic needed functionality to support basic usage -- we need to be able to query for a set of EntryType names, in order to execute local chain state validation machinery, and for that we need to be able to pass multiple query strings, and scan for them efficiently; globset::GlobSetBuilder constructs a single efficient regex matcher to do that.  And, we get good, simple and understandable EntryType name "namespacing" using standard filename-like  "globbing" for free, to support organizing the naming of EntryTypes.

Comment (zippy) [reply hdk-rust/src/api.rs]: Overall this looks fabulous as a way to not build a whole query language and get what we need in the short term.  Nice work.  One thing though: I do think that for most cases people will just want to do:
`hdk::query("post", 0, 0)` not `hdk::query( vec!["post".to_string()], 0, 0)`.  So I propose we do the same thing we did for `hdk::get_entry()` which is that at the hdk level we make the most common use case have the normal name, but then have something like `hdk::query_glob()` which uses the vec, and that's what calls through to `query()` in api.rs

I'll give a more thorough review tommorow.

Comment (sphinxc0re) [reply hdk-rust/src/api.rs]: You could also say, that that the first argument is `impl From<QueryArgsNames>`:
```rust
pub fn query(query_args: impl From<QueryArgsNames>, start: usize, limit: usize) { ... }
```
And then do the conversion implicitly in the function body.

pjkundert:
Archived; see #781

PR #742, opened 2018-12-11
[WIP] Add API function get_links_and_load

willemolding:
This aims to restore some of the functionality that was in holochain-proto for loading all the entries linked to a given address.

It also adds a function `get_links_and_load_as` which is identical but will also attempt to convert the loaded entries to a given native type. Any entries that cannot be successfully converted are dropped so this function can still be called on a link base that has different entry types linked to it.

While this does not enable any new functionality it does significant reduce the boilerplate code required to write zomes. Combined with the ability to return `ZomeApiResult<>` means many zome functions can be written in one line.

example from holo-chat:
```
pub fn handle_get_my_channels() -> ZomeApiResult<utils::GetLinksLoadResult<Channel>> {
    utils::get_links_and_load_as(&get_my_member_id().hash(), "member_of")
}

pub fn handle_get_members(address: HashString) -> ZomeApiResult<utils::GetLinksLoadResult<member::Member>> {
    utils::get_links_and_load_as(&address, "has_member")
}

pub fn handle_get_messages(address: HashString) -> ZomeApiResult<utils::GetLinksLoadResult<message::Message>> {
    utils::get_links_and_load_as(&address, "message_in")
}

pub fn handle_get_subjects(address: HashString) -> ZomeApiResult<utils::GetLinksLoadResult<Subject>> {
    utils::get_links_and_load_as(&address, "channel_subject")
}
```

We have been including this helper function in our zomes up to this point but to encourage good practices for other developers it makes sense to include this as part of the core api

### DIFF
--- a/hdk-rust/src/api.rs
+++ b/hdk-rust/src/api.rs
@@ -13,7 +13,7 @@ pub use holochain_wasm_utils::api_serialization::validation::*;
 use holochain_wasm_utils::{
     api_serialization::{
         get_entry::{EntryHistory, GetEntryArgs, GetEntryOptions, StatusRequestKind},
-        get_links::{GetLinksArgs, GetLinksResult},
+        get_links::{GetLinksArgs, GetLinksResult, GetLinksLoadElement, GetLinksLoadResult},
         link_entries::LinkEntriesArgs,
         QueryArgs, QueryResult, UpdateEntryArgs, ZomeFnCallArgs,
     },
@@ -171,15 +171,6 @@ pub enum BundleOnClose {
     Commit,
     Discard,
 }
-
-// Response for calls to get_links_and_load
-#[derive(Serialize, Deserialize, Debug, Clone)]
-pub struct GetLinksLoadElement<T> {
-    pub address: HashString,
-    pub entry: T
-}
-
-pub type GetLinksLoadResult<T> = Vec<GetLinksLoadElement<T>>;
 
 //--------------------------------------------------------------------------------------------------
 // API FUNCTIONS
--- a/hdk-rust/src/api.rs
+++ b/hdk-rust/src/api.rs
@@ -2,6 +2,7 @@ use crate::{
     error::{ZomeApiError, ZomeApiResult},
     globals::*,
 };
+use std::convert::TryFrom;
 use holochain_core_types::{
     cas::content::Address,
     crud_status::CrudStatus,
@@ -19,6 +20,7 @@ use holochain_wasm_utils::{
     holochain_core_types::{
         hash::HashString,
         json::{JsonString, RawString},
+        entry::AppEntryValue,
     },
     memory_allocation::*,
     memory_serialization::*,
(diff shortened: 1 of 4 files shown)


wasm_utils/src/api_serialization/get_links.rs
```diff
@@ -20,3 +20,11 @@ impl GetLinksResult {
         &self.addresses
     }
 }
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct GetLinksLoadElement<T> {
+    pub address: Address,
```
zippy: I'm not sure we need the `address` field because every entry has an .address() function that returns the entrie's address.  I guess there could be a santity check that when the link is loaded that the actual address matches the calculated `.address()`

thedavidmeister: @zippy yes i agree, would be a validation/sanity check rather than data, "out of the tarpit" and all :)

thedavidmeister: so it's just return address OR entry, right? an address can be "trade in" for an entry via some workflow that relies on network/storage but an entry should always dictate its own address deterministically

willemolding: Cool I actually didn't know you could just call `.address()` on an entry.

So this makes sense for `get_links_and_load` but I think `get_links_and_load_as` still needs to return the struct with a separate address field. This is because the result is no longer an `entry` type but a native struct type. What do you think?

hdk-rust/src/api.rs
```diff
@@ -896,6 +898,68 @@ pub fn close_bundle(_action: BundleOnClose) -> ZomeApiResult<()> {
...
+    .map(|address| {
+        get_entry(address.to_owned())
+        .map(|entry: Option<Entry>| {
+            GetLinksLoadElement{
+                address: address.to_owned(),
+                entry: entry.unwrap()
+            }
+        })
+    })
+    .filter_map(Result::ok)
```
AshantiMutinta: I think we should record that some of the get entries error out. Either by bubbling up the errors or returning a vec of results.

AshantiMutinta: Yes! I think that would be awesome

zippy:
The main issue that I'd like to see resolved here is that I'd much rather there just be one call at the API level one call, get_links that takes a [GetLinkOptions](https://realtimeboard.com/app/board/o9J_kyiXmFs=/?moveToWidget=3074457346349718699) structure that matches [GetEntryOptions](https://github.com/holochain/holochain-rust/blob/02f79f79ead2ea913c4480143998f4211307b3ef/wasm_utils/src/api_serialization/get_entry.rs#L25)

We are going to need to be able to also specify crud request as well for links, so we really need all this machinery, and then we can hide the details by adding simplified versions in the HDK, that match the signatures you've implemented.

zippy:
@Connoropolous I don't think there should be any performance hit if we implement it right, because the return type on the GetEntryResult uses an enum on the data returned (i.e. an EntryHistory or a single GetEntryResultItem, so it should just be the cost of an if.  I think this makes it much easier for us down the line to extend the functionality by being able to add extra fields in the GetEntryOptions and GetEntryResultItem without having to change api signatures or add extra api functions.

hdk-rust/src/api.rs
zippy: I think this is now addressed.

PR #743, opened 2018-12-11
EntryHistory -> GetEntryResult return value for get_entry_result

zippy:
get_entry_result returns GetEntryResult which now encapsulates EntryHistory, and prepares for being able to return sources, and headers with bools in GetEntryOptions.


wasm_utils/src/api_serialization/get_entry.rs
```diff
@@ -72,3 +87,82 @@ impl EntryHistory {
...
+
+    pub fn latest(&self) -> Option<Entry> {
+        if !self.found() {
+            return None;
+        }
+        let entry = match self.history {
+            None => None,
+            Some(ref h) => Some(h.entries.iter().next().unwrap().clone()),
+        };
+        entry
```
sphinxc0re: You can rewrite this and make it much cleaner:
```rust
self.history
    .filter(|_| self.found())
    .map(|h| h.entries[0].clone())
```

wasm_utils/src/api_serialization/get_entry.rs
```diff
@@ -72,3 +87,82 @@ impl EntryHistory {
...
+
+    pub fn push(&mut self, entry_with_meta: &EntryWithMeta) {
+        if !self.found {
+            self.found = true;
+            self.add_history();
+        }
+        match self.history {
+            None => unreachable!(),
+            Some(ref mut history) => history.push(entry_with_meta),
+        };
```
sphinxc0re: Is the `None` state really unreachable? Then I would advise to use `.expect("something went wrong")` and `.map` on the `self.history`.

If this shouldn't crash, I'll use an `if let Some(foo) = ...`

wasm_utils/src/api_serialization/get_entry.rs
sphinxc0re: I see this pattern of using `match` to work with an `Option<T>` way too often. Most of the matches you can and want to do on an `Option<T>` can be done with what's in the standard library: https://doc.rust-lang.org/std/option/enum.Option.html
I invite you to take a look at all those beautiful functions 😄

core/src/agent/state.rs
```diff
@@ -79,15 +79,13 @@ impl AgentState {
...
-        let agent_entry = agent_entry_history.entries.iter().next().unwrap().clone();
+        let agent_entry_result = await!(get_entry_result_workflow(context, &entry_args))?;
+        let agent_entry = agent_entry_result.latest();
         match agent_entry {
-            Entry::AgentId(agent_id) => Ok(agent_id),
+            None => Err(HolochainError::ErrorGeneric(
+                "Agent entry not found".to_string(),
+            )),
+            Some(Entry::AgentId(agent_id)) => Ok(agent_id),
             _ => unreachable!(),
```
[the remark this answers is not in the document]
zippy: which part?  I don't think so.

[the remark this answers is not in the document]
sphinxc0re: 🙈 I did a wrong 😂

wasm_utils/src/api_serialization/get_entry.rs
zippy: Sweet.  fixed and then changed.

wasm_utils/src/api_serialization/get_entry.rs
```diff
@@ -44,31 +62,209 @@ pub struct GetEntryArgs {
     pub options: GetEntryOptions,
 }
 
-#[derive(Deserialize, Debug, Serialize, DefaultJson)]
+#[derive(Deserialize, Debug, Serialize, DefaultJson, Clone)]
+pub struct EntryResultMeta {
+    pub address: Address,
+    pub entry_type: EntryType,
```
lucksus: Why is the entry type part of meta? Entry type is also part of the entry now as per discussion with @artbrock and @thedavidmeister in Montreal and implemented by @thedavidmeister's "enum branch" recently. Do we need this here?

lucksus: As per verbal discussion: it is possible to set entry, header and sources to false in `GetEntryOptions`. That would return only the `EntryResultMeta` which then also carries the information of the entry type which otherwise can only be retrieved together with the entry. Getting the type of an potentially large entry wouldn't be possible if we hadn't this field here.

core/src/network/handler/store.rs
```diff
@@ -32,13 +32,13 @@ pub fn handle_store_dht_meta(dht_meta_data: DhtMetaData, context: Arc<Context>)
...
             )
-            .expect("dht_meta_data should be EntryWithHeader");
-            let link_add = match entry_with_header.entry {
-                Entry::LinkAdd(link_add) => link_add,
-                _ => unreachable!(),
-            };
-            let link = link_add.link().clone();
-            let _ = block_on(add_link(&link, &context.clone()));
+            .expect("dht_meta_data should be EntryWithHader");
+            thread::spawn(move || {
```
sphinxc0re: You are spawning a system thread here. This can't be good. The Future should be dispatched to an executor

PR #744, opened 2018-12-11
basic ChainMigrate struct

thedavidmeister:
## changes

- port from golang + bits and pieces from https://hackmd.io/uuzBdsOKRRqS1m2Q4wyeCg?view
- rename DNA hash to address
- use `Address` type for DNA address
- `Sources` type alias
- `Signatures` type alias
- implement `AddressableContent` for `Dna`
- fix nondeterministic `test_dna()` function
- replace static constants with functions (because static constants can only return primitives and tuples)

### DIFF
--- a/core/src/network/reducers/get_entry.rs
+++ b/core/src/network/reducers/get_entry.rs
@@ -14,7 +14,7 @@ fn inner(network_state: &mut NetworkState, address: &Address) -> Result<(), Holo
         network_state,
         ProtocolWrapper::GetDht(GetDhtData {
             msg_id: "?".to_string(),
-            dna_hash: network_state.dna_hash.clone().unwrap(),
+            dna_address: network_state.dna_address.clone().unwrap(),
             from_agent_id: network_state.agent_id.clone().unwrap(),
             address: address.to_string(),
         }),
@@ -113,7 +113,7 @@ mod tests {
 
         let action_wrapper = ActionWrapper::new(Action::InitNetwork(NetworkSettings {
             config: mock_network_config(),
-            dna_hash: String::from("abcd"),
+            dna_address: "abcd".into(),
             agent_id: String::from("abcd"),
         }));
         let store = store.reduce(context.clone(), action_wrapper);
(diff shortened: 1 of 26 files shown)


core/src/network/reducers/get_entry.rs
```diff
@@ -190,7 +190,7 @@ mod tests {
         };
         let dht_data = DhtData {
             msg_id: String::from(""),
-            dna_hash: String::from(""),
+            dna_address: "".into(),
```
sphinxc0re: Please use `String::new()` if you want to create an empty string

thedavidmeister: @sphinxc0re yes and a `HashString` is not a `String`

thedavidmeister:
@zippy @lucksus 

i don't think you can lock a user out of a chain, that sounds bad, for reasons that have been documented and discussed, see the hackmd in the OP for example

```rust
pub struct ChainMigrate {
    sources: Sources,
    dna: Address,
    migrate_type: MigrateType,
    sharing: Sharing,
     #[serde(serialize_with = "serialize_data")]
    #[serde(deserialize_with = "deserialize_data")]
    data: Data,
}
```

is almost a straight port from golang version of this, that's where most of it came from

the noisy changes to dna is where the compiler took me when i set `dna: Address`

lucksus:
What was discussed as the very base version for migration is a closing entry that has the main purpose of telling other agents in the old network that we moved over to the new one.

There are several points in the hackmd that I have problems getting into a coherent picture with the mental models I perceived to be shared. My feeling is you are jumping ahead a bit. I would suggest we sync up via voice call on this before migration is due?

thedavidmeister:
@lucksus well yes of course, i already said a few times i want to sync on this but last time we tried you weren't in a good headspace for it w/ everything else that needed to land recently

core/src/network/reducers/get_entry.rs
sphinxc0re: I think it might be better to implement `Default` for `HashString` and use `Default::default()` instead of the current instantiation

thedavidmeister: mmm  that is an interesting idea

sphinxc0re: I think you can just derive it, since this will pick the default value for the internal string

thedavidmeister:
ok so this is all superceded by the docs at https://hackmd.io/5OQgaCSbSvWj4eqiHmV27Q#

PR #746, opened 2018-12-11
Node-to-node send/receive

lucksus:
# Usage
Define receive callback in `define_zome!` next to `genesis`:
```rust
define_zome! {
...
    genesis: || { Ok(()) }

    receive: |payload| {
        format!("Received: {}", payload)
    }

    functions: {
```

Use `hdk::send` to asynchronously send and receive the return value of the `receive` callback from the other node:
```rust
fn handle_send_message(to_agent: Address, message: String) -> String {
    match hdk::send(to_agent, message) {
        Ok(response) => response,
        Err(error) => error.to_string(),
    }
}
```

Yes, you can have a receive callback in each zome. `hdk::send` calls only map to the receive callback of the zome in which send gets called.

# Chunks
* `receive` added to `define_zome!` macro
* `CustomDirectMessage` added
* hc_send/invoke_send implemented
* custom_send action creator and Future that waits for the result to appear in the (network) state
* `handle_custom_direct_message` workflow that runs on the receiving end, calls the callback and dispatches an action to send the response


zippy:
tests didn't pass :-(

core/src/network/actions/custom_send.rs
```diff
@@ -0,0 +1,75 @@
...
+    task::{LocalWaker, Poll},
+};
+use holochain_core_types::cas::content::Address;
+use snowflake::ProcessUniqueId;
+use std::{
+    pin::{Pin, Unpin},
+    sync::Arc,
+};
+
+/// GetValidationPackage Action Creator
```
zippy: whoops, you forgot to update the comment!  Validation Package??

core/src/network/actions/custom_send.rs
```diff
@@ -0,0 +1,75 @@
...
+    let action_wrapper = ActionWrapper::new(Action::SendDirectMessage(direct_message_data));
+    dispatch_action(&context.action_channel, action_wrapper);
+
+    await!(SendResponseFuture {
+        context: context.clone(),
+        id,
+    })
+}
+
+/// GetValidationPackageFuture resolves to an Option<ValidationPackage>
```
zippy: ditto

core/src/network/direct_message.rs
```diff
@@ -1,4 +1,18 @@
-use holochain_core_types::{cas::content::Address, validation::ValidationPackage};
+use holochain_core_types::{
+    cas::content::Address, error::HolochainError, json::JsonString, validation::ValidationPackage,
+};
+
+/// This is direct message that got created by the zome code through hdk::send().
+#[derive(Serialize, Deserialize, Clone, PartialEq, Debug, DefaultJson)]
+pub struct CustomDirectMessage {
+    /// We have to track which zome as send the message so we can call the
```
zippy:
```suggestion
    /// We have to track which zome sent the message so we can call the
```

core/src/nucleus/ribosome/api/send.rs
```diff
@@ -0,0 +1,32 @@
...
+    network::{actions::custom_send::custom_send, direct_message::CustomDirectMessage},
+    nucleus::ribosome::{api::ZomeApiResult, Runtime},
+};
+use futures::executor::block_on;
+use holochain_core_types::error::HolochainError;
+use holochain_wasm_utils::api_serialization::send::SendArgs;
+use std::convert::TryFrom;
+use wasmi::{RuntimeArgs, RuntimeValue};
+
+/// ZomeApiFunction::CommitAppEntry function code
```
zippy: Another comment that needs updating!

core/src/workflows/handle_custom_direct_message.rs
```diff
@@ -0,0 +1,48 @@
...
+use holochain_core_types::{cas::content::Address, error::HolochainError};
+use std::sync::Arc;
+
+pub async fn handle_custom_direct_message(
+    to_agent_id: Address,
+    msg_id: String,
+    custom_direct_message: CustomDirectMessage,
+    context: Arc<Context>,
+) -> Result<(), HolochainError> {
+    println!(
```
zippy: Log instead of println?

hdk-rust/tests/integration_test.rs
```diff
@@ -574,14 +582,30 @@ fn can_remove_entry() {
...
 fn can_remove_modified_entry() {
-    let (mut hc, _) = start_holochain_instance("can_remove_modified_entry");
+    let (mut hc, _) = start_holochain_instance("can_remove_modified_entry", "alice");
     let result = hc.call("test_zome", "test_cap", "remove_modified_entry_ok", r#"{}"#);
     assert!(result.is_ok(), "result = {:?}", result);
 }
+
+#[test]
+fn can_send_and_receive() {
+    let (mut hc, _) = start_holochain_instance("can_remove_modified_entry", "alice");
```
zippy: why the "can_remove_modified_entry"  I think thats a bit confusing in the test?

core/src/network/actions/custom_send.rs
lucksus: yep

core/src/workflows/handle_custom_direct_message.rs
lucksus: actually neither. debugging left over.

lucksus:
Implemented all your points, @zippy. Yes, the timeout, was thinking about this and planning to add it as follow up. Or should I add it here and now?

zippy:
I'm ok with followup unless you think it's really easy.

One other thing I forgot was that there's no app_spec example.  That should also be a followup.

hdk-rust/wasm-test/src/lib.rs
```diff
@@ -333,6 +333,13 @@ fn hdk_test_entry() -> Entry {
     Entry::App(hdk_test_app_entry_type(), hdk_test_entry_value())
 }
 
+fn handle_send_message(to_agent: Address, message: String) -> String {
```
Connoropolous: So `send` and receive can pass/return String, can it use other types?

Like could `receive` return a ZomeApiResult?

core/src/network/direct_message.rs
```diff
@@ -1,4 +1,18 @@
-use holochain_core_types::{cas::content::Address, validation::ValidationPackage};
+use holochain_core_types::{
+    cas::content::Address, error::HolochainError, json::JsonString, validation::ValidationPackage,
+};
+
+/// This is direct message that got created by the zome code through hdk::send().
+#[derive(Serialize, Deserialize, Clone, PartialEq, Debug, DefaultJson)]
+pub struct CustomDirectMessage {
+    /// We have to track which zome sent the message so we can call the
```
Connoropolous: That's an interesting point... so an app can have one `receive` callback per zome? 

But it could handle different event types, if they were passed as data, and using a `match` in receive?

hdk-rust/wasm-test/src/lib.rs
lucksus: I tried with a `Result<String, String>` but discovered a bug in `JsonString::from<Result<>>`. Already synced with @thedavidmeister. So I reverted back to just String for now.

Basically possible but tried to keep the scope small for this one.

core/src/nucleus/ribosome/callback/receive.rs
```diff
@@ -1,17 +1,59 @@
...
 ) -> CallbackResult {
-    call(context, zome, &Callback::Receive, params)
+    let params = match params {
+        CallbackParams::Receive(payload) => payload,
+        _ => return CallbackResult::NotImplemented,
+    };
+
+    let zome_call = ZomeFnCall::new(
+        zome,
+        "no capability since this is a callback",
```
Connoropolous: I thought we weren't calling this capability anymore? Trait?

hdk-rust/src/api.rs
```diff
@@ -901,9 +906,32 @@ pub fn query(entry_type_name: &str, start: u32, limit: u32) -> ZomeApiResult<Que
...
 
-/// Not Yet Available
-pub fn send(_to: Address, _message: serde_json::Value) -> ZomeApiResult<serde_json::Value> {
-    Err(ZomeApiError::FunctionNotImplemented)
+/// Sends a node-to-node message to the given agent.
+/// This works in conjunction with the `receive` callback that has to be defined in the
+/// [define_zome!](macro.defineZome.html) macro.
+///
+/// This functions blocks and returns the result returned by the `receive` callback on
+/// the other side.
```
Connoropolous: I want to add example code in here to this description, but I am open to / happy to do that as a followup

hdk-rust/wasm-test/src/lib.rs
Connoropolous: I see, ok well knowing that we will implement a complex struct, not just a string eases my mind on this one 👍

core/src/nucleus/ribosome/callback/receive.rs
lucksus: We haven't removed "capability" from the code yet. Otherwise this function wouldn't expect a capability name as second parameter.

lucksus:
@zippy, timeout added: https://github.com/holochain/holochain-rust/pull/746/commits/5b5016998dc52ed3777276ed928774dc61dff27b

core/src/nucleus/ribosome/callback/receive.rs
Connoropolous: K

hdk-rust/src/api.rs
lucksus: Great 👍

core/src/network/actions/custom_send.rs
```diff
@@ -0,0 +1,81 @@
...
+};
+use snowflake::ProcessUniqueId;
+use std::{
+    pin::{Pin, Unpin},
+    sync::Arc,
+    thread::sleep,
+    time::Duration,
+};
+
+/// SendDiretMessage Action Creator for custom (=app) messages
```
Connoropolous:
```suggestion
/// SendDirectMessage Action Creator for custom (=app) messages
```

Connoropolous: Does the code below mean the timeout is 60 seconds? 
Can that also be added into this comment description if so?

core/src/network/direct_message.rs
thedavidmeister: @Connoropolous pretty similar to how websockets works i guess, you can send and receive strings, but dispatching logic and simulating request/response behaviour is BYO

hdk-rust/wasm-test/src/lib.rs
thedavidmeister: oh yeah i forgot about that bug

i'd better actually fix that! :sweat_smile:

Connoropolous: Is there a follow up created for this one @lucksus @thedavidmeister ?

Connoropolous:
Wooot!

PR #748, opened 2018-12-12
Get rid of bitflag

sphinxc0re:


core_types/src/crud_status.rs
```diff
@@ -24,52 +24,38 @@ pub fn create_crud_link_eav(from: &Address, to: &Address) -> EntityAttributeValu
...
+        CrudStatus::Live
     }
 }
 
-impl From<&'static str> for CrudStatus {
-    fn from(s: &str) -> CrudStatus {
-        CrudStatus::from(String::from(s))
+impl From<CrudStatus> for String {
+    fn from(status: CrudStatus) -> String {
+        let res_str = serde_json::to_string(&status).expect("failed to serialize CrudStatus enum");
```
AshantiMutinta: if it fails should we be utilzing try_from instead?

sphinxc0re: No, usually I would implement the designated `FromStr` trait which was specifically made for this.

core_types/src/crud_status.rs
```diff
@@ -24,52 +24,38 @@ pub fn create_crud_link_eav(from: &Address, to: &Address) -> EntityAttributeValu
...
-    pub struct CrudStatus: u8 {
-        const LIVE = 0x01;
-        const REJECTED = 0x02;
-        const DELETED = 0x04;
-        const MODIFIED = 0x08;
-        /// CRDT resolution in progress
-        const LOCKED = 0x10;
-    }
+/// the CRUD status of a Pair is stored as EntryMeta in the hash table, NOT in the entry itself
+/// statuses are represented as bitflags so we can easily build masks for filtering lookups
```
zippy: comment needs to be fixed!

wasm_utils/src/memory_allocation.rs
```diff
@@ -52,8 +52,6 @@ pub struct SinglePageAllocation {
     length: u16,
 }
 
-#[allow(unknown_lints)]
```
zippy: ummm why are you deleting this in this PR?  Seems like not the right place...

wasm_utils/src/memory_serialization.rs
```diff
@@ -119,8 +119,6 @@ pub fn load_json<'s, T: Deserialize<'s>>(encoded_allocation: u32) -> Result<T, H
 /// Convert a json string stored in wasm memory into a specified struct
 /// If json deserialization of custom struct failed, tries to deserialize a CoreError struct.
 /// If that also failed, tries to load a string directly, since we are expecting an error string at this stage.
-#[allow(unknown_lints)]
```
zippy: ditto?  This deletion should happen in a different PR

PR #750, opened 2018-12-12
EAV attribute validate and allowed fields

AshantiMutinta:
This is a pull request to address and implement how we want to handle attribute name validation for the EAV. At the moment I just have it accepting alphanumeric characters but that will be improved on the regex used.

Further improvements to this will require a santization rather than a validation.

Additional : DHT Reducers will need a refactor to look more like idiotmatic rust.


core/src/dht/dht_reducers.rs
```diff
@@ -231,14 +253,26 @@ fn reduce_remove_entry_inner(
...
     }
     // Update crud-link
-    let crud_link_eav = create_crud_link_eav(latest_deleted_address, deletion_address);
+    let result = create_crud_link_eav(latest_deleted_address, deletion_address);
+    if result.is_err() {
+        return Err(HolochainError::ErrorGeneric(String::from(
+            "Could not create eav",
+        )));
+    }
+    let crud_link_eav = result.expect("Should create crud link");
```
sphinxc0re: You can use the `?` operator for this.
```rust
let crud_link_eav = create_crud_link_eav(latest_deleted_address, deletion_address)
    .map_err(|_| HolochainError::ErrorGeneric(String::from("Could not create eav")))?
```

core_types/Cargo.toml
```diff
@@ -18,6 +18,7 @@ bitflags = "1.0"
 objekt="0.1.1"
 holochain_core_types_derive = { path = "../core_types_derive" }
 uuid = { version = "0.7", features = ["v4"] }
+regex="1.1.0"
```
sphinxc0re: Please put spaces around the equal sign. Makes it a little easier to read

sphinxc0re: Also don't make the version completely strict to avoid crate duplication. We can probably make it `^1.1`

core/src/dht/dht_reducers.rs
AshantiMutinta: I mentioned in heartbeat yesterday I would rather come in and do a follow up to the whole dht_reducers :), needs to be refactored so that it looks more idiomatic.

PR #752, opened 2018-12-12
Bridges DNA definition

lucksus:
## Update
I've applied @zippy's change request to have a `enum BridgeReference` and I just reused `struct dna::capabilities::Capability` for the `BridgeReference::Capability` variant.

This way we can already specify a bridge dynamically through function signatures :)

## Addition to DNA spec:
```json
"bridges": [
    {
        "presence": "required",
        "handle": "DPKI",
        "reference": {
            "dna_address": "Qmabcdef1234567890"
        }
    },
    {
        "presence": "optional",
        "handle": "Vault",
        "reference": {
            "capabilities": {
                "persona_management": {
                    "capability": {
                        "membrane": "public"
                    },
                    "functions": [
                        {
                            "name": "get_persona",
                            "inputs": [{"name": "domain", "type": "string"}],
                            "outputs": [{"name": "persona", "type": "json"}]
                        }
                    ]
                }
            }
        }
    }
]
```

As per discussion with @zippy and @artbrock on 13/12/2018 I've moved the bridge definitions into the zome.


philipbeadle:
We also need to be able to setup bridges at runtime, eg when a new private chat group is setup that has its own DHT :)

lucksus:
@philipbeadle, we talked about your point yesterday. I see setting up these bridges as container concern. The UI that deals with several apps/DNAs/DHTs needs to have admin privileges (or rely on some signalling system) to install private-channel-DNAs. Those channel DNAs define (statically) that they can or need to bridge to some other DNA (Vault or a root chat DNA). They can do that by either locking in the DNA hash of that dependency (v1) or by defining a trait, that is set of functions they expect in their bridge app (follow-up)

Does that match what you're thinking?

core_types/src/dna/dna.rs
```diff
@@ -192,6 +192,14 @@ impl Dna {
         multihash::encode(multihash::Hash::SHA2256, &s.into_bytes())
             .map_err(|error| HolochainError::ErrorGeneric(error.to_string()))
     }
+
+    pub fn get_required_bridges(&self) -> Vec<Bridge> {
+        self.zomes
+            .iter()
```
sphinxc0re:
```suggestion
            .values()
```

core_types/src/dna/dna.rs
```diff
@@ -192,6 +192,14 @@ impl Dna {
...
+            .map(|(_, zome)| zome.get_required_bridges())
```
sphinxc0re:
```suggestion
            .map(|zome| zome.get_required_bridges())
```

core_types/src/dna/zome.rs
```diff
@@ -76,6 +82,9 @@ pub struct Zome {
     /// Validation code for this entry_type.
     #[serde(default)]
     pub code: DnaWasm,
+
+    /// A list of bridges to other DNAs that this DNA can use or depends on.
+    pub bridges: Option<Vec<Bridge>>,
```
sphinxc0re:
```suggestion
    #[serde(default)]
    pub bridges: Vec<Bridge>,
```

core_types/src/dna/zome.rs
```diff
@@ -108,6 +118,21 @@ impl Zome {
...
+        match self.bridges {
+            None => Vec::new(),
+            Some(ref bridges) => bridges
+                .iter()
+                .filter(|bridge| match bridge {
+                    Bridge::Address(b) => b.presence == BridgePresence::Required,
+                    Bridge::Trait(b) => b.presence == BridgePresence::Required,
+                })
+                .cloned()
+                .collect(),
```
sphinxc0re: With the changes I requested, you can remove the match statement. Since iterators in Rust are lazy, an empty iterator will instantly return an empty `Vec<T>`.

core_types/src/dna/bridges.rs
```diff
@@ -0,0 +1,62 @@
...
+///
+/// The other DNA can either be referenced statically by exact DNA address/hash or dynamically
+/// by defining the trait that other DNA has to implement in order to be used as bridge.
+///
+/// Bridges can be required or optional. If a required bridge DNA is not installed this DNA
+/// can't run, so required bridges are hard dependencies that have to be enforced by the container.
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Hash)]
+#[serde(untagged)]
+pub enum Bridge {
+    Address(AddressBridge),
```
zippy: I as I said on the heartbeat today I do think it's better to have the `dna_address` be the enum (e.g. `DnaReference`), not the `Bridge` struct itself, but I'm ok for that to be a follow-up

lucksus: Yeah, I almost did this change just now but then I thought I would love to get the details straight on the function groups (=capabilities) which are changes to the DNA as well.. can we talk about that before finalizing this one?

lucksus:
See updated description:

> 
> I've applied @zippy's change request to have a enum BridgeReference and I just reused struct dna::capabilities::Capability for the BridgeReference::Capability variant.
> 
> This way we can already specify a bridge dynamically through function signatures :)

lucksus:
@sphinxc0re please re-review.
@zippy, I know you have change Capabilities already in your branch. I would merge this one if you agree with the latest changes. Or we merge your's first and I fix conflicts in here. Anyway, I think we should have both merged in tomorrow and pair a bit on the following steps :)

core_types/src/dna/bridges.rs
lucksus: done

lucksus:
@philipbeadle, reusing the `Capabilities` struct we had in the DNA already I got this in quickly. Do think this covers everything you need? The root chat app could have a special capability meant for the private channels. That capability gets specified in the private channel DNA like above. The container can then match and suggest (via admin UI or yet TBD signal/request system) that these two DNAs can be bridged and change the container config accordingly if the user chooses so.

core_types/src/dna/bridges.rs
```diff
@@ -0,0 +1,65 @@
...
+    /// A bridge reference that defines another DNA loosely by expecting a DNA that implements
+    /// a given set of capabilities, i.e. that has specific sets of zome functions with
+    /// matching signatures.
+    Capability {
+        capabilities: BTreeMap<String, Capability>,
+    },
+}
+
+/// Required or optional
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Hash)]
```
sphinxc0re: Could you add a `#[serde(rename_all = "lowercase")]`. That way, the naming is consistent with all other JSON fields:
```suggestion
#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Hash)]
#[serde(rename_all = "lowercase")]
```

PR #755, opened 2018-12-13
Network config struct

ddd-mtl:
For loading from file and defaults handling

### DIFF
--- a/container_api/src/holochain.rs
+++ b/container_api/src/holochain.rs
@@ -11,7 +11,7 @@
 //! extern crate holochain_cas_implementations;
 //! extern crate tempfile;
 //! use holochain_container_api::*;
-//! use holochain_net::p2p_network::P2pNetwork;
+//! use holochain_net::p2p_network::P2pNetworkNode;
 //! use holochain_core_types::{agent::AgentId, dna::Dna, json::JsonString};
 //! use std::sync::{Arc, Mutex,RwLock};
 //! use holochain_core::context::Context;
@@ -37,7 +37,7 @@
 //!     Arc::new(Mutex::new(SimplePersister::new(file_storage.clone()))),
 //!     file_storage.clone(),
 //!     Arc::new(RwLock::new(EavFileStorage::new(tempdir().unwrap().path().to_str().unwrap().to_string()).unwrap())),
-//!     JsonString::from("{\"backend\": \"mock\"}"),
+//!     JsonString::from("{\"backend_end\": \"MOCK\"}"),
 //!  ).unwrap();
 //! let mut hc = Holochain::new(dna,Arc::new(context)).unwrap();
 //!
--- a/container_api/src/holochain.rs
+++ b/container_api/src/holochain.rs
@@ -21,6 +21,7 @@
 //!        cas::file::FilesystemStorage, eav::file::EavFileStorage,
 //! };
 //! use tempfile::tempdir;
+//! use holochain_core::context::mock_network_config;
 //!
 //! // instantiate a new holochain instance
 //!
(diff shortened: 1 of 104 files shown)


container_api/src/holochain.rs
```diff
@@ -37,7 +37,7 @@
 //!     Arc::new(Mutex::new(SimplePersister::new(file_storage.clone()))),
 //!     file_storage.clone(),
 //!     Arc::new(RwLock::new(EavFileStorage::new(tempdir().unwrap().path().to_str().unwrap().to_string()).unwrap())),
-//!     JsonString::from("{\"backend\": \"mock\"}"),
+//!     JsonString::from("{\"backend_end\": \"MOCK\"}"),
```
neonphog: is this supposed to be `backend_kind`?

net/src/p2p_config.rs
```diff
@@ -0,0 +1,147 @@
...
+        {
+          "cmd":"node",
+          "env":
+          {
+            "N3H_HACK_MODE":"1",
+            "N3H_IPC_SOCKET":"tcp://127.0.0.1:*"
+          }
+        }
+      }
+    }"#;
```
neonphog: Trailing curly braces and spaces after colons tend to be more standard for json formatting:

```rust
 = r#"{
  "backend_kind": "IPC",
  "backend_config": {
    "socketType": "zmq",
    "spawn": {
      "cmd": "node",
      "env": {
        "N3H_HACK_MODE": "1",
        "N3H_IPC_SOCKET": "tcp://127.0.0.1:*"
      }
    }
  }
}"#;
```
https://jsonformatter.org/

test_bin/src/network_config.json
```diff
@@ -0,0 +1,16 @@
...
+    {
+      "cmd":"node",
+      "env":
+      {
+        "N3H_HACK_MODE":"1",
+        "N3H_IPC_SOCKET":"tcp://127.0.0.1:*"
+      }
+    }
+  }
+}
```
neonphog: https://jsonformatter.org/

container_api/src/holochain.rs
ddd-mtl: fixed

net/src/p2p_config.rs
```diff
@@ -0,0 +1,147 @@
...
+// Conversions
+impl FromStr for P2pConfig {
+    type Err = String;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        serde_json::from_str(s).map_err(|err| err.to_string())
+    }
+}
+impl P2pConfig {
+    pub fn as_str(&self) -> String {
+        serde_json::to_string(self).unwrap()
```
neonphog: `expect(...)` here instead of `unwrap()`? Or better yet, use TryFrom instead, so it can return a result?

ddd-mtl: unwrap() is fine since there is no way this can fail since this struct derives from Serialize.
Would you want me to add that as a comment?

net/src/p2p_config.rs
ddd-mtl: fixed

net/src/p2p_config.rs
ddd-mtl: fixed

test_bin/src/network_config.json
ddd-mtl: fixed

net/src/p2p_config.rs
neonphog: oh, cool :+1: - yeah comment would be good, my brain is trained to think "unwrap bad" outside of tests : )

ddd-mtl: Only time unwrap is fine is when code logic makes it impossible to fail, like here :)

net/src/p2p_network.rs
```diff
@@ -2,22 +2,19 @@
...
-use super::{ipc_net_worker::IpcNetWorker, mock_worker::MockWorker};
-
-use serde_json;
+use super::{ipc_net_worker::IpcNetWorker, mock_worker::MockWorker, p2p_config::*};
 
 /// The p2p network instance
 pub struct P2pNetwork {
-    con: NetConnectionThread,
+    connection: NetConnectionThread,
+    // config: P2pConfig,
```
zippy: why commented out?

ddd-mtl: Wasn't sure if it would be useful to keep the config around.

PR #756, opened 2018-12-13
Net Spawn - Minor fixes and more complete test_bin

neonphog:
This should be a quick review. 2 small real fixes. The rest of the change is demo code in test_bin.


neonphog:
> You probably also need to change it_can_convert_peer_connected() since you changed PeerData

good catch, updated

PR #757, opened 2018-12-14
move node container test to circle and remove clean step

thedavidmeister:
## changes

- ~~removes the clean step from publish.js~~
  - ~~as i understand, if rust ever needs a clean something is very wrong somewhere~~
  - ~~`npm` and `yarn` sometimes need cleaning because the node ecosystem doesn't make any guarantees about deterministic builds~~
  - would really like to hear from @Connoropolous on this one because i'm not sure about it
- creates a cache key in circle for nodejs container
  - checksum on both rust and yarn package config and publish.js
  - doesn't do anything because of the clean step, but discussing/fixing is a followup
- as circle caches are immutable, add a version prefix to cache keys to facilitate debugging over time
- remove a few redundant steps in circle
- split tasks into individual steps in circle to make it clearer what is happening
  - the nodejs container already builds as part of app spec test!
  - don't need to run the nodejs container twice, just make it obvious that it does run
- add docker for nix shell as it is required by circleci-cli

## timings

travis before: https://travis-ci.com/holochain/holochain-rust/builds/94762864

image

travis after: https://travis-ci.com/holochain/holochain-rust/builds/94763122

image

travis before: ~70 mins
travis after: ~60 mins
travis diff: -10 mins

circle before: https://circleci.com/gh/holochain/holochain-rust/126

image

circle after: https://circleci.com/gh/holochain/holochain-rust/144

image

circle before: ~10 mins
circle after: ~10 mins
circle diff: no change

total diff: -10 mins

## followups

- can we make node container not rebuild everything from scratch?
- can we run things in parallel?
- can we use nix to build the node container?


Connoropolous:
@thedavidmeister "can we make node container not rebuild everything from scratch?" I think yes, we can make it *not* do that

thedavidmeister:
@zippy yar yar, getting there

thedavidmeister:
the fails in the last commit are also in develop https://circleci.com/gh/holochain/holochain-rust/182

merging

PR #758, opened 2018-12-14
added test for logger

JhonatanHern:


core/src/logger.rs
```diff
@@ -31,3 +26,15 @@ impl Logger for SimpleLogger {
...
+    use crate::{
+        logger::SimpleLogger,
+        logger::Logger
+    };
+    #[test]
+    fn test_logger() {
+        let mut logger_test = SimpleLogger{};
+        logger_test.log("Example Log".to_string());
+    }
+}
```
sphinxc0re: Missing newline

core/src/logger.rs
```diff
@@ -31,3 +26,15 @@ impl Logger for SimpleLogger {
...
+#[cfg(test)]
+pub mod tests {
+    use crate::{
+        logger::SimpleLogger,
+        logger::Logger
+    };
+    #[test]
+    fn test_logger() {
+        let mut logger_test = SimpleLogger{};
+        logger_test.log("Example Log".to_string());
```
zippy: This test doesn't actually assert something so it's more of a smoke test to make sure logging doesn't panic, but it would be nice to have a test that actually tests that asserts that what was logged actually did get logged.  Maybe move the [TestLogger](https://github.com/holochain/holochain-rust/blob/a32b7db17c653f4c31539b2aae33a7c4191cac72/core/src/instance.rs#L327) over here which is probably a better place for it?

PR #759, opened 2018-12-14
improve readability and completeness of hc run docs

Connoropolous:
Note that `hc run` relies on there being a `bundle.json` file in the app root, but that's different than where `hc package` puts it by default, which is in `dist/bundle.json`. This kinda sucks because its muddy, so it'd be better if `hc run` put (and looked for) `bundle.json` in the same place as `hc package`

if that gets changed, then this should also get re-updated.

### DIFF
--- a/cmd/README.md
+++ b/cmd/README.md
@@ -188,8 +188,32 @@ where `test/test.js` is the path of the file.
 You have the flexibility to write tests in quite a variety of ways, open to you to explore.
 
 ## Running your application
-For the purpose of *testing* APIs or prototyping user interfaces, you can run your DNA using `hc run`.
-This will start the application and open a WebSocket on port `8888` or any other port specified by the `-p`/`--port` option. The `-b`/`--package` flag will build your DNA before running it. By default, none of the data your application is putting on the source chain gets persisted. To persist data onto the file system, use the `--persist` flag.
+For the purpose of *testing* APIs or prototyping user interfaces, you can run the DNA of a hApp, from the directory it's contained. The most basic way to do this is to run:
+```shell
+hc run
+```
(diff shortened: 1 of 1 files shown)


cmd/README.md
```diff
@@ -188,8 +188,32 @@ where `test/test.js` is the path of the file.
 You have the flexibility to write tests in quite a variety of ways, open to you to explore.
 
 ## Running your application
-For the purpose of *testing* APIs or prototyping user interfaces, you can run your DNA using `hc run`.
-This will start the application and open a WebSocket on port `8888` or any other port specified by the `-p`/`--port` option. The `-b`/`--package` flag will build your DNA before running it. By default, none of the data your application is putting on the source chain gets persisted. To persist data onto the file system, use the `--persist` flag.
+For the purpose of *testing* APIs or prototyping user interfaces, you can run the DNA of a hApp, from the directory it's contained. The most basic way to do this is to run:
```
maackle: This is an opportunity for us to clean up some of these concepts. What's the natural way to word this? The problem with how it's written is that there is "*the* DNA" for a hApp, where a hApp typically consists of many DNAs.

Is "run *a* DNA" how we want to say it? It's weird because DNA in biology is just "stuff", not a "thing". "run DNA" sounds more natural, but it's talking about stuff, not a thing. Are we going to (have we already) co-opted "DNA" to refer to a thing?

```suggestion
For the purpose of *testing* APIs or prototyping user interfaces, you can run a DNA from the directory it's contained. The most basic way to do this is to run:
```

Connoropolous: good question. Not sure about the bigger picture, but I like this suggestion, merged it

Connoropolous:
@maackle can you re-review?

PR #761, opened 2018-12-14
Context refactoring

lucksus:
## ContextBuilder
```rust
let context = ContextBuilder::new()
            .with_file_storage(temp_path)
            .expect("Filestorage should get instantiated with tempdir")
            .spawn();
```

Most test contexts could be created with just
```rust
let context = ContextBuilder::new().spawn();
```

Used in 
* container_api
* nodejs container

## Different CASes for chain and DHT
Context now has two CASes: `chain_storage` and `dht_storage` which still get set to the same CAS in 
all initialization code but that might change soon.. So follow-up would be to add a `ContextBuilder::with_chain_storage()` and `ContextBuilder::with_dht_storage()` function.

## Real agent IDs
The container code is doing
```rust
let pub_key = KeyBuffer::with_corrected(&agent_config.public_address)?;
context_builder.with_agent(AgentId::new(&agent_config.name, &pub_key));
```
Hence the changes to the container config to include a syntactically correct public address.
```toml
[[agents]]
id = "test agent 1"
name = "Holo Tester 1"
public_address = "HoloTester1-------------------------------------------------------------------------AHi1"
key_file = "holo_tester.key"
```


container_api/src/context_builder.rs
```diff
@@ -0,0 +1,144 @@
...
+use holochain_core_types::{
+    agent::AgentId, cas::storage::ContentAddressableStorage, eav::EntityAttributeValueStorage,
+    error::HolochainError, json::JsonString,
+};
+use holochain_net::p2p_config::P2pConfig;
+use std::sync::{Arc, Mutex, RwLock};
+
+pub struct ContextBuilder {
+    agent_id: Option<AgentId>,
+    //logger: Option<Arc<Mutex<Logger>>>,
```
Connoropolous: what's happening with logger here? Why's it commented out?

container_api/src/context_builder.rs
```diff
@@ -0,0 +1,144 @@
...
+    path::create_path_if_not_exists,
+};
+use holochain_core::{context::Context, logger::SimpleLogger, persister::SimplePersister};
+use holochain_core_types::{
+    agent::AgentId, cas::storage::ContentAddressableStorage, eav::EntityAttributeValueStorage,
+    error::HolochainError, json::JsonString,
+};
+use holochain_net::p2p_config::P2pConfig;
+use std::sync::{Arc, Mutex, RwLock};
+
```
Connoropolous: /// comment to introduce ContextBuilder?

container_api/src/context_builder.rs
```diff
@@ -0,0 +1,144 @@
...
+    //logger: Option<Arc<Mutex<Logger>>>,
+    //persister: Option<Arc<Mutex<Persister>>>,
+    chain_storage: Option<Arc<RwLock<ContentAddressableStorage>>>,
+    dht_storage: Option<Arc<RwLock<ContentAddressableStorage>>>,
+    eav_storage: Option<Arc<RwLock<EntityAttributeValueStorage>>>,
+    network_config: Option<JsonString>,
+}
+
+impl ContextBuilder {
+    pub fn new() -> Self {
```
Connoropolous: /// comment for all these methods too

nodejs_container/native/src/app.rs
```diff
@@ -1,58 +1,29 @@
...
-
-            let context = HolochainContext::new(
-                agent,
-                Arc::new(Mutex::new(NullLogger {})),
-                Arc::new(Mutex::new(SimplePersister::new(file_storage.clone()))),
-                Arc::new(RwLock::new(MemoryStorage::new())),
-                Arc::new(RwLock::new(EavMemoryStorage::new())),
-                mock_network_config(),
-            ).unwrap();
+            let context = ContextBuilder::new().with_agent(agent).spawn();
```
Connoropolous: Love the simplicity here

container_api/src/container.rs
```diff
@@ -264,21 +253,22 @@ pub fn instantiate_from_config(
...
-                }
-            }?;
+            let pub_key = KeyBuffer::with_corrected(&agent_config.public_address)?;
+            context_builder.with_agent(AgentId::new(&agent_config.name, &pub_key));
+
+            instance_config.network.map(|network_config| {
+                context_builder.with_network_config(JsonString::from(network_config))
+            });
+
+            if let StorageConfiguration::File { path } = instance_config.storage {
```
zippy: where did the Memory context go?

container_api/src/context_builder.rs
```diff
@@ -0,0 +1,165 @@
...
+        ContextBuilder {
+            agent_id: None,
+            chain_storage: None,
+            dht_storage: None,
+            eav_storage: None,
+            network_config: None,
+        }
+    }
+
+    /// Sets the agent of the context that gets build.
```
Connoropolous:
```suggestion
    /// Sets the agent of the context that gets built.
```

container_api/src/context_builder.rs
```diff
@@ -0,0 +1,165 @@
...
+        let eav = Arc::new(RwLock::new(EavMemoryStorage::new()));
+        self.chain_storage = Some(cas.clone());
+        self.dht_storage = Some(cas);
+        self.eav_storage = Some(eav);
+        self
+    }
+
+    /// Sets all three storages, chain, DHT and EAV storage, to persistent file based implementations.
+    /// Chain and DHT storages get set to the same file CAS.
+    /// Returns an error if no file storage could spawned on the given path.
```
Connoropolous:
```suggestion
    /// Returns an error if no file storage could be spawned on the given path.
```

container_api/src/container.rs
lucksus: That's the default that `ContextBuilder` sets the context to if you just do a `ContextBuilder::new().spawn()`.

PR #762, opened 2018-12-14
WIP: capabilities actually work

zippy:
- [x] convert [CapabilityType to public/transferrable/assigned](https://realtimeboard.com/app/board/o9J_kyiXmFs=/?moveToWidget=3074457346360211001)
- [ ] add system cap entry types
- [ ] cap entries created on bridging
- [ ] caps checked in `reduce_call`
- [ ] examples in app_spec


zippy:
yes but probably not in this PR...

core/src/workflows/author_entry.rs
```diff
@@ -31,9 +31,14 @@ pub async fn author_entry<'a>(
     // 2. Validate the entry
     await!(validate_entry(entry.clone(), validation_data, &context))?;
     // 3. Commit the entry
-    await!(commit_entry(entry.clone(), maybe_crud_link, &context))?;
+    let addr = await!(commit_entry(entry.clone(), maybe_crud_link, &context))?;
     // 4. Publish the valid entry to DHT. This will call Hold to itself
-    await!(publish(entry.address(), &context))
+    //TODO: missing a general public/private sharing check here, for now just manually
+    // forcing CapTokenGrant to private
+    if entry.entry_type() != EntryType::CapTokenGrant {
```
lucksus: There is already a function `EntryType::can_publish()` where currently it just distinguishes `EntryType::Dna` from everything else (see https://github.com/holochain/holochain-rust/blob/develop/core_types/src/entry/entry_type.rs#L91). At least this check should go there. When I said that this is is broken I meant for App entries where we actually have to check the DNA to see how it is defined. I had left some code commented out there that was used in the DHT reducer before adding the network.

zippy: fixed

core/src/nucleus/mod.rs
```diff
@@ -49,17 +54,31 @@ impl ZomeFnCall {
...
+        match self.cap.clone() {
+            Some(call) => call.cap_name,
+            None => panic!("null cap call unimplemented!"),
+        }
+    }
+
+    pub fn cap_token(&self) -> Address {
+        match self.cap.clone() {
+            Some(call) => call.cap_token,
+            None => panic!("null cap call unimplemented!"),
```
lucksus: Wondering if these getters should return an Option and panic should happen in the context that uses `ZomeFnCall`?

zippy: All the places that use that are going to change and disappear, so though I agree theoritically, I don't think I want to bother to fix it right now.  Will do so in next PR.

core/src/nucleus/ribosome/api/call.rs
```diff
@@ -112,38 +114,22 @@ pub(crate) fn reduce_call(
         return;
     }
     let dna = state.dna.clone().unwrap();
-    let maybe_cap = get_capability_with_zome_call(&dna, &fn_call);
-    if let Err(fn_res) = maybe_cap {
+    let maybe_public = is_fn_public(&dna, &fn_call);
```
lucksus: Hm, I'm not sure if this check is actually correct here.
`Action::Call` is only used for calls that originate from a local `hdk::call`. So this is how we call zome functions of other zomes.

`Action::ExecuteZomeFunction` is what facilitates calls coming in from the outside..

core/src/nucleus/ribosome/api/call.rs
```diff
@@ -112,38 +114,22 @@ pub(crate) fn reduce_call(
...
-            // TODO #301 - check if caller has Agent Capability
-            false
-        }
-        Membrane::ApiKey => {
-            // TODO #301 - check if caller has ApiKey Capability
-            false
-        }
-    };
-    if !can_call {
+    if !public && !check_capability(context.clone(), &fn_call.clone()) {
```
lucksus: Ah I guess my comment above gets resolved by this: guess `check_cabapility` will be true for a local call?

core/src/nucleus/ribosome/api/call.rs
zippy: ok refactored the code that validates the call in both those actions..

core/src/nucleus/mod.rs
```diff
@@ -289,77 +289,26 @@ fn reduce_execute_zome_function(
...
 
-    // Get DNA
-    let dna = match state.dna {
-        None => {
-            dispatch_error_result(
-                context.action_channel(),
-                &fn_call,
-                HolochainError::DnaMissing,
-            );
+    // 1. Validate that the call (a number of tings could go wrong)
```
lucksus: that the call... ?

core/src/nucleus/mod.rs
```diff
@@ -49,17 +54,31 @@ impl ZomeFnCall {
...
+            cap: cap,
             fn_name: function.to_string(),
             parameters: parameters.into(),
         }
     }
 
     pub fn same_fn_as(&self, fn_call: &ZomeFnCall) -> bool {
         self.zome_name == fn_call.zome_name
-            && self.cap_name == fn_call.cap_name
+            && self.cap == fn_call.cap
```
maackle: Just checking if this is really the right equality check, I don't know enough to say if it is or not.

core_types/src/entry/cap_entries.rs
```diff
@@ -0,0 +1,250 @@
...
+}
+
+/// System entry to hold a capabilities granted by the callee
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, DefaultJson)]
+pub struct CapTokenGrant {
+    assignees: Option<Vec<Address>>,
+}
+
+impl CapTokenGrant {
+    pub fn new(assignees: Option<Vec<Address>>) -> Self {
```
maackle: Does this want to be private then, since it looks like the proper constructor is `create`? Maybe even rename `create` to `new`? (Only the tests are using `new` now.)

PR #765, opened 2018-12-15
NodeJS container using container_api

maackle:


app_spec/test/test.js
```diff
@@ -1,28 +1,47 @@
...
+const dnaPath = "./dist/app_spec.hcpkg";
+const dna = ConfigBuilder.dna(dnaPath);
+
+const instanceAlice = ConfigBuilder.instance(agentAlice, dna);
+const instanceBob = ConfigBuilder.instance(agentBob, dna);
+
+const config = ConfigBuilder.habitat(instanceAlice, instanceBob);
+
+const hab = new Habitat(config);
+hab.start();
```
maackle: Please take a good look at the config building up to this point, this is likely how JS test files will all begin. Let's get this syntax right.

app_spec/test/test.js
```diff
@@ -1,28 +1,47 @@
...
+const caller = id => (zome, cap, fn, params) => (
+  hab.call(id, zome, cap, fn, params)
+)
+
+const app = {
+  call: caller('alice-' + dnaPath)
+}
+const app2 = {
+  call: caller('bob-' + dnaPath)
+}
```
maackle: These are just here to replicate the original behavior of having two objects you can run `call` on to trigger calls on different instances.

BTW nothing has changed after this point, it is all StandardJS fixes applied by my editor. Skip to the next file.

container_api/src/container.rs
```diff
@@ -302,10 +296,10 @@ impl Logger for NullLogger {
 }
 
 fn create_memory_context(
-    _: &String,
+    agent_name: &String,
     network_config: JsonString,
 ) -> Result<Context, HolochainError> {
-    let agent = AgentId::generate_fake("c+bob");
+    let agent = AgentId::generate_fake(agent_name);
```
maackle: Hooking up this param actually made all the difference when one of the JS tests was hanging. Does that make sense?

nodejs_container/index.js
```diff
@@ -4,52 +4,25 @@ const fs = require('fs');
...
+    const rawResult = this.callRaw(id, zome, trait, fn, stringInput);
+    let result;
+    try {
+        result = JSON.parse(rawResult);
+    } catch (e) {
+        console.log("JSON.parse failed to parse the result. The raw value is: ", rawResult);
+        result = { error: "JSON.parse failed to parse the result", rawResult };
+    }
+    return result;
+}
```
maackle: This is a replacement for the old `_call` / `call` switcheroo, just renamed `_call` to `callRaw`

nodejs_container/index.js
```diff
@@ -4,52 +4,25 @@ const fs = require('fs');
...
 // deals with ensuring the correct version for the machine/node version
-const binding_path = binary.find(path.resolve(path.join(__dirname,'./package.json')));
-
-const HolochainApp = require(binding_path).HolochainApp;
+const binding_path = binary.find(path.resolve(path.join(__dirname, './package.json')));
+
+const { ConfigBuilder, Habitat } = require(binding_path);
+
+Habitat.prototype.callRaw = Habitat.prototype.call
+Habitat.prototype.call = function (id, zome, trait, fn, params) {
```
maackle: I'm playing around with what it feels like to call a Container a Habitat. We can change this before merge. (Or not :smiling_imp: )

container_api/src/container.rs
Connoropolous: So, wait... we were passing in agent ids, but it wasn't using them , it was just hardcoded in here?

lucksus: I've replaced this function altogether by the `ContextBuilder` in https://github.com/holochain/holochain-rust/pull/761. Just FYI.

nodejs_container/native/src/config.rs
```diff
@@ -0,0 +1,137 @@
...
+#[derive(Serialize, Deserialize)]
+pub struct DnaData {
+    pub path: PathBuf,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct InstanceData {
+    pub agent: AgentData,
+    pub dna: DnaData,
+}
```
lucksus: Why are you using these stripped down copies instead of exposing the types in container_api/config?

lucksus:
Thinking about above statement: it might not be too bad for a monolithic Electron bundle which this first release would be anyway since there would be only one UI interface anyway and no way to install any other apps etc..

nodejs_container/native/src/config.rs
maackle: @lucksus the idea was to make it as easy as possible to create a scenario test (e.g. no need to pick and match IDs), but I think you're right and I should just expose the Configuration types and recreate the ConfigBuilder in pure JS. I can also expose the config-from-toml-string functionality to make it fully functional.

maackle:
> Interesting change. What about the use case of stopping one instance, while another stays running?

@Connoropolous Yeah good point, that will need to be added. But we won't need that for a while / until we are running an actual production service using this nodejs container.

maackle:
> I'm starting to feel ok with that name.. ;) (still like "Receptor" more)

@lucksus I won't push a new name through without a discussion. This might not be the PR to make a sweeping name change, I wouldn't want to call it Container in one place and Habitat in another.

Speaking of "Receptor", does it fit the meaning of the ceptr term? If so why don't we use that!

nodejs_container/native/src/config.rs
maackle: I just tried exposing the types (`AgentConfiguration`, `DnaConfiguration`, etc) and realized it wouldn't be very helpful, because the exposed `Habitat` constructor expects a JS object anyway, not special opaque rust-backed types. So if the user wants to build a config by hand they have to build an object that matches that shape anyway.

I also made Habitat able to accept a string of TOML config.

lucksus:
> Speaking of "Receptor", does it fit the meaning of the ceptr term? If so why don't we use that!

@artbrock suggested that name in Montreal. I think he was seeing similarities between our container and what the receptor is in ceptr.. and the idea was that it would prepare folks for ceptr terminology and concepts. @zippy had problems with that. I keep bringing it up though ;) I like that it is something completely custom.

zippy:
nice, CI passing!

maackle:
@zippy btw, when I added a logging statement it broke several tests (which were testing for presence of certain log messages).

PR #766, opened 2018-12-17
Changed parity len from 2 to 5

ddd-mtl:


core_types/src/agent.rs
```diff
@@ -73,16 +76,17 @@ impl AgentId {
     /// `s` will be used for the `nick` and included in the key string as well
     /// this agent is not cryptographically generated...
     /// it will not be able to sign / encrypt anything
-    pub fn generate_fake(s: &str) -> Self {
-        let mut buf = s.to_string();
-        while buf.len() < 84 {
+    pub fn generate_fake(nick: &str) -> Self {
```
sphinxc0re: This method obviously won't be used in production, so I think it would be good to add a
`#[cfg(test)]` attribute or don't event define it inside the `AgentId` impl

lucksus: But because we don't have ways yet to handle real agent keys this is only way to create a syntactically correct agent id object. It is used in the container and I just used it in the the ContextBuilder: https://github.com/holochain/holochain-rust/pull/761

I would suggest to leave at as it is for now and do what you suggest once we have reduced its usage to just testing..

zippy: agreed @lucksus

core_types/src/agent.rs
```diff
@@ -7,53 +7,56 @@ use crate::{
...
     pub fn with_corrected(s: &str) -> Result<KeyBuffer, HolochainError> {
         let s = s.replace("-", "+").replace("_", "/");
-        let s = base64::decode(&s)?;
-        let dec = Decoder::new(PARITY_LEN);
-        let dec = *dec.correct(s.as_slice(), None)?;
-        Ok(KeyBuffer::with_raw(array_ref![dec, 0, 64]))
+        let base64 = base64::decode(&s)?;
+        let dec = Decoder::new(KeyBuffer::PARITY_LEN);
+        let dec = *dec.correct(base64.as_slice(), None)?;
+        Ok(KeyBuffer::with_raw(array_ref![dec, 0, KeyBuffer::KEY_LEN]))
```
neonphog: Reed-Solomon will erroneously correct things if it uses up all its parity confidence. @pjkundert can give further details, but I think we need to add something like

```rust
const CORRECTION_BYTE_LIMIT = 2
```

and return a HolochainError if > this number of bytes are altered compared to the input string

PR #773, opened 2018-12-17
Handle bridge dependencies with graph theory

lucksus:
Setting up the interfaces for bridges gets a lot easier when we can assert that a bridge dependency (i.e. the *callee* of the bridge) is already instantiated when creating an interface to pass in as context when creating the *caller*.

This needs to happen on creation because we are going to need to add a genesis callback that uses the bridge (-> and thus the callee) during initialization.

With this new function `Configuration::instance_ids_sorted_by_bridge_dependencies()` I get all instances sorted in a way that I can iterate through to spawn them and set up bridges at the same time. -> next PR

### DIFF
--- a/container_api/src/config.rs
+++ b/container_api/src/config.rs
@@ -179,7 +179,7 @@ impl Configuration {
             HolochainError::ConfigError("Cyclic dependency in bridge configuration".to_string())
         })?;
 
-        // REVERSE order because we want to the instance with NO dependencies first
+        // REVERSE order because we want to get the instance with NO dependencies first
         // since that is the instance we should spawn first.
         sorted_nodes.reverse();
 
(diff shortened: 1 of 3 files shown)


container_api/src/config.rs
```diff
@@ -542,4 +629,137 @@ pub mod tests {
...
+    dna = "app spec rust"
+    agent = "test agent"
+    [instances.logger]
+    type = "simple"
+    file = "app_spec.log"
+    [instances.storage]
+    type = "file"
+    path = "app_spec_storage"
+
+    {}
```
maackle: what's this?

lucksus: This whole string block is in a `format!`, so this is where the parameter gets inserted. The following tests just change the defined bridges so I wanted to dry these tests out a bit at least.. could be more..

PR #778, opened 2018-12-19
[app-spec] New validation proposal

sphinxc0re:
This is proposal for a new way of dealing with validations. Since I'm unsure of what the requirements are for the `ValidationData<T>` enum, here is a template:

```rust
pub enum ValidationData<T> {
    Create {
        entry: T,
        ...
    },
    Modify { ... },
    Delete { ... },
}
```
Because every action requires a different set of information for the agent, validating it.


thedavidmeister:
@sphinxc0re why do we need a `T`? can't it be `Entry`?

sphinxc0re:
@thedavidmeister I'd like to ask the inverse question in return: Why should it be `Entry`? Can't it be `T`? I thought that happs should only validate app-entrys.

thedavidmeister:
@sphinxc0re still need both the `AppEntryType` and `AppEntryValue` so it should be an `Entry`, which is the thing that bundles those two together

we explicitly allow, and hApps rely on, using the same structs for `AppEntryValue` across several `AppEntryType`s

see https://github.com/holochain/holochain-rust/pull/741 for last time we tried to avoid `Entry` in hApps context

app_spec/zomes/blog/code/src/post.rs
```diff
@@ -53,9 +56,22 @@ pub fn definition() -> ValidatingEntryType {
             hdk::ValidationPackageDefinition::ChainFull
         },
 
-        validation: |post: crate::post::Post, _ctx: hdk::ValidationData| {
-            (post.content.len() < 280)
-                .ok_or_else(|| String::from("Content too long"))
+        validation: |data: ValidationData<Post>| {
```
lucksus: As discussed in person with half of the team right now, we would like to get rid of the app specific type here (`Post`) since it's defined up as the `native_type` which is currently not used.

lucksus:
Closing this PR as this got implemented already through a different PR.

PR #779, opened 2018-12-19
Bridge call roundtrip

lucksus:
## Context
This is basic bridging. But no checks and no capability tokens. If the bridge is set in the config the caller can call any zome function of the callee.

screenshot 2018-12-19 at 22 52 45

## `hdk::call` for bridge calls
Added instance handle to `hdk::call()`, which can also be `hdk::THIS_INSTANCE` to call other local zomes.

```rust
hdk::call("test-callee", "greeter", "public", "token", "hello", JsonString::from("{}"))
```

## implement bridged call via `Context::container_api`
core/src/nucleus/ribosome/api/call.rs:
```rust
    let handler = container_api.write().unwrap();

    let id = ProcessUniqueId::new();
    let request = format!(
        r#"{{"jsonrpc": "2.0", "method": "{}", "params": {}, "id": "{}"}}"#,
        method, input.fn_args, id
    );

    let response = handler
        .handle_request_sync(&request)
        .ok_or("Bridge call failed".to_string())?;
```

## Bridge roundtrip test in container_api crate
The test container config sets up a *caller* and a *callee* instance. The callee is created completely in the tests section in container.rs with WAT code. In there is a function "hello" that returns the string "Holo World" (set as ascii in WAT). The caller is the new WASM Rust project "test-bridge-caller" so I can use `hdk::call`. Test shows that "Holo World" arrives at the caller side.


container_api/src/container.rs
```diff
@@ -397,14 +413,14 @@ pub mod tests {
...
+    caller_id = "bridge-caller"
     callee_id = "test-instance-2"
     handle = "happ-store"
 
     [[bridges]]
-    caller_id = "test-instance-3"
+    caller_id = "bridge-caller"
     callee_id = "test-instance-1"
-    handle = "DPKI"
+    handle = "test-callee"
```
lucksus: This handle is the "instance_handle" that the bridge caller has to provide to `hdk::call` to reference this bridged instance.

container_api/test-bridge-caller/src/lib.rs
```diff
@@ -0,0 +1,29 @@
+#[macro_use]
+extern crate hdk;
+extern crate serde;
+#[macro_use]
+extern crate serde_derive;
+use hdk::holochain_core_types::json::JsonString;
+
+
+fn handle_call_bridge() -> JsonString {
+    hdk::call("test-callee", "greeter", "public", "token", "hello", JsonString::from("{}")).unwrap()
```
lucksus: "test-callee" is the handle set in the container config for this bridge. This actually has to be defined in the DNA but currently the container does not enforce these handles to match up (next PR). Then we will also have a way of defining the needed bridges in the define_zome! macro below. A bridge handle would then be like the declaration of binding: if declared in the zome as required, zome code can expect a call to such a bridge to work because the container would not load a config in which a matching bridge definition was missing..

Connoropolous: At this point, it kind of looks like hdk::call should accept a struct maybe, instead of so many params

wasm_utils/src/api_serialization/call.rs
```diff
@@ -1,8 +1,11 @@
 use holochain_core_types::{dna::capabilities::CapabilityCall, error::HolochainError, json::*};
 
+pub const THIS_INSTANCE: &str = "__hdk_this_instance";
```
Connoropolous: Are you open to finding a different name for "THIS_INSTANCE"?

lucksus: Sure. Do you have a suggestion?

container_api/test-bridge-caller/src/lib.rs
lucksus: yep, though it will lose at least one again soon when the capability stuff is completely switched to tokens..

container_api/src/container.rs
```diff
@@ -556,4 +572,155 @@ pub mod tests {
...
+                    Address::from("fake_token"),
+                    None,
+                )),
+                "call_bridge",
+                "{}",
+            )
+            .unwrap();
+
+        // "Holo World" comes for the callee_wat above which runs in the callee instance
+        assert_eq!(result, JsonString::from(RawString::from("Holo World")));
```
zippy: woot woot!!!

container_api/test-bridge-caller/src/lib.rs
zippy: exactly.  also the capability name is in the current call and that will go away too.

wasm_utils/src/api_serialization/call.rs
Connoropolous: What if the argument was an option instead, meaning you could pass None in the usual call case, and only pass Some in the bridging call case?

lucksus: I would prefer something more explicit. Totally not attached to THIS_INSTANCE if you have a better name.

Connoropolous: have given it some thought, but not landed on anything, so ¯\_(ツ)_/¯ carry on

PR #780, opened 2018-12-19
Remove Link

AshantiMutinta:
This pull request makes it able to remove link from local 

The files changed are this number because my cargo fmt adds windows line endings, so error on my part. Should be harmless!


cas_implementations/src/eav/file.rs
```diff
@@ -90,6 +90,24 @@ impl EavFileStorage {
         Ok(())
     }
 
+    fn delete_file(
```
Connoropolous: Am I incorrect in thinking that the EAV was append only? 
@lucksus

AshantiMutinta: We already synced in heartbeat and currently implementing eavindexing before this can get in

Connoropolous: ah, I see. Sorry to interrupt!

Do you mean that you are currently switched to implementing indexing the EAV before proceeding further with this PR?

AshantiMutinta: No problem! Yes :) Deleting will simply be changing the hash to a new one which does not contain the data! Working on that right now

Connoropolous:
@AshantiMutinta for my documentation work, I'm trying to make sense of 
https://github.com/holochain/holochain-rust/blob/develop/core_types/src/link/link_add.rs
would you take a look at it, in relation to the work that you're doing? And let me know if you can make sense of how they relate? 
It looks like LinkAdd has an action_kind, which doesn't make sense to me

app_spec/test/test.js
```diff
@@ -57,6 +57,19 @@ scenario1.runTape('create_post', async (t, { alice }) => {
   t.equal(result.Ok, "QmY6MfiuhHnQ1kg7RwNZJNUQhwDxTFL45AAPnpJMNPEoxk")
 })
 
+
+scenario1.runTape('delete_post', async (t, { alice }) => {
+
+  const content = "Holo world"
+  const in_reply_to = null
+  const params = { content, in_reply_to }
+  const result = alice.call("blog", "create_post", params)
```
lucksus: I think what you want to do here is:
```suggestion
  const result = await alice.callSync("blog", "create_post", params)
```

But doing that below on `delete_post` might mean that the `waiter` needs to learn about `LinkRemove` action. Right @maackle?

AshantiMutinta: Awesome! I will implement that! :) Thank you!

Connoropolous:
Uh, @AshantiMutinta 2500 changed files :p You checked in the node_modules?

core_types/src/eav.rs
```diff
@@ -173,6 +195,26 @@ impl EntityAttributeValueIndex {
...
         e.map(|a| a == eav).unwrap_or(true)
     }
+
+    /// this is a predicate for matching on eav values. Useful for reducing duplicated filtered code.
+    pub fn filter_on_eav_with_prefix<'a>(
+        eav: &'a String,
+        e: Option<&'a String>,
+        index_query: &'a IndexQuery<'a>,
+    ) -> bool {
+        let prefixes = if index_query.prefixes().len() > 0 {
```
sphinxc0re:
```suggestion
        let prefixes = if !index_query.prefixes().is_empty() {
```

core/src/nucleus/ribosome/callback/validation_package.rs
```diff
@@ -88,6 +88,44 @@ pub fn get_validation_package_definition(
...
+            let call = ZomeFnCall::new("", None, "__hdk_get_validation_package_for_link", params);
+
+            ribosome::run_dna(
+                &dna.name.clone(),
+                context,
+                wasm.code.clone(),
+                &call,
+                Some(call.parameters.into_bytes()),
+            )?
+        }
```
lucksus: Except for the first match, this block is exactly the same as above for `LinkAdd`, right? Can we DRY this out?

core/src/workflows/remove_link.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    context.log(format!("debug/workflow/remove_link: {:?}", link));
+    // 1. Get validation package from source
+    context.log(format!(
+        "debug/workflow/remove_link: getting validation package..."
+    ));
+    let maybe_validation_package = await!(get_validation_package(header.clone(), &context))?;
+    let validation_package = maybe_validation_package
+        .ok_or("Could not get validation package from source".to_string())?;
+    context.log(format!("debug/workflow/hold_link: got validation package!"));
```
lucksus:
```suggestion
    context.log(format!("debug/workflow/remove_link: got validation package!"));
```

core/src/workflows/remove_link.rs
```diff
@@ -0,0 +1,55 @@
...
+    // 3. Validate the entry
+    context.log(format!("debug/workflow/hold_link: validate..."));
+    await!(validate_entry(entry.clone(), validation_data, &context)).map_err(|err| {
+        context.log(format!("debug/workflow/hold_link: invalid! {:?}", err));
+        err
+    })?;
+    context.log(format!("debug/workflow/hold_link: is valid!"));
+
+    // 3. If valid store the entry in the local DHT shard
+    await!(remove_link(&link, &context))?;
```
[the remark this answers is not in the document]
lucksus: Ok, since this workflow is now used on the receiving end (of a DhtStoreMeta), this line makes sense here. 👍

core/src/dht/dht_store.rs
```diff
@@ -58,12 +58,16 @@ impl DhtStore {
...
+        let index_query = IndexQuery::new_only_prefixes(vec!["link__", "removed_link__"]);
+        let filtered =
+            self.meta_storage
+                .read()?
+                .fetch_eavi(Some(address), Some(tag), None, index_query)?;
+
+        Ok(filtered
+            .into_iter()
+            .filter(|eav| eav.attribute().starts_with("link__"))
+            .collect())
```
lucksus: I don't understand why this does or should work. Seems like you are just ignoring the removals.. ??

core/src/nucleus/ribosome/callback/validation_package.rs
AshantiMutinta: We could but I think this could be a follow up! I had some ideas about how we could DRY out the whole workflow

app_spec/zomes/blog/code/src/lib.rs
```diff
@@ -66,9 +66,15 @@ define_zome! {
...
+            inputs: |content: String|,
+            outputs: |result: ZomeApiResult<Address>|,
+            handler: blog::handle_delete_post
+        }
+
+        delete_entry_post: {
             inputs: |post_address: Address|,
             outputs: |result: ZomeApiResult<()>|,
-            handler: blog::handle_delete_post
+            handler: blog::handle_entry_delete_post
```
maackle:
```suggestion
            handler: blog::handle_delete_entry_post
```

app_spec/zomes/blog/code/src/blog.rs
```diff
@@ -137,7 +144,7 @@ pub fn handle_get_post(post_address: Address) -> ZomeApiResult<Option<Entry>> {
     hdk::get_entry(&post_address)
 }
 
-pub fn handle_delete_post(post_address: Address) -> ZomeApiResult<()> {
+pub fn handle_entry_delete_post(post_address: Address) -> ZomeApiResult<()> {
```
maackle:
```suggestion
pub fn handle_delete_entry_post(post_address: Address) -> ZomeApiResult<()> {
```

core_types/src/eav.rs
```diff
@@ -62,12 +62,13 @@ impl Ord for EntityAttributeValueIndex {
 }
 
 #[derive(Clone, Debug)]
-pub struct IndexQuery {
+pub struct IndexQuery<'a> {
     start: Option<i64>,
     end: Option<i64>,
+    prefixes: Vec<&'a str>,
```
maackle: Does this really belong in the `IndexQuery`, since it refers to the Attribute? Seems like we should have an `AttributeQuery` as well. Also, I'm going to have to rewrite before merging in #952 because we won't be referring to attributes by raw strings anymore, but enum variants. I don't mind doing that but just pointing out that this will have to change.

Perhaps `AttributeQuery` could just be something like:

```rust
enum AttributeQuery {
    Single(Attribute),
    Predicate(Box<FnOnce(Attribute) -> bool>),
}
```

core/src/action.rs
```diff
@@ -100,6 +100,9 @@ pub enum Action {
     /// Does not validate, assumes link is valid.
     AddLink(Link),
 
+    //Removes a link for the local DHT
+    RemoveLink(Link),
```
maackle: Last I heard about removing links (a few weeks ago), I thought it had to reference the hash of a previously issued `AddLink`. If so, shouldn't that be part of the action?

core_types/src/eav.rs
AshantiMutinta: That is definitely a new addition but I think this could be done as a follow up, I kind of don't mind keeping it in the index_query at this point because it is small enough

core/src/action.rs
maackle: Per heartbeat, we decided we don't need this for closed alpha and can go forward with this implementation

core_types/src/eav.rs
```diff
@@ -252,37 +294,33 @@ impl EntityAttributeValueStorage for ExampleEntityAttributeValueStorage {
...
-                    .map(|start| {
-                        println!("start {:?}", e.index().clone());
-                        println!("start condition {:?}", start.clone() <= e.index());
-                        start <= e.index()
-                    })
+                    .map(|start| start <= e.index())
                     .unwrap_or_else(|| {
-                        let latest = get_latest(e.clone(), map.clone())
+                        let latest = get_latest(e.clone(), map.clone(), index_query.clone())
                             .unwrap_or(EntityAttributeValueIndex::default());
```
sphinxc0re: You can just write `unwrap_or_default()`

core/src/network/handler/store.rs
```diff
@@ -39,6 +42,23 @@ pub fn handle_store_meta(dht_meta_data: DhtMetaData, context: Arc<Context>) {
...
+            // TODO: do a loop on content once links properly implemented
+            assert_eq!(dht_meta_data.content_list.len(), 1);
+            let entry_with_header: EntryWithHeader = serde_json::from_str(
+                //should be careful doing slice access, it might panic
+                &serde_json::to_string(&dht_meta_data.content_list[0])
+                    .expect("dht_meta_data should be EntryWithHader"),
+            )
+            .expect("dht_meta_data should be EntryWithHader");
+            thread::spawn(move || {
+                match context.block_on(remove_link_workflow(&entry_with_header, &context.clone())) {
```
sphinxc0re: You could make this much cleaner by using an `if let` expression

AshantiMutinta: Agreed! I think I found it structured like this and just added to it. In imo it would be a good candidate for a refactor since we are trying to get this story in for the closed alpha

maackle: Then again it's not so hard to refactor a little as we go when we see easy improvements that would take almost no time...

core/src/nucleus/ribosome/api/mod.rs
```diff
@@ -489,6 +497,7 @@ pub mod tests {
             ("hc_entry_address", 12),
             ("hc_send", 13),
             ("hc_sleep", 14),
+            ("hc_remove_link", 15),
```
maackle: Not suggesting a change rn, but I wonder if `hdk::remove_link` shouldn't be called `hdk::unlink_entries` to match better with `hdk::link_entries`?

AshantiMutinta: I think the "hc_" prefix comes from the hdk names rather than the functions that are being used internally? I could be wrong tho

core/src/network/handler/store.rs
AshantiMutinta: You are right, let me change that

core/src/nucleus/ribosome/api/mod.rs
AshantiMutinta: ooops didn't read that fully. I think it is definitely something to look into!

app_spec/test/test.js
```diff
@@ -79,27 +79,56 @@ scenario1.runTape('create_post', async (t, { alice }) => {
...
-  t.ok(createResult.Ok)
+  //create post
+ const alice_create_post_result = await alice.callSync("blog", "create_post",
+    { "content": "Posty", "in_reply_to": "" }
+  )
 
-  const deletionParams = { post_address: createResult.Ok }
-  const deletionResult = alice.call("blog", "delete_post", deletionParams)
+  
+  const bob_create_post_result = await alice.callSync("blog", "posts_by_agent",
```
lucksus: This should be
```suggestion
  const bob_create_post_result = await bob.callSync("blog", "posts_by_agent",
```
now since it's even calling `posts_by_agent`, right?

core/src/dht/dht_store.rs
lucksus: BTW, we clarified this per voice-chat yesterday. So `fetch_eavi` only returns the latest version of potentially multiple EAVs that match the query. With the prefixes set in line 61, all EAVs where the attribute matches either `link__<tag>` or `removed_link_<tag>` are considered the same. If we get the latest of these each, it will be either a `link__<tag>` or `removed_link_<tag>`, depending on if the last action (for this specific link!) was an add or removal. So yes, if remove all those links for which the latest action is a removal we get exactly what we want.

And, this could get more documented and tested :)

app_spec/test/test.js
AshantiMutinta: yes! Missed that part thank you!

core/src/dht/dht_reducers.rs
```diff
@@ -147,6 +148,43 @@ pub(crate) fn reduce_add_link(
...
+    // Get Action's input data
+    let action = action_wrapper.action();
+    let link = unwrap_to!(action => Action::RemoveLink);
+    let mut new_store = (*old_store).clone();
+    let storage = &old_store.content_storage().clone();
+    if !(*storage.read().unwrap()).contains(link.base()).unwrap() {
+        new_store.actions_mut().insert(
+            action_wrapper.clone(),
+            Err(HolochainError::ErrorGeneric(String::from(
+                "Base for link not found for remov",
```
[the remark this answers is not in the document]
zippy: lets fix this typo!

core/src/nucleus/ribosome/api/mod.rs
zippy: Department of Names Department List of Names That Need To Be Considered:  DNDLNTNTBC!

AshantiMutinta:
> e it marked as approved, but @AshantiMutinta please do fix the one typo. And also I'm wondering why there are read/excecute changes for the various scripts/install/*.sh files in this PR. They really should be being changed.

Hello! I think I am having some trouble with cargo fmt issues with it affecting my line endings. I will use notepad to remove them, also thanks, typo has been fixed

PR #781, opened 2018-12-19
Add regex-based "glob" pattern matching to hdk::query

pjkundert:
NOTE: Re-submitted pull request using local branch in holochain repo.

o Change the Holochain query API to accept a `Vec<String>` vs. `String`
o If multiple EntryType names (or one with "glob" wildcards) is used,
  then produce a GlobSetBuilder matcher to recognize the selected names.

There is no way to get more than a single EntryType via hdk::query.

The primary purpose of this change is to support querying multiple EntryType names in a single hdk::query request.  To do this efficiently, a single regex encompassing all of the desired EntryType names is produced.

The globset::GlobSetBuilder is used to do this, and also supports the full "glob" pattern matching on EntryType names.  This includes the concept of path-like namespaces for EntryType names.  For example, if you have 3 EntryTypes `"one"`, `"two"` and `"three"` related to the concept "tx" in your Holochain application, you can name them `"tx/one"`, `"tx/two"` and `"tx/three"`.  Then, when you want to query all of your tx-related entries from your local chain, you'd call

```
hdk::query( "tx/*".into() )
```

Later, when you add a fourth "tx/four" EntryType, the above code would automatically get all the tx-related EntryTypes.

You can also do things like get all EntryTypes except the System EntryTypes: `"[!%]*"`.

NOTE: Changes the hdk::query API!


pjkundert:
OK, TODOs ticketed.  Should be good to do!

PR #782, opened 2018-12-19
Check currently installed version of Rust, offer to un/re-install

pjkundert:
NOTE: Re-submitted pull request using local branch in holochain repo (was: #686).

o Properly export PATH (in case this is the first installation of Rust)
o Use target-specific variables to select TOOLS/CORE_RUST_VERSION

If the wrong version of Rust is installed, we can check and offer to replace it with
one compatible with holochain-rust, using `rustup override ...` (ie. for this project only).


pjkundert:
I think this upgrade to the Makefile is worth considering for inclusion in 0.0.3. It doesn't pose much risk, as it uses `rustup override ... ` to change Rust toolchain defaults for holochain-rust only, instead of `rustup default ...` (used currently in Makefile in various places).  Getting the Rust toolchain version wrong is a pretty major problem for novice builders of holochain-rust.

Prompts user *only* if both stdin and stdout are connected to a TTY; otherwise proceeds automatically to select correct Rust toolchain version; this retains the ability for a power-user to operate using a different Rust toolchain.

It also provides a very handy `test-%` target (eg. `make test-query`), to run all cargo unit tests matching `query`.  I've further tidied up the output to avoid spurious test output clutter, added a 'help' target.

pjkundert:
I agree with all of these observations, @thedavidmeister!  This Makefile is not ideal, but this commit doesn't address the structural deficiencies, it just removes the broken version checking.  Since I don't know where the Makefile is currently being used, I wasn't confident to remove any targets.

It is critical to have a working Makefile, though, as the traditional `./configure; make; make install` build method is still generally expected to work (by the graybeards, anyway).  We can move on to fixing the targets, etc. in other pull requests.

thedavidmeister:
@pjkundert cool, well i approved it anyway so carry on :)

PR #784, opened 2018-12-20
incremental compilation improvements for wasm/nix/docker

thedavidmeister:
## changes

- removes hardcoded compilation optimisation/lto config from manifests
  - setting `lto = <anything>` in the manifest breaks incremental compilation
  - can be handled at compile time very easily and more flexibly
- incremental compilation is enabled for nix/circle and improved for travis (wasm incrementally builds)
- pre-warms the docker with everything nix & rust predownloaded and precompiled
  - running nix shell commands in docker is much faster
  - docker download is much bigger
  - a lot of the time ~1 hour spent building docker is actually building/installing node related things
  - i attempted to do this on docker hub but eventually gave up and used circle instead
    - circle gives realtime logs, docker hub gives no logs
    - circle keeps build histories, docker hub only keeps about 5-10 builds (don't remember exactly)
    - circle has much easier workflow management
      - can kick off builds on a daily cron job
      - important because the build takes over 2 hours and both circle/docker hub kill builds at 2 hours
      - the docker box has become a serial chain of FROM builds, followup to stamp the "end" box as the official box and cleanup docker hub
    - circle supports docker files in subdirectories of the repo without crazy symlink/copy for makefiles etc.
- circle is split into two parallel runs for unit tests and integration tests
- circle runs `hc` tests now
- added ability to compile a target outside the main repo
  - means we can support docker volumes without blowing away/sharing targets with the host
  - means we can "bake in" compiled rust into the docker box
  - fixed the travis incremental build misses in the process it seems
- updated nix commands to match circle changes
- updated .dockerignore to not have a massive build context (3GB+ down to 40MB)
- ignore target directory in .gitignore properly
- splits the `cargo test --all` into multiple `cargo test` calls, removes the need for a separate `wasm build` step
  - honestly i'm not too sure about this one...
    - pro: not possible to forget to build wasm because it happens inline with tests (has burned us in past)
    - cons: seems to massively bloat the docker box because many things have duplicate compilations
  - main challenge to support `cargo test --all` is juggling relevant paths/environment variables across circle/travis & nix/windows & docker/metal & develop/test/release contexts, the splitting allows for simpler hardcoding of these contexts, i don't think it's impossible to get this working, it's just a bit more elbow grease
- removes all circle ci caching logic as docker/nix/rust effectively makes it redundant
- removes a lot of boilerplate in travis yml
- adds target dir to output of scaffolding tool

## timings

significant improvements for travis and docker (e.g. local dev)

not quite as good as originally thought for circle, mostly due to downloading docker adding a few minutes to startup (can probably be mitigated in followups)

#### travis

**walltime -12 mins**

current develop https://travis-ci.com/holochain/holochain-rust/builds/95920625:

- wall time: 32 mins
- total time: 4 hr 11 mins

this branch https://travis-ci.com/holochain/holochain-rust/builds/95975649:

- wall time: 20 mins
- total time: 2 hours 40 mins

the win here comes from the incremental wasm compilation, which really takes the edge off the slowest builds (slow because they have a lot of wasm)

#### circleci

**walltime -6 mins**

current develop https://circleci.com/gh/holochain/holochain-rust/824:

wall time: 16 mins
total time: 16 mins

this branch https://circleci.com/workflow-run/a537aac9-e350-4f2f-aca7-9588c2247942:

unit tests time: 10 mins
integration tests time 4.5 mins

wall time total: 10 mins
total time: 14.5 mins

note that the unit tests in this branch also run `hc` now

not a massive improvement overall, mostly because the docker box is large ~2 mins download, which is roughly par with the native circle caching

the size and build time of the docker has the potential to come down a lot by combining everything back into a single `cargo test --all` call, rather than one `cargo test` per folder

i don't expect `cargo test --all` to speed up the test run itself because of incremental compilation, but the initial download might be faster... i added a followup to investigate this, but could also be convinced to give it a try in this branch too

much of the improvement on wall time comes from the parallelisation

#### docker

**walltime -13 mins (at least... it also actually runs all the tests)**

develop ubuntu docker:

```
$ time bash ./docker/run make test

...

real    16m5.571s
user    0m0.334s
sys     0m0.473s

```

**important note: this didn't even finish!! breaks on hc**

errors on yarn coming out of the ubuntu/makefile/docker approach:

```
./scripts/build_nodejs_container.sh: 2: ./scripts/build_nodejs_container.sh: yarn: not found
/bin/sh: 1: yarn: not found
/holochain/nodejs_container/node_modules/shelljs/src/common.js:401
      if (config.fatal) throw e;
                        ^

Error: exec: /bin/sh: 1: yarn: not found
```

new circle docker (includes working `hc` tests):

```
$ time docker run holochain/holochain-rust:circle.99.end

...

real    3m5.576s
user    0m0.054s
sys     0m0.038s
```

## followup

- consolidate dockers, use the "end" circle box as the main box and remove a lot from docker hub
- documentation
- rework docker builds to be faster and smaller by reusing deps between crates

**there is a WIP on some of the docker stuff at #807**


zippy:
circle ci failed...!

thedavidmeister:
@zippy yes, back in WIP for this one, i'm "baking in" the incremental build data straight into the docker box

thedavidmeister:
@zippy yeah we can tweak the build process over time

i had to force myself to stop messing with it to put this PR up ;)

actually i'd prefer to figure out how to get the deps reused properly, if that works then the build time should collapse like 10x or something

PR #785, opened 2018-12-20
Remove two unused core_types

Connoropolous:
its not used anywhere in the codebase. It used to be, obviously :p 

found this during my rustdoc work


zippy:
I think this got superceeded by @AshantiMutinta work on "EAV attributes define and validate allowed characters for attribute names see #576"  is that correct?

AshantiMutinta:
> 
> 
> I think this got superceeded by @AshantiMutinta work on "EAV attributes define and validate allowed characters for attribute names see #576" is that correct?

Hes this is true, validation is now done at the eav level

Connoropolous:
So can y’all approve this? :)

thedavidmeister:
i'm happy to see the old agent cruft cleaned up

merge when green :+1:

Connoropolous:
thanks David

PR #786, opened 2018-12-20
Added basic (currently failing) crud tests for app-spec

sphinxc0re:


app_spec/test/test.js
```diff
@@ -75,6 +75,45 @@ test('create_post', (t) => {
...
+  const params = { content, in_reply_to }
+  const createResult = app.call("blog", "main", "create_post", params)
+
+  t.ok(createResult.Ok)
+
+  const updateParams = { post_address: createResult.Ok, new_content: "Hello Holo" }
+  const result = app.call("blog", "main", "update_post", updateParams)
+
+  t.equals(result.Ok, null)
+})
```
zippy: For this test to prove that update works it also needs one more part, which is to do a get on the address of the initial post, and show that what comes back is the updated post.

sphinxc0re: But doesn't the address change when updating the contents?

app_spec/zomes/blog/code/src/post.rs
```diff
@@ -53,9 +56,22 @@ pub fn definition() -> ValidatingEntryType {
...
+                ValidationData::Modify { .. } => Ok(()),
+                ValidationData::Delete { entry_addr, orig_entry, .. } => {
+                    let del_entry = hdk::get_entry(&entry_addr)?;
+                    let orig_entry = hdk::get_entry(&orig_entry)?;
+
+                    let del_entry_sources = entry.sources();
+                    let orig_entry_sources = entry.sources();
+
+                    (del_entry_sources == orig_entry_sources)
+                        .ok_or_else(|| "deletion is unauthorized".into())
```
lucksus: Would be cool to have a test that triggers this and verifies entries can only be deleted by their authors..

PR #787, opened 2018-12-20
Context fully owns signal channel sender

maackle:
This contains two changes:

- Remove all references to signal channel in `Holochain` and `Instance`, instead rely on `Context`
- Make `ContextBuilder::spawn()` consume the builder, removing the need to clone every parameter, but making complex build logic slightly less convenient


maackle:
@lucksus if we ever have to include something non `Clone` we would have to do this at that point. Also might find that some things are expensive to clone down the line. Anyway there is only one place where it requires the very slight inconvenience of reassigning the builder to a `mut let`, all others are simple chaining operations. If you think it's worth the tradeoff of convenience I can revert that commit.

lucksus:
@maackle If we put something non-cloneable into the context it would break the context because that has to be cloned throughout 90% of the code base. So I would say lets stick to having a cloneable builder as well..

maackle:
@lucksus It also keeps things clean to not give anyone the chance to leave the Builder lying around. For instance in ContextBuilder, if someone does hold onto the builder after `spawn`ing a context, then all `Arc`s cloned will have an extra reference so its may never be deallocated, and an extra `Sender` might leave a channel open that was supposed to be closed when all other `Senders` went out of scope.

maackle:
There's nothing stopping this builder from being `Clone` itself, but I still think it shouldn't clone all of its params when it spawns.

Also it's possible that the Builder might accept a non-cloneable param but use it to compute a cloneable value that goes into the eventual instance.

I'll hold off on this until tomorrow, I'd like to talk about it a little and I know you're done for the day.

lucksus:
I anticipate more complex usages of `ContextBuilder` to come in when we start refactoring our tests. I don't think the extra Arcs or Senders actually cause any problems. But also the inconvenient doesn't either .. ;) that's why I approved. Would want to hold off a merge but would prefer the clean usage. Would also be interested in discussing this tomorrow to see if there is another solution to this problem that could also be applied to the api builder...

PR #791, opened 2018-12-21
capabilities-2: separate from zome function declaration

zippy:
The PR is all pre-work for capabilities fully working which means tokens being passed in an checked.  To get there we first need to:
- [x] Complete the change over function definitions not happening in capability definitions
- [x] update `define_zome!` macro to separate capability definition from function declarations eg:
``` rust
define_zome! {
    entries: []

    genesis: || {
        Ok(())
    }

    functions: [
        call_bridge: {
            inputs: | |,
            outputs: |result: JsonString|,
            handler: handle_call_bridge
        }
    ]

    capabilities: {
        public (Public) [call_bridge]
    }
}
```
- [x] implement checking of `Public` capability type, i.e. if you don't declare your functions as public, you'll see a `CapabilityCheckFailed` error
- [x] Update testing code to not be based on creating a capability, but instead function declarations
- [x] remove capability from the http/websocket interface path
- [x] update `define_zome!` macro to produce a __list_functions() zome function (like the __list_capabilities() function
- [x] change testing and example names for capabilities to be less confusing and more informative (from `main` -> `public` for capability and from `main` -> `public_test_fn` for function)
- [x] capabilities and declarations are moved into their own files.


zippy:
@maackle I don't know why this is failing the app_spec tests.  I think it might get fixed when scenario API gets merged in and the regression tests get fixed?

core_types/src/dna/capabilities.rs
```diff
@@ -89,61 +87,16 @@ pub enum CapabilityType {
...
 pub struct Capability {
     /// capability type enum
     #[serde(rename = "type")]
     pub cap_type: CapabilityType,
 
-    /// "fn_declarations" array
+    /// "functions" array
     #[serde(default)]
-    pub functions: Vec<FnDeclaration>,
+    pub functions: Vec<String>,
```
maackle: `function_names` maybe clearer?

core_types/src/dna/fn_declarations.rs
```diff
@@ -0,0 +1,115 @@
...
+impl FnDeclaration {
+    /// Allow sane defaults for `FnDecrlaration::new()`.
+    pub fn new() -> Self {
+        Default::default()
+    }
+}
+
+/// Represents an trait definition for bridging
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Hash)]
+pub struct Trait {
```
maackle: So here's my understanding from the code: `Trait` is a lot like a `Capability`, but the only difference is that `Trait` contains the full function signature because it's referring to functions that might exist in other DNAs, whereas `Capability` is entirely inward-facing, and only holds the function names because it can just look up the function signatures from its own DNA. Is that right?

core_types/src/dna/fn_declarations.rs
```diff
@@ -0,0 +1,115 @@
...
+    /// capability type enum
+    #[serde(rename = "type")]
+    pub cap_type: CapabilityType,
+
+    /// "functions" array
+    #[serde(default)]
+    pub functions: Vec<FnDeclaration>,
```
maackle: should this be called `fn_declarations` also?

nodejs_container/native/src/container.rs
```diff
@@ -0,0 +1,141 @@
...
+    container: RustContainer,
+    _signal_rx: SignalReceiver,
+}
+
+declare_types! {
+
+    /// A Container can be initialized either by:
+    /// - an Object representation of a Configuration struct
+    /// - a string representing TOML
+    pub class JsContainer for NodeContainer {
```
maackle: Why this duplication of node Container? js_test_container.rs still exists.

core_types/src/dna/fn_declarations.rs
zippy: Yep...  Trait is what's used at the composition level of the container for bridging.  Maybe we use a different name for it:  https://realtimeboard.com/app/board/o9J_kyiXmFs=/?moveToWidget=3074457346285220939

nodejs_container/native/src/container.rs
zippy: I didn't add that, I just made things work for what was there.

zippy:
So, to answer your Trait vs Capability question.  It's really two separate concerns that we probably need to clarify because they appear to overlap.  Capabilities are strictly for the security concern.  The identify the kind of capability token that's necessary to call sets of functions.  Traits are for composibility.  They identify groups of functions (plus signatures) that you may want to declare as existing.  It _may_ be the case that these overlap, but note that in this PR I'm mostly trying to land the security concern not the composibility concern.

core_types/src/dna/zome.rs
```diff
@@ -79,6 +79,10 @@ pub struct Zome {
     #[serde(default)]
     pub capabilities: ZomeCapabilities,
 
+    /// An array of functions declared in this this zome.
+    #[serde(default)]
+    pub fn_declarations: ZomeFnDeclarations,
```
lucksus: Hm, I remember that the field `functions` in Capability once was called `fn_declarations` as well and we renamed it to `functions`.

I would actually prefer if this was `functions`, and functions in ZomeCapabilities would be renamed `function_names` as @maackle suggested above.

lucksus:
```suggestion
    pub functions: ZomeFnDeclarations,
```

nodejs_container/native/src/container.rs
lucksus: Was this file renamed from `container.rs` to `js_test_container.rs` in develop and accidentally added here after a merge, @zippy?

core_types/src/dna/fn_declarations.rs
zippy: So the reason I didn't call it declarations here is that it isn't where the functions are declared, its where the trait is saying that "I need a function that looks like this".  There's only one place where it's declared, and that's in the zome.  Possible we could make a distinctions between declared and defined...  I think I'm pushing this to our DND session.

zippy:
ok, spuriously added file deleted.  Re naming, I'd like to deal with this in our DND session, because I asked a few other people and got a different answer... :-P

PR #792, opened 2018-12-21
fix mdbook api func links and content

Connoropolous:
- [x] fix existing links


Connoropolous:
merging with one review because Zippy and I agreed it's ok to do so today, and merging with a failing test because it's a test that's also failing in develop branch, with someone else working on a fix

PR #794, opened 2018-12-21
add temporary checks in network handler to filter messages to others

zippy:
Currently a single process instance of the networking doesn't multi-plex correctly, and our code is assuming one spawn per systems.  This pr adds checks to filter messages so an instance doesn't processes messages intended for others.

### DIFF
--- a/core/src/network/handler/mod.rs
+++ b/core/src/network/handler/mod.rs
@@ -9,6 +9,30 @@ use crate::{
 use holochain_net_connection::{net_connection::NetHandler, protocol_wrapper::ProtocolWrapper};
 use std::{convert::TryFrom, sync::Arc};
 
+// FIXME: Temporary hack to ignore messages incorrectly sent to us by the networking
+// module that aren't really meant for us:
+fn is_me(c: &Arc<Context>, dna_hash: &str, agent_id: &str) -> bool {
+    // TODO: we also need a better way to easily get the DNA hash!!
+    let state = c
+        .state()
+        .ok_or("is_me could not get application state".to_string())
+        .unwrap();
+    let dna = state
+        .nucleus()
+        .dna()
+        .ok_or("is_me called without DNA".to_string())
+        .unwrap();
+    let my_dna_hash = base64::encode(&dna.multihash().unwrap());
+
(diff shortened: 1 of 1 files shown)


core/src/network/handler/mod.rs
```diff
@@ -9,6 +9,30 @@ use crate::{
 use holochain_net_connection::{net_connection::NetHandler, protocol_wrapper::ProtocolWrapper};
 use std::{convert::TryFrom, sync::Arc};
 
+// FIXME: Temporary hack to ignore messages incorrectly sent to us by the networking
+// module that aren't really meant for us:
+fn is_me(c: &Arc<Context>, dna_hash: &str, agent_id: &str) -> bool {
```
Connoropolous: @zippy finding the name 'is_me' is a little confusing... what "is me"? what "is us"? as per the comment. Who is me, who is us?

core/src/network/handler/mod.rs
```diff
@@ -18,20 +42,42 @@ pub fn create_handler(c: &Arc<Context>) -> NetHandler {
...
-            Ok(ProtocolWrapper::StoreDht(dht_data)) => handle_store_dht(dht_data, context.clone()),
+            Ok(ProtocolWrapper::StoreDht(dht_data)) => {
+                // NOTE data in message doesn't allow us to confirm agent!
+                if !is_me(&context, &dht_data.dna_hash, "") {
+                    return Ok(());
+                }
+                handle_store_dht(dht_data, context.clone())
+            }
             Ok(ProtocolWrapper::StoreDhtMeta(dht_meta_data)) => {
+                if !is_me(&context, &dht_meta_data.dna_hash, &dht_meta_data.agent_id) {
```
neonphog: I "think" this is the from agent_id, so maybe should be `""` here too.

core/src/network/handler/mod.rs
Connoropolous: Should/could this be called "message_is_for_me"?

zippy:
moving this into `real-networking` branch

zippy:
and actuall the function really is "is me" because there's no message in the function, it's just checking that the dna and the agent_id is the same as in the context.

PR #796, opened 2018-12-22
add globals to app_spec

zippy:
we have: [We can access the app dna hash in the zome](https://realtimeboard.com/app/board/o9J_kyiXmFs=/?moveToWidget=3074457346312836527)

and also

[runtime globals are actually correct values  (See invoke_init_globals)](https://realtimeboard.com/app/board/o9J_kyiXmFs=/?moveToWidget=3074457346367734561) in the SoA.  
as if they aren't implemented.

It actually looks to me like they are, at least partially, working.  As these app_spec tests seem to show.

Feedback?


zippy:
well, something in the build on circle CI causes it to come out with a different dna hash, then on my machine...

zippy:
hmmm dna address still coming back different that what happens when I run on my machine.  I'm using the nix-shell command `nix-shell --run hc-test-app-spec` so I really don't know what's different from what's happening on circleci when the wasm gets compiled.  @thedavidmeister @lucksus any ideas?

lucksus:
@zippy, I'm not sure but I guess any discrepancy in the toolchain will result in a different binary output. Maybe the hash should not be hardcoded in the test but rather taken from the DNA file under test?

core/src/nucleus/ribosome/api/init_globals.rs
```diff
@@ -20,7 +19,8 @@ pub fn invoke_init_globals(runtime: &mut Runtime, _args: &RuntimeArgs) -> ZomeAp
         dna_address: Address::from(""),
         agent_id_str: JsonString::from(runtime.context.agent_id.clone()).to_string(),
         // TODO #233 - Implement agent pub key hash
-        agent_address: Address::encode_from_str("FIXME-agent_address", Multihash::SHA2256),
+        agent_address: Address::from(""),
```
lucksus: Hm? Isn't this just:
```suggestion
        agent_address: Address::from(runtime.context.agent_id.address()),
```
?

core_types/src/dna/dna.rs
```diff
@@ -19,6 +19,11 @@ fn empty_object() -> Value {
     json!({})
 }
 
+//use uuid::Uuid;
+//fn new_uuid() -> String {
+//    Uuid::new_v4().to_string()
+//}
+
```
lucksus: Remove?

PR #798, opened 2018-12-22
Eav indexing

AshantiMutinta:
Implemented eav indexing using a hash version to denote versions. Makes use of immutable datastructures in which you can only have one copy of each for a given key. For the file storage, a hashfile which will depict versioning by storing addresses in it


core_types/src/eav.rs
```diff
@@ -28,6 +31,9 @@ pub type Attribute = String;
 /// Address of AddressableContent representing the EAV value
 pub type Value = Address;
 
+#[derive(PartialEq, Eq, Debug, Clone, Hash, PartialOrd, Ord)]
+pub struct Key(pub i64, pub Action, pub Uuid);
```
lucksus: Can we get a rustdoc on this central struct?

I am wondering why we need this complex key, and why a simple unsigned integer for the index is not enough?

AshantiMutinta: If you wanted to get a record of the eav at any point in time, you probably are not able to do that with the index only. Also, if we are looking at an agent centric POV, we have to look at it from the point that collisions are handled well. If the UUID is put there it wont overide a time + action enum

core_types/src/eav.rs
```diff
@@ -47,6 +55,60 @@ pub struct EntityAttributeValue {
...
+            Action::None
+        }
+    }
+}
+
+pub fn create_key(action: Action) -> Result<Key, HolochainError> {
+    Ok(Key(Utc::now().timestamp_nanos(), action, Uuid::new_v4()))
+}
+
+pub fn from_key(key: HashString) -> Result<Key, HolochainError> {
```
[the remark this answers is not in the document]
AshantiMutinta: This was left over from the old eavi implementation, it could just be a string, it was for when I was a hashmap instead of an ordmap

core_types/Cargo.toml
```diff
@@ -18,6 +18,9 @@ objekt="0.1.1"
 holochain_core_types_derive = { path = "../core_types_derive" }
 uuid = { version = "0.7", features = ["v4"] }
 regex = "^1.1"
+chrono = "0.4"
+guid-create = "^0.1.1"
```
thedavidmeister: unneccessary dep?

core_types/src/eav.rs
```diff
@@ -122,88 +176,117 @@ pub trait EntityAttributeValueStorage: objekt::Clone + Send + Sync + Debug {
...
+impl ExampleEntityAttributeValueStorage {
+    pub fn new() -> ExampleEntityAttributeValueStorage {
+        ExampleEntityAttributeValueStorage {
+            storage: Arc::new(RwLock::new(BTreeSet::new())),
         }
     }
+}
+
+pub fn increment_key_till_no_collision(
+    mut eav: EntityAttributeValueIndex,
```
thedavidmeister: why is eav mutable?

core_types/src/eav.rs
```diff
@@ -122,88 +176,117 @@ pub trait EntityAttributeValueStorage: objekt::Clone + Send + Sync + Debug {
...
+    if map
+        .iter()
+        .filter(|e| e.index == eav.index())
+        .collect::<BTreeSet<&EntityAttributeValueIndex>>()
+        .len()
+        > 0
+    {
+        let timestamp = eav.clone().index + 1;
+        eav.set_index(timestamp);
+        increment_key_till_no_collision(eav, map)
```
thedavidmeister: how does this preserve ordering better than generating a random squuid in a threaded situation?

core_types/src/lib.rs
```diff
@@ -32,6 +32,9 @@ extern crate maplit;
 
 extern crate uuid;
 
+extern crate chrono;
+extern crate guid_create;
```
thedavidmeister: extra dep

core_types/src/lib.rs
```diff
@@ -32,6 +32,9 @@ extern crate maplit;
...
+
 pub mod cas;
 pub mod chain_header;
 pub mod crud_status;
```
thedavidmeister: why the permissions and line ending changes in the files below?

core_types/src/eav.rs
```diff
@@ -122,88 +176,117 @@ pub trait EntityAttributeValueStorage: objekt::Clone + Send + Sync + Debug {
...
-#[derive(Clone, Debug)]
-pub struct ExampleEntityAttributeValueStorage {
-    content: Arc<RwLock<ExampleEntityAttributeValueStorageNonSync>>,
-}
-
-impl ExampleEntityAttributeValueStorage {
-    pub fn new() -> HcResult<ExampleEntityAttributeValueStorage> {
-        Ok(ExampleEntityAttributeValueStorage {
-            content: Arc::new(RwLock::new(ExampleEntityAttributeValueStorageNonSync::new())),
+pub fn get_latest(
```
thedavidmeister: can get_latest just be fetch_eav with take 1 or first or last?

is this duplicated logic or different for some reason?

core_types/src/eav.rs
```diff
@@ -70,17 +97,33 @@ fn validate_attribute(attribute: &Attribute) -> HcResult<()> {
...
-        Ok(EntityAttributeValue {
+        Ok(EntityAttributeValueIndex {
             entity: entity.clone(),
             attribute: attribute.clone(),
             value: value.clone(),
+            index: Utc::now().timestamp_nanos(),
+        })
+    }
+
+    pub fn new_with_timestamp(
```
thedavidmeister: i don't like that we're mixing terminology like "timestamp" and "index"

should just be "index" everywhere

core_types/src/eav.rs
```diff
@@ -31,23 +32,49 @@ pub type Value = Address;
...
     attribute: Attribute,
     value: Value,
-    // index: Index,
+    index: Index,
     // source: Source,
 }
 
-impl AddressableContent for EntityAttributeValue {
+pub struct IndexQuery {
+    start_time: Option<i64>,
```
thedavidmeister: just start and end i think, not "time"

core_types/src/cas/storage.rs
```diff
@@ -327,47 +329,66 @@ impl EavTestSuite {
...
+            let eav =
+                EntityAttributeValueIndex::new(&one.address(), &attribute.clone(), &many.address())
+                    .expect("Could not create eav");
+            expected_one.insert(eav);
+            let fetch_set = eav_storage
+                .fetch_eav(
+                    None,
+                    Some(attribute.clone()),
+                    Some(many.address()),
+                    IndexQuery::default(),
```
thedavidmeister: do we have tests for index queries other than `default`, e.g. range?

cas_implementations/src/lib.rs
```diff
@@ -7,13 +7,15 @@
...
 
 extern crate serde;
 extern crate serde_json;
 
+extern crate chrono;
+extern crate glob;
+
 pub mod cas;
 pub mod eav;
 pub mod path;
```
thedavidmeister: unrelated permissions changes in the files below

cas_implementations/src/eav/file.rs
```diff
@@ -94,88 +122,131 @@ impl EavFileStorage {
...
+            Ok(BTreeSet::new())
+        }
     }
 }
 
 impl EntityAttributeValueStorage for EavFileStorage {
-    fn add_eav(&mut self, eav: &EntityAttributeValue) -> Result<(), HolochainError> {
+    fn add_eav(
+        &mut self,
+        eav: &EntityAttributeValueIndex,
```
thedavidmeister: eavi?

cas_implementations/src/eav/file.rs
```diff
@@ -94,88 +122,131 @@ impl EavFileStorage {
...
-            .and_then(|_| self.write_to_file(ATTRIBUTE_DIR.to_string(), eav))
-            .and_then(|_| self.write_to_file(VALUE_DIR.to_string(), eav))
+        let new_eav = increment_key_till_no_collision(eav.clone(), fetched.clone())?;
+        self.write_to_file(ENTITY_DIR.to_string(), &new_eav)
+            .and_then(|_| self.write_to_file(ATTRIBUTE_DIR.to_string(), &new_eav))
+            .and_then(|_| self.write_to_file(VALUE_DIR.to_string(), &new_eav))?;
+        Ok(Some(new_eav.clone()))
     }
 
     fn fetch_eav(
```
thedavidmeister: `fetch_eavi`?

cas_implementations/src/eav/file.rs
```diff
@@ -21,6 +26,7 @@ const VALUE_DIR: &str = "v";
 pub struct EavFileStorage {
     dir_path: String,
     id: Uuid,
+    current_hash: HashString,
```
thedavidmeister: current_hash?

core_types/src/eav.rs
```diff
@@ -9,9 +9,10 @@ use crate::{
     error::{HcResult, HolochainError},
     json::JsonString,
 };
+use chrono::offset::Utc;
 use objekt;
 use std::{
-    collections::HashSet,
+    collections::BTreeSet,
```
thedavidmeister: why BTreeSet instead of HashSet?

core_types/Cargo.toml
AshantiMutinta: oh yes left over from previous, will remove that

core_types/src/eav.rs
AshantiMutinta: recursive function that returns non mutable eav. It is only mutable in the scope of this function

core_types/src/eav.rs
AshantiMutinta: Guards on eav so only one write per time, timestamps are sorted by i64 thus ordered

core_types/src/lib.rs
AshantiMutinta: That needs to come out for sure

core_types/src/lib.rs
AshantiMutinta: cargo fmt so line endings are affected differently on different systems. Windows vs linux

cas_implementations/src/lib.rs
AshantiMutinta: Same as above

cas_implementations/src/eav/file.rs
AshantiMutinta: yes those will need to be changed to that

cas_implementations/src/eav/file.rs
AshantiMutinta: remenants of old code, will change that

core_types/src/eav.rs
AshantiMutinta: Was a decision based on preserving order, two eavis won't have the same hash so we need a datastructure that implements an ordered trait to preserve order because of the index

core_types/src/eav.rs
AshantiMutinta: This is obtaining get latest for each and every entry in the list, it looks duplicated because it is being called while doing an fetch_eav. what we need is a fetch_eav_lazy which gets you the iterator so that you can be able to perform operations on it but that is more of a follow up. We are getting the latest timestamp of each entry that passes through that iterator

core_types/src/eav.rs
AshantiMutinta: I can change that for sure

core_types/src/eav.rs
AshantiMutinta: I can change that as well

core_types/src/cas/storage.rs
AshantiMutinta: I can definitely do add a ranged timestamp for sure, I was using the get_latest as a proof but a range as well would be good

core_types/src/eav.rs
thedavidmeister: kk

core_types/src/lib.rs
thedavidmeister: kk

core_types/src/eav.rs
thedavidmeister: so then i have the inverse question, why do we need to call it recursively

if this is one write at a time we should just be able to to `+ 1` against the current highest value if it equals the proposed index

core_types/src/cas/storage.rs
thedavidmeister: great

get_latest doesn't prove a range ;)

core_types/src/eav.rs
AshantiMutinta: This is a more fireproof way of incase you have multiple collisions on the nanoseconds.
value 1 = 0001, value 2 =0001, value 3 = 0001

If both of these are created at the same time, value 2 would be 0002 if I add only one. If it happens to value 3 and we add only 1 it would happen override value 2 because the timestamp would be 0002. That is why we need to iterate until we find no collisions :)

PR #800, opened 2018-12-28
Update for futures-preview alpha 11 and nightly-2018-12-26

zippy:
currently this comments out the async timeouts in:
`core/src/network/actions/get_entry.rs` and `core/src/network/actions/custom_sent.rs` because I'm not sure how to get them to work given the new changes, but I don't think we have any tests that test them either.

Adding that to the SoA as a follow up.

### DIFF
--- a/core/src/dht/actions/hold.rs
+++ b/core/src/dht/actions/hold.rs
@@ -14,10 +14,7 @@ use holochain_core_types::{
     entry::Entry,
     error::HolochainError,
 };
-use std::{
-    pin::{Pin, Unpin},
-    sync::Arc,
-};
+use std::{pin::Pin, sync::Arc};
 
 pub async fn hold_entry<'a>(
     entry: &'a Entry,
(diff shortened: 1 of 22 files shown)


core/src/dht/actions/hold.rs
```diff
@@ -37,8 +34,6 @@ pub struct HoldEntryFuture {
     address: Address,
 }
 
-impl Unpin for HoldEntryFuture {}
```
maackle: @zippy why is this removed? Assuming the compiler thinks it's no longer necessary?

core/src/dht/actions/remove_entry.rs
```diff
@@ -40,8 +40,6 @@ pub struct RemoveEntryFuture {
     action: ActionWrapper,
 }
 
-impl Unpin for RemoveEntryFuture {}
```
maackle: Same, why removed?

core/src/dht/actions/hold.rs
zippy: I had to.  Apparently Unpin became a private element (the use statement broke and I had to remove it), and gets auto-implemented in the default case.  I didn't dig too deep, just got it to work.

PR #803, opened 2018-12-28
WIP: Proper shutdown of network and instances when stopping Container

maackle:
TODO:
- [ ] tests


hdk-rust/tests/integration_test.rs
```diff
@@ -665,7 +665,7 @@ fn can_send_and_receive() {
     assert!(result.is_ok(), "result = {:?}", result);
     let agent_id = result.unwrap().to_string();
 
-    let (mut hc2, _) = start_holochain_instance("can_remove_modified_entry", "bob");
+    let (mut hc2, _) = start_holochain_instance("can_send_and_receive", "bob");
```
maackle: Wasn't sure about this one, seemed like the test shouldn't even pass without this change, can someone explain if it was actually supposed to be the other way?

maackle:
@neonphog I was also excited to use `Drop`, you'll see a few commit messages where I tried it and then abandoned it. I tried to use `Drop` to the utmost, but I learned some things:

- for joining, the signatures don't match: `drop(&mut self)` vs. `join(self)` so it's impossible to join during the drop without a last-second `mem::replace`. (I don't get why drop doesn't consume itself :man_shrugging:)
- I tried implementing shutdown of `Instance` and `P2pNetwork` through Drop, but several tests were going crazy because of a cloned `NetworkState` that was getting dropped repeatedly. That was only a problem because I took `NetworkState` out of its `Arc`, but I needed to do that to experiment with closing the network via `Drop` in the first place (because otherwise we can't be compile-time sure that all references have died out)

If you have any insight into how it can be effectively included, it would be easy to add in again.

core/src/network/reducers/shutdown.rs
```diff
@@ -0,0 +1,31 @@
+use crate::{action::ActionWrapper, context::Context, network::state::NetworkState};
+use holochain_net_connection::{
+    net_connection::NetConnection,
+    protocol_wrapper::{ProtocolWrapper, TrackAppData},
+};
+use std::sync::Arc;
+pub fn reduce_shutdown(
```
thedavidmeister: is lifecycle management for a stateful entity in our system part of the state of the entity (which is what reducers are for)?

i'm imagining if this was a database, then this would be like needing an `INSERT` query every time we open/close a connection to the database

could we use the topological evaluation that nicho added for lifecycle management?

i'm thinking something like this https://github.com/stuartsierra/component

maackle: @thedavidmeister I would say yes only because the current architecture requires us to put the entity into an "is_shut_down" state rather than completely destroying it, because there may still be live references to it in other threads. This is the case both for Arc references in core and for references to the instance in the mock network.

thedavidmeister: @maackle what does "is shut down" mean if there are "live references to it"?

could you explain a simple example of how that works?

maackle: @thedavidmeister by "live references" I mainly mean `Arc`s, but also value-based references like in `MockSingleton::senders` where instances are indirectly referenced by a HashMap of Senders, the channels of which point to the associated NetWorker for the Instance (so even if the Instance is dropped, this can be a sort of dangling reference leading to a `SendError`).

So the problem is that we have the possibility of instances being in this invalid "shutdown" state where we really want them to be completely gone, but other things are still referring to them. That invalid state is still a state we need to track, as long as we can't be 100% sure that we've got no dangling references to the thing. 

I guess the only question is if we want to track that state in the reducer or not. Here's one way to argue for it: `Shutdown` is the state transition that specifies that all other future state transitions are invalid. (If we could ensure that there would be no future attempts to transition state, we wouldn't need to think of it this way.) I also might be interpreting the intent of the whole redux state system wrongly too, so if this ultimately seems wrong I'll change it.

Here are the main the things that can be "shutdown" but not fully terminated (dropped):

* **`Instance`**: is_shutdown means the action loop is broken and the thread is joined, so no more actions are processed. (But the `Instance` may still live on in other `Arc`s).
* **`NetworkState`**: is_shutdown means the `P2pNetwork` and `NetWorker` (part of the state) are stopped. However, `NetworkState` is part of the `State`, which is referred to by the `Context` (via `Arc`) which is referred to by the `Holochain` (via `Arc`), therefore we can't dispose of the `NetworkState` either.
* therefore **`Holochain`**'s is_shutdown means the contained `Instance` is shutdown (meaning no `call` can be done), and the network is stopped. However there will still be `Arc<Holochain>`s hanging around so we can't dispose of the `Holochain` or `Instance` entirely.

All of these end-of-life changes are stateful because until we can fully dispose of these entities, they may be acted upon by other threads and reducers.

thedavidmeister: "we really want them to be completely gone, but other things are still referring to them"

this bit i don't totally understand, why are "we" not also the "other things"? what other things exist outside "our" control?

if we truly want something to be completely gone, then we should track everything referencing it and do a cleanup, working through dependencies right?

if something has a dependency on something that is "shutdown", how can it continue to operate? isn't the dependant thing now operating in an invalid state, regardless of how "live" it is?

maackle: @thedavidmeister and I had a good chat about this which convinced me to remove the shutdown reducer and deal with shutdown outside of the action loop. We also talked about possible approaches for tracking component dependencies to have more certainty about how to shut things down, which we might consider down the road when things get more complex.

hdk-rust/tests/integration_test.rs
lucksus: I've added this parameter that gets set as the DNA's UUID in order to not have tests interfere with each other since most of our tests create the same boilerplate DNA that then happens to have the same DNA hash which results in instances of different tests communicating with each other through the mock network.. 

So yes, this change is good. But the problem only appears if tests creating the same DNA hash run simultaneously.

core/src/network/reducers/shutdown.rs
lucksus: Ah, I was a bit confused since I saw the shutdown reducer yesterday when glancing over this PR. Interesting, what convinced you to remove the shutdown reducer? That was the part that made sense to me.. ;) Without that you have to do bad things like writing the state from the outside.

The main reason (currently) to have any shutdown process is the network, right? Telling the network module that we unregister from the network is a state transition that should be facilitated by an action reduction.

lucksus:
I don't see why you would want to overwrite the state with a fresh one, or `None`.

If you want to pause the instance or kill it without caring about cleaning up stuff, then we could just stop the action loop, yes, and just drop all the things.

But isn't it true that the main reason for having any shutdown code is to have something like the network clean up after itself by unregisering with the network process? That is a state change and we have to keep the action loop running and wait for this process to complete. On top of that we could always add a timeout and kill the instance the hard way...

But regarding writing the state from outside, why do you think you need that? The `Context` is just for distributing references to the things that always exist (once) in the context of a Holochain instance. If you want to change/write the state, that really should happen in `Instance`, either through a reducer or if it really turns out to be a special case where we don't want to reduce an action, it should be special purpose code in a `self`-consuming function in `Instance` itself.. 

```rust
impl Instance {
  fn shutdown(self) {
    self.state = None;
  }
}
```

maackle:
@lucksus I want to replace the `State` (and everything else) so that I can drop the old one. Dropping things after shutting them down is a way of ensuring that nothing else refers to them. I admit I went a little wild with that notion, trying to add a `Drop` instance to everything and shutting it down by just making it go out of scope, but our architecture doesn't really allow that. So some of these `mem::replace` instances are relics of that attempt.

I will probably do another pass over this and figure out what's realistic to drop. In the case of State, it doesn't make much sense to drop it because it's the `Arc` that matters, even if the backing state is swapped out then there will be still be dangling references to that new state by whoever holds the original Arc clones.

The big problem that I'm trying to pave the way for is handling all the places where other code assumes the existence of a running Instance and doesn't have the logic to deal with the case where it's shut down. Every Sender of an Instance-related channel and every `Arc<Holochain | Instance | State>`.

Back to the reducers, maybe it's appropriate to do the shutdown in two phases? The redux-stateful part and then another part? Network shutdown is stateful, but the actual shutdown of the action loop should not be initiated by a reducer.

thedavidmeister:
@maackle is this branch still "live"?

thedavidmeister:
@maackle feel free to reopen if you need this

PR #805, opened 2019-01-01
change dna_hash to dna_address

thedavidmeister:
from #744 

## changes

- renames `dna_hash` to `dna_address`
- represents dna addresses as `Address` not `String` (it was never a `HashString` despite the name)
- implements `AddressableContent` for `Dna` so that `dna.address()` works (no more ad-hoc multihashing)
- implements feedback from parent issue


Connoropolous:
@thedavidmeister do we still have the challenge where really big DNA entries can be passed over the network and written to the DHT, etc?

core/src/workflows/hold_link.rs
```diff
@@ -76,10 +76,11 @@ pub mod tests {
             create_test_dna_with_wat("test_zome", "test_cap", Some(&test_wat_always_invalid()));
         dna.uuid = String::from("test_reject_invalid_link_on_hold_workflow");
 
-        // Hash of the original DNA
-        let dna_hash = base64::encode(&dna.multihash().unwrap());
+        // Address of the original DNA
+        let dna_address = dna.address();
```
Connoropolous: much better to have this consistent function than calling base64... repeatedly. Sweet!

net_connection/src/protocol_wrapper.rs
```diff
@@ -34,13 +34,13 @@ pub struct ConfigData {
 
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, DefaultJson)]
 pub struct ConnectData {
-    pub address: String,
+    pub address: Address,
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, DefaultJson)]
 pub struct PeerData {
```
Connoropolous: @thedavidmeister @neonphog perhaps a change on this side would require a change on the n3h side?

thedavidmeister:
@Connoropolous this doesn't help anything with the size of DNA unfortunately, it's just tightening up the address/hash handling

core/src/workflows/hold_link.rs
thedavidmeister: @Connoropolous yar, based on the experience where we decided Q3 last year to change the way addresses work for agents, i think it's pretty important to leverage `.address()` as much as possible - it keeps this as flexible as possible

net_connection/src/protocol_wrapper.rs
thedavidmeister: @Connoropolous i just chatted to @neonphog about these changes in teal, so he's aware, all good for merge i think :)

Connoropolous:
thanks for the info David

PR #806, opened 2019-01-01
Sodium crypto pkg

zo-el:


neonphog:
Looks like you need to run `make fmt` - https://circleci.com/gh/holochain/holochain-rust/1325

sodium/src/secbuf.rs
```diff
@@ -53,7 +58,7 @@ struct SodiumBuf {
 impl Bufferable for SodiumBuf {
     /// warning: funky sizes may result in mis-alignment
     fn new(s: usize) -> Box<Bufferable> {
-        if s != 16 && s != 32 && s != 64 {
+        if s != 8 && s != 16 && s != 32 && s != 64 {
```
neonphog: mmm, unless we want to get fancy and actually check the alignment, I *think* 16 is the minimum on a 64 bit machine... 8 might be mis-aligned. (Somebody correct me if I'm wrong, alignment is a little out of my depth - https://en.wikipedia.org/wiki/Data_structure_alignment)

zo-el:
Following issues have been addressed and changes have been made:
- make sure check_init() is called in every function
- shrink the unsafe {} blocks to just wrap the unsafe ffi functions
- run make fmt
- buf() function a little confusing when it is used in a module. We should expand the name to be random_secbuf()

sodium/src/aead.rs
```diff
@@ -0,0 +1,242 @@
...
+
+/// Generate symmetric cipher text given a message, secret, and optional auth data
+///
+/// @param {SecBuf} message - data to encrypt
+///
+/// @param {SecBuf} secret - symmetric secret key
+///
+/// @param {SecBuf} adata - optional additional authenticated data
+///
+/// @param {SecBuf} nonce - Empty Buffer to be used as output
```
neonphog: This is actually input. It is pub data that must be sent along with the message (or derived through some sequence scheme) in order to decrypt it.

sodium/src/hash.rs
```diff
@@ -0,0 +1,88 @@
...
+/// @param {Buffer} input - the data to hash
+///
+/// @param {SecBuf} output - Empty Buffer to be used as output
+pub fn sha512(input: &mut SecBuf, output: &mut SecBuf) -> SodiumResult<()> {
+    check_init();
+    let input = input.read_lock();
+    let mut output = output.write_lock();
+    let input_len = input.len() as libc::c_ulonglong;
+    unsafe {
+        rust_sodium_sys::crypto_hash_sha256(
```
neonphog: lol, was wondering why there were so many zeroes in the unit test output below : ) this should be `sha512`.

sodium/src/kdf.rs
```diff
@@ -0,0 +1,104 @@
...
+    context: &mut SecBuf,
+    parent: &mut SecBuf,
+) -> SodiumResult<()> {
+    check_init();
+    {
+        let out = out.read_lock();
+        let o = out.len();
+        let context = context.read_lock();
+        let c = context.len();
+        if check_buf_len(o) {
```
neonphog: the output size doesn't need to be aligned (unless you are using a secure secbuf)... but it does have to be > `crypto_kdf_BYTES_MIN` and < `crypto_kdf_BYTES_MAX`

sodium/src/kdf.rs
```diff
@@ -0,0 +1,104 @@
...
+    #[test]
+    fn it_should_return_error_on_bad_output_buffer() {
+        let mut context = SecBuf::with_secure(8);
+        let mut parent = SecBuf::with_secure(32);
+        random_secbuf(&mut context);
+        random_secbuf(&mut parent);
+        let mut out = SecBuf::with_insecure(2);
+        {
+            let mut out = out.write_lock();
+            match derive(&mut out, 3, &mut context, &mut parent) {
```
neonphog: should be able to

```rust
derive(&mut out, 3, &mut context, &mut parent).expect_err("should have failed");
```

will panic on `Ok` result with the passed text.

sodium/src/pwhash.rs
```diff
@@ -0,0 +1,145 @@
...
+    let my_salt_locker;
+    let mut random_salt = SecBuf::with_insecure(SALTBYTES);
+    random_secbuf(&mut random_salt);
+    let random_salt = random_salt.read_lock();
+    let mut my_salt = raw_ptr_char_immut!(random_salt);
+
+    if let Some(s) = salt {
+        my_salt_locker = s.read_lock();
+        my_salt = raw_ptr_char_immut!(my_salt_locker);
+    }
```
neonphog: I think we need to take the `Option<>` off the salt... If we are using a random salt, we need to return it to the user somehow, or they'll never be able to repeat the hashing. API-wise it will be simpler to always take a salt, and if the user wants a random one, they can radomize it themselves.

sodium/src/sign.rs
```diff
@@ -0,0 +1,145 @@
...
+/// @param {Buffer} signature
+///
+/// @param {Buffer} message
+///
+/// @param {Buffer} publicKey
+pub fn verify(signature: &mut SecBuf, message: &mut SecBuf, public_key: &mut SecBuf) -> i32 {
+    check_init();
+    let mut signature = signature.write_lock();
+    let mut message = message.write_lock();
+    let mut public_key = public_key.write_lock();
```
neonphog: these should all be `read_lock`?

sodium/src/aead.rs
zo-el: Yes, this is sent back because it is needed while decrypting. The statement should have said that the implementor needs to pass in an empty buffer in which the output value will be stored.
Am I getting that right?

zo-el:
Yes I think that `panic!`, if the state is currently anything other than `NoAccess`, should be included

sodium/src/aead.rs
neonphog: no, this is a read param, you can see it does a read_lock() and passes into the ffi encrypt immutable. As you mention in the decrypt, it is the initialization vector, but it is still user specified even in the encrypt.

PR #810, opened 2019-01-02
WIP on making json string to JsonString conversion explicit

thedavidmeister:
## changes

- converting from `String` to `JsonString` is no long silent
  - `JsonString::from_json(&str)` replaces `JsonString::from(String)`
  - allows use of generic traits to convert to `JsonString`, e.g. `Result<T, E>`
    - could not be done before because `String` always gets in the way
  - highlights a lot of places that `JsonString::from()` was used for `String` values that are definitely not valid json :(
- implements generic `impl From<Result<T, E>> for JsonString {}` and `String` equivalents


thedavidmeister:
this work has moved to a more targetted PR at https://github.com/holochain/holochain-rust/pull/1159

thedavidmeister:
fixed by @willemolding elsewhere

PR #812, opened 2019-01-03
NodeJS Container ready for 0.0.3

maackle:
This gets the NodeJS Container stuff ready for 0.0.3, since the actual Scenario API stuff won't be ready for that release:

- Backports some future improvements to the nodejs container API from #797
- Adds `makeCaller` convenience function to `Container.prototype`
- Changes instance ID delimiter from `-` to `::` (hey why not?)
- Updates README

### DIFF
--- a/nodejs_container/index.js
+++ b/nodejs_container/index.js
@@ -6,9 +6,10 @@ const path = require('path');
 // deals with ensuring the correct version for the machine/node version
 const binding_path = binary.find(path.resolve(path.join(__dirname, './package.json')));
 
-const { ConfigBuilder, Container } = require(binding_path);
+const { makeConfig, Container } = require(binding_path);
 
 Container.prototype.callRaw = Container.prototype.call
+
 Container.prototype.call = function (id, zome, trait, fn, params) {
     const stringInput = JSON.stringify(params);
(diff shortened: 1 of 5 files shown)


nodejs_container/index.js
```diff
@@ -22,7 +23,26 @@ Container.prototype.call = function (id, zome, trait, fn, params) {
     return result;
 }
 
-module.exports = {
-    ConfigBuilder: new ConfigBuilder(),
-    Container: Container,
-};
+// Convenience function for making an object that can call into the container
+// in the context of a particular instance. This may be temporary.
+Container.prototype.makeCaller = function (agentId, dnaPath) {
```
lucksus: Why is this `makeCaller` and not `makeInstance` ?

maackle: @lucksus it's just making an interface into the instance -- a lightweight object with `call` (just a combinator) and `agentId`. The actual instance is made during `new Container`.

maackle:
@Connoropolous thanks for your original README update! I was originally considering ripping a lot of this out and going back to the old `instanceFromNameAndDna` way temporarily but since you documented this well, it felt solid enough to touch up and release this way.

PR #815, opened 2019-01-04
[DON'T MERGE YET] Make MockSingleton multiply instantiable

maackle:
### part 2 (based on #816)

- rename `MockSingleton` -> `MockSystem`
- maintain a Map of them, using a unique string to diferentiate them

### DIFF
--- a/net/src/mock_worker.rs
+++ b/net/src/mock_worker.rs
@@ -281,7 +281,7 @@ impl NetWorker for MockWorker {
     fn receive(&mut self, data: Protocol) -> NetResult<()> {
         if let Ok(wrap) = ProtocolWrapper::try_from(&data) {
             if let ProtocolWrapper::TrackApp(app) = wrap {
-                // Couldn't figure out lifetimes here, so accessing the mock is inlined.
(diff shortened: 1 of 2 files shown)


net/src/mock_worker.rs
```diff
@@ -283,17 +279,23 @@ impl NetWorker for MockWorker {
...
+                let mut mock = map_lock
+                    .entry(self.network_name.clone())
+                    .or_insert_with(|| Mutex::new(MockSystem::new()))
+                    .lock()
+                    .unwrap();
+
                 let (tx, rx) = mpsc::channel();
                 self.mock_msgs.push(rx);
                 mock.register(&app.dna_address, &app.agent_id, tx)?;
-                return Ok(());
```
ddd-mtl: I don't understand the reason for removing this return statement.

maackle: @ddd-mtl that was a mistake, good catch. I updated it in #830, could you take a look over there?

maackle:
I made a new PR for this at #830 rather than rebasing and force-pushing, please check that.

PR #816, opened 2019-01-04
P2pConfig enum rather than serde_json::Value

maackle:
### part 1

Use an enum of structs instead of a `serde_json::Value` for P2pConfig. The hope is that the struct will give documentation of valid configuration and provide total compatibility with existing network configs. This should make it easier to safely add new config later (which I'll be doing in #815)

### DIFF
--- a/cmd/src/cli/run.rs
+++ b/cmd/src/cli/run.rs
@@ -49,7 +49,7 @@ pub fn run(package: bool, port: u16, persist: bool) -> DefaultResult<()> {
         agent: AGENT_CONFIG_ID.into(),
         logger: Default::default(),
         storage,
-        network: Some(P2pConfig::default_mock().as_str()),
+        network: Some(P2pConfig::default_mock("mock@cmd::run").as_str()),
     };
 
     let interface_config = InterfaceConfiguration {
--- a/container_api/src/config.rs
+++ b/container_api/src/config.rs
@@ -342,7 +342,7 @@ pub mod tests {
     use holochain_core::context::mock_network_config;
 
     pub fn example_serialized_network_config() -> String {
-        String::from(mock_network_config())
+        String::from(mock_network_config("TODO give unique string"))
     }
 
     #[test]
--- a/container_api/src/container.rs
+++ b/container_api/src/container.rs
@@ -23,7 +23,6 @@ use std::{
     thread,
 };
 
-use holochain_net::p2p_config::P2pConfig;
 use interface::{ContainerApiBuilder, InstanceMap, Interface};
 /// Main representation of the container.
 /// Holds a `HashMap` of Holochain instances referenced by ID.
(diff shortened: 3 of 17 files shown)


ddd-mtl:
I'm really upset about this PR and really wished you talked to me about it before doing this.
With @neonphog we explicitly decided NOT to do structs and use a Json string since network configs are not set in stone yet.
Again, people are rewritting the code I did for no added value and instead will slow me down by having me handle more merge conflicts.

i'm trying to hold coherence for the net_* crates so please don't touch any code there without talking to me about it.

maackle:
@ddd-mtl sorry, I should have definitely give more context. This was just an experiment that took me about an hour or so to do and I'm happy to throw it out. I need to add a piece of config specific to Mock networking and it started to feel gnarly to have separate valid config states for different kinds of networks, so then I made an enum, and then I got carried away. But in the course of doing this, I understand how the config works better and feel comfortable making the change I need with invalidated JSON.

maackle:
I don't plan to merge this now, but want this to be open for discussion, and because we may use structs for network config some day.

maackle:
See #820 for a minimally invasive alternative

thedavidmeister:
OK to be fair, @maackle is just implementing what is documented in the mdbook so can we not jump on him for it?

i don't understand how strings can possibly make things faster to refactor in a strongly typed language, but let's not get lost in the weeds over that point...

merge conflicts are a PITA, but the whole team needs to be able to interact with the networking code on a BAU basis which does mean proposing changes from time to time

e.g. i would like to extend the config in #813 to implement a dummy network to help fix tarpaulin and bring back codecov, and the OP states network config will be extended again somehow in #815 

i would expect an enum backed approach to be the fastest way to deliver #813 and #815 together safely, is there any chance of achieving that in the short term, without derailing what @ddd-mtl is doing?

zippy:
One of the big issues with the networking code is that changes on one side (rust) will break code on the other (javascript) and this wont get detected by CI.  So part of the reason for @ddd-mtl request is to prevent that from happening.

But I also think it's ok and appropriate to use PR as a communication tool, i.e. it is a "request" after all and, I don't think any of us would spend our time coding something that feels like it has no value, the question that may need to be resolved in the communication is if the value one person saw (enums vs strings) is worth the hassle or other decisions another person is holding for, and that's what the communication is about, and those are fair questions.

zippy:
Also, I recognize that I approved this without remembering Damien's explicit request in chat to be careful about making changes in the net code at this stage.

ddd-mtl:
@thedavidmeister FYI discussions on this topic has continued on other channels. At the last core dev team standup we all agreed that we need to clarify responsibilities between hc-core and network libs as well as communications channels between the teams.
Also note that in those discussions nobody ever mentioned the mdbook, so I invite you to not use the mdbook as some form of justification of anything as its just documentation for explaining what the code is doing and not the opposite.

maackle:
Assuming we won't be doing this for a while, closing

PR #817, opened 2019-01-04
[DON'T MERGE YET] Each NodeJS container uses a unique MockSystem

maackle:
### part 3 (based on #815)

Use unique MockSystem for each NodeJS container, automatically


maackle:
Irrelevant now that network config comes directly from the container (#833) and containers automagically create a unique container-wide mock (#835)

PR #818, opened 2019-01-04
[WIP]Check available toolchain and upgrade if needed

AshantiMutinta:
This branch upgrades the toolchain if needed with the appropriate default-host.

Issues & Things still need to do
-Still need to write tests
-Uses rustup so it is dependent on it's structure 
-Needs feedback of downloading process to screen
-Need somewhere to put pin repository information like a file
-Takes a while to download toolchain so I do not know if we need an hc setup command? only takes long on the initial download of the toolchain
-rustup dependency on our tests


pjkundert:
This uses `rustup show` and `rustup toolchain install`, much like the (original) Makefile.  This changes the global "Rust" version for all projects.

I would recommend the approach taken by the Makefile in the feature-rustup-version-check branch in https://github.com/holochain/holochain-rust/pull/782; using `rustup override list` and `rustup override set ...`, to change the Rust version for just the holochain-rust project...

AshantiMutinta:
> 
> 
> This uses `rustup show` and `rustup toolchain install`, much like the (original) Makefile. This changes the global "Rust" version for all projects.
> 
> I would recommend the approach taken by the Makefile in the feature-rustup-version-check branch in #782; using `rustup override list` and `rustup override set ...`, to change the Rust version for just the holochain-rust project...
> 
> Otherwise, looks like a promising approach.

Oh yes it uses rustup override 

> This uses `rustup show` and `rustup toolchain install`, much like the (original) Makefile. This changes the global "Rust" version for all projects.
> 
> I would recommend the approach taken by the Makefile in the feature-rustup-version-check branch in #782; using `rustup override list` and `rustup override set ...`, to change the Rust version for just the holochain-rust project...
> 
> Otherwise, looks like a promising approach.


Thank you, this uses rustup override in the end to set for just the project in the last line of the update code :) but yes you are right about override list too! Thank you!

pjkundert:
So it does!  Sorry, I missed that.

maackle:
Cool! I like the idea of `hc package` doing nothing but outputting a message suggesting you run `hc setup` or something, which does the actual upgrade.

cmd/src/workflow/toolchain.rs
```diff
@@ -0,0 +1,129 @@
+use crate::error::DefaultResult;
+use std::process::Command;
+
+pub fn upgrade_toolchain_if_outdated() -> DefaultResult<()> {
+    //todo move this to a file
+    let pinned_toolchain = "nightly-2018-12-06";
```
sphinxc0re: This should be a `const` somewhere

cmd/src/workflow/toolchain.rs
```diff
@@ -0,0 +1,129 @@
...
+        upgrade_toolchain(pinned_toolchain)
+    }
+}
+fn run_rustup_command<'a>(
+    args: &'a Vec<&str>,
+    on_success: &'a Fn() -> DefaultResult<()>,
+    on_fail: &'a Fn() -> DefaultResult<()>,
+) -> DefaultResult<()> {
+    let mut command = Command::new("rustup");
+    args.iter().for_each(|s| {
```
sphinxc0re: Did you try:
```rust
args.iter().for_each(command.arg);
```

cmd/src/workflow/toolchain.rs
```diff
@@ -0,0 +1,129 @@
...
+    }
+}
+
+fn upgrade_toolchain(pinned_toolchain: &str) -> DefaultResult<()> {
+    //get default host
+    let default_host = get_default_host()?;
+    println!("Default host : {:?}", default_host);
+
+    //get complete concat of toolchain + default_host
+    let complete_pinned_toolchain = &*vec![pinned_toolchain, &*default_host].join("-");
```
sphinxc0re: Are all the `&*` necessary?

cmd/src/workflow/toolchain.rs
```diff
@@ -0,0 +1,129 @@
...
+    run_rustup_command(
+        &override_rustup_args,
+        &|| Ok(println!("target added")),
+        &|| Err(format_err!("Failed to add target")),
+    )
+}
+
+fn get_default_host() -> DefaultResult<String> {
+    let default_host_unsanitized = extract_from_rustup("Default host")?;
+    let split_hosts = default_host_unsanitized.split(" ").collect::<Vec<&str>>();
```
sphinxc0re: Please take a look at the `stdlib` docs for `str`: https://doc.rust-lang.org/std/primitive.str.html#method.split_whitespace

You should always look at the `stdlib` before solving a problem yourself.

sphinxc0re: Also, no need for turbofish:
```suggestion
    let split_hosts: Vec<_> = default_host_unsanitized.split(" ").collect();
```

cmd/src/workflow/toolchain.rs
```diff
@@ -0,0 +1,129 @@
...
+    let mut split_string = split_hosts.iter();
+    split_string.next();
+    split_string.next();
```
sphinxc0re: Look into the `stdlib`:
There is the `.nth` method on Iterator

cmd/src/workflow/toolchain.rs
```diff
@@ -0,0 +1,129 @@
...
+    Ok(String::from(*split_string.next().clone().ok_or_else(
+        || format_err!("Could not get default toolchain"),
+    )?))
+}
+
+fn get_override_toolchain() -> DefaultResult<String> {
+    let default_host_unsanitized = extract_from_rustup("(directory override")?;
+    let split_hosts = default_host_unsanitized.split(" ").collect::<Vec<&str>>();
+
+    if split_hosts.len() > 0 {
```
sphinxc0re:
```suggestion
    if !split_hosts.is_empty() {
```

PR #819, opened 2019-01-04
Better logging

zippy:
This PR adds:
- [x] ChanelLogger: a generic implementation of Logger that sends the log message over a channel.
- [x] a DebugLogger at the container level which uses ChannelLogger for collect log messages
- [x] a LogRules which allow you to match for log messages you want (or don't want) and colorize them
- [x] colorization by instance id
- [x] updating of current log messages in the code to follow a convention of classifying them by type (debug/notify/warn/error) and code section.
- [x] removes configuration of logger at the instance level and promotes it to the container level


container_api/src/config.rs
```diff
@@ -505,6 +504,13 @@ pub mod tests {
     file = "/tmp/holochain.sock"
     [[interfaces.instances]]
     id = "app spec instance"
+
+    [logger]
+    type = "debug"
+    [[logger.rules.rules]]
```
maackle: awkward but ok :)

container_api/src/logger.rs
```diff
@@ -0,0 +1,169 @@
...
+    for b in text.to_string().into_bytes() {
+        total += b;
+    }
+    ID_COLORS[(total as usize) % ID_COLORS.len()]
+}
+
+// renders a log message, using the id color if no color specified for the message.
+pub fn render(msg: LogMessage) {
+    let id_color = pick_color(&msg.id);
+    let msg_color = match msg.color == "" {
```
maackle: Why not `Option<String>` for color?

container_api/src/config.rs
```diff
@@ -43,8 +44,22 @@ pub struct Configuration {
...
+/// There might be different kinds of loggers in the future.
+/// Currently there is a "debug" and "simple" logger.
+/// TODO: make this an enum
+#[derive(Deserialize, Serialize, Clone, Default)]
+pub struct LoggerConfiguration {
+    #[serde(rename = "type")]
+    pub logger_type: String,
+    #[serde(default)]
+    pub rules: LogRules,
+    //    pub file: Option<String>,
```
maackle: remove?

container_api/src/logger.rs
zippy: fixed

ddd-mtl:
How about these standard logging levels:
https://www.tutorialspoint.com/log4j/log4j_logging_levels.htm

ddd-mtl:
@zippy There are inconsistencies between "err" and "error". Might be easier with an enum?

PR #820, opened 2019-01-04
Add "network name" as mock network config

maackle:
This is a rewrite of #816 as a minimally invasive alternative. It doesn't convert network config to structs, it just adds another way to interpret the json blob, and lets `MockWorker` accept that blob of config. It's needed for #815 to allow multiple independent `MockSingletons` (soon to be called `MockSystem`s, or whatever else we want).

Main changes: 
- `MockWorker::new` has to take a `JsonString` of config (just like `IpcWorker`), and the `MockWorker` struct stores the network name.
- All places which create a mock network for testing or otherwise also need to pass a network name as a string.

### DIFF
--- a/net/src/p2p_config.rs
+++ b/net/src/p2p_config.rs
@@ -85,8 +85,8 @@ impl P2pConfig {
             .expect("file is not a proper JSON of a P2pConfig struct")
     }
 
-    pub fn default_mock() -> Self {
-        P2pConfig::from_str(P2pConfig::DEFAULT_MOCK_CONFIG)
+    pub fn default_mock(network_name: &str) -> Self {
+        P2pConfig::from_str(&P2pConfig::default_mock_config(network_name))
             .expect("Invalid backend_config json on P2pConfig creation.")
     }
 
@@ -98,10 +98,17 @@ impl P2pConfig {
 
 // statics
 impl P2pConfig {
-    pub const DEFAULT_MOCK_CONFIG: &'static str = r#"{
+    pub fn default_mock_config(network_name: &str) -> String {
+        format!(
+            r#"{{
     "backend_kind": "MOCK",
-    "backend_config": ""
-    }"#;
+    "backend_config": {{
+        "networkName": "{}"
+    }}
+}}"#,
+            network_name
+        )
+    }
 
     pub const DEFAULT_IPC_CONFIG: &'static str = r#"
     {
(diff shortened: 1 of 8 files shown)


net/src/p2p_network.rs
```diff
@@ -84,7 +86,11 @@ mod tests {
 
     #[test]
     fn it_should_create_mock() {
-        let mut res = P2pNetwork::new(Box::new(|_r| Ok(())), &P2pConfig::default_mock()).unwrap();
+        let mut res = P2pNetwork::new(
+            Box::new(|_r| Ok(())),
+            &P2pConfig::default_mock("TODO give unique string"),
+        )
+        .unwrap();
         res.send(Protocol::P2pReady).unwrap();
```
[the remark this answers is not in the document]
maackle: I guess through all the other tests that rely on a mock network. Also we don't use mock network outside of tests anyway so...

ddd-mtl:
@maackle  hy not use snowflake to generate an unique id instead?

net/src/p2p_config.rs
```diff
@@ -85,22 +85,41 @@ impl P2pConfig {
...
+        )
+    }
 }
 
 // statics
 impl P2pConfig {
-    pub const DEFAULT_MOCK_CONFIG: &'static str = r#"{
-    "backend_kind": "MOCK",
-    "backend_config": ""
+    pub const DEFAULT_MOCK_CONFIG: &'static str = r#"
```
ddd-mtl: might as well remove this, since its not used anymore.

maackle:
`DEFAULT_MOCK_CONFIG` is still used in one place.

I also like the idea of using snowflake, but not for the default, because I feel that if you invoke default_mock multiple times, you should get the same network. For instance, if you spin up several instances in a loop and call `default_mock` each time.

I would propose three ways of creating a mock network then:

- `default_mock` which uses the string "default-mock" for the name
- `unique_mock` which uses snowflake
- `named_mock` which lets the user refer to a particular name

Thoughts @ddd-mtl @lucksus?

net/src/p2p_config.rs
maackle: `DEFAULT_MOCK_CONFIG` is still used in one place, but doesn't show up in the diff

ddd-mtl:
@maackle Do you need `named_mock` if you have `unique_mock` ?

maackle:
@ddd-mtl I would say we only need `named_mock` and `unique_mock`, and can drop `default_mock`. In the majority of cases, you would want to use `unique_mock`, but `named_mock` can be useful for testing and in situations where you actually intend to refer to the same mock network in two different places (this happens in at least one test). So I think we need `named_mock`. However `default_mock` is now just a named mock with a hardcoded name.

maackle:
@ddd-mtl in the previous commit I remove the only remaining reference to default_mock and replace it with named_mock, if you don't like that I will revert

ddd-mtl:
That also works for me 👍

PR #822, opened 2019-01-06
Scenario API

maackle:
Framework for running deterministic multi-agent scenario tests. Based on #830, **merge that first**.

This branch includes none of the shutdown behavior I was working on (#803). It seems the main thing preventing me from running non-overlapping tests with independent Containers was the `MockSingleton` which only allowed for one mock network. In this branch, each new test spins up a new Container which also has its own independent mock network. 

The instances and networks never shut down, but it's OK because the networks are independent. We should still address the shutdown behavior later, but it turns out it's not necessary for this branch to work.

### TODO

- [x] more extensive tests for `callSync` and `callWithPromise`
- [x] update README
- [x] more comments
- [x] add nodejs_container/test to CI tests
- [x] update CHANGELOG


zippy:
CI failing though...

maackle:
Whoa @zippy that was a quick review, there is a lot in here. 

@zippy will we have a container-renaming conversation any time soon? I kept it Habitat because that's what I had in the previous branch. I'll change it to Container or whatever we decide to name it before merging.

maackle:
Nevermind, did the rename now.

maackle:
Rebased on #835 and force-pushed 'cause the commit history was getting gnarly

maackle:
@thedavidmeister can you check the CI config to see if I hooked up these new tests right? (writing this message before seeing if CI even passes...)

zippy:
oh, also @maackle please remember to update the changelog before  merge.

maackle:
For people reading this...I know this is a big PR, but honestly you can safely skim through huge chunks of this. You don't have to read the pure additions like waiter.rs, though that is the heart of this PR. Also there are many routine changes to the app spec tests and other files that you can visually see are not interesting, just noisy. I think you'll be able to see which parts are important to focus on.

nodejs_container/index.js
```diff
@@ -23,26 +69,97 @@ Container.prototype.call = function (id, zome, trait, fn, params) {
...
+     *          assert(resultAlice === resultBob)
+     *          stop()
+     *      })
+     */
+    run(fn) {
+        const container = Container.withInstances(this.instances, this.opts)
+        container.start()
+        const callers = {}
+        this.instances.forEach(instance => {
+            const id = makeInstanceId(instance.agent.name, instance.dna.name)
```
lucksus: I‘m not fully on board with hard-coding the instance ID from agent and DNA. I had this case recently where I wanted to have multiple (separate) instances in separate tests (instances that wouldn’t even communicate). I had to add another agent because if this.

I guess you would say with these changes I could/should create separate scenarios for such a case?

nodejs_container/native/src/js_test_container.rs
```diff
@@ -0,0 +1,226 @@
...
+};
+use holochain_core_types::{
+    cas::content::Address, dna::capabilities::CapabilityCall, entry::Entry,
+};
+use holochain_node_test_waiter::waiter::{CallBlockingTask, ControlMsg, MainBackgroundTask};
+
+/// Block until Hold(agent.public_address) is seen for each agent in the container.
+/// NOTE that this consumes a bunch of signals related to initialization!
+/// The `Waiter` currently doesn't care about these, but beware.
+fn await_held_agent_ids(config: Configuration, signal_rx: &SignalReceiver) {
```
lucksus: Nice!

nodejs_container/native/src/js_test_container.rs
```diff
@@ -0,0 +1,226 @@
...
+                        let background_task = MainBackgroundTask::new(signal_rx, sender_rx, tc.is_running.clone());
+                        background_task.schedule(js_callback);
+                        tc.is_started = true;
+
+                    })
+                })
+            };
+
+            result.or_else(|e| {
+                println!("Indeed an error: {:?}", e);
```
lucksus: Removed println?
```suggestion
```

nodejs_container/native/src/js_test_container.rs
```diff
@@ -0,0 +1,226 @@
...
+                cx.throw_error(format!("unable to start container: {}", e))
+            }).map(|_| {
+                println!("No error whatsoever");
```
lucksus: Remove println?
```suggestion
```

nodejs_container/index.js
lucksus: ... and I guess you‘d be right. It’s just that I think it might be worthwhile to also expose the ability to explicitly set the instance id.. if you know what you‘re doing..

maackle: @lucksus I agree, actually I'm not sure if the "name" part of the Config.instance is or isn't used as the instance ID, but it probably should be. Happy to make that change in a future revision.

Also the thing about scenario tests now is that *every* test is separate from each other test, no instances communicate across tests. So would you still need to create multiple instances as you describe in that case?

thedavidmeister:
@maackle @willemolding  just reiterating for context (we talked about it but i don't think we wrote anything down) that the waiter approach is only "deterministic" in the sense that passing tests continue to pass "eventually" (still an improvement over the false negatives we suffered from previously)

bugs can still cause flaky (false positives) behaviour ("sometimes" errors/panics before the final success condition is reached) or hanging tests (the success condition is never reached)

the waiter itself does nothing (other than spamming debug logs) to reliably reproduce nor debug nor provide reliable evidence of the absence of these types of bugs

deterministic tests can only be achieved via externally driven ticks of both the action and network loops "stepping through" various edge cases

PR #823, opened 2019-01-06
821 container http interface

### LINKED ISSUE #821: Rust container should support HTTP interface
Most of the work is in place. Should be a small job to use jsonrpc-http-server crate as well as jsonrpc-ws-server crate with ability to select from container-config.toml

willemolding:
Closes #821 

Most of the groundwork already done for this one. It really just activates the ability to make HTTP json-rpc calls


container_api/src/container.rs
```diff
@@ -293,9 +293,10 @@ impl<'a> TryFrom<&'a Configuration> for Container {
 
 /// This can eventually be dependency injected for third party Interface definitions
 fn make_interface(interface_config: &InterfaceConfiguration) -> Box<Interface> {
-    use interface_impls::websocket::WebsocketInterface;
+    use interface_impls::{http::HttpInterface, websocket::WebsocketInterface};
     match interface_config.driver {
         InterfaceDriver::Websocket { port } => Box::new(WebsocketInterface::new(port)),
+        InterfaceDriver::Http { port } => Box::new(HttpInterface::new(port)),
         _ => unimplemented!(),
```
maackle: Let's remove this `_` catch-all now that we're handling every variant.

willemolding: what about the `DomainSocket` and `Custom` variants? I think we need to keep the default

PR #824, opened 2019-01-07
noodling with libzmq linking

thedavidmeister:
## changes

- includes our own `zmq-sys` so that `LIBZMQ_PREFIX` is never needed
  - effectively a simplified, hardcoded version of what the environment variables do upstream
  - sits the zmq vendor file inside this
- adds an install script for ubuntu that includes a working libzmq
- runs against a cold `ubuntu` box on circle to demonstrate installation (develop branch only)
- removes `LIBZMQ_PREFIX` from windows builds on travis
- copies the zmq dll into the `CARGO` dir so it can be found on the path by the compiler build scripts
- removes the `PATH` logic for libzmq on travis/windows
- adds `--release` to cargo tests on travis to alleviate pressure on the cache (was timing out uploads)
- updates docs
- shifts `test cmd` to circle because it breaks on travis (doesn't run standard installation)


zmq-sys/build.rs
```diff
@@ -0,0 +1,34 @@
...
+        .and_then(|path| path.to_str().map(|p| p.to_owned()))
+}
+
+fn main() {
+    println!("cargo:rerun-if-env-changed=CARGO_MANIFEST_DIR");
+
+    #[cfg(windows)]
+    // hack to get the libzmq dll on the PATH
+    // copies it next to cargo
+    // kind of messy, is less than 1MB
```
neonphog: :+1: I like this as an interim solution until we get static linking working... since this is where cargo install will put the binaries in any case.

Connoropolous:
this is a big improvement for usability, externally. Yay! 

Will seek to bring any related documentation in line with this

PR #827, opened 2019-01-08
NodeJS API touchups

maackle:
Just a small change to make the nodejs container more consistent with the Scenario API work (#822). Let's leave this open until it's time for 0.0.3 in case I find more, but this should definitely go into 0.0.3.

### DIFF
--- a/cmd/README.md
+++ b/cmd/README.md
@@ -40,6 +40,10 @@ Run `hc --version` to confirm.
 
 Run `hc help` for help.
 
+### Networking
+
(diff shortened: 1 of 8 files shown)


zippy:
hey @maackle when you update these things in the test framework, they also need to be updated in the scaffold that gets created when we do `hc init`.  Currently I've been keeping track of this in #706.  So please take a quick look at that to approve, and then we can make these kinds of updates in both places at the same time.

PR #829, opened 2019-01-08
Signature refactor

sphinxc0re:


sphinxc0re:
CI should be green but is erroring for some unknown reason

Connoropolous:
No at this point I don’t believe so. Now I know of the change I’ll keep my eyes on it. Why the change?

zippy:
The change just makes the hdk a little more ergonomic and efficient.  No need clone() when calling because the address of the Address is plenty to do the work that needs doing.

PR #830, opened 2019-01-08
Make MockSingleton multiply instantiable (rebased on #820)

maackle:
I cherry-picked commits from #815 onto `develop` to build this branch, rather than dealing with merge conflicts of a true rebase. It also has one extra commit to fix a mistake.

### DIFF
--- a/net/src/mock_worker.rs
+++ b/net/src/mock_worker.rs
@@ -280,7 +280,11 @@ impl NetWorker for MockWorker {
     /// forward to our mock singleton
     fn receive(&mut self, data: Protocol) -> NetResult<()> {
         let map_lock = MOCK_MAP.read().unwrap();
-        let mut mock = map_lock.get(&self.network_name).unwrap().lock().unwrap();
+        let mut mock = map_lock
+            .get(&self.network_name)
+            .expect("MockSystem should have been initialized by now")
+            .lock()
+            .unwrap();
         if let Ok(wrap) = ProtocolWrapper::try_from(&data) {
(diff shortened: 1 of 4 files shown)


zippy:
CI broken...

net/src/mock_worker.rs
```diff
@@ -284,17 +279,24 @@ impl NetWorker for MockWorker {
...
+
                 let (tx, rx) = mpsc::channel();
                 self.mock_msgs.push(rx);
                 mock.register(&app.dna_address, &app.agent_id, tx)?;
                 return Ok(());
             }
         }
-
+        let map_lock = MOCK_MAP.read().unwrap();
+        let mut mock = map_lock.get(&self.network_name).unwrap().lock().unwrap();
```
ddd-mtl: I don't believe the "get" is guaranteed to work so I'd put an expect instead of unwrap here.

maackle:
So far I'm stumped why the last app_spec test fails here. It passed on develop. This is a one-file change. Does anyone have an idea?

maackle:
I wrote #835 which changes an aspect of container network config initialization and makes this PR work. I rebased on that PR and force-pushed.

maackle:
@neonphog yeah I'm punting on shutdown. Shutdown would get us to the point of running independent tests in nodejs, but we still need multiple mock networks if we want parallel tests, and multiple mocks also gets us to independent tests (I hope)

maackle:
to david's last point, we picked up the conversation here: https://github.com/holochain/holochain-rust/issues/841

PR #835, opened 2019-01-09
Refactor Container P2pConfig generation

maackle:
So far, the mock network case has been treated as a default fallback. Now we need a little more fine-grained control over mock network config and creation. Currently, the only way to have a Container use a mock network is to pass it `None` network config. The problem is that it generates the mock config every time an instance is created, which works fine when there is only one mock network, but now that there can be multiple, this has the effect of putting each instance in isolation, since it uses `unique_mock_config` as its default. We need to generate the mock config once for the container rather than once per instance, so each instance shares the same one.

This PR shuffles the code around significantly, but basically the main thing it does is replace `Container::network_ipc_uri` with `Container::p2p_config` which stores the entire config. So, rather than store the one small piece of state needed to generate an IPC-based config, and then generating the config from scratch for each new instance, it generates the config once, stores that as state, and uses that "memoized" config for each new instance. This accomplishes the desired outcome of having instances share the same mock network. I assume that no matter what kind of config we create, we want each instance to share the exact same one, so hopefully this will help with future changes to config.

### DIFF
--- a/container_api/src/container.rs
+++ b/container_api/src/container.rs
@@ -1,7 +1,5 @@
 use crate::{
-    config::{
-        Configuration, InterfaceConfiguration, InterfaceDriver, NetworkConfig, StorageConfiguration,
-    },
+    config::{Configuration, InterfaceConfiguration, InterfaceDriver, StorageConfiguration},
     context_builder::ContextBuilder,
     error::HolochainInstanceError,
     logger::DebugLogger,
@@ -50,7 +48,7 @@ pub struct Container {
     dna_loader: DnaLoader,
     signal_tx: Option<SignalSender>,
     logger: DebugLogger,
-    network_ipc_uri: Option<String>,
+    p2p_config: Option<JsonString>,
     network_child_process: NetShutdown,
 }
 
@@ -82,7 +80,7 @@ impl Container {
             dna_loader: Arc::new(Box::new(Self::load_dna)),
             signal_tx: None,
             logger: DebugLogger::new(rules),
-            network_ipc_uri: None,
+            p2p_config: None,
             network_child_process: None,
         }
     }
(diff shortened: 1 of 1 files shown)


maackle:
By the way, I'm kind of annoyed at how we have no choice but to use opaque JsonStrings to feed to `P2pConfig`. Having a `Option<JsonString>` as struct state makes me sad twice, once for the Option, once for the JsonString. I'm concerned that it's going to get unwieldy fast as the config spec grows and we reference it more and more. If that trend continues and the network crate maintainers don't want to use structs soon, I am considering porting #816 as a `core_types` struct that serializes to valid P2pConfig JsonStrings that core crates can use instead of `JsonString::from(json!([hand-coded json here]))` to pass off to network crates.

container_api/src/container.rs
```diff
@@ -191,49 +189,60 @@ impl Container {
...
+            Some(ref net_config) => {
+                let uri = self
                     .config
                     .clone()
                     .network
                     .unwrap() // unwrap safe because of check above
                     .n3h_ipc_uri
                     .clone()
                     .or_else(|| self.spawn_network().ok());
+                JsonString::from(json!(
```
thedavidmeister: can we make this a struct in core types?

neonphog:
> By the way, I'm kind of annoyed at how we have no choice but to use opaque JsonStrings to feed to P2pConfig

@maackle @thedavidmeister @ddd-mtl 

My feeling on this is that there are several categories of configuration here:

- `truly opaque` - config that does not need to be interpreted by rust at all, and should just be passed along as a string (probably internally json) to whatever networking module is being used
- `truly non-opaque` - things like the `network_kind` that MUST be interpreted by rust, and therefore absolutely should be structs / strongly typed
- `in-between` - this is where I think much of the controversy comes in. There is some configuration that doesn't *really* matter to rust... it is just stuff that needs to be passed along to the *specific* networking module that needs it ... it needs to be in the container config somewhere, but holochain-rust core does *not* care about it. I don't see any particular reason that `n3h_ipc_uri` should be a struct in core_types... The ultimate incarnation of networking is not going to care about that... maybe the `net_ipc` incarnation of the networking module should have a struct to parse it, but why does core_types need to do that?

container_api/src/container.rs
maackle: I just moved the json! from one place to another, let's deal with it in another PR. I made an issue: https://github.com/holochain/holochain-rust/issues/841

thedavidmeister:
@neonphog @maackle 

the "opaqueness" doesn't alleviate my concerns much at all but this isn't the place for that discussion

i'm approving so we can move on

PR #840, opened 2019-01-09
Container Admin Functions (DNA/instance management)

lucksus:
Everything in grey is part of this PR:
screenshot 2019-01-12 at 01 40 20

## Added container API functions
* admin/dna/install_from_file({id, path})
* admin/dna/uninstall({id})
* admin/dna/list()
* admin/instance/add({id, dna_id, agent_id})
* admin/instance/remove({id})
* admin/instance/start({id})
* admin/instance/stop({id})
* admin/instance/list()
* admin/instance/running()

Which get added to an interface if `admin = true`.

## Prework / Refactoring

* Had to make the container live in a static place so the interface threads can get a mutable reference for calling admin functions. *Note though that I left this lazy_static! CONTAINER reference as something optional to use so tests and holochain_nodejs can own a Container instance and don't run into problems as @maackle pointed out..*

* Interfaces can now be shutdown which happens now in `Container::shutdown()` but not (yet !) between admin calls. We will need that to have changes (like new installed DNA) be reflected in the registered JSON RPC functions but I've left this as a follow-up since this PR starts to get big..

## Follow-ups
1. The big thing that is missing from this is updating interfaces after changes to the config. I almost added that to this PR already but want to think about a good solution first though. Easy catch would be to completely reload the config after any change but I think we can be more intelligent than that. I currently favor adding functions for handling interfaces (which we need anyway - like adding/removing interfaces and adding/removing instances from/to interfaces) and include a restart/update function there. *Note that this would transfer responsibility for deciding when to update which interface over to the UI, i.e. HCadmin*. (@zo-el, @JettTech)

2. File/storage management. Currently the path to the DNA file to be installed is left as it is and added to the config. I would like to have admin/dna/install_from_file copy the the DNA file into a managed directory (~/.holochain/dnas/, or that is <config-path>/dnas). Next to that we could then also configure instances to have their storage in ~/.holochain/instances/<id>/storage (or <config-path>/instance/<id>/storage) - *currently the storage is hard-coded to memory (!).*


maackle:
I'm concerned about introducing a singleton for this. Singletons and tests do not mix well. As long as we can keep this singleton completely out of tests, it should be OK. But the moment we want two Containers in one process, we are screwed.

I think the way to do this without a singleton would be to build the IoHandler and the Container at the same time. ContainerApiBuilder might become ContainerBuilder. The ContainerBuilder would build the Container and return it as an `Arc<Mutex<Container>>`, then clone that Arc to construct the IoHandler. Then the two become tightly coupled, and you can only interact with the Container through the mutex, but you can have multiple containers in multiple mutexes.

willemolding:
Making this visible, from Lucksus:

"`admin/dna/install_from_file` should work and I tried to confirm by actually using hc-web-client to connect to an admin interface (used container/example-configs/basic.toml for that) but I get "Method not found" for the admin function and also for info/instances which is strange. Tried to debug a bit but couldn't find the reason yet."

willemolding:
In response to the above I can call `info/instances` and `admin/dna/install_from_file` using cURL with no problems. To recreate
```
holochain_container -c ./example-config/empty-container.toml
```
and 
```
curl -X POST -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","id":"0","method":"info/instances"}' http://localhost:4000

curl -X POST -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","id":"0","method":"admin/dna/install_from_file"}' http://localhost:4000
```

willemolding:
`admin/dna/install` is not working great. I started on `admin/instance/start` but I've come up against an issue. It is possible to add a new instance to the config and get the container to start the instance but it also needs to rebuild its rpc interface to allow calls to the new instance. I am not sure of the best way to go about that yet.

lucksus:
@willemolding, yeah, after the config was changed, the container actually needs to reload the config in a proper way.. I was anticipating that.

container_api/src/error.rs
```diff
@@ -29,6 +31,7 @@ impl Error for HolochainInstanceError {
             HolochainInstanceError::InternalFailure(ref err)  => Some(err),
             HolochainInstanceError::InstanceNotActiveYet => None,
             HolochainInstanceError::InstanceAlreadyActive => None,
+            HolochainInstanceError::NoSuchInstance => None,
```
maackle: How hard would it be to add an ID to these three errors? (InstanceNotActiveYet, InstanceAlreadyActive, NoSuchInstance)? These parameterless errors are tough to track. I could believe that it's hard to add, but just asking.

container_api/src/interface_impls/websocket.rs
```diff
@@ -12,12 +13,12 @@ impl WebsocketInterface {
...
 impl Interface for WebsocketInterface {
-    fn run(&self, handler: IoHandler) -> Result<(), String> {
+    fn run(&self, handler: IoHandler, kill_switch: Receiver<()>) -> Result<(), String> {
         let url = format!("0.0.0.0:{}", self.port);
-        let server = ServerBuilder::new(handler)
+        let _server = ServerBuilder::new(handler)
             .start(&url.parse().expect("Invalid URL!"))
             .map_err(|e| e.to_string())?;
-        server.wait().map_err(|e| e.to_string())?;
+        let _ = kill_switch.recv();
```
maackle: Nice!

container_api/src/interface.rs
```diff
@@ -156,10 +189,155 @@ impl ContainerApiBuilder {
...
+        });
+
+        self.io.add_method("admin/instance/stop", move |params| {
+            let params_map = Self::unwrap_params_map(params)?;
+            let id = Self::get_as_string("id", &params_map)?;
+            container_call!(|c| c.stop_instance(&id))?;
+            Ok(serde_json::Value::String("success".into()))
+        });
+
+        self.io.add_method("admin/instance/list", move |_params| {
```
maackle: Feel free to remove `info/instances` if it hasn't been already, I think this is pretty much the same.

Although, `info/instances` needs to be accessible from every UI, i.e. without admin privileges, so this should be taken out of the admin namespace?

container_api/src/container.rs
```diff
@@ -42,10 +72,11 @@ use interface::{ContainerApiBuilder, InstanceMap, Interface};
...
 pub struct Container {
-    instances: InstanceMap,
-    config: Configuration,
-    interface_threads: HashMap<String, InterfaceThreadHandle>,
-    dna_loader: DnaLoader,
+    pub(crate) instances: InstanceMap,
+    pub(crate) config: Configuration,
+    pub config_path: PathBuf,
+    interface_threads: HashMap<String, Sender<()>>,
+    pub(crate) dna_loader: DnaLoader,
```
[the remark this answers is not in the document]
lucksus: Well, they are *just* crate public.. I guess we also make them only `ContainerAdmin` public.. but that, I guess, would be easier to reason about than having setters.


I regard `ContainerAdmin` as part of the container but wanted to create another bucket instead of putting all new functions into the already quite big `struct Container` and file container.rs.

container_api/src/interface.rs
lucksus: I first mistook those two for the same too, but actually `info/instances` is something different:

* Yes, `info/instances` needs to be accessible in every interfaces as it lists all the instances available here.
* `admin/instance/list` is only available to admins and lists all instances that are configured.

maackle: Ahh yes OK

container_api/src/container.rs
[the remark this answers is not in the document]
maackle: `Yeah `pub(crate)` helps, but it still means we have to consider every file in the crate (better than every file in the code base at least) to understand how those members are being used. IMO it's always useful to spend the small bit of energy upfront to avoid this, if it is indeed a small amount of energy. It's so much harder to go back to private later once we've leaked a member out into the public. IMO ideally every public member should have a *very* good reason for being publicly mutable.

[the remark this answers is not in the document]
lucksus: I tried `pub(in crate::container_admin::ContainerAdmin)` but that didn't work, I think because the other module needs to be parent.. So the alternative would be merging `ContainerAdmin` into container which I think is making it harder to read than this pub(crate).. I'm very open to suggestions :)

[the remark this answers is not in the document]
maackle: I haven't looked into why these things need to be referenced outside of the container but would it be possible to just write getters and setters?

container_api/src/config.rs
```diff
@@ -34,9 +34,11 @@ use toml;
 pub struct Configuration {
     /// List of Agents, this mainly means identities and their keys. Required.
     pub agents: Vec<AgentConfiguration>,
-    /// List of DNAs, for each a path to the DNA file. Required.
+    pub default_agent_id: Option<String>,
```
willemolding: This doesn't get used. I added it and forgot to remove it

container_api/src/container_admin.rs
```diff
@@ -0,0 +1,528 @@
...
+
+        let add_result = container.add_instance(InstanceConfiguration {
+            id: String::from("new-instance"),
+            dna: String::from("new-dna"),
+            agent: String::from("test-agent-1"),
+            storage: StorageConfiguration::Memory,
+        });
+
+        assert_eq!(add_result, Ok(()))
+    }
```
willemolding: Should include a more comprehensive test that checks the config was actually modified

container_api/src/interface.rs
```diff
@@ -156,10 +189,155 @@ impl ContainerApiBuilder {
...
+        self.io
+            .add_method("admin/dna/install_from_file", move |params| {
+                let params_map = Self::unwrap_params_map(params)?;
+
+                let id = Self::get_as_string("id", &params_map)?;
+                let path = Self::get_as_string("path", &params_map)?;
+
+                container_call!(|c| c.install_dna_from_file(PathBuf::from(path), id.to_string()))?;
+
+                Ok(serde_json::Value::String("success".into()))
```
willemolding: From the call this morning it sounds like returning a JSON string messes with the hc-web-client or hc-redux-middleware. We could change this to a full json object like
```
{success: true}
```
but this seems kind of redundant since it is already returning `Ok`. 

I believe problem comes from this line https://github.com/holochain/hc-redux-middleware/blob/e50540d24466c9ef1b4ac9d5dd4c1384b8565197/src/lib/middleware.ts#L25

I will test @thedavidmeister 's suggestion of using a RawString and will see if this can be parsed successfully

willemolding: There is also some inconsistency between 'info/instances' and 'admin/dna/list' as to if they return a stringified result or a regular json result. For an empty config we have
'info/instances': `{"jsonrpc":"2.0","result":"[]","id":"0"}`
'admin/dna/list': `{"jsonrpc":"2.0","result":[],"id":"0"}`

container_api/src/config.rs
willemolding: removed it now

container_api/src/interface.rs
```diff
@@ -156,10 +189,158 @@ impl ContainerApiBuilder {
...
+        match params {
+            Params::Map(map) => Ok(map),
+            _ => Err(jsonrpc_core::Error::invalid_params("expected params map")),
+        }
+    }
+
+    /// To be consistent with holochain function calls the expected output of the admin calls
+    /// is stringified JSON. This performs that stringification and then wraps it in the format expected
+    /// by the jsonrpc add_method closure
+    fn format_response(response: &serde_json::Value) -> jsonrpc_core::Result<serde_json::Value> {
```
willemolding: Added this to produce the json string format that is required

container_api/src/interface.rs
```diff
@@ -156,10 +189,158 @@ impl ContainerApiBuilder {
...
+    }
+
+    pub fn with_admin_dna_functions(mut self) -> Self {
+        self.io
+            .add_method("admin/dna/install_from_file", move |params| {
+                let params_map = Self::unwrap_params_map(params)?;
+                let id = Self::get_as_string("id", &params_map)?;
+                let path = Self::get_as_string("path", &params_map)?;
+                container_call!(|c| c.install_dna_from_file(PathBuf::from(path), id.to_string()))?;
+                Self::format_response(&json!({"success": true}))
```
willemolding: This produces a result as expected by hc-web-client/hc-redux-middleware and consistent with the other container functions.

container_api/src/interface.rs
lucksus: As I've pointed out in another comment thread, we want to keep `info/instance` as it is different to `admin/instance/list`. `info/instances/ is available in every interface and listing the instances that are registered in this interface. As long as the interface is not restarted that won't change, yes, but `info/instances` is always correctly showing which instances are there.. ;)

container_api/src/container_admin.rs
```diff
@@ -0,0 +1,615 @@
...
+[[interfaces.instances]]
+id = "test-instance-2"
+
+[interfaces.driver]
+port = 3000
+type = "websocket"
+
+[logger]
+type = ""
+[[logger.rules.rules]]
```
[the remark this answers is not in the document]
lucksus: This needs to be here because the defaults get added when saving the config. So we need to add it to the fixture. I will post my 2nd PR on ContainerAdmin in a few minutes that includes a DRY out of these fixtures..

container/example-config/basic.toml
```diff
@@ -1,46 +1,80 @@
...
+key_file = "holo_tester.key"
 name = "Holo Tester 2"
 public_address = "HoloTester2-----------------------------------------------------------------------AAAGy4WW9e"
-key_file = "holo_tester.key"
 
 [[dnas]]
-id = "app spec rust"
 file = "example-config/app_spec.hcpkg"
 hash = "Qm328wyq38924y"
+id = "app spec rust"
```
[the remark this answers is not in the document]
lucksus: 👍

container_api/src/container.rs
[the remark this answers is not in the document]
lucksus: Setters would make those things writable to everybody which IMHO is worse than making them crate public.

Needed because I've put the container admin functions in a trait `ContainerAdmin` that gets implemented for `Container`.

container_api/src/interface.rs
```diff
@@ -18,6 +22,35 @@ pub trait DispatchRpc {
...
+                        new
+                    })
+            }
+            None => {
+                println!("Admin container function called without a container mounted as singleton!");
+                // If interfaces are supposed to work, the container needs to be mounted to a static place
+                // with container_api::container::mount_container_from_config(config: Configuration).
+                // There are cases in which we don't want to treat the container as a singleton such as
+                // holochain_nodejs and tests in particular. In those cases, calling admin functions via
+                // interfaces (websockt/http) won't work, but also we don't need that.
```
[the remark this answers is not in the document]
maackle: The named mock network thing was a total hack, because MockSingleton was already firmly embedded in the architecture and would have taken too much work to dislodge it. I would have preferred that entities that need a mock network would own their own reference to the network they're on, rather than referring to a string. The container<->interface relationship is still new enough that it would be possible to represent the relationship with proper references, and I hope that if and when we want to have multiple containers, that's still the case, and we don't have to resort to magic strings.

> From perspective of the interface (thread) there should clearly be just one container. Hence singleton.

Agree with the first sentence but not the second. Yes, from the perspective of any interface there should be one container. Interfaces and Containers are many-to-one. In Rust types that means the interface should own a single reference to a container, nothing more. A singleton is sufficient but not necessary to represent a many-to-one relationship. It just says that every Interface refers to the one singleton Container, rather than one arbitrary container.

willemolding:
Format for all admin calls is not unstringified json and all holochain calls return just a string. I am happy now and will update hc-web-client/hc-redux-middleware

container_api/src/container.rs
```diff
@@ -23,17 +26,44 @@ use std::{
...
+    /// the interfaces. That code needs this static variable to be set in order to be able to
+    /// call ContainerAdmin functions.)
+    /// In order to call from interface threads Container admin functions that change
+    /// the config and hence mutate the Container, we need something that owns the Container
+    /// and is accessible from everywhere (esp. those container interface method closures
+    /// in interface.rs).
+    pub static ref CONTAINER: Arc<Mutex<Option<Container>>> = Arc::new(Mutex::new(None));
+}
+
+/// Container constructor that makes sure the instance is mounted in above static CONTAINER.
```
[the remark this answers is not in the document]
lucksus: Yep. Clarified.

lucksus:
@maackle I've implemented the refactoring we talked about. Visibility of Container structs is now restricted to a sub-module:
```rust
pub struct Container {
    instances: InstanceMap,
    pub(in crate::container) instances: InstanceMap,
    config: Configuration,
    pub(in crate::container) config: Configuration,
    interface_threads: HashMap<String, InterfaceThreadHandle>,
    pub(in crate::container) config_path: PathBuf,
    dna_loader: DnaLoader,
    interface_threads: HashMap<String, Sender<()>>,
    pub(in crate::container) dna_loader: DnaLoader,
```

Please go ahead and merge if you approve :)

container_api/src/container_admin.rs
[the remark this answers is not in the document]
maackle: @zippy @lucksus In my scenario-tests branch I make "debug" the default logger_type because I assume that's what we want, would you agree?

maackle:
That's kind of a cool pattern, module-level pub. As long as the module doesn't get huge.

container_api/src/container/base.rs
```diff
@@ -142,11 +195,12 @@ impl Container {
     }
 
     /// Stop and clear all instances
-    pub fn shutdown(&mut self) -> Result<(), HolochainInstanceError> {
-        self.stop_all_instances()?;
-        // @TODO: also stop all interfaces
+    pub fn shutdown(&mut self) {
+        let _ = self
+            .stop_all_instances()
+            .map_err(|error| notify(format!("Error during shutdown: {}", error)));
```
AshantiMutinta: Why did we change the this from returning a result? stop_all_instances() seems to be mapping an error

container_api/src/container/base.rs
```diff
@@ -246,8 +300,7 @@ impl Container {
         }
 
         let config = self.config.clone();
-        self.shutdown().map_err(|e| e.to_string())?;
-        self.instances = HashMap::new();
+        self.shutdown();
```
AshantiMutinta: shutdown could propagate the error up here

container_api/src/container_admin.rs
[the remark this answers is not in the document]
zippy: Yeah probably but this indicates that we need to think through different default conditions.  For running scenario tests, and end user probably wants the default logging just to show `debug/dna` messages, and maybe `err/*`  For the container, the default is likely no debug messages.   For us as holochain core devs, the default is what it currently is:  All messages.

container_api/src/container/base.rs
lucksus: Because if one instance errors during shutdown this function would exit early and not reset `self.instances`.

Logically, shutdown should always work. If an instance complains when being shutdown, it should only show in the log.

PR #846, opened 2019-01-10
added http interface to hc run

zippy:
this just adds an env var so people can use the http interface with `hc run`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Changed
 ### Added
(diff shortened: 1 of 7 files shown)


zippy:
This is a good point about documentation.  Motivation is that I find them a very clean easy and expected way to get data into commands that lends itself to temporary settings via `export` and so on.

zippy:
ok @Connoropolous , I got all the tests refactored and working, locally at least, we'll see what happens with CI.  But I will disable them again before merge because they still do take way too long.

Connoropolous:
@zippy is this for after the release?

zippy:
Yep not part of 0.0.3.  I just pushed turning off that run test, because it takes too long for travis.  You can turn it on and test locally if you like.  Otherwise please re-review so we can merge using http in hc.

PR #850, opened 2019-01-11
WIP: cold os x for travis

thedavidmeister:


zippy:
I merged this into the release branch of 0.0.3 because I needed it for running the release scripts.  But it currently doesn't actually work for the cold os build because of missing yarn and node, and we don't have a good non-destructive way of doing that install as a part of a "install pre-reqs" script that's polite to a users existing setup.  Thus, this will need work.

thedavidmeister:
> we don't have a good non-destructive way of doing that install as a part of a "install pre-reqs" script that's polite to a users existing setup

@zippy what are you thinking for this? (outside of nixos)

Issue #836, opened 2019-01-09
spike: memory


thedavidmeister:
https://hacks.mozilla.org/2018/10/calls-between-javascript-and-webassembly-are-finally-fast-%F0%9F%8E%89/

https://github.com/rustwasm/wee_alloc

thedavidmeister:
https://stackoverflow.com/questions/52022998/how-does-wee-alloc-a-malloc-alternative-know-where-to-start-the-heap-in-webass

thedavidmeister:
http://fitzgeraldnick.com/2018/02/09/wee-alloc.html

thedavidmeister:
https://mnt.io/2018/08/22/from-rust-to-beyond-the-webassembly-galaxy/

thedavidmeister:
ok so based on the above articles, general research (reading code, etc.) and what i already knew, here's some context and things to consider

# memory in wasm has 2 sides

wasm functions are exposed by wasm to be called by "something else"

wasm functions only accept as input, and return as output, integer values

wasm memory can be exposed to the "something else" through pointers

# inside wasm

wasm has linear memory with pages that are 64kb each

wasm memory pages can be added (grown) but cannot be removed dynamically

logic to manage what is allocated to each page, deallocate and then reuse the freed memory is BYO, this is part of the wasm spec and nothing "outside" can help with this AFAIK, other than by calling predefined, exported functions inside wasm (e.g. to allocate/deallocate a string in the outside scope so it can be read inside, processed inside, then the result read outside and then cleaned up outside)

rust has its own memory management as a key selling point, as soon as something goes out of scope it is not in memory any more - this default behaviour makes it _seem_ impossible to use wasm memory _as_ rust memory

wee alloc is a rust crate that swaps the default rust memory management for a wasm memory backed equivalent

benefits of wee alloc:

- tiny compiled size vs. rust default
- wasm memory aware
- native rust compiler support (e.g. can use `Box` to interact with wasm memory as the rust heap)
- supports pages, O(1) memory operations, defragmentation
- well tested and benchmarked (uses quickcheck, etc.)

drawbacks:

- slower than default rust memory management

it seems the key "trick" is to expose dedicated allocate/deallocate functions from wasm to "outside" that call the native rust `mem::forget()` function in `alloc` so that rust "forgets" to clean up the allocated bytes when they go out of scope, then rebinding the "forgotten" memory to a variable that immediately goes out of scope in the `dealloc`

[...]

PR #854, opened 2019-01-12
decoupling wasm memory and ribosome codes

thedavidmeister:
## context

this is prework for the changes researched in https://github.com/holochain/holochain-rust/issues/836

there are two main goals here:

- convert all the relevant ad-hoc usages of `u16` and `u32` into something semantic the compiler understands
- decouple wasm memory stack (de)allocation from (de)encoding ribosome input/output codes

these two changes should pave the way for a smooth upgrade of the stack to `u32` based `MemoryInt` allocation values and `u64` ribosome input/output codes

## changes

naming:

- an `i32` to be handled by wasmi (i.e. `wasmi::RuntimeValue::I32`) is a `RibosomeRuntimeBits`
- a `u32` to be treated as `RibosomeReturnCode` is a `RibosomeEncodingBits`
- a `u16` to be treated as tracking memory (e.g. offset or length) is a `MemoryInt`
- a `u32` to interface with memory encodings and perform "big" calculations is a `MemoryBits`
- new types for `Offset`, `Length` and `Top` in allocation/stacks
- `SinglePageStack` becomes `WasmStack`
- `SinglePageAllocation` becomes `WasmAllocation`
- `RibosomeReturnCode` becomes `RibosomeEncodedValue`
  - because these codes are both input and output to zome functions due to allocations
- functions `min` and `max` added to set bounds for allocation/stacks
- `allocation.encode()` becomes `allocation.as_ribosome_encoding()`
- `RibosomeErrorCode::from_offset` becomes `RibosomeErrorCode::from_code_int`

encodings:

- `RibosomeEncodedValue::Allocation` is a new variant
  - holds `RibosomeEncodedAllocation`
  - `RibosomeEncodedAllocation` should always be a valid `WasmAllocation`
  - implements `Into` for `RibosomeEncodingBits`
- standard `JsonString` and `String` handling in `load_ribosome_encoded_json` and `load_ribosome_encoded_string` for encoded integers

**side effect of this is that the `use` prerequisites of DefaultJson apply to all rust zomes**

@philipbeadle @willemolding ^^

reads:

- all reads are `impl` directly on `WasmAllocation`
- json reads use standard `JsonString` handling

writes:

- all write/(de)allocate functionality is `impl` directly on a `WasmStack`
  - `store_as_json()` => `stack.write_json()`
  - `store_as_json_into_encoded_allocation` => `stack.write_json()`
- json writes use standard `JsonString` handling

hdk:

- all hdk logic that just dispatches to lower level fns + memory/encoding handling is standardised and wrapped in a single `Dispatch` enum/function
  - fixes various places where (de)allocation was forgotten or inconsistent
  - removes special casing for `debug`
  - removes special casing for `remove_entry`

conversions:

- `AllocationResult` can return `AllocationError` when failing to represent a valid `WasmAllocation`
- `WasmAllocation::try_from_ribosome_encoding(RibosomeEncodingBits)` attempts to jump straight from an integer to an allocation, fails if the ribosome encoding doesn't represent a valid offset + length
- `WasmStack::try_from(WasmAllocation)` attempts to create a new stack from an allocation, can fail if the allocation is outside the bounds the stack can handle
- `return_code_for_allocation_result()` jumps from `AllocationResult` to `RibosomeEncodingBits`
- adds various `From` implementations to satisfy the compiler

hygiene:

- removes every unwrap i found along the way, which does result in some boilerplate but i think that can be fixed with macros or similar later
- replaces some types that seem wrong to me (e.g. `usize` ribosome encodings) and would be fragile in the future and/or edge case bugs
- all low level zome fns accept and return `RibosomeEncodingBits`
- a lot of `as N` type conversions disappear when using the named types consistently
- generally `match` against explicit `RibosomeEncodedValue` variants is now preferred, rather than trying to guess what some integer is supposed to be ad-hoc
- init global memory no longer can panic
- rust tests are run for app spec


thedavidmeister:
@maackle so as we discussed in chat, i kind of ran out of time this week for this PR, will be in transit pretty soon

things to wrap up:

- looks like some tests are flaky since i pulled develop in recently, at this point i think that is unrelated to the changes here but  we should confirm that
- app spec tests are definitely broken and need updating
- everything in `wasm_utils::memory::*` needs unit tests
  - there was a good set of unit tests in memory allocation/serialization files, pull them out of git as a decent starting point, definitely don't reinvent the wheel here :)
  - might want a few new tests, but don't go crazy, the old tests were pretty good

hdk-rust/src/api.rs
```diff
@@ -183,38 +183,80 @@ pub enum BundleOnClose {
...
+                Dispatch::GetEntry => hc_get_entry,
+                Dispatch::GetLinks => hc_get_links,
+                Dispatch::LinkEntries => hc_link_entries,
+                Dispatch::InitGlobals => hc_init_globals,
+                Dispatch::EntryAddress => hc_entry_address,
+                Dispatch::UpdateEntry => hc_update_entry,
+                Dispatch::RemoveEntry => hc_remove_entry,
+                Dispatch::Query => hc_query,
+                Dispatch::Send => hc_send,
+            })(encoded_input)
```
lucksus: 🥇

wasm_utils/src/memory/stack.rs
```diff
@@ -0,0 +1,237 @@
...
+
+    pub fn allocate(&mut self, allocation: WasmAllocation) -> Result<Top, AllocationError> {
+        if MemoryInt::from(self.top()) != MemoryInt::from(allocation.offset()) {
+            Err(AllocationError::BadStackAlignment)
+        } else if MemoryBits::from(self.top()) + MemoryBits::from(allocation.length())
+            > WasmStack::max()
+        {
+            Err(AllocationError::OutOfBounds)
+        } else {
+            // @todo i don't know why we return the old top instead of new one?
```
maackle: seems like the few places that use .allocate() (`WasmStack::write_in_wasm_memory` and `SinglePageManager::allocate`) do the allocation in 3 parts:
- `next_allocation`
- then `allocate`
- then `WasmAllocation::new` with the old top and new length

That could probably be consolidated into one function that just returns a `WasmAllocation` in one step...but not a big deal. I was curious too.

thedavidmeister: separate = easier to test

thedavidmeister:
> Awesome work!
> 
> > side effect of this is that the use prerequisites of DefaultJson apply to all rust zomes
> 
> Elaborate on what that means? Will this be an issue if we change the canonical serialization format to bson or something?

@maackle 

iirc much easier/safer to support different serialization consistently with `JsonString` everywhere

compare:

```
serde_json::to_str(Foo)
```

to

```
JsonString::from(Foo)
```

if you wanted to change the former to optional bson then you'd do something like (ad hoc):

```
match serializer_type {
  SerializeType::Json => serde_json.to_str(foo),
  SerializeType::Bson => serde_bson.to_str(foo),
}
```

if you want to support bson in the latter you'd just rename the type to something more semantic like:

```
SerializedValue::from(foo)
```

with the match that decides the concrete serialization format living inside `impl From<Foo> for SerializedValue`

the most painful transition would be from a hybrid solution with raw `serde_json` calls in some places and `JsonString::from()` calls elsewhere

thedavidmeister:
@maackle furthermore, mixed serialisation schemes seem high risk of double serialization

in theory i have some hope of inspecting a "maybe json" string to see if it is already valid json and try to avoid serializing it a second time (in a world where i'm not handling this with the type system)

if i'm supporting a half dozen serialization options then i have to check whether every string is valid anything at runtime every time i want to defensively avoid double serialization

thedavidmeister:
uh, the elaboration of what it means is that anything that can go `Into<JsonString>` and `TryFrom<JsonString>` can allocate onto the stack automatically at the hdk level @maackle 

i.e. supporting a new hdk function is a 1-liner + enum variant

wasm_utils/src/memory/allocation.rs
```diff
@@ -0,0 +1,253 @@
+use holochain_core_types::{error::HolochainError, json::JsonString};
```
zippy: Can we have a file level comment that describes what this file does?

wasm_utils/src/memory/ribosome.rs
```diff
@@ -0,0 +1,330 @@
...
+            RibosomeErrorCode::ZeroSizedAllocation,
+        ))?,
+        RibosomeEncodedValue::Failure(err_code) => Err(HolochainError::Ribosome(err_code))?,
+        RibosomeEncodedValue::Allocation(ribosome_allocation) => {
+            Ok(WasmAllocation::try_from(ribosome_allocation)?.read_to_string())
+        }
+    }
+}
+
+pub fn load_ribosome_encoded_json<J: TryFrom<JsonString>>(
```
zippy: could use a comment?

Issue #203, opened 2018-08-09
real uuids for snowflakes that hit the persister save/load round trip


thedavidmeister:
```rust
    // @TODO how does save/load work with snowflake IDs?
    // snowflake is only unique across a single process, not a reboot save/load round trip
    // we'd need real UUIDs for persistant uniqueness
```

PR #857, opened 2019-01-14
GetLinks bug (Adds a failing test (with @maackle))

willemolding:
This PR adds a test to the app spec that reveals a pretty major bug in the 0.0.3 that was recently introduced.

When calling zome functions that get and create links on a particular base it seems there is some unexpected side effect of calling get_links such that the order of when they are called matters.

Calling [get] -> [create] -> [get] results in the second call to `get` incorrectly returning the same result as the first (e.g. no results).

Calling [create] -> [get] works as expected and will return the entry linked in the create zome function.

This has been replicated in both the nodejs container and the holochain_container. We are not sure if it is a problem just with the result being returned from the zome function or an issue with `get_links` itself.

By observing the CAS and EAV directories in the storage folder when using the holochain_container it was possible to observe that the links were correctly created in BOTH cases on the call to `link_entries` suggesting the the problem is only with the `get_links` call


maackle:
@thedavidmeister sez:

> #203 maybe this, if it hasn't been fixed already?

maackle:
I added another test. It's identical to the first failing test, except we call get_links twice in a row instead of once, and that causes the links to come through.

** TL;DR** It seems that get_links is not waiting for DHT results to come back at all.

I noticed this behavior showing up in holofuel tests as well -- it wasn't apparent before because of the way the tests are polling, but if you take polling out, you see the same thing, where the first get_links response will come back empty, and then calling it again will return the expected result.

There's one place where Alice adds a link, then Bob wants to get links. I turned polling on and watched the logs as Bob repeatedly tries to get those links. Here's the salient points of what I observed:

* bob does his first `GetLinks` 
* and then immediately does `ReturnZomeFunction`, returning an empty list.
* bob is now waiting for the next polling interval (1 second)
* meanwhile, bob and alice are subconsciously talking about those links. They're sending direct messages about it, exchanging validation packages, etc.
* amidst all the hubbub, bob's conscious mind wakes up and asks for those links again with another `GetLinks`
* again his next action is an immediate `ReturnZomeFunction` with an empty list
* in the next polling interval, alice happens to get to the point of `RespondGetLinks`.
* bob picks it up with `GetDhtMetaResult` and their long subconscious conversation ends
* next time bob wakes up, he calls `GetLinks` for the third time, and now he gets those coveted links.

It seems that get_links is not waiting for DHT results at all. It fires off the request for metadata and then returns whatever it's got at the moment. It should be awaiting. I don't know how to do that yet but I can look into it.

maackle:
@lucksus and I paired on this, he identified and fixed the problem but then had to go to a meeting so I'll try to describe the fix:

`NetworkState::get_links_results` was keyed only by address and tag, which does not differentiate between subsequent calls, so network effects from the past were leaking into future calls. We added a snowflake::PUID to the key to make each call unique, so the `GetLinksFuture` won't prematurely respond to any old get_links data with the same address and tag. The unique ID gets threaded through the network and comes back to the link requester via `GetDhtMeta`.

maackle:
Now that I think about it, I'm wondering if we shouldn't add an extra field to `GetDhtMetaData` for this. With this fix we're hijacking `msg_id` to store data we actually care about, rather than just using it as an arbitrary UID. If we lose sight of this we'll break things again. What do you think? An extra "data" string field?

core/src/network/reducers/get_links.rs
```diff
@@ -1,27 +1,23 @@
...
-    tag: &String,
-) -> Result<(), HolochainError> {
+fn inner(network_state: &mut NetworkState, key: &GetLinksKey) -> Result<(), HolochainError> {
     network_state.initialized()?;
 
     send(
         network_state,
         ProtocolWrapper::GetDhtMeta(GetDhtMetaData {
-            msg_id: "?".to_string(),
+            msg_id: key.id.clone(),
```
maackle: This makes me uneasy. Now that I think about it, I'm wondering if we shouldn't add an extra field to GetDhtMetaData for this. With this fix we're hijacking msg_id to store data we actually care about, rather than just using it as an arbitrary UID. If we lose sight of this we'll break things again. What do you think? An extra "data" string field?

maackle: Or am I over thinking this?

lucksus: Well, I thing this usage (a unique local (session?) ID) is what this field was intended for. Right @neonphog?

maackle: If so then I approve!

neonphog: The networking module just clones this value into the response. It is intended for core to be able to link asynchronous request / responses.

PR #858, opened 2019-01-14
Investigation into agent ID problems [BUGS]

### LINKED ISSUE #223: better implementation for display of holochain error
currently uses debug implementation for display, which is weird

can stack overflow if not careful

```rust
impl fmt::Display for HolochainError {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        // @TODO seems weird to use debug for display
        // replacing {:?} with {} gives a stack overflow on to_string() (there's a test for this)
        // what is the right way to do this?
        // @see https://github.com/holochain/holochain-rust/issues/223
        write!(f, "{:?}", self)
    }
}
```

maackle:
This PR focuses on three tests in a new file, app_spec/test/regressions.js, which are all completely similar, but only one passes. The summaries of the three tests are:

### test 1 (pass)

- Alice creates a post,
- Alice polls until she can get it
- Alice recommends it to herself (links it to her agent id with a tag), 
- Alice successfully gets that recommendation (get_links on her agent id with same tag) 

Note that this is all a single agent

### test 2 (fail)

- Alice creates a post
- Tash polls until she can get it
- Tash recommends it to herself
- Tash attempts to get the recommendation, **but this fails**

### test 3 (fail)

- Alice creates a post
- Tash polls until she can get it
- Alice attempts to recommend the post to Tash, **but this fails**

- - -

I talked with @lucksus a bit about the behavior of the 3rd case. It appears that an agent cannot `get_entry` when the address is another agent's address. An agent can get their own address but not another's. That would explain why linking fails.

I was surprised by the 2nd case failing, it seems to be failing for another reason.

Clearly there are some problems with multi-agent sharing of agent IDs, sharing links, or both.

- - -

Note that I made a bunch of changes to a bunch of other files. I was running into a `NotImplemented` error in the debug output, and since there are a couple dozen instances of that error across the codebase, I had to add a unique string to each one to see which one it was. (Turned out the problem was that I forgot to the new tag I needed to the zome definition, but thought this might be a helpful change anyway.)

Also: fixes #223


maackle:
@thedavidmeister sez:

> https://github.com/holochain/holochain-rust/issues/203 maybe this, if it hasn't been fixed already?

core/src/network/actions/initialize_network.rs
```diff
@@ -20,14 +21,18 @@ pub async fn initialize_network(context: &Arc<Context>) -> HcResult<()> {
...
     };
     let action_wrapper = ActionWrapper::new(Action::InitNetwork(network_settings));
     dispatch_action(context.action_channel(), action_wrapper.clone());
 
     await!(InitNetworkFuture {
         context: context.clone(),
-    })
+    })?;
+
+    await!(publish(agent_id.clone().into(), context))?;
```
lucksus: Hm, yeah, I can see how that fixes this. 

And this also raises some question for me, that I would like to discuss with @neonphog:
This has to be a temporary hack, right? A real DHT and world model implementation would handle this intrinsically, right?

I guess the problem is that we first initialize the chain and then the network, so the committing of the agent_id happens before the network can get notified of this entry.

BUT: in the reducer of above `Action::InitNetwork` we register with the network with our agent_id. Shouldn't that be enough? ...and jumping a few months into the future: what about membranes? creating and publishing our agent_id is something that should also be validated eventually...

nodejs_container/native/src/config.rs
```diff
@@ -71,7 +71,7 @@ fn make_config(instance_data: Vec<InstanceData>) -> Configuration {
         bridges: Vec::new(),
         logger: LoggerConfiguration {
             logger_type: "debug".to_string(),
-            rules: LogRules::new(),
```
zippy: In general I don't like this kind of thing slipped into an unrelated PR.  Feels like an amendment rider on a bill that must pass.  Not that I disagree with it, but it makes it hard to think about the PR in question...  I think we should just have a class of PR that's "Cleanup" or something.

maackle: Yeah sorry @zippy, I think this actually reverts the reverse thing that slipped into a previous PR of mine, so it should be back to how it was before...

I guess I wasn't expecting this to get merged right away, anyway.

PR #867, opened 2019-01-15
Pin exact version of futures-preview

maackle:


maackle:
I feel that the safest way to handle this is actually to use a fork of futures-preview until we do migrate to alpha-12.

Note that in alpha-11 they use loose version requirements in their [main Cargo.toml](https://github.com/rust-lang-nursery/futures-rs/blob/0.3.0-alpha.11/futures/Cargo.toml#L26). This has been updated in [alpha-12's Cargo.toml](https://github.com/rust-lang-nursery/futures-rs/blob/0.3.0-alpha.12/futures/Cargo.toml#L26) to properly use exact versions. So, even if we pin to alpha-11, future-preview's internal dependencies will still immediately update themselves to alpha-12. We could probably pin each internal dependency explicitly in our own Cargo.toml files, but this prevents that. When we move to alpha-12 we can go back to the main dependency since they fixed this.

maackle:
BTW I like how they have their repo structured. They use many small crates like we do, but they have a single crate at the same level as the sub-crates which `pub use`s everything needed from the smaller crates. We should totally do this once we publish to crates.io. 

https://github.com/rust-lang-nursery/futures-rs/blob/master/futures/src/lib.rs

PR #868, opened 2019-01-16
Provide hdk::query_result API to allow return of Headers if desired

pjkundert:
The existing `hdk::query("post".into(), 0, 0)` returning `Vec<Address>` is maintained.

A new `hdk::query_result("post.into(), QueryArgsOptions::default())` is provided, returning an enum containing Vec<Addresses> or Vec<ChainHeader>.  If `QueryArgsOptions { headers: Some(true),... }` is provided it returns a `Vec<ChainHeader>`; otherwise, a `Vec<Addresses>`.

I'd like to collect commentary on the API.  It's somewhat like `hdk::get_entry_result`, but the return type is an enum, instead of a struct w/ a `result` member.

I've also cleaned up the .skip and .take a bit to avoid cloning more than I have to, since `ChainHeaders` are somewhat bigger than just an Address to copy.

Also, instead of `hdk::query` taking `u32` values for start and limit, I've switched to `usize` for consistency with what they're used for internally.  Is that a problem for JSON serialization or anything?


maackle:
@pjkundert does this mean you can only query for headers or the entries, but not both?

I'm wondering how this fits into the work I am about to start, which was to flesh out `GetEntryOptions`, which instructs `get_entry_results` whether or not to return headers as well as entries. Let's sync up tomorrow.

pjkundert:
We need both to be improved; I couldn't quite figure out how to get the headers in the get_entry flow, but I noticed we were just throwing them away in the query, so I took a crack at obtaining them there.

The APIs should be improved in both cases to make them symmetrical; the get_entry's GetEntryOptions had an extra unnecessary option for query (the StatusRequestKind), so I made a QueryArgsOptions, but perhaps we could just re-use the same struct and document that the ...Kind is ignored?  Also, the return types differ, and query_result would probably need to be made consistent with get_entry_result; I tried to retain the basic hdk::query API (returns a simple Vec<Address> only), but we'd have to sacrifice that and change it; not too big a deal, at this point.

zippy:
I think it's fine that Query would return something different then Get so it has it's own options.

core/src/agent/chain_store.rs
```diff
@@ -29,6 +29,19 @@ impl PartialEq for ChainStore {
...
 
+#[derive(Default, Debug, Clone)]
+pub struct ChainStoreQueryOptions {
+    pub start: usize,
+    pub limit: usize,
+    pub headers: bool,
+}
+
+#[derive(Debug)]
+pub enum ChainStoreQueryResult {
```
zippy: I think I agree with @maackle that this shouldn't be an enum, but instead a struct with Option<Vec<Address>> and Option<Vec<Chanheader>> that have Some() or None() based on the query so that you don't have to do two queries if you want both.  Then in the request we would have an enum for "Entries, Headers, Both".

It would be fine to have helper functions in the hdk that assume you only want one or the other and do the ChainStoreQueryOptions setting correctly and unrwap to just one or the other..  But the core function should allow both to be returned.

pjkundert: @maackle and I just paired; We'll be able to Addresses, Headers, and Entries without too much trouble.  Should have it implemented for review by end of day, hopefully.

pjkundert:
OK, the hdk::query_results API has been enhanced to optionally return Entry data.  I'd like feedback on the API and return types.

Not yet ready for merging; hasn't been heavily tested (ie. at all).

core/src/nucleus/ribosome/api/query.rs
```diff
@@ -59,31 +74,71 @@ pub fn invoke_query(runtime: &mut Runtime, args: &RuntimeArgs) -> ZomeApiResult
...
+                    .iter()
+                    .map(|header| // -> Result<Entry, HolochainError>
+                         Ok((header.to_owned(),
+                             get_entry_from_dht(&runtime.context, header.entry_address().to_owned())? // -> Result<Option<Entry>, HolochainError>
+                             .ok_or(HolochainError::ErrorGeneric(
+                                 format!("Failed to obtain Entry for Address {}", header.entry_address())))?)))
+                    .collect();
+                match maybe_headers_with_entries {
+                    Ok(headers_with_entries) => QueryResult::HeadersWithEntries(headers_with_entries),
+                    Err(_e) => return ribosome_error_code!(UnknownEntryType), // TODO: return actual error?
```
maackle: If it works, this is probably fine now, but would rather see the entry-getting stuff DRYed up and refactored into a function

core/src/nucleus/ribosome/api/query.rs
```diff
@@ -59,33 +69,88 @@ pub fn invoke_query(runtime: &mut Runtime, args: &RuntimeArgs) -> ZomeApiResult
...
+            }
+        }),
         Err(_code) => return ribosome_error_code!(UnknownEntryType),
     };
 
     runtime.store_result(result)
 }
+
+/// Get an Entry via the provided context, returning Entry or HolochainError on failure
+fn get_entry_from_context(
```
lucksus: Isn't query supposed to operate on the chain only? Why `from_context` which actually is do a get `from_dht`? This only works because we currently reuse the same CAS for both, chain and dht, but we do have to separate CASes in the context to keep them separate. This function should be `get_entry_from_chain` which only operates on the source chain.

pjkundert: Done.  I've augmented core/src/nucleus/actions/get_entry.rs to provide crate-local APIs to access Entry values from either the either the context's dht or the agent.

lucksus: 👍

PR #874, opened 2019-01-16
Use tap-summary instead of faucet

maackle:
Now that we have two test files, faucet sometimes complains with: `not ok 7 assert out of order`.

According to [this issue](https://github.com/substack/tape/issues/215#issuecomment-271470048), tap-summary can handle this better.


maackle:
This is no good either, tap-summary is picky about the output, and all of our debug output breaks the test reporting

PR #875, opened 2019-01-16
holochain_net: protocol update - HandleGetDht Data & Meta + renames

ddd-mtl:
(note: must sync with n3h PR: https://github.com/holochain/n3h/pull/27)

Added HandleGetDhtData and HandleGetDhtMeta to follow SendMessage pattern. 
Renamed ProtocolWrapper to JsonProtocol.
Renamed some of its enum variants like TrackApp -> TrackDna

Proposed naming convention:
```Rust
/// Enum holding all message types that serialize as json 'hc-core <-> P2P network module' protocol.
/// There are 4 categories of messages:
///  - Command: An order from the local node to the p2p module. Local node expects a reponse. Starts with a verb.
///  - Handle-command: An order from the p2p module to the local node. The p2p module expects a response. Start withs 'Handle' followed by a verb.
///  - Result: A response to a Command. Starts with the name of the Command it responds to and ends with 'Result'.
///  - Notification: Notify that something happened. Not expecting any response. Ends with verb in past form, i.e. '-ed'.
```

### Misc

Clearer comments and names in test_bin


core/src/network/reducers/mod.rs
```diff
@@ -86,7 +86,7 @@ pub fn reduce(
 /// that lives in the NetworkState.
 pub fn send(
     network_state: &mut NetworkState,
-    protocol_wrapper: ProtocolWrapper,
+    json_message: JsonProtocol,
```
zippy: this makes more sense, easier to read

net_connection/src/json_protocol.rs
```diff
@@ -64,7 +64,7 @@ pub struct MessageData {
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, DefaultJson)]
-pub struct TrackAppData {
+pub struct TrackDnaData {
```
zippy: thank you!  Getting consistent on DNA instead of App is a good thing.

net_connection/src/json_protocol.rs
```diff
@@ -165,120 +165,134 @@ pub struct DhtMetaData {
     pub content: serde_json::Value,
 }
 
-/// Enum holding all Message types in the 'hc-core <-> P2P Network Module' protocol.
+/// Enum holding all message types that serialize as json in the 'hc-core <-> P2P network module' protocol.
```
zippy: yes, logical naming!

net/src/mock_system.rs
```diff
@@ -162,38 +165,38 @@ impl MockSystem {
...
+            JsonProtocol::HandleSendMessage(msg.clone()).into(),
         )?;
         Ok(())
     }
 
     /// we received a SendResult message...
     /// normally this would travel over the network, then
     /// show up as a SendResult message, fabricate that message && deliver
-    fn priv_handle_send_result(&mut self, msg: &MessageData) -> NetResult<()> {
+    fn priv_handle_handle_send_message_result(&mut self, msg: &MessageData) -> NetResult<()> {
```
Connoropolous: priv_handle_handle ?

ddd-mtl: Yep. Handle the HandleSendMessageResult message.
Would be clearer if we could do `priv_handle_HandleSendMessageResult()` but rustfmt would bitch about naming conventions...

Connoropolous: Ah, ok! Looked like one of the accidental double up find and replaces that's caught me up sometimes

PR #881, opened 2019-01-17
MDBOOK: Add Zome Functions explainer

Connoropolous:


doc/holochain_101/src/zome/zome_functions.md
```diff
@@ -120,11 +120,146 @@ Imagine that there are many DNA instances running within one Container, and each
...
+        Ok(())
+    }
+
+    functions: {}
+}
+```
+
+`functions` is where the Capabilities, and function declarations will be made.
+
+### Adding a Capability
```
zippy: This will no longer be correct after #791 gets merged.  As a matter of process I'm happy to do the update to this text as part of that PR if this gets merged first.

Connoropolous: We can work together to make the relevant updates @zippy sounds good

PR #885, opened 2019-01-18
Container - Static file server for UI bundles

willemolding:
*(This is @lucksus, wrapping up @willemolding's PR with description and last polish)*

The two grey items in the center are part of this PR:
screenshot 2019-01-18 at 21 43 30

## HTTP server
Uses [hyper](https://github.com/hyperium/hyper) and [hyper staticfile](https://github.com/stephank/hyper-staticfile) to create a light-weight HTTP server to serve front-end files.

## Additions to container config
Two structs get added to the set of container config types:
* UiBundleConfiguration
* UiInterfaceConfiguration
Which look like this in the toml config file:
```toml
[[ui_bundles]]
id = "bundle1"
root_dir = "" # serves the current directory
hash = "Qm000"

[[ui_interfaces]]
id = "ui-interface-1"
bundle = "bundle1"
port = 3000
dna_interface = "<not used>"
```

**UI bundles** are analogous to DNAs: adding one means installing it and making it available inside the container under a given ID. This does not yet instantiate this UI (same as with the DNA which needs to be used in an instance).

A UI bundle has to be connected to a DNA/zome interface (the websocket thing that connects to the container JSON RPC functions) in order to create a **UI interface**, kind of like an instance of the bundle. This is what gets served by the HTTP server on a given port.

The main reason for this is not part of this PR but coming up next: the UI needs to call back to the container to open a websocket connection. It needs to know which of potentially many websocket interfaces is meant for it. The container will insert this information into the package it is serving for each UI...


container_api/src/static_file_server.rs
```diff
@@ -0,0 +1,123 @@
...
+                // .with_graceful_shutdown(rx)
+                .map_err(|e| eprintln!("server error: {}", e));
+
+            println!("Listening on http://{}", addr);
+            rt::run(server)
+        });
+        Ok(())
+    }
+
+    // pub fn stop(&mut self) -> HolochainResult<()> {
```
willemolding: This is the only part that needs to be finished. I guess this can wait until UI bundle CRUD interface though

container_api/src/container/admin.rs
```diff
@@ -607,7 +607,11 @@ pattern = ".*""#
         file.read_to_string(&mut config_contents)
             .expect("Could not read temp config file");
 
-        let mut toml = String::from("bridges = []");
+        let mut toml = String::from(
```
zippy: This looks like it's ripe for refactoring as another `add_block` unit?

PR #887, opened 2019-01-18
fixed duplicate named test container

zippy:
In my haste to get scenario api merged into develop, I missed CI failing because of a duplicated named test container.  This PR should fix that.


zippy:
cheating to get this fix in, blocks a bunch of other PRs tests from passing.

PR #888, opened 2019-01-18
Renamed 'Mock' to 'Memory' everywhere

ddd-mtl:
P2pConfig backend-kind changed from MOCK to MEMORY
Renamed MockSystem to InMemoryServer
Renamed MockWorker to InMemoryWorker
...

Added name field to InMemoryServer. This makes debugging easier.
Fixed potential bug in InMemoryWorker if TrackDna was called twice on same DNA.


net/src/memory_worker.rs
```diff
@@ -58,34 +67,59 @@ impl NetWorker for MockWorker {
...
+        // println!("InMemoryWorker::new() config = {:?}", config);
+        let server_name = config["serverName"]
             .as_str()
             .unwrap_or("(unnamed)")
             .to_string();
-
-        let mut map_lock = MOCK_MAP.write().unwrap();
-        if !map_lock.contains_key(&network_name) {
-            map_lock.insert(network_name.clone(), Mutex::new(MockSystem::new()));
+        println!("InMemoryWorker::new() server_name = {}", server_name);
```
neonphog: What's our logging strategy these days? There's a lot of `println!`s, either active or commented out in here. It'd be nice to have tagged logging statements at various levels that could all always be in here, but configure-time enabled / disabled.

ddd-mtl: Those printfs are temporary. I'll remove them once the PR is green.
As for logging strategy, 'core' has its own logger system, but I don't think 'net' should be dependent on core. I'm thinking of rolling our own in 'net', but no steps have been made in that direction yet.

PR #891, opened 2019-01-18
Added a test for hdk::send, that is currently failing

juliolrmonteiro:
I am adding a test of `hdk::send` to **app_spec** so we can start debugging it.

Currently when you run the test it will fail with:
```JS
⨯ send
  not ok 4 should be equivalent
    ---
      operator: deepEqual
      expected: |-
        'Received : ping'
      actual: |-
        { error: 'JSON.parse failed to parse the result', rawResult: '{"Ok":pong}' }
      at: scenario2.runTape (.../src/holochain-rust/app_spec/test/test.js:35:5)
      stack: |-
        Error: should be equivalent
            at Test.assert [as _assert] (.../src/holochain-rust/app_spec/test/node_modules/tape/lib/test.js:224:54)
            at Test.bound [as _assert] (.../src/holochain-rust/app_spec/test/node_modules/tape/lib/test.js:76:32)
            at Test.tapeDeepEqual (.../src/holochain-rust/app_spec/test/node_modules/tape/lib/test.js:421:10)
            at Test.bound [as deepEqual] (.../src/holochain-rust/app_spec/test/node_modules/tape/lib/test.js:76:32)
            at scenario2.runTape (.../src/holochain-rust/app_spec/test/test.js:35:5)
            at run (.../src/holochain-rust/nodejs_container/index.js:155:23)
            at Scenario.run (.../src/holochain-rust/nodejs_container/index.js:146:9)
            at Test.Scenario._tape.t (.../src/holochain-rust/nodejs_container/index.js:154:18)
            at Test.bound [as _cb] (.../src/holochain-rust/app_spec/test/node_modules/tape/lib/test.js:76:32)
            at Test.run (.../src/holochain-rust/app_spec/test/node_modules/tape/lib/test.js:95:10)
    ...
  Stopping instance "bob::app-spec"...
  Stopping instance "alice::app-spec"...
  MainBackgroundTask shut down
```

I just followed the most simple example at:
https://developer.holochain.org/api/latest/hdk/api/fn.send.html

It seems there is an unmarshalling problem on the return of `hdk::send`.

Thank you very much for all the assistance! :)


maackle:
@zippy that error will show up even if the container attempted to start but failed. I haven't been able to figure out how to get the actual `TestContainer::start()` function to throw an error when it fails. There's something I don't understand about Neon there. So, maybe the container is failing to start and we're not seeing the error.

But yeah, the other problem is clearly a deserialization issue. @juliolrmonteiro I just pushed a commit that fixes the test, but we probably shouldn't have to do this, `receive()` shouldn't be able to return something that's not deserializable. 

Also it's not at all intuitive to me that the return value of `hdk::send` should be the result of running the argument through `receive`! Is that even right? I would expect `hdk::send` to return at most `Result<(), HolochainError>`. @lucksus ?

lucksus:
@maackle `hdk::send` is blocking, waiting for and returning the result of receive since send/receive is meant to implement a synchronous request-response pattern.

Connoropolous:
Oooh, @juliolrmonteiro ... so the result is actually passing through. It's just a serialization issue. That's good! 

I need to fix the example in the docs.

I think if you pass a struct which uses DefaultJson it'd be fine, or also just use `json!` macro with serde

Connoropolous:
@juliolrmonteiro I pushed an update to the code that makes it run...

Because we expect JSON strings in various parts of the stack, its better to pass them. This case is pretty weird, and I don't like it, but it's how it is for now and hopefully this example unblocks you

app_spec/zomes/blog/code/src/lib.rs
```diff
@@ -28,6 +29,13 @@ define_zome! {
         Ok(())
     }
 
+
+    receive: |message| {
+        json!({
```
Connoropolous: create json, but convert to string

app_spec/zomes/blog/code/src/lib.rs
```diff
@@ -28,6 +29,13 @@ define_zome! {
...
+            "key": "value"
+        }).to_string()
```
Connoropolous: create json, but convert to string

Connoropolous:
Also, @juliolrmonteiro I will update the example doc.

app_spec/zomes/blog/code/src/lib.rs
juliolrmonteiro: Ok, I got it, thanks!

app_spec/zomes/blog/code/src/blog.rs
```diff
@@ -32,6 +32,10 @@ pub fn handle_check_sum(num1: u32, num2: u32) -> ZomeApiResult<JsonString> {
     )
 }
 
+pub fn handle_send(to_agent: Address, message: String) -> ZomeApiResult<String> {
```
maackle: would be more consistent to call this `handle_check_send`

app_spec/zomes/blog/code/src/lib.rs
```diff
@@ -36,6 +44,12 @@ define_zome! {
                 handler: blog::handle_check_sum
             }
 
+            check_send: {
+               inputs: |to_agent: Address, message: String|,
+               outputs: |response: ZomeApiResult<String>|,
+               handler: blog::handle_send
```
maackle:
```suggestion
               handler: blog::handle_check_send
```

juliolrmonteiro: Excellent, agreed.

maackle:
Also we might consider adding a test for the case that's currently breaking, i.e. actually doing networking in the receive callback

marcusnewton1:
> Currently there's even a different problem in the CI. @maackle any ideas? Check out the failing CI first...
> 
> It's also happening for me locally:
> 
> ```
> /home/eric/code/metacurrency/holochain/holochain-rust/nodejs_container/index.js:143
>                 agentId: container.agent_id(id)
>                                    ^
> 
> Error: internal error in Neon module: TestContainer: cannot use agent_id() before start()
> ```
> But when I look at that part of the code, it does look like the container was started...

Not sure if you've resolved this, but removing `container.stop()` fixes this. I think Tape runs as a separate process outside the main test script, so the container will stop before the tests actually run.

PR #893, opened 2019-01-18
DOC: Container Admin RPCs

lucksus:
* [x] Changelog
 * [x] rustdoc describing all admin RPCs


container_api/src/interface.rs
```diff
@@ -259,6 +259,126 @@ impl ContainerApiBuilder {
...
+    /// It is guaranteed that the config is either valid after the change or the change
+    /// does not get applied but instead an error is reported back.
+    ///
+    ///  Full list of functions:
+    ///
+    ///  * `admin/dna/install_from_file`:
+    ///     Installs a DNA from a given local file.
+    ///     Params:
+    ///     * `id`: [string] internal handle/name of the newly created DNA config
+    ///     * `path`: [string] local file path to DNA file
```
zippy: Just realizing that this is a bit of a security hole.  We should probably limit it to a path within a dna folder, so that there isn't a way for a remote attacker to get the code to load an arbitrary path on the device.  Not that it's an immediate danger because it shouldn't parse correctly, but it seems like its a hole to exploit based on some other weakness.

container_api/src/interface.rs
```diff
@@ -259,6 +259,126 @@ impl ContainerApiBuilder {
...
+    ///     all instances this DNA is used in.
+    ///     Params:
+    ///     * `id`: [string] handle of the DNA to be deleted.
+    ///
+    ///  * `admin/dna/list`
+    ///     Returns an array of all configured DNAs.
+    ///
+    ///  * `admin/instance/add`
+    ///     Creates a new instance and adds it to the config.
+    ///     Does not start the instance nor adds it to an interface
```
zippy:
```suggestion
    ///     Does not start the instance nor add it to an interface
```

container_api/src/interface.rs
```diff
@@ -259,6 +259,126 @@ impl ContainerApiBuilder {
...
+    ///     (see `admin/instance/start` and `admin/interface/add_instance`).
+    ///     Params:
+    ///     * `id`: [string] Name for the new instance
+    ///     * `agent_id`: [string] Agent to run this instance with
+    ///     * `dna_id`: [string] DNA to run in this instance
+    ///
+    ///  * `admin/instance/remove`
+    ///     Removes an instance. Does also remove its mentions in interfaces.
```
zippy:
```suggestion
    ///     Removes an instance. Also remove its any uses of it in interfaces.
```

container_api/src/interface.rs
```diff
@@ -259,6 +259,126 @@ impl ContainerApiBuilder {
...
+    ///     * `id`: [string] Which instance to remove?
+    ///
+    ///  * `admin/instance/start`
+    ///     Starts a not running instance or reports an error if the given instance is
```
zippy:
```suggestion
    ///     Starts a stopped instance or reports an error if the given instance is
```

container_api/src/interface.rs
```diff
@@ -259,6 +259,126 @@ impl ContainerApiBuilder {
...
+    ///  * `admin/instance/list`
+    ///     Returns an array of all instances that are configured.
+    ///
+    ///  * `admin/instance/running`
+    ///     Returns an array of all instances that are running.
+    ///
+    ///  * `admin/interface/add`
+    ///     Adds a new DNA / zome / container interface (that provides access to zome functions
+    ///     of selected instances and container functions, depending on the interfaces config).
+    ///     This also automatically starts the interface. Different to instances, there are no
```
zippy:
```suggestion
    ///     This also automatically starts the interface. Different from instances, there are no
```

container_api/src/interface.rs
```diff
@@ -259,6 +259,126 @@ impl ContainerApiBuilder {
...
+    ///     *not running* interfaces - every interface that is configured is also active.
```
zippy:
```suggestion
    ///     *stopped* interfaces - every interface that is configured is also active.
```

container_api/src/interface.rs
```diff
@@ -259,6 +259,126 @@ impl ContainerApiBuilder {
...
+    ///     * `type`: [string] Either "websocket" or "http"
+    ///     * `port`:  [number] Port to bind the server to.
+    ///
+    ///  * `admin/interface/remove`
+    ///     Remove an interface from config. This automatically stops the interface as well.
+    ///     Params:
+    ///     * `id`: [string] Which interface to stop?
+    ///
+    ///  * `admin/interface/add_instance`
+    ///     Make a given instance available in a given DNA interface.
```
zippy:
```suggestion
    ///     Make a given DNA instance available via a given interface.
```

container_api/src/interface.rs
lucksus: Admin interfaces shouldn’t be visible to remote machines anyway. I would like to switch to domainsockets for the admin interface asap.

PR #894, opened 2019-01-18
WASM Container Skeleton - Build for Browser Light Clients

neonphog:
Baby steps to getting us a build target that will work for Holo light client. Currently just imports core_types and exposes a test function that we should remove later.

Mostly just want everyone's opinion on:
 - [x] naming - `conductor_wasm` ok?
 - [x] build-strategy - wasm-bindgen ok? also, `--nodejs` ok? our options are: browser-only or nodejs only... but there are tools (like rollup / webpack) to turn nodejs projects into web projects, so I started with that... it makes it easier to do unit tests too, since we don't have to spin up a web-server
 - [x] not committing the generated wasm to git - we could totally also commit it
 - [x] Makefile - are we still using the root Makefile? what about the one in Docker? Something else?

before we start working through the real problems.

Example real problem... I suspect we will run into a lot of wasm incompatibilities as we try to include all of the crates needed:

```
$ make build_container_wasm
error[E0433]: failed to resolve: use of undeclared type or module `imp`
   --> /home/neonphog/.cargo/registry/src/github.com-1ecc6299db9ec823/same-file-1.0.3/src/lib.rs:105:19
    |
105 | pub struct Handle(imp::Handle);
    |                   ^^^ use of undeclared type or module `imp`

$ cargo tree -p same-file -i
same-file v1.0.3                                                                
└── walkdir v2.2.5
    └── holochain_cas_implementations v0.0.3 (/home/neonphog/projects/holochain-rust/cas_implementations)
        ├── holochain_container_api v0.0.3 (/home/neonphog/projects/holochain-ru
```


neonphog:
@lucksus 

> Did you try core? That should actually work now

Not yet, I'm excited to iterate on this once it's a priority!

> Would this crate container_wasm be a surrogate for container_api or do we need that there as well? I guess there is code in container_api that we might want to use in WASM as well

Yeah, I was, at one point, thinking we'd want to just wrap core, but I don't think that's correct... with micro-service apps bridging / capabilities etc being managed at the container level, I'm pretty sure we'll need that in the browser as well.

thedavidmeister:
@neonphog @lucksus this is great

- lets try using compiler flags to drop fs implementations at the compiler level
- what i'd really like to see, that would help "sell" bindgen for me in the short term, is the ability to get console logs out of the wasm (especially on panics) and unit test wasm (e.g. headless firefox) https://rustwasm.github.io/book/game-of-life/testing.html
- looks like the bindgen installation process is via cargo, so it would work on nix just like `hc-install-tarpaulin` in `shell.nix`

lucksus:
Can we have Compiler flags and bindgen as follow ups, @thedavidmeister?
Adding this build to the nix scripts thought is something we should add here since currently this crate isn’t even build in CI.

thedavidmeister:
@lucksus yeah followup sounds fine, i think what we need is some space in the SoA to track all of this

container_wasm/src/lib.rs
```diff
@@ -0,0 +1,19 @@
...
+
+use holochain_core_types::agent::KeyBuffer;
+
+use wasm_bindgen::prelude::*;
+
+/// There isn't really any reason to export this
+/// but we need something to prove out the wasm build
+#[wasm_bindgen]
+pub fn parse_agent_id(agent_id: &str) -> Result<Vec<u8>, JsValue> {
+    let kb = match KeyBuffer::with_corrected(agent_id) {
```
sphinxc0re: You can replace the `match` by using the `?` operator:
```rust
let kb = KeyBuffer::with_corrected(agent_id)
    .map_err(|e| JsValue::from_str(&format!("{:?}", e))?;

thedavidmeister:
- cleaned up @sphinxc0re 's comments
- renamed container to conductor
- resolved merge xflx

but am wondering if actually we want this in a dedicated repo, much like moving the node conductor to a different repo

thedavidmeister:
nixified and added tests to ci

PR #900, opened 2019-01-21
rename entry! and entries to include types

Connoropolous:
This is a proposal to rename 
`entries` in the define_zome macro to `entry_types` and the `entry!` macro to `entry_type!` because it's confusing to developers to conflate the 'class' and the 'instance'

This also brings the macros themselves further inline with the underlying DNA spec, which lists them as `entry_types`
https://github.com/holochain/holochain-rust/blame/develop/doc/specs/dna_manifest_specification.md#L112

### DIFF
--- a/app_spec/zomes/blog/code/src/lib.rs
+++ b/app_spec/zomes/blog/code/src/lib.rs
@@ -20,7 +20,7 @@ use hdk::{
 };
 
 define_zome! {
-    entries: [
+    entry_types: [
         post::definition()
     ]
 
--- a/app_spec/zomes/blog/code/src/post.rs
+++ b/app_spec/zomes/blog/code/src/post.rs
@@ -34,7 +34,7 @@ impl Post {
 }
 
 /// This is what creates the full definition of our entry type.
-/// The entry! macro is wrapped in a function so that we can have the content
+/// The entry_type! macro is wrapped in a function so that we can have the content
 /// in this file but call it from zome_setup() in lib.rs, which is like the
 /// zome's main().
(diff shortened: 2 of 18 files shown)


zippy:
Hmm this is interesting.  In [Capabilites-2](https://github.com/holochain/holochain-rust/pull/791/files) you'll see that I in the DNA i created `fn_declarations` as the functions go moved out of the `capabilites` objects, but in the Rust `define-zome!` macro I left it as `functions`   We might want make these match.  Which means it could be `entry_declarations`  instead of `types`  The issue with type for me is that I think of that as a the name of the entry, not its declaration.  Thoughts?

maackle:
What zippy said makes sense, we already have an `EntryType` type which is just a string, the name, so "entry type" feels more like just the name. This is more like a "declaration". However, it's even more like a "definition" since you provide the entirety of the entry info in this macro. Unlike the `fn_declaration`, which only specifies the signature, and then the body is "defined" elsewhere.

But yeah, I'm also feeling something along the lines of declaration or definition. Definitely better than just `entry!`

lucksus:
So it is actually an `entry_type_definition`, no?

Connoropolous:
Sure... `entry_type_definition` is just getting really long :s

Connoropolous:
I'm ok with waiting for a rename, if it feels unresolved :) I just put up this PR to get people to take the change request seriously, and get us talking about it

Connoropolous:
going to close this now... I think if changes to the HDK are made, they'll be in a different way than this

PR #901, opened 2019-01-22
Helper function to get `Vec<Entry>` from `Vec<Address>`

maackle:
Can simplify the return type of `get_links_and_load`, and can also be used for `query`

### DIFF
--- a/hdk-rust/src/api.rs
+++ b/hdk-rust/src/api.rs
@@ -905,7 +905,7 @@ pub fn get_links<S: Into<String>>(base: &Address, tag: S) -> ZomeApiResult<GetLi
 /// #    get_links::GetLinksOptions};
 ///
 /// # fn main() {
-/// fn hangle_get_links_result(address: Address) -> ZomeApiResult<Vec<ZomeApiResult<GetEntryResult>>> {
+/// fn handle_get_links_result(address: Address) -> ZomeApiResult<Vec<GetEntryResult>> {
 ///    hdk::get_links_result(&address, "test-tag", GetLinksOptions::default(), GetEntryOptions::default())
 /// }
 /// # }
--- a/hdk-rust/src/api.rs
+++ b/hdk-rust/src/api.rs
@@ -915,21 +915,13 @@ pub fn get_links_result<S: Into<String>>(
     tag: S,
     options: GetLinksOptions,
     get_entry_options: GetEntryOptions,
-) -> ZomeApiResult<Vec<ZomeApiResult<GetEntryResult>>> {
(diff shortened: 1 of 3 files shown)


hdk-rust/src/api.rs
```diff
@@ -915,40 +915,42 @@ pub fn get_links_result<S: Into<String>>(
...
             GetEntryResultType::Single(elem) => Ok(elem.entry.unwrap().to_owned()),
             GetEntryResultType::All(_) => Err(ZomeApiError::Internal("Invalid response. get_links_result returned all entries when latest was requested".to_string()))
         }
     })
-    .collect();
+    .collect()
+}
 
-    Ok(entries)
+/// Helper function to get all EntryResults from a Vec of addresses
```
zippy: I see how this is a bit cleaner, but doesn't mean there's a loss of information?  It means that if something failed you can't get the partial information.  Where is the motivation for making this change coming from?

maackle: @zippy I was abstracting this part out because `hdk::query` needs the same ability, and thought I'd try cleaning it up and reducing the amount of information at the same time, as a proposal. I figure it's a rare case that you would see an internal ZomeApiError on some addresses and not others (seems like it would typically be all or nothing), and even if that were the case, it seems unusual that the consumer would actually want to pick out the good parts and move on -- it seems appropriate that any internal error should infect the whole batch.

I could see good arguments for both ways, just floating this idea.

zippy: Yah, I can too.  I guess I'd want this to be data driven some-how, i.e. to know what actually happens in the wild.  But that's hard to figure out.  I guess maybe if we logged the error situation so at least there's a record in the log if only some errored and why.

willemolding: Total preservation of data would be to return a `ZomeApiResult<Vec<ZomeApiResult<Option<Entry>>>` right? This encodes from left to right: If `get_links` could be called on the base, if a `get_entry` for a given address could be called, and if an entry actually exists at that address. 

I agree that it would typically be all or nothing for the ZomeApiError but there is still the possibility that one of the target addresses does not link to anything. I believe there is no requirement that the target of a link must exist at creation time, only the base. Linking an empty address would therefore break this function for all future calls to that base.

I think there correct return type is therefore `ZomeApiResult<Vec<Option<Entry>>` and the calling code can easily filter out the `None` variants using `get_links_and_load(...).iter().filter_map(|x| x).collect()`

Connoropolous:
Does this get addressed in #1006?

Connoropolous:
looks like this is prbly closable @maackle ?

maackle:
:+1:

PR #903, opened 2019-01-22
 Add Timeout option to get_entry, get_links, and send

maackle:
Not sure if we want the timeout for all 3 calls, I just noticed that they all had similar hard-coded timeouts.

Also not sure how to effectively test this timeout. Tried adding a zome function in app_spec that used a timeout of 0 but it still succeeded. @thedavidmeister ideas on unit testing this? Or even integration testing this?

### DIFF
--- a/core/src/network/actions/custom_send.rs
+++ b/core/src/network/actions/custom_send.rs
@@ -11,16 +11,16 @@ use futures::{
 };
 use holochain_core_types::{cas::content::Address, error::HolochainError, time::Timeout};
 use snowflake::ProcessUniqueId;
-use std::{pin::Pin, sync::Arc, thread::sleep};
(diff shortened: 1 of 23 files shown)


core/src/network/actions/custom_send.rs
```diff
@@ -32,11 +33,11 @@ pub async fn custom_send(
     let action_wrapper = ActionWrapper::new(Action::SendDirectMessage(direct_message_data));
     dispatch_action(context.action_channel(), action_wrapper);
 
-    /* async {
-        sleep(Duration::from_secs(60));
+    let _ = async {
```
lucksus:
```suggestion
    thread::spawn(move || {
```

thedavidmeister:
@maackle timeouts would be easy to test if ticks were externally driven

PR #905, opened 2019-01-22
Seed Derivation for the dpki

zo-el:
(Damien)
Added Seed concept.
It is a struct and a trait, in order to implement the Role Object Pattern.
Trait is used for having 'TypedSeed's.
Which are seeds that can perform type specific actions. Mainly derive keys or seeds of a different type.
Currently a RootSeed can derive into a DeviceSeed.
A DeviceSeed can derive into a DevicePinSeed by addind a PIN.
A DevicePinSeed can generate Application Keys.

A Seed can mnemonic roundtrip. You'll lose its type though.

Seed is blobbable the same way as KeyBundle.
Took the opportunity to make a Blobbable trait since its the same process.


hc_dpki/src/seed.rs
```diff
@@ -0,0 +1,558 @@
...
+        passphrase: String,
+        hint: String,
+        config: Option<PwHashConfig>,
+    ) -> Result<bundle::KeyBundle, HolochainError> {
+        Ok(self.s.get_seed_bundle(passphrase, hint, config)?)
+    }
+
+    /// Get a new, completely random root seed
+    pub fn new_random() -> Self {
+        let mut s = SecBuf::with_insecure(32);
```
neonphog: I vote for deleting the `new_random()` function. We want the implementor to be able to decide if it should be a secure or insecure buffer, which is provided by the `new` function below. They can randomize that buffer before passing it in.

hc_dpki/src/seed.rs
```diff
@@ -0,0 +1,558 @@
...
+        RootSeed {
+            s: Seed::new(&"hcRootSeed".to_string(), SeedInit(s)),
+        }
+    }
+
+    /// Generate Device Seed
+    pub fn get_device_seed(&mut self, index: u64) -> Result<DeviceSeed, HolochainError> {
+        (index >= 1).ok_or(HolochainError::ErrorGeneric("Invalid index".to_string()))?;
+
+        let mut out_seed = SecBuf::with_insecure(32);
```
neonphog: most of the time this should be `secure` but, again, would be nice to let the implementor choose, because sometimes (like unit tests) it's nice if it's insecure.

Maybe pass in an already created DeviceSeed, and fill it's internal buffer?

hc_dpki/src/seed.rs
```diff
@@ -0,0 +1,558 @@
...
+        }
+    }
+
+    /// generate an application keypair given an index based on this seed
+    /// @param {number} index
+    /// @return {Keypair}
+    pub fn get_application_keypair(&mut self, index: u64) -> Result<Keypair, HolochainError> {
+        (index >= 1).ok_or(HolochainError::ErrorGeneric("Invalid index".to_string()))?;
+
+        let mut out_seed = SecBuf::with_insecure(32);
```
neonphog: ... i'm less certain about this one. At the very least, this should be made `with_secure`.

the seed isn't exported, and Keypair::new_from_seed below always returns secure secrets... so maybe we can get away without letting the implementor decide here?

hc_dpki/src/seed.rs
```diff
@@ -0,0 +1,558 @@
...
+    /// Get the proper seed type from a persistence bundle
+    ///
+    /// @param {object} bundle - the persistence bundle
+    ///
+    /// @param {string} passphrase - the decryption passphrase
+    ///
+    /// @return Type FromBundle
+    pub fn from_seed_bundle(
+        bundle: bundle::KeyBundle,
+        passphrase: String,
```
neonphog: passphrase should be secure SecBuf

hc_dpki/src/seed.rs
```diff
@@ -0,0 +1,558 @@
...
+    ///  generate a persistence bundle with hint info
+    ///
+    ///  @param {string} passphrase - the encryption passphrase
+    ///
+    ///  @param {string} hint - additional info / description for persistence
+    ///
+    /// @return {KeyBundle} - bundle of the seed
+    pub fn get_seed_bundle(
+        &mut self,
+        passphrase: String,
```
neonphog: passphrase should be secure SecBuf

hc_dpki/src/seed.rs
```diff
@@ -0,0 +1,558 @@
...
+    pub fn new(stype: &String, sm: InitializeSeed) -> Self {
+        match sm {
+            SeedInit(s) => Seed {
+                seed_type: stype.clone(),
+                seed_buf: s,
+            },
+            MnemonicInit(phrase) => {
+                let mnemonic = Mnemonic::from_phrase(phrase, Language::English).unwrap();
+                let entropy: &[u8] = mnemonic.entropy();
+                let mut buf = SecBuf::with_insecure(entropy.len());
```
neonphog: `secure` here.

plus... erm, is entropy here immutable? *shudder* we might not have a way of zeroing this secret in memory. See if there's any way you can clear it out, otherwise add a comment to come back and address it... we might have to make our own mnemonic parser.

hc_dpki/src/seed.rs
```diff
@@ -0,0 +1,558 @@
...
+            }
+        }
+    }
+
+    /// Generated a mnemonic for the seed.
+    pub fn get_mnemonic(&mut self) -> Result<String, HolochainError> {
+        let entropy = self.seed_buf.read_lock();
+        let e = &*entropy;
+        let mnemonic = Mnemonic::from_entropy(e, Language::English).unwrap();
+        Ok(mnemonic.phrase().to_string())
```
neonphog: here again, we need some way of zeroing the internal memory used by mnemonic if possible... see if you can, or add a comment to come back to it.

hc_dpki/src/seed.rs
```diff
@@ -0,0 +1,558 @@
...
+        passphrase: String,
+        hint: String,
+        config: Option<PwHashConfig>,
+    ) -> Result<bundle::KeyBundle, HolochainError> {
+        let mut passphrase = SecBuf::with_insecure_from_string(passphrase);
+        let seed_data: bundle::ReturnBundleData =
+            util::pw_enc(&mut self.seed_buf, &mut passphrase, config)?;
+
+        // convert -> to string -> to base64
+        let seed_data_serialized = json::encode(&seed_data).unwrap();
```
neonphog: use `expect` maybe, or promote the error with `?`

hc_dpki/src/seed.rs
```diff
@@ -0,0 +1,558 @@
...
+    /// @return Type FromBundle
+    pub fn from_seed_bundle(
+        bundle: bundle::KeyBundle,
+        passphrase: String,
+        config: Option<PwHashConfig>,
+    ) -> Result<FromBundle, HolochainError> {
+        let mut passphrase = SecBuf::with_insecure_from_string(passphrase);
+
+        let seed_data_decoded = base64::decode(&bundle.data).unwrap();
+        let seed_data_string = str::from_utf8(&seed_data_decoded).unwrap();
```
neonphog: no unwraps... promote errors

hc_dpki/src/seed.rs
```diff
@@ -0,0 +1,558 @@
...
+
+        let seed_data_deserialized: bundle::ReturnBundleData =
+            json::decode(&seed_data_string).unwrap();
```
neonphog: `?`

hc_dpki/src/seed.rs
zo-el: I agree. It makes sense

zo-el: Done :+1:

hc_dpki/src/seed.rs
zo-el: HI @neonphog, But should the seed in DeviceSeed be mutable? Because if that's not done you cant update the secret when passed into this fn.
what do you think?

zo-el: we could pass in a var to state if the implementor need `secure` or `insecure`

hc_dpki/src/seed.rs
zo-el: I am working on this. But I have put a comment for now.

core_types/Cargo.toml
```diff
@@ -21,6 +21,7 @@ futures-util-preview = "=0.3.0-alpha.12"
 
 reed-solomon = "0.2.1"
 rust-base58 = "0.0.4"
+rustc-serialize = "0.3.24"
```
sphinxc0re: What is this needed for? It really shouldn't be used. This crate has been deprecated in favor of `serde` about 2 years ago.

hc_dpki/src/seed.rs
```diff
@@ -0,0 +1,499 @@
...
+            seed_buf_in[0] = 12;
+            seed_buf_in[1] = 70;
+            seed_buf_in[2] = 88;
+        }
+        let seed_type = "hcRootSeed".to_string();
+
+        let mut s = Seed::new(&seed_type, SeedInit(seed_buf_in));
+
+        let m = s.get_mnemonic().unwrap();
+        println!("Menemenoc: {:?}", m);
```
sphinxc0re:
```suggestion
        println!("Mnemonic: {:?}", m);
```

core_types/Cargo.toml
zo-el: Thanks for catching that. I made the changes to update to `serde` :)

zippy:
Hey @zo-el what's the status on this branch?  Is it still in progress?

zo-el:
> Hey @zo-el what's the status on this branch? Is it still in progress?

Just need approval.

hc_dpki/src/seed.rs
```diff
@@ -0,0 +1,558 @@
...
+        &mut self,
+        pin: String,
+        config: Option<PwHashConfig>,
+    ) -> Result<DevicePinSeed, HolochainError> {
+        (pin.len() >= 4).ok_or(HolochainError::ErrorGeneric("Invalid PIN Size".to_string()))?;
+
+        // let pin_encoded = base64::encode(&pin);
+        let mut pin_buf = SecBuf::with_insecure_from_string(pin);
+
+        let mut hash = SecBuf::with_insecure(pwhash::HASHBYTES);
```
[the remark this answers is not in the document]
ddd-mtl: ???

hc_dpki/src/seed.rs
ddd-mtl: are you talking about the 'phrase' String stored within the Mnemonic struct? or maybe the entropy vec?

ddd-mtl: I changed the return into a `into_phrase()`. Would that suffice?

neonphog: maybe? hard to know without looking into the internal memory management of the crate

hc_dpki/src/seed.rs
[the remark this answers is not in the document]
neonphog: For now, I guess just change this to `with_secure`... I still like the pattern of taking the outputs as parameters, but in this case, we pretty much always want to protect seeds.

hc_dpki/Cargo.toml
```diff
@@ -9,7 +9,8 @@ lazy_static = "~1.2.0"
 base64 = "0.10.0"
 holochain_core_types = { path = "../core_types" }
 holochain_sodium = { path = "../sodium" }
-serde = "1.0"
+serde = "1.0.87"
```
zippy: Maybe we should update to the same version of serde everywhere so there aren't lots of extra versions compiled into the code?

hc_dpki/src/seed.rs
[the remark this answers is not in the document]
lucksus: Should this be
```suggestion
        let mut hash = SecBuf::with_secure(pwhash::HASHBYTES);
```
since hash is what stores the seed?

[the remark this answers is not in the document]
ddd-mtl: Ok, done.
I'd suggest later we add `SecureBuf` and `InsecureBuf` to sodium crate, since with what you described it would be a lot more error proof to have the `Seed` struct hold a `SecureBuf` instead of a `SecBuf`. Or maybe have a `SeedBuf` which is secure and hardcoded to 32 bytes ?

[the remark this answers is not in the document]
ddd-mtl: Its also kind of weird to have an "insecure" SecBuf.

hc_dpki/Cargo.toml
ddd-mtl: Reverted

PR #907, opened 2019-01-22
[WIP] Proper rejection when `.call()` throws an exception

maackle:
Fixes some bad exception handling in the JS for nodejs Container. Also will fix https://github.com/holochain/holochain-rust/pull/891#issuecomment-455753867


maackle:
Redoing this and adding more in another branch.

PR #908, opened 2019-01-23
add nightly version to container readme

Connoropolous:


lucksus:
Once https://github.com/holochain/holochain-rust/pull/904 gets merged, this needs to be changed again..

Connoropolous:
@lucksus that's alright... at least it'll be right, and we can do a full find-and-replace on the old values to do it well

PR #909, opened 2019-01-23
Container virtual file

willemolding:
- Adds HTTP endpoint to retrieve the dna interfaces that are mapped to the UI interface
- Adds ability to start and stop containers on the fly (I would appreciate a review on this part)


container/example-config/static-only.toml
```diff
@@ -13,3 +13,13 @@ hash = "Qm000"
...
+dna_interface = "http interface"
+
+
+[[interfaces]]
+admin = true
+id = "http interface"
+
+[interfaces.driver]
+port = 4000
+type = "http"
```
lucksus: Since hc-web-client can only talk websocket and since this is mainly catered for that use case, would only use websocket interfaces as examples here..

container_api/src/static_file_server.rs
```diff
@@ -1,31 +1,68 @@
...
+const DNA_CONFIG_ROUTE: &str = "/_dna_connections.json";
 
 fn redirect_request_to_root<T>(req: &mut Request<T>) {
     let mut original_parts: uri::Parts = req.uri().to_owned().into();
     original_parts.path_and_query = Some("/".parse().unwrap());
     *req.uri_mut() = uri::Uri::from_parts(original_parts).unwrap();
 }
 
+fn dna_connections_response(config: &Option<InterfaceConfiguration>) -> Response<Body> {
+    let intefaces = match config {
```
lucksus:
```suggestion
    let interfaces = match config {
```

container_api/src/static_file_server.rs
```diff
@@ -1,31 +1,68 @@
...
+        Some(config) => json!([config]),
+        None => json!([]),
+    };
+    Builder::new()
+        .body(json!({ "dna_interfaces": intefaces }).to_string().into())
```
lucksus:
```suggestion
        .body(json!({ "dna_interfaces": interfaces }).to_string().into())
```

lucksus: why is this plural (and an array) if there can ever only be one?

PR #917, opened 2019-01-24
Enhance holochain_core_types::time::Iso8601 with DateTime conversions

pjkundert:
The Iso8601 type is currently a simple wrapper around String, with no checking.  For Holofuel, we need to be able to actually operate on these as DateTime timestamps.  Handling such conversions and comparisons gracefully is non-trivial.

I have enhanced Iso8601 to maintain un-checked From traits, and have added String, in addition to static &str.  This supports importation of Iso8601 data from various external sources.

I have added checked TryFrom for &Iso8601 -> DateTime<FixedOffset>, which supports direct RFC 3339 conversion, as well as supports a greater subset of ISO 8601 timestamps, at the expense of a Regex invocation.  So, you pay for what you get; quick conversion for straight, simple RFC 3339 timestamps, and support for richer ISO 8601 from external sources.  Particularly, it defaults timestamps w/ no Time Zone specification to UTC, or "Zulu" time.

This validation is done at initial `Iso8601::new` time, and stored for future use for fmt::Display, PartialOrd, etc.  Serialization and fmt::Display always displays the canonicalized ISO 8601 form of the original timestamp (except if invalid; then the original String is produced).

Finally, the PartialOrd and PartialEq have been implemented.  Since not all Iso8601 are comparable (because they may be invalid), we treat the comparison of invalids for Equality as always false, and present None for partial_cmp (somewhat like float NaN are treated).


thedavidmeister:
ok cool, the string type for iso8601 was only ever intended as a quick placeholder until something more comprehensive can be done

core_types/src/time.rs
```diff
@@ -42,35 +42,135 @@ impl From<usize> for Timeout {
...
 impl From<&'static str> for Iso8601 {
     fn from(s: &str) -> Iso8601 {
         Iso8601(s.to_owned())
     }
 }
 
-pub fn test_iso_8601() -> Iso8601 {
-    Iso8601::from("2018-10-11T03:23:38+00:00")
+impl From<String> for Iso8601 {
+    fn from(s: String) -> Iso8601 {
```
thedavidmeister: why not put the validity check in the construction of the `Iso8601`?

pjkundert: I experimented with that a bit. We could, but that was a more drastic step than I wanted to embark on at this time -- I'm trying to get the Holofuel transactor working!  I think a reasonable next step would be to add an `is_valid` method, though, and look at what fails when we invoke that in the infallible `From` trait implementations, which would have to panic, which I don't like.  A lot of tests failed, to start with.  This is a first step...

My feeling is that allowing incoming data with whatever timestamps is OK, as long as you don't have to compare them -- then be more strict.  It's now easy to filter out and report/reject items with bad Iso8601s and produce a less catastrophic HolochainError, for example.

pjkundert: Since we pass a lot of Iso8601 back and forth across external APIs, I think causing these APIs to fail in deserialization would be worse than validating the timestamps in the target function, according to the rules of the application.

thedavidmeister: @pjkundert `TryFrom`?

zippy: @pjkundert look at our [jsonstring](https://github.com/holochain/holochain-rust/blob/develop/doc/holochain_101/src/json_string.md) docs for how to do this...

pjkundert: OK, I've implemented JSON serialization via both serde::Serialize/Deserialize, and via TryFrom<JsonString> for Iso8601 and From<Iso8601> for JsonString.  We do not want to fail invalid Iso8601 values in transit, I believe, right now. Too many tests to fix, for one thing.

I've added a simple check for validity via `Iso8601.validated().is_ok()`, and optimized the whole process to only attempt conversion once on creation.  We round-trip the canonicalized, valid ISO 8601 timestamp if possible, or retain the original invalid one -- JSON round-tripping still only involves one String, but it'll be the canonicalized version, if possible.

lucksus:
Cool! Thanks @pjkundert! @thedavidmeister, can you have a look again. Think it should be fine now. But tests are failing in a strange way. Is this a bug in the change to how we call tape?

pjkundert:
Yes, these failures are weird.  I've change the app_spec/test/run.sh to dump the raw output of the tape tests on failure, but I can't find any indication of why they are failing -- they don't fail for me when I run them...

lucksus:
Yes, some app-spec tests are flaky and we are looking into this right now. We think we found the reason. I suggest we merge this PR right now anyway.

lucksus:
I understand and share @thedavidmeister's request to use `TryFrom<String>` for `Iso8601` and have that type conversion fail if the given string is not an Iso8601 date time. If there are cases where you don't need to parse a string as a date time then you don't need to convert it. Having an `Iso8601` type that does not enforce the underlying data to actually be a date time at all feels like something that can lead to much confusion.

I see that all the tests here are build around the assumption that an `Iso8601` is just a string and then have conversions from `Iso8601` to `DateTime`. Is there code in other repos that is build around the same assumption?

lucksus:
@pjkundert, what unit tests are you talking about? The ones here in this PR?

I don't fully understand why you say you couldn't validate incoming dates in zome code.. If you expect something to sometimes not be a date then why would you use an `Iso8601` type to represent it? You could (and I would say should) leave it as a string and then `try_from` convert it if you need it to be a date. You can then explicitly handle both cases:
```rust
fn handle_xxx(input: String) {
  match Iso8601::try_from(input) {
    Ok(date) => ..
    Err(_) => ...
  }
  ...
}
```
which is I think what you want, no?
The main point we're trying to get across is that the above is way better because idomatic rust, than:
```rust
fn handle_xxx(input: Iso8601) {
  if input.is_valid() {
  }
  ...
}
```

lucksus:
If this would break a lot of code in your zomes I'm fine with merging it. What do you say @thedavidmeister?

pjkundert:
I'm saying that a lot of Zomes pass Iso8601's as arguments in within structs that are arguments.  If you pass an invalid one in from (say) Javascript, and the try_into `Result<...>` is an `Err`, it'll fail the deserialization (won't it?) and return an error saying that the Zome call failed.  A lot of Rust unit tests presently create invalid Iso8601's, so those will need to be fixed, too.  If not, then this is not as bad; I though it would fail the deserialization, not just pass a `Result` w/ an `Err`...

The idiomatic Rust that you show, above, is what the present pull request implemnents, sort of: `Iso8601::from("blah").validated().is_ok()`.  We could certainly add the `.is_valid()` method (I'm not super familiar with what the idiomatic validation API for Rust is).

thedavidmeister:
@pjkundert @lucksus 

> If you pass an invalid one in from (say) Javascript, and the try_into `Result<...>` is an `Err`, it'll fail the deserialization (won't it?) and return an error saying that the Zome call failed

yes, invalid data is always an invalid call/result

> A lot of Rust unit tests presently create invalid Iso8601's

dang... but that sounds like an important-to-fix-bug in the tests to me :(

> so those will need to be fixed, too

yes please!

> I though it would fail the deserialization, not just pass a `Result` w/ an `Err`...

returning an error result from a `try_from` native trait is what both @lucksus and i are asking for at the point of creation of any date struct (iso8601 or otherwise) from any string

> The idiomatic Rust that you show, above

i'm pretty sure @lucksus is saying that the `input.is_valid()` is NOT idiomatic rust, providing it as a counter example (what to avoid)

read over the "runtime impossible" docs that i wrote up for more info:

https://hackmd.io/QPgBfOk4TwOMrZnCCE46ZA?both#Runtime-impossible

> If this would break a lot of code in your zomes I'm fine with merging it.

feels like kicking the can down the road

the longer we leave a correct implementation, the more work/LOC it will be ultimately as we build more infra on top of it

fixing this RN is the easiest/fastest it will ever be IMO

pjkundert:
Thanks, @thedavidmeister -- I agree with everything you said.  I'll work toward that.  The failing serialization on Zome API calls needs some serious, serious work -- basically (last time I checked), it was an opaque, valueless error, giving no indication of what you did wrong.  Sort of like filling out pages of paperwork, and getting it back with a "rejected" stamp, with no indication of what you did wrong... :)

thedavidmeister:
@pjkundert 
> The failing serialization on Zome API calls needs some serious, serious work

yerp

pjkundert:
Ok, @lucksus, @thedavidmeister, @zippy -- at long last, I think this Iso8601 might be ready to look at again.  There wasn't many Rust unit tests that failed; I was certain that previously there were more tests that used Iso8601::from( &str ), but maybe I was mistaken.  Anyway, all the Rust unit tests pass cleanly.

There are a few other changes, mostly around returning errors that help diagnose Zome API call failures.

core_types/src/time.rs
```diff
@@ -42,78 +42,490 @@ impl From<usize> for Timeout {
...
+}
+
+impl Iso8601 {
+    fn new(secs: i64, nsecs: u32) -> Self {
+        Self(FixedOffset::east(0).timestamp(secs, nsecs))
+    }
+}
+
+/*
+ * Note that the WASM target does not have a reliable and consistent means to obtain the local time,
```
zippy: This is important. @Connoropolous take note for documentation!!

core_types/src/time.rs
```diff
@@ -42,78 +42,490 @@ impl From<usize> for Timeout {
...
+        Iso8601::from_str(s)
+    }
+}
+
+impl FromStr for Iso8601 {
+    type Err = HolochainError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        lazy_static! {
+            static ref ISO8601_RE: Regex = Regex::new(
```
zippy: Yowza!  That's one hecka regular expression!

core_types/src/time.rs
```diff
@@ -1,12 +1,12 @@
-//! The Iso8601 type is defined here. It is used in particular
-//! within ChainHeader to enforce that their timestamps
-//! are defined in a useful and consistent way.
+//! The Iso8601 type is defined here. It is used in particular within ChainHeader to enforce that
```
zippy: This is a superb file @pjkundert .  I think really we should be abstracting this out as its own rust crate for contribution to the community, and then using it.  That would also increase surface area for removing any bugs in it, if others are using it for orthogonal purposes.

zippy:
Haven't heard from @thedavidmeister so I'm merging.  We can fix thing later...

thedavidmeister:
great result, soz for the delayed response

PR #918, opened 2019-01-24
Node Container upgrades for proper exception handling

maackle:
    - (this can be considered part of the existing Scenario API changelog item)

The most notable changes here in terms of API are:
- a new `Container.run` function, which  starts a container and immediately runs a closure. This is necessary for `Scenario.run` to actually handle exceptions.
- when using `Config.instance`, the default instance name was `[agentname]::[dnapath]`. Now the default instance name is simply `[agentname]`. If needed, an explicit instance name can be passed as the third argument to `Config.instance`


Connoropolous:
What does exception handling really mean in this instance Michael? Like, which exceptions?

maackle:
@Connoropolous if the container fails to start up the test runs anyway, and fails for a misleading reason. Usually something along the lines of "can't run agent_id() before start()". So this lets that exception come through. There was another case where exceptions would get covered up but I forget at the moment. Basically exception handling was kind of a mess, it assumed that everything worked smoothly.

PR #919, opened 2019-01-24
Container UI bundle manipulation RPC

willemolding:
Depends on #876

- Adds new trait ContainerUiAdmin for Container that allows for manipulation of the configuration of ui bundles and ui interfaces
-  Adds a further set of functions to the container RPC for managing 
   static UI bundles and HTTP interfaces to these.
   This adds the following RPC endpoints:
   
   * `admin/ui/install`
   * `admin/ui/uninstall`
   * `admin/ui/list`
   * `admin/ui_interface/add`
   * `admin/ui_interface/remove`
   * `admin/ui_interface/list`
   * `admin/ui_interface/start`
   * `admin/ui_interface/stop`

This does enforce some assumptions about using the filesystem to manage the static UI bundles. I think this is OK for now since the container API already assumes the existence of the filesystem for persisting configs between executions. In future all of this needs to be plugable to allow a lite-client container.


container_api/src/container/ui_admin.rs
```diff
@@ -0,0 +1,389 @@
...
+        assert_eq!(config_contents, toml,)
+    }
+
+    #[test]
+    fn test_add_ui_interface() {
+        let mut container = create_test_container("test_add_ui_interface", 3000);
+        assert_eq!(
+            container.add_ui_interface(UiInterfaceConfiguration {
+                id: "test-ui-interface-id".into(),
+                port: 3000,
```
zippy: I think you need to use a different port number for each test, because on some platforms (travis/windows) they run asynchronously and we get a port in-use error.

container_api/src/container/ui_admin.rs
```diff
@@ -0,0 +1,389 @@
...
+    use std::{fs::File, io::Read};
+
+    #[test]
+    fn test_install_ui_bundle_from_file() {
+        let mut container = create_test_container("test_install_ui_bundle_from_file", 3000);
+        let bundle_path = PathBuf::from(".");
+        assert_eq!(
+            container.install_ui_bundle_from_file(bundle_path, &"test-bundle-id".to_string()),
+            Ok(())
+        );
```
lucksus: This test is not testing if the installed UI bundle is actually there after running the call. Just a smoke test?

container_api/src/container/ui_admin.rs
lucksus: Yeah, and in this case, you are using the same port for the websocket and the UI interface in the same test (3000 in line 207 above). This won't work.

container_api/src/container/ui_admin.rs
willemolding: Actually at this stage the UI dir isn't copied at all just served from wherever it already is. I think we do actually want the copy option though so I will add that in

container_api/src/container/ui_admin.rs
willemolding: Great to know to keep this in mind. All tests now use different ports for both UI static servers and interfaces

zippy:
hey @willemolding it looks like this is very close to merging, I see one test breaking.  Are there any other blockers?

PR #920, opened 2019-01-24
Synchronize various crate versions being used to latest supported

pjkundert:
Several sub-modules use various and sundry crate versions.  Update to latest versions, to reduce compilation time, size, and improve maintainability.  These are all minor version number changes, so APIs should be compatible.

o base64 = "0.10"
o lazy_static = "1.2"
o syn = "0.15"


lucksus:
...well, if tests would pass ;)

zippy:
I like this too but tests do need to pass, and it's weird breakage.  @thedavidmeister ?

PR #922, opened 2019-01-24
holochain_net: Changes and upgrade for publish & hold list protocol

ddd-mtl:

**Note: Start review by looking in changes in JsonProtocol.**

# JsonProtocol update

## Messages

**Added `HandleGet*List` message types**

Renamed some fields in most messages, for clarity:
 - `msg_id` -> `request_id`
 - `from_agent_id` -> `requester_agent_id`
- `agent_id` -> `provider_agent_id`
- `data` -> `entry`

Added (or renamed field) requester and provider agent_id when passing data around.

## Protocol
On TrackDna, network module will reply with 4 list requests (publish & hold, data & meta).
Core does NOT HAVE TO reply but if it does, it will allow network module to do proper bookkeeping.
If the list responses shows that the network module is missing some data, it will initiate fetchData/Meta requests to its core.
Upon receiving responses to those requests, it will do a publishData/Meta itself.

# memory_server
Now keeps tracks of "who has what", i.e. bookkeeping.
Data is stored by "bucket" in a "book". A bucket beeing 'dna+agent_id'
A book being a Hashmap bucketId -> Vec of data.
There is also a book of connected peers.
Since server also initiates requests. It keeps track of its own requests by request_id in a request book.
Renamed 'handle()' to 'serve()' to distinguish message handling by core from message handling by network module.

# test_bin

**test_bin** now has a CI test, that runs all tests with the inMemory network.

P2pNode is now mocking core by storing data and having its one handler of received Protocol messages. 
In this handler, it will basically log messages and store data when requested by HandleStore* requests.
P2pNode now also requires an agent_id.

Separated basic workflow tests from lists workflow tests.
Made separate file for constants and added lazy_statics.

# Misc
Renamed 'Get' to 'Fetch' when appropriate:
Fetch = Request between node and the network (other nodes)
Get    = Request within a node between network module and core


net_connection/src/json_protocol.rs
```diff
@@ -223,44 +335,75 @@ pub enum JsonProtocol {
     HandleSendMessageResult(MessageData),
 
     /// Request data from the dht network
-    #[serde(rename = "getDht")]
-    GetDhtData(GetDhtData),
```
neonphog: if we're changing all these to be suffixed with `Data` on the rust side, should we make the json rename match as well?

net_connection/src/json_protocol.rs
```diff
@@ -102,75 +100,189 @@ pub struct FailureResultData {
...
+/// Data Request from own p2p-module
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, DefaultJson)]
+pub struct DhtMetaRequestData {
+    pub attribute: String,
+
+    // DhtDataRequestData
+    #[serde(rename = "_id")]
+    pub request_id: String,
+    #[serde(rename = "dnaAddress")]
+    pub dna_address: Address,
```
neonphog: I feel like we might want to add some consistency to the ordering of properties on these structs. Since we almost always have a `request_id` and a `dna_address`, if they were always first, it would be easier to gloss over them when scanning through this file.

(if only we could extend a base struct so we didn't have to duplicate the code)

net_connection/src/json_protocol.rs
```diff
@@ -102,75 +100,189 @@ pub struct FailureResultData {
...
 
     #[serde(rename = "fromAgentId")]
-    pub from_agent_id: String,
+    pub provider_agent_id: String,
 
-    pub address: String,
+    #[serde(rename = "address")]
+    pub data_address: Address,
     pub attribute: String,
+    pub content: serde_json::Value,
```
neonphog: I wish there was a way these could be define in-line with the enum so we could more easily see what was going on...

I know you can do struct-style enums, but they are not very ergonomic to work with, and cause some serialization problems...

neonphog:
dude, this is amazing! I didn't go line-by-line or anything since it's a WIP, but I think you nailed what I was imagining for the protocol messages. Nice work!

net_connection/src/json_protocol.rs
ddd-mtl: Yeah, I tried to order fields from the bottom, but from the top seems more natural.
We could extend the struct by composition, but that would make things harder on the javascript side, no?

net_connection/src/json_protocol.rs
ddd-mtl: I'm having a hard time discerning what need you are expressing here.
The github diff might be confusing, as I actually did not add this 'content' field but moved it around a bit.
I went with the idea that the `*Data` struct are reusable and not make one struct per enum variant. I see how can this add a bit of confusion but in my mind it outweighs the tedious duplication of the same fields in every struct if we did one struct per enum variant.

ddd-mtl:
@neonphog thanks! Happy to know you like it and that I didn't go off-topic :)

ddd-mtl:
@lucksus I invite you to take a look of the changes I propose in `core/src/network/handler/mod.rs` and would like to know how you feel about it.
Mainly, the breaking up of `is_me()` until several functions. It seemed necessary since some message types don't carry an agent_id or a dna_address required by the former `is_me()`.

net_connection/src/json_protocol.rs
ddd-mtl: just pushed a re-ordering and a renaming of the structs and some fields.

net_connection/src/json_protocol.rs
neonphog: I want to be able to:

```rust
pub enum MyEnum {
  pub VarOne {
    pub prop_one: String,
    pub prop_i_dont_care: String,
  }
}
```

without having to match painfully like:

```rust
match v {
  VarOne { prop_one, prop_i_dont_care: _ } => { /* ... */ },
}
```

or whatever the exact syntax is

net_connection/src/json_protocol.rs
```diff
@@ -2,6 +2,8 @@
 //! basically handles serialization / deserialization from / to the core
 //! protocol message types (NamedBinary and Json).
 
+#![allow(non_snake_case)]
```
neonphog: why are we doing this? i don't see any non-snake-case variables in here. shouldn't we stick with default rust style stuff?

net/src/memory_server.rs
```diff
@@ -1,20 +1,31 @@
 //! provides fake in-memory p2p worker for use in scenario testing
 
-use holochain_core_types::cas::content::Address;
+#![allow(non_snake_case)]
```
neonphog: ditto snake case question

neonphog: ah, in this file i do see non-snake-case symbols though : )

net_connection/src/json_protocol.rs
neonphog: I'm good with this. I kinda like the composition idea... since I like to think of this rust protocol as the source-of-truth protocol spec, I don't mind requiring the javascript to conform to whatever's easiest to maintain here.

This PR's rather large already, though, if we want to undertake that change, we may want to do it as a follow-on.

net/src/memory_worker.rs
```diff
@@ -183,316 +183,4 @@ mod tests {
 
         memory_worker_1.tick().unwrap();
     }
-
-    #[test]
-    #[cfg_attr(tarpaulin, skip)]
-    fn can_memory_network_flow() {
```
neonphog: I absolutely understand wanting to take these out so we don't have to maintain two sets of basically duplicated unit tests.

Since `test_bin` is not yet in CI, though, I worry about unrelated work accidentally breaking networking if CI is not covering it.

What do you think about adding a unit `#[test]` to the `test_bin` crate that executes the full suite, but only for the "In-Memory" module, as a baby step to getting this in CI?

net_connection/src/json_protocol.rs
ddd-mtl: Its for the unit test function names like `it_can_convert_HandleSendMessageResult()`
maybe I can move it in the test mod.

net/src/memory_server.rs
ddd-mtl: need it for `priv_serve_HandleGetHoldingMetaListResult()`

net/src/memory_worker.rs
ddd-mtl: good idea! Let me try that.

ddd-mtl:
@neonphog CI is green, can you approve?

net_connection/src/json_protocol.rs
ddd-mtl: Ok, next time there is a big protocol update I'll change to compositions.

PR #923, opened 2019-01-24
WIP: block_on -> ThreadPool executor

lucksus:
- [x] Added a `futures::executor::ThreadPool` to `Instance` and a clone of it (which is a reference) to `Context`
- [x] Cleaned up initialization of `Instance` into `Instance::initialize()` so that the thread pool can already be used there for running the `workflow::application::initialize` to completion
- [x] Changed every usage of `futures::executor::block_on` to us the new `Context::block_on()` which uses the instances `ThreadPool`.
- [ ] Confirm that this enables using `hdk::commit` and `hdk::get_entry` etc in `genesis` and receive`


core/src/context.rs
```diff
@@ -184,6 +191,13 @@ impl Context {
             .as_ref()
             .expect("Observer channel not initialized")
     }
+
+    pub fn block_on<F: Future>(&self, future: F) -> <F as Future>::Output  {
+        match self.future_executor {
+            None => ThreadPool::new().expect("Default ThreadPool has to be spawnable"),
```
sphinxc0re: So, when there is no executor, you are just spawning new ones into the void?

lucksus: Yeah, I've replaced a `panic!` with this. Should only happen in tests where we use a context without an instance.

PR #929, opened 2019-01-28
Futures refactoring -> enable nested block_on

lucksus:
-[ ] I have added a summary of my changes to the changelog

# Fixing `genesis` and `receive`

This is a set of changes. The main intention is to allow nested root futures and hence nested Future executors like `block_on`. 

## We need this because 
* we call `genesis` from within the initialization workflow which gets polled to completion by `block_on` (or `receive` from within the handling of a received `send` message which also needs `block_on` to be synchronous)
* most API functions (like `hdk::commit`) are starting several asynchronous tasks, but need to await/block_on the future to be synchronous
* we can't chain futures with `async/await!` syntax through WASM -> `invoke_commit_entry` needs to use `block_on` and can not use `await!`
* using anything reasonable like `hdk::commit` within genesis/receive leads to nested `block_on`s and a panic.

## Solution
In this PR I suggest a custom `block_on` implementation in `Context::block_on` that is compatible with (and taking advantage of) our Redux action loop. Our futures, in most cases action creators, poll the Redux state for a result anyway. We do not actually need a separate executor for that since we already have the thread running the main action loop.

## Separate steps taken in this PR
* [PREP] Initialization of `Instance` cleaned up
* [PREP] Observers refactored and simplified -> just a Sender/Receiver<()> sending a tick after every state mutation
* `Context::block_on` using the new observers wait for state mutations and (re-)poll the future
* Changing all future resolving from `block_on` to `context.block_on`


lucksus:
@juliolrmonteiro Can you check if these changes fix your problem with `receive`? I see locally that a commit in genesis works with this. There is still a hanging app-spec test which is why the tests don't pass, so will need to look into that more tomorrow. But wanted to ping you early about this..

juliolrmonteiro:
For some reason I can't compile `futures-util-preview` using `nightly-2019-01-24-x86_64-unknown-linux-gnu`, I get some 9 errors complaining about `std::pin::Pin`, here's a sample error:

```rust
   Compiling futures-util-preview v0.3.0-alpha.12
error[E0308]: mismatched types
   --> /home/monteiro/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-util-preview-0.3.0-alpha.12/src/future/maybe_done.rs:110:18
    |
110 |         Pin::set(&mut self, MaybeDone::Done(res));
    |                  ^^^^^^^^^
    |                  |
    |                  expected struct `std::pin::Pin`, found mutable reference
    |                  help: consider removing the borrow: `self`
    |
    = note: expected type `std::pin::Pin<_>`
               found type `&mut std::pin::Pin<&mut future::maybe_done::MaybeDone<Fut>>`
```

lucksus:
@juliolrmonteiro are you sure you‘re nightly-2019-01-24? This is the error you‘d get if you‘re on older version. Compiled for me and on CI. Try make and maybe even cargo update..?

core/src/instance.rs
```diff
@@ -235,56 +269,17 @@ impl Instance {
...
+pub fn dispatch_action_and_wait(context: Arc<Context>, action_wrapper: ActionWrapper) {
+    let observer_rx = context.create_observer();
+    dispatch_action(context.action_channel(), action_wrapper.clone());
+
+    loop {
+        if context.state().unwrap().history.contains(&action_wrapper) {
+            return;
         } else {
-            false
+            observer_rx.recv().expect("Local channel must work");
```
maackle: Could abstract these loops out with something like:

```rust
fn block_with<F>(rx: Receiver<()>, f: F)
where F: FnMut() -> bool {
    loop {
        if f() {
            return
        } else {
            rx.recv().unwrap();
        }
    }
}
```

juliolrmonteiro:
Got it @lucksus, there was a local override for the current folder that was forcing the old nightly.
I managed to get it to compile and run with that `futures-refactor` and holofuel branch `feature-transactions-sendreceive-julio`, and looks like we have some advancements!

Now it's failing because the DHT can't find the commit that was just done by the other peer, so, perhaps by adding some wait it will be able to work.

I saw in the logs:
```
2019-01-29 11:05:48:alice::dist/bundle.json: debug/net/handle: GetDht: GetDhtData { msg_id: "puid-95-1", dna_address: HashString("QmdpyCNYKgnUXN2SkS5Jug2mXpjgRcZzy8o8FV8f2r3SWZ"), from_agent_id: "alice-----------------------------------------------------------------------------AAAIuDJb4M", address: "\"QmbJaMuJonzYfzww9sVufBv6QgctVSLG2PgBz21WKovHqd\"" }
2019-01-29 11:05:49:alice::dist/bundle.json: debug/net/handle: GetDhtResult: DhtData { msg_id: "puid-95-1", dna_address: HashString("QmdpyCNYKgnUXN2SkS5Jug2mXpjgRcZzy8o8FV8f2r3SWZ"), agent_id: "alice-----------------------------------------------------------------------------AAAIuDJb4M", address: "\"QmbJaMuJonzYfzww9sVufBv6QgctVSLG2PgBz21WKovHqd\"", content: Null }
done puid-0-5c
(node:9286) UnhandledPromiseRejectionWarning: unable to call zome function: InternalFailure(RibosomeFailed("Trap: Trap { kind: Unreachable }"))
(node:9286) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:9286) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

This probably means the `hdk::get_entry` call is returning, and not freezing anymore! Even though it's returning an error, it's good news.

Here's the branch I am working on, in case you want to reproduce it:
https://github.com/Holo-Host/holofuel/pull/14

core/src/instance.rs
lucksus: Well, some of those need to match `Some(result)` and some don't. I would actually like to get rid of most of those loops in favor of futures that then use the loop in `Context::block_on`, but as a follow up.

core/src/instance.rs
```diff
@@ -34,12 +42,53 @@ impl Instance {
         100
     }
```
Connoropolous: rustdoc

core/src/context.rs
```diff
@@ -184,6 +191,28 @@ impl Context {
             .as_ref()
             .expect("Observer channel not initialized")
     }
+
+    pub fn create_observer(&self) -> Receiver<()> {
```
Connoropolous: rustdoc

core/src/context.rs
```diff
@@ -184,6 +191,28 @@ impl Context {
...
+        let (observer_tx, observer_rx) = channel();
+        self.observer_channel()
+            .send(Observer {
+                ticker: observer_tx,
+            })
+            .expect("Observer channel not initialized");
+        observer_rx
+    }
+
+    pub fn block_on<F: Future>(&self, future: F) -> <F as Future>::Output {
```
Connoropolous: rustdoc

core/src/nucleus/mod.rs
```diff
@@ -92,35 +93,6 @@ impl EntrySubmission {
     }
 }
 
-/// Dispatch ExecuteZoneFunction to and block until call has finished.
```
Connoropolous: I did some searches for this and I guess it wasn't even being called anywhere? If so, good you removed it.

core/src/nucleus/mod.rs
```diff
@@ -129,22 +101,24 @@ pub fn call_and_wait_for_result(
 ) -> Result<JsonString, HolochainError> {
     let call_action = ActionWrapper::new(Action::ExecuteZomeFunction(call.clone()));
 
+    let id = snowflake::ProcessUniqueId::new().to_string();
+    println!("call_and_wait_for_result {}", id);
```
zippy: println!

core/src/nucleus/mod.rs
```diff
@@ -129,22 +101,24 @@ pub fn call_and_wait_for_result(
...
+    let (tick_tx, tick_rx) = channel();
+    instance
+        .observer_channel()
+        .send(Observer { ticker: tick_tx })
+        .expect("Observer channel not initialized");
+    instance.dispatch(call_action);
+
+    loop {
+        if let Some(result) = instance.state().nucleus().zome_call_result(&call) {
+            println!("done {}", id);
```
zippy: another

PR #932, opened 2019-01-29
GetEntryOptions

maackle:
Question:

How do we deal with the difference between storing one's own header vs. storing others' headers?

- During Action::Commit, the header gets added to the chain CAS.
- During Action::Hold, the entry and header get added to the DHT CAS, and metadata is added to DHT EAV

The question is do we also want to add our own header to the DHT CAS and EAV, so we can get them all from the same place? Or do we want to look in both places, adding our own header to the collection of stored headers from other agents? (This is because in general we don't Hold our own Entry)

### Answer:

We look in both places: search our chain for our own header to this entry, and also check the DHT meta storage for other headers we've seen, merging the two results into the total set of headers.

- [x] Add headers to CAS and EAV when seen during Hold
- [x] Add `headers` to GetEntryOptions
- [x] Understand the difference between AgentState.chain.content_storage and DhtStore.content_storage

## changes

- header "sources" is now "provenances"
- added triple provenance app spec test
  - added carol
  - carolpeletier
- adds `pretty_assertions` crate as a core dep
- `Action::Hold` now operates on `EntryWithHeader`
- chain header no longer optional in validation package
- DHT can find headers by checking meta/CAS
- DHT can add headers to CAS/meta
- fix flaky waiter logic in testing that made invalid assumptions about memory network

## followups

- impossible states: `EntryWithHeader` to `ChainItem` #946
- impossible states: "provenance" as real struct #947


core/src/agent/state.rs
```diff
@@ -95,6 +95,12 @@ impl AgentState {
             .iter_type(&self.top_chain_header(), &entry.entry_type())
             .find(|h| h.entry_address() == &entry.address())
     }
+
+    pub fn get_header_for_entry_address(&self, entry_address: &Address) -> Option<ChainHeader> {
```
thedavidmeister: `get_first_header_for_entry_address`?

there can be many headers for a single entry, even within a single chain

also i think we used to have this but i removed it because it boils down to applying a `find` to an `iter` that already exists

also `self.chain().iter(&self.top_chain_header())` feels wrong somehow :/

core/src/dht/actions/hold.rs
```diff
@@ -4,28 +4,28 @@ use crate::{
...
 ) -> Result<Address, HolochainError> {
-    let action_wrapper = ActionWrapper::new(Action::Hold(entry.clone()));
+    let action_wrapper = ActionWrapper::new(Action::Hold(entry_wh.clone()));
     dispatch_action(context.action_channel(), action_wrapper.clone());
 
     await!(HoldEntryFuture {
-        context: context.clone(),
-        address: entry.address(),
+        context: context,
+        address: entry_wh.entry.address(),
```
thedavidmeister: don't collapse names of variables like this, write the whole thing out :P

core/src/dht/dht_store.rs
```diff
@@ -61,6 +66,51 @@ impl DhtStore {
...
+                Some(entry_address),
+                Some(ENTRY_HEADER_ATTRIBUTE.to_string()),
+                None,
+                Default::default(),
+            )?
+            .into_iter()
+            // get the header addresses
+            .map(|eavi| eavi.value())
+            // fetch the header content from CAS
+            .map(|a| self.content_storage().read().unwrap().fetch(&a))
```
thedavidmeister: any way to not unwrap here?

core/src/dht/dht_store.rs
```diff
@@ -61,6 +66,51 @@ impl DhtStore {
...
+        &self,
+        entry: &Entry,
+        header: &ChainHeader,
+    ) -> Result<(), HolochainError> {
+        let eavi = EntityAttributeValueIndex::new(
+            &entry.address(),
+            &ENTRY_HEADER_ATTRIBUTE.to_string(),
+            &header.address(),
+        )?;
+        self.content_storage().write().unwrap().add(header)?;
```
thedavidmeister: unwrap

core/src/dht/dht_store.rs
```diff
@@ -61,6 +66,51 @@ impl DhtStore {
...
+        self.meta_storage().write().unwrap().add_eavi(&eavi)?;
```
thedavidmeister: unwrap

core_types/src/entry/mod.rs
```diff
@@ -137,6 +137,8 @@ pub struct EntryWithMeta {
     pub maybe_crud_link: Option<Address>,
 }
 
+pub type EntryWithMetaWithHeaders = (EntryWithMeta, Vec<ChainHeader>);
```
thedavidmeister: :fearful:

core/src/agent/state.rs
maackle: Thanks for pointing this out, it was a real bug.

I introduced `iter_chain` to  `AgentState` for this purpose, which just iterates from the top header, what do you think?

core/src/dht/dht_store.rs
maackle: @thedavidmeister Didn't we agree that it's OK to unwrap when locking a mutex or rwlock? All the unwraps are of that variety.

core/src/dht/actions/hold.rs
maackle: How about we deal with this after the `ChainItem` refactor?

core/src/dht/dht_store.rs
maackle: I'd like to defer on all the unwraps until we get our error handling story straight (https://github.com/holochain/holochain-rust/issues/949)

PR #941, opened 2019-02-01
[WIP] - New macro HDK 

willemolding:
# HDK Improvements 

Rust edition 2018 allows for defining procedural macros. These allow for arbitrary transformations to be made to modules, functions and structs/enums at compile time. These were not available at the time the current HDK was created.

This is used by popular rust crates (notable Serde and Rocket.rs) to allow simple DSLs. For example in Rocket.rs it is used to define which functions act as callbacks to HTTP reqests:
```rust 
#[get("/")] // example of an attribute-like macro
fn hello() -> &'static str {
    "Hello, world!"
}
```

---

The flexibility of proc macros means there almost unlimited ways to define a DSL that makes zome development in rust as productive as possible with minimum boiler-plate. This PR serves as a place to discuss both what the HDK should look like from a developer perspective but also what is easy/possible to implement.

## Initial Proposal

```rust
#![hdk::zome] // this attribute-like macro operates on the whole module. This is equivalent to define_zome! in the old HDK

#[hdk:entry] // instructs the top level macro that a entry should be created that maps to this struct
struct Post {
    content: String,
}

#[hdk:entry(validation = "validate_blog", validation_package = ChainFull)]
// provides custom validation callback and also an idea of how links could be defined
struct Blog {
    owner: Agent,
    #[hdk::links("posts")]
    posts: vec<Post>,
}

fn validate_blog(entry: BlogEntry, validation_data: ValidationData) -> ZomeApiResult<()> {

}

/////////////////////////////////////////////////////////////////

#[hdk::genesis] // genesis callback
fn genesis() -> bool {
	true
}

#[hdk::receive] // receive callback
fn receive() -> ZomeApiResult<()> {
}

// a zome api function that belongs to these two traits 
// (this maybe needs some discussion)
#[hdk::zome_function(traits = ["zome", "bridge"])]
fn get_posts() -> ZomeApiResult<String> {
}

#[hdk::zome_function(traits = ["plumbus"])]
fn delete_all_posts_idc() -> ZomeApiResult<> {
}
```

Considerations:
- It should still be possible to split entries/callbacks between multiple files
- All of the customization of define_zome! must be supported
- The GraphQL schema generation should be an extension that works on top of what is defined here


jmday:
Suggestion:

Consider providing an abstracted `LinkList` instead of `vec` so that:
```
struct Blog {
    owner: Agent,
    #[hdk::links("posts")]
    posts: vec<Post>,
}
```
becomes
```
struct Blog {
    owner: Agent,
    #[hdk::links("posts")]
    posts: LinkList<Post>,
}
```

jmday:
Is there / will there be a default for validation proc macro parameters (when not specified)?

For example, does/will `#[hdk:entry]` get interpreted as something like `#[hdk:entry(validation = "__validate_default", validation_package = None)]`?

jmday:
Could there be a default genesis callback that returns true without the need to define one at all unless a non-default is desired?

ddd-mtl:
How about, instead of having a tag per callback (#[hdk::genesis], #[hdk::receive)] having a tag for callbacks in general, and have the name of the function informe what callback it is:
```Rust
#[hdk::zome_callback]
fn genesis() -> bool {
	true
}
```

If it thats now possible, how about adding it as a parameter:
```Rust
#[hdk::zome_callback(name="genesis")]
fn genesis() -> bool {
	true
}
```

pospi:
If it's possible, I would love to see the `#[hdk::links("...")]` macro be contextual, based on the following struct field. If the field is a vector-like type, the relationship is 1:many. Otherwise it is 1:1.

Something else we are interested in for [HoloREA](https://github.com/holo-rea) is authoring GraphQL schemas using the [GraphQL SDL](https://graphql.org/learn/schema/). In ValueFlows we are [looking in to](https://github.com/valueflows/valueflows/issues/447) a reference schema authored in a language-agnostic format that can be used to both validate the spec and create an implementation; as there are ValueFlows implementations on a handful of other platforms and we want them all to eventually be compatible. It's quite likely Juniper likely already has methods to parse the SDL to generate a schema, most GraphQL implementations seem to allow that. Not sure if this would be a common use-case though, or just an "us" thing. Still, SDL is a super simple language and the option of using it may be a big benefit to adoption.

willemolding:
> If it's possible, I would love to see the `#[hdk::links("...")]` macro be contextual, based on the following struct field. If the field is a vector-like type, the relationship is 1:many. Otherwise it is 1:1.
> 
> Something else we are interested in for [HoloREA](https://github.com/holo-rea) is authoring GraphQL schemas using the [GraphQL SDL](https://graphql.org/learn/schema/). In ValueFlows we are [looking in to](https://github.com/valueflows/valueflows/issues/447) a reference schema authored in a language-agnostic format that can be used to both validate the spec and create an implementation; as there are ValueFlows implementations on a handful of other platforms and we want them all to eventually be compatible. It's quite likely Juniper likely already has methods to parse the SDL to generate a schema, most GraphQL implementations seem to allow that. Not sure if this would be a common use-case though, or just an "us" thing. Still, SDL is a super simple language and the option of using it may be a big benefit to adoption.

GraphQL integration is something we discussed in depth at the retreat just passed. It seems like a really great fit for the way we structure and retrieve data. Using Juniper it is already possible to handle GraphQL queries but it does not support the SDL. What we settled on is a SDL-like macro DSL. This PR is the first step toward that.

willemolding:
> the ergonomics of this seem nicer and more idiomatic at a high level than the current set of macros
> 
> i don't want to bike shed anything here as real world usage will quickly provide feedback
> 
> i agree with @AshantiMutinta that the many to one relationship between entries and structs is a common gotcha to consider in hdk design

@thedavidmeister Any ideas for resolving the many-to-one entry-to-struct dilemma? I remember this came up last time when we were talking about a new HDK

Connoropolous:
@willemolding I'm wondering whether we could add some info/thoughts here about what kinds of custom Holochain validation could be done on a Zome file in this PR, and further spec out the proc macro?

Connoropolous:
Also, I want to note that I think we should be strategic and well thought out about the timing of
a. giving any dev resources to this, and
b. when we would release it

willemolding:
> Could there be a default genesis callback that returns true without the need to define one at all unless a non-default is desired?

In response to this and the above comment it is super easy to add sensible defaults and that seems to be a common pattern in other libs that use proc macros such as Serde

willemolding:
> Suggestion:
> 
> Consider providing an abstracted `LinkList` instead of `vec` so that:
> 
> ```
> struct Blog {
>     owner: Agent,
>     #[hdk::links("posts")]
>     posts: vec<Post>,
> }
> ```
> 
> becomes
> 
> ```
> struct Blog {
>     owner: Agent,
>     #[hdk::links("posts")]
>     posts: LinkList<Post>,
> }
> ```

This seems to make sense I think. If the collection is to be represented using links then this whole struct field is removed by the macro anyway so it can be whatever type we like.

philipbeadle:
> @willemolding I'm wondering whether we could add some info/thoughts here about what kinds of custom Holochain validation could be done on a Zome file in this PR, and further spec out the proc macro?

I'd love it if we could add attributes to fields such as [0-143] and generate validation code that checks the value is between 0-143.

maackle:
Just throwing an idea into the mix: I'd prefer to see this as Rusty as possible. For instance I consider it more idiomatic to replace

```rust
#[hdk:entry(validation = "validate_blog", validation_package = ChainFull)]
struct Blog {
    owner: Agent,
    #[hdk::links("posts")]
    posts: vec<Post>,
}

fn validate_blog(entry: BlogEntry, validation_data: ValidationData) -> ZomeApiResult<()> {

}
```

with:

```rust
#[hdk:entry]
struct Blog {
    owner: Agent,
    #[hdk::links("posts")]
    posts: vec<Post>,
}

impl Validation for Blog {
    type ValidationData = ChainFull;

    fn validate(entry: Self, validation_data: Self::ValidationData) -> ZomeApiResult<()> {
        
    }
}
```

if possible. i.e. use the type system as much as possible rather than strings. Then we can require all entries to implement `Validation` and the compiler helps us and we don't have to add extra checks in the macro.

thedavidmeister:
@willemolding 

> Any ideas for resolving the many-to-one entry-to-struct dilemma? I remember this came up last time when we were talking about a new HDK

i don't see it as an issue that it exists, it's an issue that people will probably keep trying to make it not exist w/o understanding the implications

the entries work the way that we want/need them to right now

rust enums/types are a verbose tool and macrology isn't a silver bullet

( ͡~ ͜ʖ ͡°)

thedavidmeister:
@willemolding 

> In response to this and the above comment it is super easy to add sensible defaults and that seems to be a common pattern in other libs that use proc macros such as Serde

very true

serde hides an absurd amount of boilerplate and complexity behind sane defaults

willemolding:
> Just throwing an idea into the mix: I'd prefer to see this as Rusty as possible. For instance I consider it more idiomatic to replace
> 
> ```rust
> #[hdk:entry(validation = "validate_blog", validation_package = ChainFull)]
> struct Blog {
>     owner: Agent,
>     #[hdk::links("posts")]
>     posts: vec<Post>,
> }
> 
> fn validate_blog(entry: BlogEntry, validation_data: ValidationData) -> ZomeApiResult<()> {
> 
> }
> ```
> 
> with:
> 
> ```rust
> #[hdk:entry]
> struct Blog {
>     owner: Agent,
>     #[hdk::links("posts")]
>     posts: vec<Post>,
> }
> 
> impl Validation for Blog {
>     type ValidationData = ChainFull;
> 
>     fn validate(entry: Self, validation_data: Self::ValidationData) -> ZomeApiResult<()> {
>         
>     }
> }
> ```
> 
> if possible. i.e. use the type system as much as possible rather than strings. Then we can require all entries to implement `Validation` and the compiler helps us and we don't have to add extra checks in the macro.

I believe because it is a macro you still get all the type validation even when using a string to specify the validation function. In fact it isn't really a string it is just the way to write literals inside attributes. serde_json uses the same pattern for specifying custom serializers for fields and they will correctly do type inference/compile time errors.

That said I like the `impl Validation` WAAAAAAy more! I think we should definitely do it this way.

willemolding:
> @willemolding
> 
> > Any ideas for resolving the many-to-one entry-to-struct dilemma? I remember this came up last time when we were talking about a new HDK
> 
> i don't see it as an issue that it exists, it's an issue that people will probably keep trying to make it not exist w/o understanding the implications
> 
> the entries work the way that we want/need them to right now
> 
> rust enums/types are a verbose tool and macrology isn't a silver bullet
> 
> ( ͡~ ͜ʖ ͡°)

Do you have a suggestion of how we could modify the proposed macro DSL to permit single-struct-multi-entry or are you suggesting we leave it as is?

thedavidmeister:
@willemolding @maackle  `impl Validation` is nice and rusty... i have questions...

- validation is done on an entry, not a struct, so it would have to be `impl Validation for Entry` which is kind of pointless if we just end up with an internal `match` to replicate the current fn handlers, how do we maintain the current fn level dispatch logic? would it work something like the dispatch enum in the hdk?
- `AppEntryValue` is anything that implements `JsonString` which is more than structs, it can also be raw `String`s or `RawString`s, such as base64 encoded binary data as highlighted by the community, so implementing validation against something like `RawString` is hard and maybe impossible as you may not know how to dispatch validation logic without the app entry type for it, how does the trait impl look for non-struct data?
- i'm not exactly sure how the `ValidationData` type of the `Validation` trait is used in practise? i thought you could only use that in fn signatures, i didn't realise you could use it to dispatch/match against in order to send the right thing through at runtime like an enum or whatever? i think i just need to see a worked example for that bit...

thedavidmeister:
@willemolding hmm on closer/fresh inspection i'm not sure i understand how this is supposed to work

`Entry` is an enum with many variants, one happens to be `Entry::App` that contains a tuple of type/value which is a string/json pair

i don't understand exactly what it means to "create an entry that maps to a struct"

```rust
let e1 = Entry::App("foo".into(), Foo{ .. }.into());
let e2 = Entry::App("bar".into(), Foo{ .. }.into());
```

how do i unambiguously get from a given `Foo` to both/either `e1` and/or `e2`?

afaics, the only way to map from a `Foo` to an `Entry` either as a standalone data structure or when dispatching validation etc. is to force a 1:1 relationship between structs and app entry types managed/generated by macros

maybe this could be alleviated through use of `From`?

if the macros were doing something like...

```rust
let e1 = Entry::App("foo".into(), Foo{ .. }.into());
let e2 = Entry::App("bar".into(), Bar { .. }.into());

struct Baz { .. }
impl From Baz for Foo { .. }
impl From Baz for Bar { .. }

let e1 = Entry::App("foo".into(), Baz{ .. }.into().into());
let e2 = Entry::App("bar".into(), Baz{ .. }.into().into());
```

maackle:
> validation is done on an entry, not a struct, so it would have to be `impl Validation for Entry`

Then let's say that validation is done on a `Validation + Entry`

thedavidmeister:
> > validation is done on an entry, not a struct, so it would have to be `impl Validation for Entry`
> 
> Then let's say that validation is done on a `Validation + Entry`

@maackle could you give a code example showing how that works re: my last comment?

maackle:
@thedavidmeister no, because my `Validation + Entry` idea was half-baked. Entry isn't a trait so of course that won't work.

Something that's coming up for me in this convo is that it would be a lot simpler if we really could "create an entry that maps to a struct". To me that means a 1:1 relationship between a struct and an entry type is established, so that that struct cannot be used to back any other entry type. Seems it would have to be done through macros/annotations. Then we could ambiguously round-trip between struct and entry type trivially.

You might say this is a loss of expressiveness, but I can't imagine there being many situations where you would actually want to re-use a struct for multiple entry types. Even if there were such a situation, I can't imagine it being terribly onerous to redefine the struct N times. In fact, I imagine if such a situation arose, it would probably be coincidence that two entry types had the exact same data shape (maybe for some trivial entries like `Comment { message: String }` and `DirectMessage { message: String }`), in which case it kinda makes sense to have separate structs so that they are allowed to drift apart if the spec changes later.

If we had that 1:1 relationship, then we really could attach validation to the struct itself.

willemolding:
After discussion with David we settled on using function-like macros instead of attribute macros for entries. This totally decouples the entries from the structs (or anything that implements To/From JsonString.

Something like this

```
#[derive(EntrySerializable)]
struct Post {
...
}

define_entry! {
   name: "entry1",
   type: Post,
   validationCallback: validationFunction, 
   links: []
}

```

willemolding:
Art raised an interesting point today that whatever we do should be compatible with code generation tools such as https://github.com/holochain/hcrs. There needs to be a way to keep generated code and user code seperate

thedavidmeister:
context for "after discussion with david":

- leaky/subtle abstractions are often worse to work with and harder to fix later than explicit/verbose code with no abstractions
- entry value is a json string at the end of the day, so leaning in too hard to the coupled entry/struct ideas makes it harder to "dial back" to regular string handling in the case of binary data etc.
- concerns around the ability for zome devs to "progressively go deeper" when using the HDK, rather than being forced with a binary choice between "full macro" or "full wasm" if we over-sugar
- right now app entries are the only entry variants exposed to zomes, which seems to be a source of confusion and desire for coupling (because they are the same thing, right?), but the migrations work will likely expose other system entry variants to zome devs https://hackmd.io/5OQgaCSbSvWj4eqiHmV27Q# that won't work like app entry variants at all (no json and direct interaction with DNA addresses)
- locking a struct to an entry type so that it cannot be used elsewhere runs counter to rust philosophy (e.g. all the discourse around why types cannot be negated), which is verbose but highly decoupled - this idea makes zome code not play well in the world of crates and commons based coding because nobody else can re-use your structs with their own entries during zome composition
- the worst immediate pain that we are feeling that is causing us to repeatedly reach for tight entry/struct coupling seems to be the awkwardness of the `#[derive(..)]` syntax and implementation of json handling, so that's what @willemolding's suggestion aims to address

willemolding:
Update: I think for the time at least I will use the existing entry! macro and allow for adding attributes to functions that signify they are callbacks that return entry defs. 

e.g. 
```rust

define_zome!{
entries: [
    post_definition()
]

...


pub fn post_definition() -> EntryDefinition {
   ...
}

```

becomes 

```
#[zome]
mod my_zome {

#[entry_def]
pub fn post_definition() -> EntryDefinition {
   ...
}

}
```

willemolding:
I have also decided to change the way functions are added to traits. The previous proposal was 
```
[zome_fn(traits = ["hc_public", "other_trait"])]
fn some_func() {
```

but I plan to implement it as
```
[zome_fn("hc_public", "other_trait")]
fn some_func() {
```
which is slightly more succinct and also more consistent with other proc-macro crates such as Rocket.rs

maackle:
Looks amazing! I think this new `zome_fn` is a big win in every way...except traits.

Personally I think I prefer the old style of defining traits as a list of functions. While the new way is more writeable, the old way seems more readable. It's easier to grasp what a particular trait is at a glance by seeing the closed set of functions that comprise it, rather than looking around for every function definition  that references that trait name.

But, I'm also a little late to the conversation, and feel bad for saying so after it's already implemented! I'd still be happy to see it go in this way.

jmday:
I, too, and loving the direction this is going. It always felt odd to essentially define a zome function twice and this change eliminates that issue.

I'm going to offer a different take (than maakle) on the trait definition style: While it would be nice to have a convenient way to see all the functions associated with a trait, I sense it is also important to be able to quickly see all the traits associated with a specific function, which the current hdk format makes easy. While the ability to have both quick-references would be great, requiring duplication would not and my sense is that the list of traits by function is of equal or greater value than a list of functions by trait.

willemolding:
> Looks amazing! I think this new `zome_fn` is a big win in every way...except traits.
> 
> Personally I think I prefer the old style of defining traits as a list of functions. While the new way is more writeable, the old way seems more readable. It's easier to grasp what a particular trait is at a glance by seeing the closed set of functions that comprise it, rather than looking around for every function definition that references that trait name.
> 
> But, I'm also a little late to the conversation, and feel bad for saying so after it's already implemented! I'd still be happy to see it go in this way.

Nah all good this is definitely the time to be iterating. I want to make sure we are absolutely happy with everything before we show it to the public.

I think you are right about the traits. The way we are planning to use them it probably makes more sense to group functions under a trait rather than traits on a function. I think the main reason I did it this way is I couldn't figure out a syntax I like haha. Any ideas? The only constraint is it needs to be valid rust code. How about

```rust
#[zome]
mod some_zome {

    #[zome_fn]
    pub fn my_func() {
    }

   #![traits {hc_public: [my_func]}]
    
}
```

maackle:
Oof, yeah, I can't think of a good syntax for that yet either. Are macros off the table? :)

Maybe this?

```rust
#[zome]
mod some_zome {

    #[zome_fn]
    pub fn my_func() {

    }

    #[zome_trait]
    static hc_public: &[&str] = &[
    	"my_func"
    ]

    #[zome_trait]
    static another_trait: &[&str] = &[
    	"my_func"
    ]
}
```

zippy:
That last one @maackle doesn't work for me because it makes you think that a trait is an actual low level rust thing what with the static and the &[] stuff, and it's really not.  It's really quite meta, so it feels better to me to have it inside the "#[ .. ]"

If we were to go with the associate functions with trait it makes more sense to future proof this and make it more explicit that the params are traits with something like `#[zome_fn(traits:["hc_public"], other_meta:"value")]` 

But I do think I agree with @jmday that we want to declare the trait as a unity, so that to understand what a trait is, you don't have to look all over the place in code.  I actually think your suggestion @willemolding  (`#![traits hc_public(my_func), another_trait(my_func)]`) is perfectly adequate.  The only problem would be if this has to be on one line or something which renders undreadable for lots of traits or for complex traits.  Remember that it's actually not likely that `hc_public` will be full of lots of functions, and not every function has to be part of a trait.

pdaoust:
This is exciting stuff. Trying to track the traits conversation, and I don't claim to know enough about idiomatic Rust and what's possible with macros. But to me, defining the trait entirely in a `#[]` style macro, either to list out the traits explicitly or to mark a static which lists out the traits, feels stylistically out of step with the rest of the design. The entry def is returned from a function, so it'd feel more consistent if the traits were returned from a function too (or alternatively, if the entry def were instantiated as a static). What's the design rationale behind having entry def as a function?

maackle:
```rust
#![traits 
    hc_public(
        my_func,
        another_func
    ), 
    another_trait(
        my_func
    )
]
```

That could work. It's still a little wild though. I also don't know if if newlines are significant in a construction like that.

If we're gonna get that custom with it, what's the benefit of using a free-floating attribute like this, vs. a macro?

willemolding:
Actually it's not a free floating attribute. The bang(!) means it is
applied to the containing code so it is an additional attribute on the zome
module.

I'm still not happy with it though. This seems like one of those cases
where there is too much freedom and it is going to be really hard to make a
choice

maackle:
Going back to [the idea of using static slices of strings](https://github.com/holochain/holochain-rust/pull/941#issuecomment-484573926)... @zippy said that doesn't work for him because it makes a trait look too low-level. However the *definition* of a trait really is just that: a collection of strings that correspond to function names. The underlying trait may be complex but the genesis is that list of strings, in the context of a zome.

I think this just comes down to the fact that Rust requires explicit types for static data, and the fact that syntax for slices of slices is noisy. What if we provide a simple macro that is syntactic sugar for creating a static slice of slices? It wouldn't be doing anything more than hiding that ugly syntax.

willemolding:
Closing this for discussion. The implementation can be found in #1326

PR #944, opened 2019-02-01
pass opaque network configuration via stdin on child process spawn

neonphog:

this is related to n3h proxy / relay - https://github.com/holochain/n3h/pull/35

We need to pass end-user configured configuration to the networking module (all of which should be opaque to rust core... i.e. it doesn't need to parse / alter / read / write / anything the data).

This is the first step, ended up passing it to the child process over stdin rather than creating a file as I had originally thought, since it was much simpler this way.

Hopefully @ddd-mtl or someone else with more knowledge of how the composer loads config files can take it from here. (right now there are just `let opaque_net_config = String::from("{}");` stubs in the locations that need this data for subprocess spawning).

### DIFF
--- a/container_api/src/container/base.rs
+++ b/container_api/src/container/base.rs
@@ -235,6 +235,8 @@ impl Container {
                 "attempt to spawn network when not configured".to_string(),
             ))?;
 
+        let opaque_net_config = String::from("{}");
+
         println!(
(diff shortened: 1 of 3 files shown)


neonphog:
PS, tested this out with n3h master - it is backward / forward compatible. master n3h will just ignore the stdin data.

PR #945, opened 2019-02-02
Fix dead edit link - now links correctly to github

jonathan-s:

As described in the title. A minor change like this probably does not need to be in the changelog.


jonathan-s:
Yup, I built it too.

PR #952, opened 2019-02-05
[DON'T MERGE] Rewrite Attribute as an enum

maackle:
**Wait until #780 is merged before merging this**

- [x] Rewrites Attribute as an enum, no more stringly typed attributes
- [x] Removes Default impl for EAVI, requiring some major logic rewriting
- [x] ~I have added a summary of my changes to the changelog~ (nothing to add)


core/src/nucleus/actions/get_entry.rs
```diff
@@ -119,12 +119,7 @@ pub fn get_entry_with_meta<'a>(
         Ok(Some(entry)) => entry,
     };
     // 2. try to get the entry's metadata
-    let maybe_meta = get_entry_crud_meta_from_dht(context, address);
-    if let Err(err) = maybe_meta {
-        return Err(err);
-    }
-    let (crud_status, maybe_crud_link) = maybe_meta
-        .unwrap()
+    let (crud_status, maybe_crud_link) = get_entry_crud_meta_from_dht(context, address)?
```
Connoropolous: `?.expect(...)` that's a thing?

core_types/src/eav.rs
```diff
@@ -19,13 +19,36 @@ use std::{
...
+
+    #[serde(rename = "link")]
+    Link,
+
+    LinkTag(String),
+}
+
+impl fmt::Display for Attribute {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
```
Connoropolous: Here we have some dashes and some underscores, anything we can do about that?

core/src/nucleus/actions/get_entry.rs
maackle: Sure, it's a `Result<Option<_>>` so the `?` unwraps the Result and the `expect` unwraps the Option.

core_types/src/eav.rs
maackle: This is just what we've always had, but now would be a good time to standardize the names since they're all in one place now.

maackle:
FYI @ddd-mtl and @neonphog, we are standardizing Attribute names in core through this enum, I didn't touch networking code but feel free to use this enum (instead of `String`) and add whatever variants you need, if you want.

core_types/src/eav.rs
```diff
@@ -19,13 +19,41 @@ use std::{
...
 /// Address of AddressableContent representing the EAV entity
 pub type Entity = Address;
 
-/// Using String for EAV attributes (not e.g. an enum) keeps it simple and open
-pub type Attribute = String; // BUT we totally could make it an enum with a Custom variant!
+/// All Attribute values are pre-defined here. If ever a user-defined Attribute is needed,
+/// just add a new Custom variant for it with a String parameter
+#[derive(PartialEq, Eq, PartialOrd, Hash, Clone, Debug, Serialize, Deserialize, DefaultJson)]
+pub enum Attribute {
+    #[serde(rename = "crud-status")]
```
sphinxc0re: you can simply write
```rust
#[serde(rename_all = "lowercase")]
pub enum Attribute {
...
```

core_types/src/eav.rs
sphinxc0re: I agree and we should also choose an automation method for this

maackle: @sphinxc0re how?

sphinxc0re: I thought there might be a derive macro for Display and FromStr somewhere

sphinxc0re: https://crates.io/crates/strum

core/src/nucleus/actions/get_entry.rs
Connoropolous: Cool, k!

core_types/src/eav.rs
sphinxc0re: More docs: https://github.com/Peternator7/strum#additional-attributes

ddd-mtl:
@maackle Thanks for the heads up! I believe networking should be agnostic to this, but if we ever need to have different behaviors depending on attribute type, I'll surely use this enum. 👍

ddd-mtl:
Just found this:
```Rust
lazy_static! {
    static ref LINK: Regex =
        Regex::new(r"^link__(.*)$").expect("This string literal is a valid regex");
}
```
in get.rs
Maybe do a search all for "link__" and make sure that the enum is used everywhere.

core_types/src/eav.rs
```diff
@@ -19,13 +19,34 @@ use std::{
...
 
 /// Address of AddressableContent representing the EAV entity
 pub type Entity = Address;
 
-/// Using String for EAV attributes (not e.g. an enum) keeps it simple and open
-pub type Attribute = String; // BUT we totally could make it an enum with a Custom variant!
+/// All Attribute values are pre-defined here. If ever a user-defined Attribute is needed,
+/// just add a new Custom variant for it with a String parameter
+#[derive(PartialEq, Eq, PartialOrd, Hash, Clone, Debug, Serialize, Deserialize, DefaultJson)]
+#[serde(rename_all = "lowercase")]
```
sphinxc0re: Sorry, I actually meant:
```suggestion
#[serde(rename_all = "snake_case")]
```

core_types/src/eav.rs
thedavidmeister: relevant context, fyi only - https://github.com/holochain/holochain-rust/pull/750/files - we previously broke windows via A names that weren't file system friendly

core_types/src/eav.rs
```diff
@@ -19,13 +20,74 @@ use std::{
...
-use std::fmt::Debug;
+use std::{
+    convert::TryFrom,
+    fmt::{self, Debug},
+};
 
 /// Address of AddressableContent representing the EAV entity
 pub type Entity = Address;
 
-/// Using String for EAV attributes (not e.g. an enum) keeps it simple and open
```
thedavidmeister: i wrote this originally and i approve this change :+1:

lucksus:
@maackle, this could now be merge, right? (after resolving conflicts)
@sphinxc0re, I see your change request implemented. Could you approve?

maackle:
This has moved over to https://github.com/holochain/holochain-rust/pull/1024 because a lot of extra groundwork (#1018) had to be laid and included in order to merge develop and resolve conflicts.

PR #953, opened 2019-02-05
Switched to standard Path api

sphinxc0re:

This removes the need for `MAIN_SEPARATOR` and makes the code much more durable


sphinxc0re:
@zippy The tests fully run for me locally so it seems like a weird CI issue to me

PR #958, opened 2019-02-06
holochain_net: Test config file + tweetlogger + Meta as a list

ddd-mtl:

## JsonProtocol Update: Meta content_list

**DhtMetaData.content changed to content_list (a vec).**
So now we can handle many different values for the same attribute, and also aggregate them.

Added MetaTuple and MetaKey.
MetaTuple is used for identifying unique meta values, and can be converted into a smaller form: MetaId.
meta_list is now a vec of MetaTuple.

**New test: many_meta_test()**
Makes sure that content_list is working properly.

### MetaStore
Added an abtraction for storing metadata in p2p_node.
Does the hashing of a meta_content.

JSON is pretty explicit:
Define network modes, test suites, and loggers to use when launching the test program.

Rust version of the tweetlog in n3h: https://github.com/holochain/n3h/tree/master/packages/tweetlog
Tried to keep the same api as much as possible.
Added macros for ease of use.
Changed all the commented printf to use the tweetlog instead.


net_connection/src/json_protocol.rs
```diff
@@ -206,7 +214,9 @@ pub struct DhtMetaData {
     pub entry_address: Address,
 
     pub attribute: String,
-    pub content: serde_json::Value,
+    // single string or list of hashs
+    #[serde(rename = "contentList")]
```
sphinxc0re: I don't know why you choose this formatting but this is the only place in the project where we choose camelCase.

neonphog: camelCase is far more standard as a convention for json properties. Changing from this at this point would be a much larger discussion, as all the networking modules, including n3h have been built to this protocol spec with camelCase.

zippy:
I don't understand why we add yet another logging abstraction on top of the one that already exists and which is meant to be configurable at the container level.  Can you explain?  Does the networking code not have access to the logger?

lucksus:
Why are you using `TweetLog` instead of the already existing logger? Because you need it in the memory network which does not have a `Context`?

lucksus:
Ha, didn't see @zippy's comment before posting mine.. ;)

test_bin/src/main.rs
```diff
@@ -61,70 +66,134 @@ lazy_static! {
...
+        );
+    } else {
+        config_path = args[1].clone();
     }
-    let n3h_path = args[1].clone();
-    if n3h_path == "" {
-        usage();
+    if config_path == "" {
+        println!(
+            "Usage: No config file supplied. Using default config: test_bin/data/test_config.json"
```
neonphog: looks like we might end up printing this `Usage` line twice with the `args.len()` check above.

zippy:
FYI the test that failed in windows is something @lucksus and I are looking into right now.  I don't think it has anything to do with you.  We also see that occasionally in the linux build.

test_bin/src/main.rs
```diff
@@ -61,70 +66,134 @@ lazy_static! {
...
+        );
+        config_path = format!(
+            "test_bin{}data{}test_config.json",
+            std::path::MAIN_SEPARATOR,
+            std::path::MAIN_SEPARATOR
+        )
```
neonphog: think you can do

```rust
format!("test_bin{0}data{0}test_config.json", std::path::MAIN_SEPARATOR)
```

test_bin/data/test_config.json
```diff
@@ -0,0 +1,24 @@
+{
+  "N3H_PATH": "C:\\github\\n3h",
+  "modes": {
+    "IN_MEMORY": false,
+    "IPC_MOCK": false,
+    "HACK_MODE": true
```
neonphog: if this is the default one, do we need `IN_MEMORY` to be the only one that is true for CI?

test_bin/src/main.rs
ddd-mtl: Good call. Indeed I got a bit sloppy there.
Fixed.

test_bin/src/main.rs
ddd-mtl: Nice. thanks!

test_bin/data/test_config.json
ddd-mtl: The default file is not used by CI.
CI is calling IN_MEMORY for all suites directly.
This is more like the default file for testing locally and debugging.
I did update some values since your comment.

ddd-mtl:
@zippy @lucksus net crate is agnostic from Core and Conductor so I don't have access to the other Logging abstractions.
The goal of this one is to ease debugging of the network code, so its designed to be more in-tune with the logger in n3h, than the one in Core/Conductor.
Maybe later we can adapt this abstraction so it can fit the one given by Conductor.

PR #960, opened 2019-02-06
added mute env var for simple logger to decrease CI log size

zippy:

Loogging output during CI has made it so you can't see all the test results without downloading a file.  This PR adds an environment variable that mutes that can mute that output and the nix CI is set to use it.

### DIFF
--- a/core/src/logger.rs
+++ b/core/src/logger.rs
@@ -21,8 +21,14 @@ pub struct SimpleLogger {
 #[cfg_attr(tarpaulin, skip)]
 impl Logger for SimpleLogger {
     fn log(&mut self, msg: String) {
-        let date = Local::now();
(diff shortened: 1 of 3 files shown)


core/src/logger.rs
```diff
@@ -21,8 +21,14 @@ pub struct SimpleLogger {
 #[cfg_attr(tarpaulin, skip)]
 impl Logger for SimpleLogger {
     fn log(&mut self, msg: String) {
-        let date = Local::now();
-        println!("{}:{}", date.format("%Y-%m-%d %H:%M:%S"), msg);
+        match std::env::var("HC_SIMPLE_LOGGER_MUTE") {
```
sphinxc0re: You don't have to match here. Please use `.is_err()`

PR #961, opened 2019-02-06
[WIP] Pending links

maackle:
## Validating out-of-order entries

This PR lays the groundwork for validating and holding out-of-order entries, and uses that groundwork to solve the particular problem of a Link arriving at a node before its base.

### Proposed changes for groundwork:
- Add a `Attribute::PendingEntry` variant which refers to the address of the entry being waited for
- Add a `CrudStatus::Pending` variant, to help prevent accidentally serving out an unvalidated Entry
    - This pairs well with the as-of-yet unused `CrudStatus::Rejected`

### Implementation for out-of-order Links fix
- In `reduce_add_link`, always add the Link, but use the new CrudStatus and Pending attribute to mark it as pending if base not found
- In the hold_entry workflow, check for Pending EAV entries with the same base, and queue them for re-validation (TODO: how?)


maackle:
Starting over at #1026

PR #966, opened 2019-02-07
make hash optional

Connoropolous:
- [x] I have added my changes to the CHANGELOG


conductor_api/src/config.rs
```diff
@@ -314,11 +314,13 @@ impl From<AgentConfiguration> for AgentId {
...
 /// A DNA is represented by a DNA file.
-/// A hash has to be provided for sanity check.
+/// A hash can optionally be provided, which could be used to validate that the DNA being installed
+/// is the DNA that was intended to be installed.
 #[derive(Deserialize, Serialize, Clone, Debug, PartialEq)]
 pub struct DnaConfiguration {
     pub id: String,
     pub file: String,
+    #[serde(default)]
     pub hash: String,
```
lucksus:
```suggestion
    pub hash: Option<String>,
```

conductor_api/src/config.rs
```diff
@@ -314,11 +314,13 @@ impl From<AgentConfiguration> for AgentId {
...
 
 /// A DNA is represented by a DNA file.
-/// A hash has to be provided for sanity check.
+/// A hash can optionally be provided, which could be used to validate that the DNA being installed
+/// is the DNA that was intended to be installed.
 #[derive(Deserialize, Serialize, Clone, Debug, PartialEq)]
 pub struct DnaConfiguration {
     pub id: String,
     pub file: String,
+    #[serde(default)]
```
lucksus: Should this rather be an `Option<String>`, since that would be explicitly saying what we mean here..

Connoropolous:
I'm feeling like I'm agreeing with David's comments, but I will take that as a followup to this PR, which we could keep discussing, even though at the moment I'm going to merge, since I'm working towards a stable state for the mdbook

maackle:
I agree with @thedavidmeister's comments but would like to point out that now that UI hash *is* optional, there is no danger.

I'd also like to propose the idea that Conductors shouldn't even be aware of UIs at all. As David points out, we should expect any and all UIs to connect to running instances, we should not attempt to sandbox them into those specified in the Conductor config. The current ability of conductors to serve static files is just a convenience that should be eventually be decoupled and removed from the conductor API.

Until last week, I have been assuming that the Conductor would do everything, and users would just point their browsers directly to the UI interface ports opened by the conductor config, but I don't think that anymore after working on Holo. In general we should expect to see something sitting in between the browser and the conductor, some other server that talks to the conductor directly. Therefore we should make APIs that make it easy to build such go-betweens, rather than baking into the Conductor API, which is already getting bloated.

thedavidmeister:
> I have been assuming that the Conductor would do everything, and users would just point their browsers directly to the UI interface ports opened by the conductor config

exactly, this seems like it should be the holochain model and if holo wants an additional layer it would be a decoupled concern @maackle

PR #968, opened 2019-02-07
Signing 1 - Core can create signatures via Conductor callback

lucksus:
-> No, since this is prep work. Will add later when the whole signing story is implemented.

# Context
This is the first PR of several to follow within the *Real Crypto* frame:
screenshot 2019-02-08 at 14 54 52

# What
These are all the changes needed in core to get signatures into chain headers. But instead of calling `Keypair::sign` directly, core uses the conductor API to call `agent/sign` via the conductor <-> core JSON-RPC interface. This allows us to:
1. Easily replace the signing method with a remote call as planned as interim for Holo closed alpha
2. Implement a mock of this function with a static key so that we have deterministic signatures in unit tests that don't depend on the conductor or real key files.

The latter is part of this PR.


core/src/context.rs
```diff
@@ -32,6 +39,68 @@ use std::{
...
+    seed.write(0, mock_seed.as_slice())
+        .expect("SecBuf must be writeable");
+
+    // Create keypair from seed:
+    let mut keypair = Keypair::new_from_seed(&mut seed).unwrap();
+
+    // Convert payload string into a SecBuf
+    let payload = payload.as_bytes();
+    let mut message = SecBuf::with_insecure(payload.len());
+    message.write(0, payload).expect("SecBuf must be writeable");
```
neonphog:
```rust
message = SecBuf.with_insecure_from_string(payload);
```

core/src/context.rs
```diff
@@ -32,6 +39,68 @@ use std::{
...
+
+    // Create signature
+    let mut message_signed = SecBuf::with_insecure(64);
```
neonphog: is there a constant on the crate for signature size so we don't have to use magic number 64?

hc_dpki/src/keypair.rs
```diff
@@ -385,6 +385,26 @@ mod tests {
...
+    #[test]
+    fn it_should_invalidate_altered_message() {
+        let mut seed = SecBuf::with_insecure(SEEDSIZE);
+        random_secbuf(&mut seed);
+        let mut keypair = Keypair::new_from_seed(&mut seed).unwrap();
+
+        let mut message = SecBuf::with_insecure(16);
+        random_secbuf(&mut message);
+
+        let mut message_signed = SecBuf::with_insecure(64);
```
neonphog: magic 64 here

hc_dpki/src/keypair.rs
```diff
@@ -385,6 +385,26 @@ mod tests {
...
+
+        keypair.sign(&mut message, &mut message_signed).unwrap();
+
+        random_secbuf(&mut message);
+
+        let check: i32 =
+            Keypair::verify(keypair.pub_keys, &mut message_signed, &mut message).unwrap();
```
neonphog: why does this return an i32? wouldn't a bool be more appropriate?

core/src/context.rs
lucksus: 👍

core/src/context.rs
lucksus: Apparently there is not.

hc_dpki/src/keypair.rs
lucksus: Sorry, I copied the whole test above and just changed the message to assert that for a wrong message `check` is `!= 0`.

Yes, I agree, this should be a bool, but also this function is not coming through this PR. Just added a test that I felt missing..

lucksus:
64 cleaned up. Also added real timestamps in headers since Iso8601 get merged in between and I had to touch `ChainHeader` anyway..

core/src/context.rs
```diff
@@ -31,6 +38,66 @@ use std::{
...
+/// This is a local mock for the `agent/sign` conductor API function.
+/// It creates a syntactically equivalent signature using dpki::Keypair
+/// but with key generated from a static/deterministic mock seed.
+/// This enables unit testing of core code that creates signatures without
+/// depending on the conductor or actual key files.
+pub fn mock_signer(payload: String) -> String {
+    // Create deterministic seed:
+    let mut seed = SecBuf::with_insecure(SEEDSIZE);
+    let mut mock_seed: Vec<u8> = Vec::new();
+    for i in 1..SEEDSIZE {
```
sphinxc0re: Keep it simple:
```rust
let mock_seed: Vec<u8> = (1..SEEDSIZE).map(|e| e as u8).collect();
```

thedavidmeister:
i can repro the app spec tests issue locally by running `hc-install-node-conductor`

thedavidmeister:
actually, can repro by `cd nodejs_conductor/native` then `cargo build`

this is much faster as it avoids all the neon rebuilding

nodejs_conductor/native/Cargo.toml
```diff
@@ -8,7 +8,7 @@ exclude = ["artifacts.json", "index.node"]
 
 [lib]
 name = "holochain_node"
-crate-type = ["dylib"]
+# crate-type = ["dylib"]
```
thedavidmeister: dropping this line got `cargo build` to run (was previously throwing linking error)

let's see if tests pass...

sphinxc0re: We need that though. The node native module has to be compiled into a dynamic c library.

thedavidmeister: why? @sphinxc0re

PR #973, opened 2019-02-08
re-allow empty array of zome functions

Connoropolous:
Zippy changed this to mean only 1 or more zome functions was allowed, but it should allow zero

No because Zippy updated this since 0.0.3 and I'm just updating it BACK


Connoropolous:
Merging because I'm blocked by this and I had @lucksus verbal approval to make this change, plus how to make this change, plus Zippys implementation approval

PR #974, opened 2019-02-08
Signing 2 - hc keygen

lucksus:

`hc keygen` asks for passhprase and writes encrypted key bundle file to `~/.holochain/keys`.


cli/src/cli/keygen.rs
```diff
@@ -0,0 +1,50 @@
...
+            Some(PwHashConfig(
+                pwhash::OPSLIMIT_INTERACTIVE,
+                pwhash::MEMLIMIT_INTERACTIVE,
+                pwhash::ALG_ARGON2ID13,
+            )),
+        )
+        .unwrap();
+
+    let path = match directories::UserDirs::new() {
+        Some(user_dirs) => user_dirs.home_dir().join(".holochain").join("keys"),
```
Connoropolous: Based on what I added into the tree about directories, I propose that we create a new crate, or shareable file, that keeps track of and documents all the “magic strings” representing paths on devices. I find that they get scattered in the code and the act as high risk points of failure for our system I think. If we could start in that direction in this Pr that would be amazing

sphinxc0re: Also I'm in favor of creating these directories with the XDG_CONFIG convention in mind. The `directories` crate already does that for you, we just have to use it.

common/src/paths.rs
```diff
@@ -0,0 +1,23 @@
...
+
+/// Returns the path to the root config directory for all of Holochain.
+/// If we can get a user directory it will be an XDG compliant path
+/// like "/home/peter/.config/holochain".
+/// If it can't get a user directory it will default to "/etc/holochain".
+pub fn config_root() -> PathBuf {
+    directories::ProjectDirs::from(QUALIFIER, ORGANIZATION, APPLICATION)
+        .and_then(|dirs| Some(dirs.config_dir().to_owned()))
+        .or(Some(PathBuf::new().join("/etc").join(APPLICATION)))
+        .unwrap()
```
sphinxc0re: Better make this an `unwrap_or_else`:
```rust
unwrap_or_else(|| PathBuf::from("something"));
```
also the `/etc` path won't work on windows either

common/src/paths.rs
```diff
@@ -0,0 +1,23 @@
...
+pub const APPLICATION: &'static str = "holochain";
+pub const KEYS_DIRECTORY: &'static str = "keys";
+
+/// Returns the path to the root config directory for all of Holochain.
+/// If we can get a user directory it will be an XDG compliant path
+/// like "/home/peter/.config/holochain".
+/// If it can't get a user directory it will default to "/etc/holochain".
+pub fn config_root() -> PathBuf {
+    directories::ProjectDirs::from(QUALIFIER, ORGANIZATION, APPLICATION)
+        .and_then(|dirs| Some(dirs.config_dir().to_owned()))
```
sphinxc0re: This can be a `map`
```suggestion
        .map(|dirs| dirs.config_dir().to_owned())
```

common/src/paths.rs
sphinxc0re: I think we should bubble up the error instead of choosing a dummy path

hc_dpki/src/lib.rs
```diff
@@ -4,10 +4,11 @@ extern crate holochain_sodium;
 #[macro_use]
 extern crate arrayref;
 extern crate base64;
-extern crate rustc_serialize;
-
 extern crate bip39;
 extern crate boolinator;
+extern crate rustc_serialize;
```
sphinxc0re: Why is this needed again?

Connoropolous:
Also, this needs to be added to the readme of CLI, plus the mdbook.

hc_dpki/src/lib.rs
lucksus: I don't know. I've just moved this by two lines for alphanumeric sorting.

common/src/paths.rs
```diff
@@ -0,0 +1,22 @@
...
+pub const KEYS_DIRECTORY: &'static str = "keys";
+
+/// Returns the path to the root config directory for all of Holochain.
+/// If we can get a user directory it will be an XDG compliant path
+/// like "/home/peter/.config/holochain".
+/// If it can't get a user directory it will default to "/etc/holochain".
+pub fn config_root() -> PathBuf {
+    directories::ProjectDirs::from(QUALIFIER, ORGANIZATION, APPLICATION)
+        .map(|dirs| dirs.config_dir().to_owned())
+        .unwrap_or_else(|| PathBuf::new().join("/etc").join(APPLICATION))
```
sphinxc0re:
```suggestion
        .unwrap_or_else(|| PathBuf::from("/etc").join(APPLICATION))
```
Also I still think, this is a bad idea

lucksus: What is a bad idea?

sphinxc0re: Btw, proof, that these are the same:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=978badfa745d4fa9b744dbf8229b28e9

Connoropolous:
Ok, this is great I think, AND it takes us for now to a point where there is a `~/.hc` `~/.holochain` and now additionally `~/.config/holochain` ... I guess this is with the intention to press forward on making this consistent, and the XDG compliant approach would be the one we'd consolidate to?

cli/src/main.rs
```diff
@@ -142,6 +146,12 @@ enum Cli {
         #[structopt(long = "skip-package", short = "s", help = "Skip packaging DNA")]
         skip_build: bool,
     },
+    #[structopt(
+        name = "keygen",
+        alias = "k",
+        about = "Creates a new agent key pair, asks for a passphrase and writes an encrypted key bundle to ~/.holochain/keys"
```
```suggestion
        about = "Creates a new agent key pair, asks for a passphrase and writes an encrypted key bundle to ~/.config/holochain/keys"
```

lucksus:
@sphinxc0re, ok to merge?

Connoropolous:
@lucksus can you comment on https://github.com/holochain/holochain-rust/pull/974#issuecomment-462373731?

lucksus:
@Connoropolous, yes, this only makes sense if we now switch all config path handling to `holochain_common::paths`. But please let us do that in a follow-up PR.

PR #975, opened 2019-02-08
fix nodejs_conductor config bug: couldn't set to not-debug logging

Connoropolous:

since `Default::default` for LoggerConfiguration picks "debug", and the other picks "debug", there's no way to pick simple

This then allows muting the log output of the tests, which is a firehose


Connoropolous:
I was also looking at the TODO in the code of changing this to an enum yesterday. Not done, but it's really worth it soon I think

PR #976, opened 2019-02-08
holochain_net: end-user network config

ddd-mtl:
(Follow-up of https://github.com/holochain/holochain-rust/pull/944)

### New config setting
**Adding `pub n3h_end_user_config_filepath: Option<String>` to Conductor Configuration.**
So the user can configure some network settings used by n3h.

**cli** loads a new env variable: `HC_N3H_SETTINGS_FILE`

This new setting will be passed to the P2pConfig.
The P2pConfig constructor will try to load this json file. On failure. it will use a default config instead.
P2pConfig will pass around the loaded end-user config to the IpcNetWorker which sends it to n3h via stdin.

#### P2pConfig
Added `maybe_end_user_config` to struct.
Its an option since its not required for all types of backend kinds.

### Refactors
Now passing around a P2pConfig struct instead of its JsonString form.
Added a new constructor for P2pConfig.


ddd-mtl:
@Connoropolous I changed the env var name to `HC_N3H_SETTINGS_FILE`

The rest seems out of scope for this PR, but why not.

Connoropolous:
when you say 'why not?' you mean to say you will add, or no?

ddd-mtl:
@Connoropolous Sorry for the confusion. I meant I'm okay with your proposal but I don't want to add it in this PR. I don't have much knowledge on those folders so I prefer not touch things I'm not confortable with.

Connoropolous:
Ok, I see. I'm thinking I'll be doing some coding on that stuff very soon

Connoropolous:
@ddd-mtl can you offer up an example of what this file should look like?

cli/src/cli/run.rs
```diff
@@ -100,13 +100,17 @@ pub fn run(
             n3h_bootstrap.push(n3h_bootstrap_node.unwrap())
         }
 
+        // Load end_user config file
+        let n3h_end_user_config_filepath = env::var("HC_N3H_SETTINGS_FILE").ok();
```
neonphog: We started down this path of naming things n3h, but I'd actually like to undo that. The things in this config are going to need to exist even when we write the embedded rust version of the networking module. So, maybe `"NETWORKING_CONFIG_FILE"`?

conductor_api/src/config.rs
```diff
@@ -625,6 +627,7 @@ pub mod tests {
     bootstrap_nodes = ["/ip4/127.0.0.1/tcp/45737/ipfs/QmYaEMe288imZVHnHeNby75m9V6mwjqu6W71cEuziEBC5i"]
     n3h_path = "/Users/cnorris/.holochain/n3h"
     n3h_persistence_path = "/Users/cnorris/.holochain/n3h_persistence"
+    n3h_end_user_config_filepath = "/Users/cnorris/.holochain/n3h_end_user_config.json"
```
neonphog: similarly `"network_config.json"`?

net/src/p2p_config.rs
```diff
@@ -142,7 +170,45 @@ impl P2pConfig {
...
+            },
+            "wssAdvertise": "auto",
+            "wssRelayPeers": null
+          }
+        })
+    }
+
+    pub fn load_end_user_config(
+        maybe_end_user_config_filepath: Option<String>,
+    ) -> serde_json::Value {
```
neonphog: Should this just return a `String` so we don't have to do the json parsing / have the `json!` macro in the default, etc?

test_bin/data/end_user_net_config.json
```diff
@@ -0,0 +1,12 @@
...
+    "connection": {
+      "rsaBits": 1024,
+      "bind": [
+        "wss://0.0.0.0:8443"
+      ]
+    },
+    "wssAdvertise": "auto",
+    "wssRelayPeers": null
+  }
+}
```
neonphog: @Connoropolous ^^

net/src/p2p_config.rs
ddd-mtl: I believe `serde_json::Value` is better as we want to make sure as early as possible that the config is valid JSON.

conductor_api/src/config.rs
ddd-mtl: fixed

cli/src/cli/run.rs
ddd-mtl: changing just this one for now.

Connoropolous:
Thanks for the example. Where can info be found as to what those properties
are/do?

neonphog:
@Connoropolous - those configs are not part of n3h master yet, they will be included in https://github.com/holochain/n3h/pull/35 - I don't think there is any documentation for the parameters anywhere yet.

Connoropolous:
We'll need some docs for them, on this side or that

PR #982, opened 2019-02-10
Signals (part 1)

maackle:
This adds a notion of `Broadcaster`, which is something that knows how to send a `Signal` back to the client. Each Interface implementation's `run` method returns a `Broadcaster`. Then, if the Conductor is set up for it, it will start a new thread which continually consumes the signal channel and sends each signal over every interface via its Broadcaster.

In order to maintain backwards compatibility with nodejs_conductor, the conductor can be configured in one of three states, with regard to signals:
1. Signal Receiver is owned externally -- this is what nodejs_conductor does, managing its own signal stream
2. Signal Receiver is owned internally by the Conductor API -- this is the new behavior which allows signals to be broadcast over interfaces automatically
3. Uninitialized

This PR also includes a new module `core_types::ugly`, which contains a type `Initable`, representing a value that may or may not be initialized. It is isomorphic to `Option`. With this type, we can better keep track of which values are truly Optional, and which ones we just used Option for to represent invalid or uninitialized state.

## Question

Seems that signals need to be serializable (into JsonString), and currently Action is a big ol' non-serializable struct which gets passed as a signal. Should we bite the bullet and actually implement Serialize for Action and allll its sub-structs, or somehow punt on that since we don't need that for Holo?

If we do want `Action` to be serializable eventually, let's do it now. Otherwise, we can punt, but that would also be a bit of work, probably meaning we have to manually implement `Serialize` for Signal and just return some dummy value for the case of `Signal::Internal`.

- [x] ~I have added a summary of my changes to the changelog~ no new user-facing functionality, only groundwork

## Nico's additions
fixes https://github.com/holochain/holochain-rust/issues/739


core/src/signal.rs
```diff
@@ -1,14 +1,47 @@
...
 
-#[derive(Debug)]
+#[derive(Clone, Debug, DefaultJson)]
 pub enum Signal {
     Internal(ActionWrapper),
     User(JsonString),
+    Holo(JsonString),
+}
+
+impl Serialize for Signal {
```
sphinxc0re: Do we have to implement this by hand? If so, why do we?

core_types/src/ugly.rs
```diff
@@ -0,0 +1,35 @@
...
+/// Represents a value which may not have been initialized.
+/// Isomorphic to `Option`, but semantically different.
+#[derive(Clone)]
+pub enum Initable<T> {
+    Init(T),
+    Uninit,
+}
+
+impl<T> Initable<T> {
+    pub fn to_option(val: Initable<T>) -> Option<T> {
```
sphinxc0re: You can just "mount" `self` as in:
```suggestion
    pub fn into_option(self) -> Option<T> {
```

core/src/signal.rs
maackle: @sphinxc0re because we can't immediately derive `Serialize` for `Action`, which `Signal::Internal` contains. This is a temporary way to move forward without implementing Serialize for the very big `Action` enum.

core/src/instance.rs
```diff
@@ -197,7 +197,7 @@ impl Instance {
             *state = new_state;
         }
 
-        // context.log(format!("trace/reduce: {:?}", action_wrapper.action()));
+        context.log(format!("trace/reduce: {:?}", action_wrapper.action()));
```
Connoropolous: Are you playing tag on this line of code? Didn't we uncomment it once now it's recommented and now you're uncommenting it again?

conductor_api/src/conductor/broadcaster.rs
```diff
@@ -0,0 +1,36 @@
...
+    fn drop(&mut self) {
+        match self {
+            Broadcaster::Ws(sender) => sender.close(ws::CloseCode::Normal).unwrap_or(()),
+            Broadcaster::Noop => (),
+        }
+    }
+}
+
+impl Broadcaster {
+    pub fn send<J>(&self, msg: J) -> Result<(), HolochainError>
```
Connoropolous: code comment on this fn?

core/src/instance.rs
```diff
@@ -212,12 +212,15 @@ impl Instance {
...
     /// emitted as a `Signal::Internal`, and if so, send it
     fn maybe_emit_action_signal(&self, context: &Arc<Context>, action: ActionWrapper) {
-        if let Some(ref tx) = context.signal_tx {
+        if let Some(tx) = context.signal_tx() {
             // @TODO: if needed for performance, could add a filter predicate here
             // to prevent emitting too many unneeded signals
             let signal = Signal::Internal(action);
-            tx.send(signal).unwrap_or(())
-            // @TODO: once logging is implemented, kick out a warning for SendErrors
+            tx.send(signal).unwrap_or_else(|_e| {
```
Connoropolous: 👍

conductor_api/src/conductor/base.rs
```diff
@@ -514,23 +540,40 @@ impl Conductor {
...
-                let iface = make_interface(&interface_config);
-                iface.run(dispatcher, rx).map_err(|error| {
-                    let message = format!(
-                        "err/conductor: Error running interface '{}': {}",
-                        interface_config.id, error
-                    );
-                    let _ = log_sender.send((String::from("conductor"), message));
-                    error
-                })
+        let (kill_switch_tx, kill_switch_rx) = channel();
```
Connoropolous: can you say something (and possibly in a code comment documented) about the kill_switch? where/when does it get called?

conductor_api/src/interface_impls/http.rs
```diff
@@ -13,12 +15,21 @@ impl HttpInterface {
...
+    ) -> Result<(Broadcaster, thread::JoinHandle<()>), String> {
         let url = format!("0.0.0.0:{}", self.port);
-        let _server = ServerBuilder::new(handler)
+
+        let server = ServerBuilder::new(handler)
             .start_http(&url.parse().expect("Invalid URL!"))
             .map_err(|e| e.to_string())?;
-        let _ = kill_switch.recv();
-        Ok(())
+        let broadcaster = Broadcaster::Noop;
```
Connoropolous: So HTTP implements a broadcaster but it's really just a no-op. So you've also set up a pattern that's extendable into different interfaces being implemented...

conductor_api/src/conductor/base.rs
```diff
@@ -113,28 +119,52 @@ impl Conductor {
...
         }
-        self.signal_tx = Some(signal_tx);
-        self
     }
 
     pub fn config(&self) -> Configuration {
         self.config.clone()
     }
 
+    pub fn start_signal_broadcast(&mut self, signal_rx: SignalReceiver) -> thread::JoinHandle<()> {
```
zippy: this function I think needs some rusdoc comments because it's a `pub` function and feels like an important place where people will look.

conductor_api/src/conductor/base.rs
```diff
@@ -306,13 +337,15 @@ impl Conductor {
     /// The first time we call this, we also initialize the conductor-wide config
     /// for use with all instances
     ///
+    /// Note that the `signal_tx` parameter represents an important bifurcation of signal handling functionality.
+    /// if None, then the signal channel will be instantiated and the receive will be owned by the `Conductor`,
+    /// allowing it to automatically handle signals and push them out across the Interfaces via Broadcasters.
+    /// if it is Some, then the signal receiver is externally owned, and signals will not be sent over Interfaces.
```
zippy: It seems important to document the Why not just the What here.  Suggestion: 
```suggestion
    /// if it is Some, then the signal receiver is externally owned, and signals will not be sent over Interfaces.  This is for the use-case of the nodejs conductor that we use for testing, which expects not to have those interfaces
```

conductor_api/src/interface_impls/http.rs
zippy: This sounds right.  @maackle it seems like this same method could be used to implement a `LongPollHttpInterface` that could do broadcasting over http, it just waits to "send" the messages when it gets a poll http request.

conductor_api/Cargo.toml
```diff
@@ -23,8 +23,9 @@ serde_json = { version = "1.0", features = ["preserve_order"] }
 toml = "0.4.8"
 boolinator = "2.4"
 tiny_http = "0.6.0"
-jsonrpc-ws-server = { git = "https://github.com/paritytech/jsonrpc" }
-jsonrpc-http-server = { git = "https://github.com/paritytech/jsonrpc" }
+jsonrpc-core = { git = "https://github.com/holochain/jsonrpc", branch = "broadcaster-getter" }
```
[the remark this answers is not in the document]
maackle: https://github.com/holochain/jsonrpc/pull/1/files

conductor_api/src/conductor/base.rs
maackle: Nothing new here, just gave it a more descriptive name. I'll find a place to add a comment.

conductor_api/src/conductor/base.rs
```diff
@@ -222,6 +252,7 @@ impl Conductor {
     }
 
     /// Stop and clear all instances
+    /// @QUESTION: why don't we care about errors on shutdown?
```
maackle: @lucksus I put this in randomly to remind me to ask: I remember this got changed in a PR, but it feels wrong, why are we ignoring the possibility of shutdown errors?

conductor_api/src/interface_impls/http.rs
maackle: Exactly. Each new Interface has the opportunity to also specify a Broadcaster, or use Noop if it doesn't support two-way messaging.

core/src/instance.rs
maackle: I added an "exclude: `^trace/`" rule to LogRules::default, which will allow this line to stay in, what do you think @Connoropolous @zippy ?

conductor_api/src/conductor/broadcaster.rs
maackle: @Connoropolous it seems a bit redundant since this is the only method on this impl and seems clear that this is the function that "makes it go"

conductor_api/src/logger.rs
```diff
@@ -167,13 +170,6 @@ pub mod tests {
         assert_eq!(m.msg, "xboy");
     }
 
-    #[test]
-    fn test_log_rules_default() {
-        let rules = LogRules::default();
-        assert_eq!(rules.rules.len(), 3);
-        assert_eq!(format!("{:?}",rules),"LogRules { rules: [LogRule { pattern: ^err/, exclude: false, color: Some(\"red\") }, LogRule { pattern: ^debug/dna, exclude: false, color: Some(\"white\") }, LogRule { pattern: .*, exclude: false, color: None }] }".to_string());
-    }
-
```
maackle: @zippy how do you feel about this test being removed? I feel this is too brittle and is not a useful test. It just caused my previous commit's build to break because I added another default rule.

conductor_api/Cargo.toml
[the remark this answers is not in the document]
maackle: We'll have to remember to keep this fork up to date with jsonrpc upstream

conductor_api/src/conductor/base.rs
```diff
@@ -172,6 +179,52 @@ impl Conductor {
         self.config.clone()
     }
 
+    /// Starts a new thread which monitors each instance's signal channel and pushes signals out
+    /// all interfaces the according instance is part of.
+    /// TODO: add kill-switch to this thread to close it on shutdown / also make sure to recreate if interfaces get changed..
```
maackle: @lucksus maybe this part is important? You've been looking at the code more recently, so you might have a better sense. Do we need to do anything extra to handle the case when interfaces get restarted?

maackle:
@lucksus I was the original author so I can't approve, but most of it looks good. I did notice the TODO in there that says we need to address the case when interfaces get restarted, what do you think?

conductor_api/src/logger.rs
zippy: seems ok.

conductor_api/src/conductor/base.rs
lucksus: I've added the kill-switch which gets triggered on shutdown. Also added shutdown to the Drop implementation.

lucksus:
@Connoropolous, the usage is actually not included. Nor is a test. My intention here was to get this old branch merged before we accumulate more conflicts again. The client-code of this would live UI side, i.e. in hc-web-client or hQuery, since it is about the actual sending of signals via websocket interface from conductor to UI. The whole point of this first PR was to solve the pragmatic problem of pushing something down the other direction which is something the JSON_RPC crate does not support natively. That's why @maackle had to fork it. I would prefer keeping this PR (as part 1) scoped to that.

@thedavidmeister's anticipated switch from the nodejs_conductor to a JS client using the Rust conductor depends on this as well. We will have to re-implement the waiter (well basically everything that holochain_nodejs does right now), and in order to wait for a zome call's consequences to be completed, the JS testing client will have to wait for signals of all agents in a scenario and those signals will be passed back to the client with this code. Once we have that, we will switch app-spec over and this will be a basic component of our integration test suite.

core/src/signal.rs
```diff
@@ -1,20 +1,54 @@
...
+                "Signal",
+                0,
+                "Internal",
+                "(Internal signal serialization not yet implemented)",
+            ),
+            Signal::User(msg) => {
+                serializer.serialize_newtype_variant("Signal", 1, "User", &msg.to_string())
+            }
+            Signal::Holo(msg) => {
+                serializer.serialize_newtype_variant("Signal", 2, "Holo", &msg.to_string())
```
Connoropolous: What is this Holo Signal?

Connoropolous:
I don’t want to just rush the branch in cause it’s old. It needs the same quality as any other PR

lucksus:
We currently have almost no tests for the conductor zome interfaces. The reason is that we can only test those as integration test that spawns a conductor instance and then runs real network connections to that process. The plan is to do exactly that with the whole app-spec suite. But in order to do that we need to replicate the waiter functionality. This PR is a dependency for that and a first step in a set of several PRs to come. I would not like to do all of that in just one PR because it will be way to big.

Connoropolous:
If we can't unit test that, isn't that a sign that it's not written in the best way? Or could we unit test it and we just aren't?

conductor_api/src/conductor/base.rs
```diff
@@ -172,6 +182,56 @@ impl Conductor {
...
+                                        interface_config
+                                            .instances
+                                            .iter()
+                                            .find(|instance| instance.id == *instance_id)
+                                            .is_some()
+                                    })
+                                    .collect();
+                                for interface in interfaces_with_instance {
+                                    broadcasters.get(&interface.id).map(|broadcaster| {
+                                        broadcaster.send(signal.clone()).expect("TODO: result");
```
Connoropolous: I don't get this "TODO: result"?

should this be a response from the client?

lucksus:
@Connoropolous, the unit under test would be the code that facilitates sending data "backwards" through the websocket. We could test that, as I said, by spawning a conductor and talking to it via a websocket. I actually added some more commits in preparation for writing such a test but I'm a bit stuck finding the right websocket client crate that we could use in such a test. IMHO it would make more sense to test that from JS, which is exactly what we want to do and what this is a preparation step for.

conductor_api/src/interface.rs
```diff
@@ -968,7 +971,11 @@ impl ConductorApiBuilder {
 }
 
 pub trait Interface {
```
Connoropolous: @lucksus or @maackle can we get a description of this trait here, since it could be important in the future, for implementing other interfaces?

It can be pulled roughly from the description of the PR I think

conductor_api/src/signal_wrapper.rs
```diff
@@ -0,0 +1,8 @@
+use holochain_core::signal::Signal;
+use holochain_core_types::{error::HolochainError, json::JsonString};
+
+#[derive(Serialize, Deserialize, Debug, DefaultJson)]
```
Connoropolous: code comment

conductor_api/src/conductor/base.rs
```diff
@@ -1373,9 +1464,64 @@ pub mod tests {
...
         assert_eq!(
-            conductor.boot_from_config(None),
+            conductor.boot_from_config(),
             Err("Error while trying to create instance \"test-instance-1\": Key from file \'holo_tester1.key\' (\'HcSCI7T6wQ5t4nffbjtUk98Dy9fa79Ds6Uzg8nZt8Fyko46ikQvNwfoCfnpuy7z\') does not match public address HoloTester1-----------------------------------------------------------------------AAACZp4xHB mentioned in config!"
                 .to_string()),
         );
     }
+
+    #[test]
+    fn test_signals_through_admin_websocket() {
```
lucksus: @Connoropolous, I've added this test that shows signals received via websocket.

core/src/signal.rs
maackle: Good catch, this can be removed. It was from back when I was assuming that the wormhole hack would be implemented via Signals

Connoropolous:
Oh good!

conductor_api/src/interface.rs
Connoropolous:
```suggestion
/// A Broadcaster is something that knows how to send a Signal back to a client.
/// Each Interface implementation's `run` method must return a Broadcaster, even if it's just the No-op.
/// Then, if the Conductor is set up for it, it will start a new thread which continually consumes the signal channel and sends each signal over every interface via its Broadcaster.
pub trait Interface {
```

core/src/instance.rs
zippy: Yes, approved.

conductor_api/src/interface_impls/http.rs
pdaoust: You could use it to do HTTP/2 server-side events too...

PR #985, opened 2019-02-11
Added CORS header

sphinxc0re:

### DIFF
--- a/conductor_api/src/static_file_server.rs
+++ b/conductor_api/src/static_file_server.rs
@@ -57,9 +57,10 @@ impl Future for MainFuture {
 
         response_result.map(|res| {
             res.map(|mut response| {
-                response
-                    .headers_mut()
-                    .insert("Access-Control-Allow-Origin", "*".parse().unwrap());
+                response.headers_mut().insert(
+                    "Access-Control-Allow-Origin",
+                    "http://localhost:*".parse().unwrap(),
+                );
 
                 response
             })
--- a/conductor_api/src/static_file_server.rs
+++ b/conductor_api/src/static_file_server.rs
@@ -27,11 +27,12 @@ fn redirect_request_to_root<T>(req: &mut Request<T>) {
     *req.uri_mut() = uri::Uri::from_parts(original_parts).unwrap();
 }
 
-fn dna_connections_response(config: &Option<InterfaceConfiguration>) -> Response<Body> {
-    let interface = match config {
-        Some(config) => json!(config),
-        None => serde_json::Value::Null,
-    };
+fn dna_connections_response(config: Option<&InterfaceConfiguration>) -> Response<Body> {
+    let interface = config
+        .cloned()
+        .map(|config| json!(config))
+        .unwrap_or(serde_json::Value::Null);
+
     Builder::new()
         .body(json!({ "dna_interface": interface }).to_string().into())
         .expect("unable to build response")
@@ -47,10 +48,22 @@ impl Future for MainFuture {
     type Error = Error;
 
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
-        match *self {
-            MainFuture::Config(ref config) => Ok(Async::Ready(dna_connections_response(config))),
+        let response_result = match self {
+            MainFuture::Config(config) => {
+                Ok(Async::Ready(dna_connections_response(config.as_ref())))
+            }
             MainFuture::Static(ref mut future) => future.poll(),
-        }
+        };
+
+        response_result.map(|res| {
+            res.map(|mut response| {
+                response
+                    .headers_mut()
+                    .insert("Access-Control-Allow-Origin", "*".parse().unwrap());
+
+                response
+            })
+        })
     }
 }
 


conductor_api/src/static_file_server.rs
```diff
@@ -47,10 +48,22 @@ impl Future for MainFuture {
...
+            }
             MainFuture::Static(ref mut future) => future.poll(),
-        }
+        };
+
+        response_result.map(|res| {
+            res.map(|mut response| {
+                response
+                    .headers_mut()
+                    .insert("Access-Control-Allow-Origin", "*".parse().unwrap());
```
sphinxc0re: This rule is still up to debate. I don't really know what the sweet spot here is.

sphinxc0re:
@Connoropolous I didn't think so. I thought the only problem was the statically served files. Where else do we use HTTPRequests?

Connoropolous:
There is HTTP based json-rpc interface

Connoropolous:
@sphinxc0re https://github.com/holochain/holochain-rust/blob/develop/conductor_api/src/interface_impls/http.rs

sphinxc0re:
@Connoropolous The small doesn't seem to be clear on where the header should be set. Also why are there two http interfaces that are completely detached from one another?

Connoropolous:
One serves the UI bundles, the other is a JSON-RPC interface to call Zome functions

Connoropolous:
https://developer.holochain.org/guide/latest/conductor_ui_interfaces.html
and
https://developer.holochain.org/guide/latest/json_rpc_http.html

sphinxc0re:
That still doesn't answer the question whether I need to add this in order to solve the problem.

sphinxc0re:
... or maybe the problem is already solved by this PR?

Connoropolous:
Actually, this makes me realize that I don't believe this is quite what's needed. This PR, I believe, doesn't solve the CORS problem, because say the UI interface is running over `localhost:3000`, then JS files in the browser served at localhost:3000 can make calls to that server without CORS. 

Where a CORS header is required is when calling HTTP from `localhost:3000` to say `localhost:8000` which might be where the DNA JSON-RPC interface is running

Connoropolous:
I believe rather you need this: https://docs.rs/jsonrpc-http-server/10.0.1/jsonrpc_http_server/cors/index.html

for this:
https://github.com/holochain/holochain-rust/blob/develop/conductor_api/src/interface_impls/websocket.rs
and
https://github.com/holochain/holochain-rust/blob/develop/conductor_api/src/interface_impls/http.rs

willemolding:
As I understand it the CORS header in the static file allows the javascript embedded in it to make calls to another origin than the one it was served on. Therefore I think it makes sense the way it is done now.

Should we pull in someone else to sanity check?

willemolding:
My change suggestion related to the allow list being too broad (*) and we could tighten it to the ports with the dna interfaces this UI has been explicitly allowed to use

Connoropolous:
And I agree with cross referencing the allowed ports, as Willem mentions
"My change suggestion related to the allow list being too broad (*) and we could tighten it to the ports with the dna interfaces this UI has been explicitly allowed to use"

Connoropolous:
Ok I did some digging. 

A UI served over port 8888 CAN, by default, make WS and HTTP calls (with JS `fetch` function at least) to a DNA API interface being served over port 3000. 

This made me very curious, why is this? Because the default for the servers is to not block origins, which generally makes sense. It's something you want to configure on a case by case basis, but a different default wouldn't make sense. 
See: 
https://github.com/paritytech/jsonrpc/blob/master/http/src/lib.rs#L224
https://github.com/paritytech/jsonrpc/blob/master/http/src/lib.rs#L237

The question then becomes do we want to make it MORE secure by constraining allowed origins?
Either way, the solution implemented in this PR isn't a solution that helps anything in particular for our situation at the moment, from what I can tell.

Connoropolous:
From localhost:8888
screen shot 2019-02-12 at 10 58 28 am

via websockets
screen shot 2019-02-12 at 10 59 50 am

and the conductor config:
```toml
[network]
n3h_path = "/Users/connor/code/holochain/n3h"
n3h_persistence_path = "/tmp"
bootstrap_nodes = []

[[agents]]
id = "test_agent1"
name = "HoloTester1"
public_address = "sandwich--------------------------------------------------------------------------AAAEqzh28L"
key_file = "holo_tester.key"

[[agents]]
id = "test_agent2"
name = "HoloTester2"
public_address = "bagel-----------------------------------------------------------------------------AAADAq0mCu"
key_file = "holo_tester.key"

[[agents]]
id = "test_agent3"
name = "HoloTester3"
public_address = "croissant-------------------------------------------------------------------------AAAEuHzxV4"
key_file = "holo_tester.key"

[[dnas]]
id = "simple_app_dna"
file = "bundle.json"


[[instances]]
id = "test-instance1"
dna = "simple_app_dna"
agent = "test_agent2"
[instances.storage]
type = "file"
path = "./tmp-storage"


[[interfaces]]
admin = true
id = "websocket_interface"
[interfaces.driver]
type = "websocket"
port = 3000
[[interfaces.instances]]
id = "test-instance1"

[[ui_bundles]]
id = "bundle1"
root_dir = "./ui"

[[ui_interfaces]]
id = "ui-interface-1"
bundle = "bundle1"
port = 8888
dna_interface = "websocket_interface"

[logger]
type = "debug"
  [[logger.rules.rules]]
    pattern = "^err/"
    color = "red"
  [[logger.rules.rules]]
    pattern = "^debug/dna"
    exclude = true
    color = "white"
  [[logger.rules.rules]]
    pattern = "^debug/reduce"
    color = "magenta"
  [[logger.rules.rules]]
    pattern = "^debug/net/handle"
    exclude = true
  [[logger.rules.rules]]
    pattern = ".*"
```

peeech:
From Holo perspective:
- [here](https://hackmd.io/YHpcl0KgRuO4UzxVOxLZkg#Serve-Dynamic-Headers-for-static-files-Holo) is the full specification of the headers on static files received from the Host
- in short we need "access-control-allow-origin:*" (or tighter - see specs) and "Cache-control: 31536000"
- if static files are passed via Holo-Director then we can rewrite headers there. If not, they need to be set by Conductor

willemolding:
Ok I got the low down from PJ. We actually don't need to implement this anymore at all. There is zero need for it for the non-Holo use case and Holo is now going to manage headers themselves in the Holo Director.

PR #986, opened 2019-02-11
Remove dead code related to auto-generated instance IDs

maackle:

### DIFF
--- a/nodejs_conductor/index.js
+++ b/nodejs_conductor/index.js
@@ -23,18 +23,8 @@ const defaultOpts = {
 
 const Config = {
     agent: name => ({ name }),
-    dna: (path, name) => {
-        if (!name) {
-            name = path
-        }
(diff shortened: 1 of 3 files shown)


nodejs_conductor/index.js
```diff
@@ -137,7 +126,7 @@ Conductor.withInstances = function (instances, opts=defaultOpts) {
 Conductor.run = function (instances, opts, fn) {
     if (typeof opts === 'function') {
         fn = opts
-        opts = {}
+        opts = undefined
```
Connoropolous: this one because if you pass in an empty object, it won't pick the default args, whereas it will if you pass `undefined`

Connoropolous:
@maackle instead of making code tweaks in my docs branch, I just added them here, so they get in faster and cleaner anyways

PR #989, opened 2019-02-12
Add empty API function stubs to HDK to enable zome unit testing 

lucksus:

Instead of [adding these definitions in the zome](https://github.com/Holo-Host/holofuel/pull/15), we can also add them in the HDK directly.
Pro: zome developers don't need to get this right
Con: if the zome developer wants to provide a more sophisticated mock of the API to be used in Rust native tests, it will collide with these functions.

Thoughts?


lucksus:
Done, @zippy.

thedavidmeister:
imma try to fix up these test fails

thedavidmeister:
is there a way to "snooze" a github issue?

i'd like to pick this up again once we flush some current bugs out

lucksus:
I've added the empty function definitions back to hdk-rust/integration_test.rs which now compiles and links on windows, but now we see the same problem with the examles in api.rs. So obviously the windows toolchain gets rid of those empty functions before linking. We could now add those redundant definitions back in and merge just the added ones under `#[cfg(test)]` in api.rs, which is the main thing that enables zome devs to run rust tests.. I would suggest. Or does anybody feel inclined to find out why the windows rust toolchain is doing that? @thedavidmeister, @Connoropolous, @AshantiMutinta, @sphinxc0re?

thedavidmeister:
i would like to know what is going on if possible...

lucksus:
@thedavidmeister, can we merge this?
I also would like to understand in what way the Windows Rust toolchain differs, but the main scope of this PR is enabling zome devs to run `cargo test` on their zome code, which is possible with these changes on non Windows OSes. I was just jumping ahead in trying to remove all redundant definitions. If you can't make it work on windows without those now, let's just move this PR as it is now. You can still find out what's going on later.. ;)

thedavidmeister:
@lucksus i approved

it's marginal benefit to not support windows as the tests will just fail on windows without the function definitions anyway....

but still it's a step in the right direction so go ahead and merge

PR #990, opened 2019-02-12
consolidate env vars into one file

Connoropolous:

Relates to this in the SoA tree
https://realtimeboard.com/app/board/o9J_kyiXmFs=/?moveToWidget=3074457346447871323

Creates a nice docs page like this
screen shot 2019-02-12 at 2 53 50 pm


common/src/env_vars.rs
```diff
@@ -0,0 +1,31 @@
...
+            EnvVar::N3hMode => &OsStr::new("HC_N3H_MODE"),
+            EnvVar::N3hWorkDir => &OsStr::new("HC_N3H_WORK_DIR"),
+            EnvVar::N3hBootstrapNode => &OsStr::new("HC_N3H_BOOTSTRAP_NODE"),
+            EnvVar::SimpleLoggerMute => &OsStr::new("HC_SIMPLE_LOGGER_MUTE"),
+            EnvVar::TargetPrefix => &OsStr::new("HC_TARGET_PREFIX"),
+        }
+    }
+}
+
+pub fn get_env_var_value(env_var: EnvVar) -> Result<String, VarError> {
```
sphinxc0re: It would be way easier to use this if you defined this on `EnvVar`:
```rust
impl EnvVar {
    pub fn value(&self) -> Result<String, VarError> {
        std::env::var(self)
    }
}
```

common/src/env_vars.rs
```diff
@@ -0,0 +1,31 @@
...
+    Interface,
+    N3hPath,
+    N3hMode,
+    N3hWorkDir,
+    N3hBootstrapNode,
+    SimpleLoggerMute,
+    TargetPrefix,
+}
+
+impl AsRef<OsStr> for EnvVar {
```
sphinxc0re: You don't have to actually implement `AsRef` for the `var` function to work. You can just map to `&'static str` values. That would still work.

common/src/env_vars.rs
Connoropolous: I like it, will do

common/src/env_vars.rs
Connoropolous: I was trying... but I couldn't figure it out, can you show me?

sphinxc0re: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&code=enum%20Var%20%7B%0A%20%20%20%20A%2C%0A%20%20%20%20B%2C%0A%20%20%20%20C%0A%7D%0A%0Aimpl%20Var%20%7B%0A%20%20%20%20pub%20fn%20as_str(%26self)%20-%3E%20%26str%20%7B%0A%20%20%20%20%20%20%20%20match%20self%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20Var%3A%3AA%20%3D%3E%20%22A%22%2C%0A%20%20%20%20%20%20%20%20%20%20%20%20Var%3A%3AB%20%3D%3E%20%22B%22%2C%0A%20%20%20%20%20%20%20%20%20%20%20%20Var%3A%3AC%20%3D%3E%20%22C%22%2C%0A%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%7D%0A%7D%0A%0Afn%20main()%20%7B%0A%20%20%20%20println!(%22%7B%3A%3F%7D%22%2C%20std%3A%3Aenv%3A%3Avar(Var%3A%3AA.as_str()))%3B%0A%7D

Connoropolous:
@sphinxc0re made your change requests

Connoropolous:
@ddd-mtl I know there was a change to the name of the env var you used in your PR, but I noticed while doing this that now it is the only one without an `HC` prefix. Can we re-add that in this PR?

Connoropolous:
@lucksus @zippy I determined failing tests were because the `directories` crate couldn't be included in any wasm builds as a subdirectory because it doesn't build to WASM. It was coming in through holochain_common because it handles paths to. Tricky stuff

PR #991, opened 2019-02-12
Refactor Nucleus and Zome Calls to pure reducers

lucksus:

# Purify call reducers
The main thing here is to replace `ExecuteZomeFunction` (the action and the reducer) by an action creator that returns a future. Just as we did everywhere else except here. This was overdue and happened now.

## Clean `hdk::call`
The side effect is that we can use this action creator, the future return value and our home-grown `block_on` executor in `hdk::call`, which hopefully gets rid of the flaky race condition. (and if not, we now have a much better overview of what is really happening)

## Steps roughly in commit order

- [x] Break down `nucleus/mod.rs` that contained all reducer functions into separate files under `nucleus/reducer`
- [x] Add action creator `call_zome_function` that spins of a WASM instance without any State change prior and that only dispatches `Action::ReturnZomeFunctionResult` after the fact
- [x] Remove `ExecuteZomeFunction` (the action and the reducer)
- [x] Refactor `hdk::call` to also use above action creator
- [x] Remove `Call` (the action and the reducer)

## SignalZomeFunctionCall
After discussing this in our heartbeat with @maackle, I've readded an action that gets dispatched when a zome function gets called. Since there is no reducer for this action (i.e. no state change) I've named it `Signal..`. It is used by the nodejs test waiter as a signal to attach waiters..


zippy:
Looking good!  I just renabled the call_test_assigned test and it's passing for me

core/src/nucleus/actions/call_zome_function.rs
```diff
@@ -0,0 +1,144 @@
...
+        let dna = nucleus_state
+            .dna
+            .as_ref()
+            .ok_or(HolochainError::DnaMissing)?;
+
+        // 1. Validate the call (a number of things could go wrong)
+        // 1.a make sure the zome and function exists
+        let _ = dna
+            .get_function_with_zome_name(&zome_call.zome_name, &zome_call.fn_name)
+            .map_err(|e| HolochainError::Dna(e))?;
```
sphinxc0re: This can be shorthanded to:
```suggestion
            .map_err(HolochainError::Dna)?;
```

nodejs_waiter/src/waiter.rs
```diff
@@ -162,7 +162,7 @@ impl Waiter {
                 let aw = aw.clone();
                 match (self.current_checker(), aw.action().clone()) {
                     // Pair every `ExecuteZomeFunction` with one `ReturnZomeFunctionResult`
```
maackle:
```suggestion
                    // Pair every `SignalZomeFunction` with one `ReturnZomeFunctionResult`
```

zippy:
@lucksus I ran app-spec locally and it works just fine on my ubuntu system.  Not sure why it's failing here.  In any case this is much better than before...

PR #993, opened 2019-02-12
holochain_net: UntrackDna JsonProtocol

ddd-mtl:
Can now send an "UntrackDna" message to network module which will effectively make the agent leave the network of that DNA.

Network module will check if core is tracking the DNA it is sending messages to.
If its not the case, the network module will send back a FailureResult Message.


net/src/memory_server.rs
```diff
@@ -253,140 +253,248 @@ impl InMemoryServer {
...
-                        })
-                        .into(),
-                    )?;
-                    // Request all data lists from this agent
-                    self.priv_request_all_lists(&msg.dna_address, &msg.agent_id);
-                }
-
-                JsonProtocol::SendMessage(msg) => {
-                    self.priv_serve_SendMessage(&msg)?;
+        // serve only JsonProtocol
```
ddd-mtl: Only change here is dealing with the Error case first so we dont have to indent the rest of the code.

ddd-mtl:
@neonphog FYI, The `block_on`s are only for the tests where you have to manually call things to have a proper test case.

net/src/memory_server.rs
```diff
@@ -253,140 +272,251 @@ impl InMemoryServer {
...
+                bucket_id
+            ));
+            return Err(NetworkError::GenericError {
+                error: "DNA not tracked by agent and no sender info.".to_string(),
+            }
+            .into());
+        }
+        let sender_info = maybe_sender_info.unwrap();
+        let sender_agent_id = sender_info.0;
+        let sender_request_id = match sender_info.1 {
```
sphinxc0re:
```rust
let sender_request_id = sender_info.1.unwrap_or_default();
```

net/src/memory_server.rs
```diff
@@ -40,14 +40,33 @@ fn into_bucket_id(dna_address: &Address, agent_id: &str) -> BucketId {
...
+        HashString::from(vec[0]),
+        vec[1].to_string(),
+        serde_json::from_str(vec[2]).expect("metaId not holding valid json content"),
+    )
+}
+
+/// Tells if metaId is of given entry and attribute
+fn _metaId_is(metaId: &Address, entry_address: Address, attribute: String) -> bool {
+    let meta_tuple = _undo_meta_id(metaId);
+    return meta_tuple.0 == entry_address && meta_tuple.1 == attribute;
```
sphinxc0re: Unnecessary return statement.

net/src/memory_server.rs
```diff
@@ -40,14 +40,33 @@ fn into_bucket_id(dna_address: &Address, agent_id: &str) -> BucketId {
...
+/// return a unique identifier out of an entry_address and attribute
+fn into_meta_id(meta_tuple: &MetaTuple) -> Address {
     HashString::from(format!(
         "{}||{}||{}",
         meta_tuple.0, meta_tuple.1, meta_tuple.2
     ))
 }
 
+/// unmerge meta_id into a tuple
+fn _undo_meta_id(meta_id: &Address) -> MetaTuple {
```
sphinxc0re: Unnecessary underscore in front of function name. There are multiple of these in this change

core/src/workflows/respond_validation_package_request.rs
```diff
@@ -19,7 +19,9 @@ fn get_entry(address: &Address, context: &Arc<Context>) -> Result<Entry, Holocha
         .read()
         .unwrap()
         .fetch(address)?
-        .ok_or(HolochainError::ErrorGeneric("Entry not found".to_string()))?;
+        .ok_or(HolochainError::ErrorGeneric(
```
sphinxc0re: Use `ok_or_else` here of lazy evaluation and better performance

net/src/memory_server.rs
```diff
@@ -555,34 +761,103 @@ impl InMemoryServer {
...
+                    let (_k, r) = &e
+                        .get_mut()
+                        .iter()
+                        .next()
+                        .expect("senders_by_dna.entry does not hold any value");
                     r.send(JsonProtocol::HandleFetchMeta(msg.clone()).into())?;
                     return Ok(());
                 }
             }
+            // TODO: Once linking meta is properly implemented in Core
```
lucksus: What exactly does this mean?
Links (and the according meta messages and handling thereof) are implemented already.

ddd-mtl: If you look into `publish_link_meta()` you'll see that Core is storing the whole AddLinkEntry with its meta as the value of the link and the attribute of this meta is always "link".  I'm pretty sure thats not how we want links to work. We would want the attribute to be the type of the link (as defined by the zome code) and the value to be the hash of the other Entry.

lucksus: No, we can't do it the way you say because that would lock-in (or equate) the meta system with links. But we have more meta items, like CRUD status and links, and potentially more to come. We need to send the whole `AddLinkEntry` as meta value since the receiving node needs that info, esp. the header of that entry, to run the validation on that StoreMeta request.

So, the way it is implemented right now is pretty much the way it should be, I think.

lucksus: The reason we need `PublishMeta` as a network method at all is that the routing (to the nodes that receive the `HandleStoreMeta`) needs to look at the address of a separately defined `base` and not the thing that gets send. - It was not meant as a 1-to-1 match of the EAVs or even links.

ddd-mtl: I think we should talk about this. The meta system is already locked in for links since linking is the main feature of Holochain's DHT. In the current way, the network module and thus DHT cannot optimize requests on link type/tag since this is hidden within the meta value. For example, when requesting a link of a certain type/tag, the DHT will have to fetch ALL links on that entry from the network and let Core do the triage.
Also some other parts of the code seem to be expecting what I'm saying: Look at `reduce_get_links_inner()`. Its asking for meta with the attribute `link__{}`. This attribute name is never used when doing a PublishMeta, so I'm pretty sure the 'get_links' isn't getting anything back or if it is, its not based on that attribute.

lucksus: What do you mean with „it is already locked in for links“? I can’t see how this could be the case as we are also storing (and publishing) CRUD statuses as meta items. We recently discussed that we might have to store headers as well as meta items on links. The last point I see (there was no explicit discussion nor decision around that yet, but I can’t imagine how we could get around this) is validation packages that would need to be stored as meta.

lucksus: I‘m pretty sure that links get published with ˋlink__<tag>ˋ.

lucksus: Yeah, I think we should sync ASAP. When you talk about optimization, you seem to be asserting things that go beyond the design @neonphog and I discussed.

I expect the network implementation to tunnel through the meta attribute and content to the „right“ other instance, where core understands the attribute and deals with it accordingly. Same as with the entries in a regular publish. The only difference (from the network implementation) between Publish and PublishMeta is that the latter routes according to a base address, different to the content that is published. And it has an attribute that gets passed through as, so that the receiving core instances can understand different meta items like links, CRUD and the others mentioned above. @ddd-mtl

PR #997, opened 2019-02-13
Capabilities->Traits

zippy:
Context: capabilities-3 got too big, and there are a bunch of bits in it that I needed to do as pre-work that should really be split up for individual review AND would be good to get in sooner rather than later.

- [x] complete convert of capabilities to trait (removing capability type)

The most visible change here is the renaming of the function grouping from "capabilities" to "traits"  This PR is partly about completing coming to agreement around that name.  Remember this is about the fact that the "traits" i.e. the function groupings, are what's available at bridge time.  Note that other possible names include "aspect"  "interface" "conduit" "connector".

Currently this PR is on top of #992 which is on top of #990, which is the correct merge order.


core/src/nucleus/ribosome/fn_call.rs
```diff
@@ -0,0 +1,1271 @@
...
+
+    pub fn cap_token(&self) -> Address {
+        self.cap.cap_token.clone()
+    }
+}
+
+/// Reduce ExecuteZomeFunction Action
+/// Execute an exposed Zome function in a separate thread and send the result in
+/// a ReturnZomeFunctionResult Action on success or failure
+pub(crate) fn reduce_execute_zome_function(
```
lucksus: 😱 
Is this whole file an accidental add?

doc/holochain_101/src/zome/capabilities.md
```diff
@@ -0,0 +1,3 @@
+# Capabilities
+
+Holochain uses a modified version of the [capabilities](link_to_wikipedia) security model. Instead of the more familiar access control list (ACL) security model, Holochain DNA instances grant revokable capability tokens which can then be given out as a credential to be provided back when access to some function is desired.  These capability grants are recorded as private entries on an agent's chain, and are checked for every access.  We currently have implemented three different levels of capability grants:  Public, Transferable, and Assigned.  Public grants are meant to be used by the Conductor for functions which are available for anyone to call.  Transferable grants provide tokens that are valid from any source.  Assigned grants generate tokens that are only valid for specific set of agents.
```
lucksus:
```suggestion
Holochain uses a modified version of the [capabilities](https://en.wikipedia.org/wiki/Capability-based_security) security model. Instead of the more familiar access control list (ACL) security model, Holochain DNA instances grant revokable capability tokens which can then be given out as a credential to be provided back when access to some function is desired.  These capability grants are recorded as private entries on an agent's chain, and are checked for every access.  We currently have implemented three different levels of capability grants:  Public, Transferable, and Assigned.  Public grants are meant to be used by the Conductor for functions which are available for anyone to call.  Transferable grants provide tokens that are valid from any source.  Assigned grants generate tokens that are only valid for specific set of agents.
```

lucksus: That's the wikipedia link

doc/holochain_101/src/zome/zome_functions.md
```diff
@@ -99,9 +73,22 @@ The `name` is the most important thing here, because when a function call to an
...
+Traits provide a way to group functions by name.  The primary use of this feature is for creating a composibility space where DNA creators can implement different DNAs to emergent function interfaces and then compose with them in the conductor by matching on the function group names and signatures.  Additionally Holochain may reserve a few special trait names that have specific side-effects.  The first of such reserved names is `hc_public`.  Functions grouped in this name will have automatically added to a public capability grant that happens at genesis time, thus making them accessible to any caller.  For more details on the Holochain security model please see the [Capbilities](capabilities.md) section.
+
+Here is an example of what a trait definition using the public reserved trait name might look like:
+
+```json
+"traits": {
+    "hc_public": {
+        "functions": ["get_task_list"]
+    }
+}
```
lucksus: This doesn't match the code. Shouldn't it be:
```
traits: {
  hc_public [get_task_list]
}
```
?

core/src/nucleus/ribosome/fn_call.rs
zippy: yep.

doc/holochain_101/src/zome/zome_functions.md
zippy: That file shows the json that comes out at the end.  Not the rust code.

zippy:
Yes you are right "hc_public" having the purpose of marking functions for which to create the public grant, makes it operate not fully as a trait.  We can indeed consider this a bootstrapping solution for now.

core/src/network/test_utils.rs
```diff
@@ -196,10 +196,10 @@ pub fn test_wat_always_invalid() -> String {
 
     (func
         (export "__list_functions")
-        (param $allocation i32)
-        (result i32)
+        (param $allocation i64)
+        (result i64)
 
-        (i32.const 0)
+        (i64.const 0)
```
Connoropolous: what happened here? Why were there still i32s?

doc/holochain_101/src/zome/define_zome.md
```diff
@@ -12,10 +12,11 @@ A Zome has
 - `config`: Not Implemented Yet
 - validating entry types: definition may vary based on the language
 - a `genesis` function: a callback that Holochain expects and requires, defined in the code itself
-- `capabilities`: a collection of custom functions, which share configuration of their visibility and permissions, a.k.a who/what is allowed to call those functions
-    - custom functions: the core application logic of a Zome, written in a language that compiles to WASM, which Holochain interprets through that compiled WASM
+- `fn_declarations`: a collection of custom functions declarations,
+- `traits`: sets of named function groups used for composibility
```
Connoropolous:
```suggestion
- `traits`: sets of named function groups used for composability
```

core_types/src/dna/traits.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    /// used for declaring functions that will auto-generate a public grant during genesis
+    Public,
+}
+
+impl FromStr for ReservedTraitNames {
+    type Err = &'static str;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "hc_public" => Ok(ReservedTraitNames::Public),
```
Connoropolous: can you define "hc_public" as a const that you reference everywhere you currently use the magic string?

core/src/network/test_utils.rs
zippy: I don't know, it slipped through cracks with David.  I talked to him about it to confirm.

core_types/src/dna/traits.rs
Connoropolous: Adding it that way instead also serves as a reminder to document the nature of that string, and what it represents, whereas the use of the raw string throughout the code doesn't

zippy: The problem with this is there is an issue with how this works in macros.  In the macros the name can't have quotes.  I.e. the macro adds quotes:
```
    traits: {
        hc_public [call_bridge]
    }
```
So you couldn't use any static like HC_PUBLIC because that would turn into "HC_PUBLIC"   We would have to change the way the macro works.  I could put that in for internal use, but then it would be a half measure.. so I would prefer to leave it "hc_public" so you can search and find them the same way.

Connoropolous: K. happy to let this one go, just holding for this to happen more going forward

doc/holochain_101/src/zome/capabilities.md
```diff
@@ -0,0 +1,3 @@
+# Capabilities
+
+Holochain uses a modified version of the [capabilities](https://en.wikipedia.org/wiki/Capability-based_security) security model. Instead of the more familiar access control list (ACL) security model, Holochain DNA instances grant revokable capability tokens which can then be given out as a credential to be provided back when access to some function is desired.  These capability grants are recorded as private entries on an agent's chain, and are checked for every access.  We currently have implemented three different levels of capability grants:  Public, Transferable, and Assigned.  Public grants are meant to be used by the Conductor for functions which are available for anyone to call.  Transferable grants provide tokens that are valid from any source.  Assigned grants generate tokens that are only valid for specific set of agents.
```
Connoropolous:
```suggestion
Holochain uses a modified version of the [capabilities](https://en.wikipedia.org/wiki/Capability-based_security) security model. Instead of the more familiar access control list (ACL) security model, Holochain DNA instances grant revokable capability tokens which can then be given out as a credential to be provided back when access to some function is desired. 

These capability grants are recorded as private entries on an agent's chain, and are checked for every access.  We currently have implemented three different levels of capability grants:  `Public`, `Transferable`, and `Assigned`.  

`Public` grants are meant to be used by the Conductor for functions which are available for anyone to call.

`Transferable` grants provide tokens that are valid from any source.

`Assigned` grants generate tokens that are only valid for specific set of agents.
```

doc/holochain_101/src/zome/capabilities.md
```diff
@@ -0,0 +1,11 @@
+# Capabilities
+
+Holochain uses a modified version of the [capabilities](https://en.wikipedia.org/wiki/Capability-based_security) security model. Instead of the more familiar access control list (ACL) security model, Holochain DNA instances grant revokable capability tokens which can then be given out as a credential to be provided back when access to some function is desired. 
+
+These capability grants are recorded as private entries on an agent's chain, and are checked for every access.  We currently have implemented three different levels of capability grants:  `Public`, `Transferable`, and `Assigned`.
```
Connoropolous: I'm a little lost. How do I use Capabilities as a developer now? Or that's not really a thing, all that comes in Capabilites 3?

zippy: updated

PR #1000, opened 2019-02-13
Missing hc_dpki serde_drive crate and macros prevents compilation

pjkundert:
Without crate serde_derive and its macros, hc_dpki doesn't compile.  I'm not certain why everyone isn't having this problem, but it seems to be a requirement for clean compilation...

### DIFF
--- a/hc_dpki/src/lib.rs
+++ b/hc_dpki/src/lib.rs
@@ -7,8 +7,9 @@ extern crate base64;
 extern crate bip39;
 extern crate boolinator;
 extern crate rustc_serialize;
-#[macro_use]
 extern crate serde;
+#[macro_use]
+extern crate serde_derive;
 
 pub mod bundle;
 pub mod error;


sphinxc0re:
The `hc_dpki` crate is a 2018 edition crate. This issue is fixed in #1001

sphinxc0re:
This is only the first step in removing all the `extern crate` statements in that crate

maackle:
@sphinxc0re I'm pretty sure this is a dependency slippage, everyone who has `cargo update`d lately has seen this, and the solution has been to pull the macros from `serde_derive`, probably because of a change to serde. Reopening until we know for sure.

(Also congrats on PR 1000 @pjkundert!)

maackle:
Sorry @sphinxc0re actually didn't see the line that added serde_derive in your PR, and thought you hadn't addressed this issue.

PR #1001, opened 2019-02-13
Small adjustment

sphinxc0re:


sphinxc0re:
@Connoropolous That change might be especially interesting for you.

Connoropolous:
That is so interesting! I was trying that and it wasn’t working for me, and I can’t see how what I had tried is different than what you wrote. 

I was getting getting EnvVar::Something is not a module

pjkundert:
It cleans up the problem for me, too.  I did delete all my Cargo.lock files before trying it, though -- I don't know if that had an effect.  Is serde_json no longer a (separate) crate that needs to be included, then?  Perhaps we should clean up everywhere else then, too...

lucksus:
`hc_dkpi` is an `edition = 2018` crate. In Edition 2018 you don't need to declare `extern crate ..`, you can just use it directly. @pjkundert

PR #1003, opened 2019-02-13
The Signature type should implement both PartialEq and Eq

pjkundert:
o When working with structures containing Signatures, the concept of
  ensuring that Signatures are equivalent is (and should be):

  reflexive: a == a;
  symmetric: a == b implies b == a; and
  transitive: a == b and b == c implies a == c.

  This implies the Eq relationship.  There are no Signatures that
  require (only) the weaker partial-equivalence relationship implied
  by specifying only the PartiaEq Trait; just a == b and a != b
  being strict inverses.  By not supplying Eq, we are prevented
  from performing Eq comparisons on (derivative) structs and/or
  other Signatures.


pjkundert:
What are the chances of getting this into 0.0.4-alpha?

zippy:
I'm sorry this didn't get into 0.0.4 just an oversight.  I don't see any reason we can't merge to develop at least now @lucksus ?  @maackle ?

core_types/src/signature.rs
```diff
@@ -5,7 +5,7 @@
 /// Signature is meant in the classic cryptographic sense,
 /// as a string which can be validated as having been signed
 /// by the private key associated with a given public key
-#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
```
sphinxc0re: To be clear, `Eq` is just really just a marker. The actual implementation is in `PartialEq`:
https://doc.rust-lang.org/1.32.0/core/cmp/trait.Eq.html

pjkundert: Yes, and without this marker, we are stating that PartialEq does *not* implement a "total order", and thus the type cannot be used within any other type/struct requiring, for example, `std::cmp::AssertParamIsEq`:
```
error[E0277]: the trait bound `holochain_core_types::signature::Signature: std::cmp::Eq` is not satisfied
  --> src/entries.rs:48:5
   |
48 |     pub proposal_sig:    Signature,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::cmp::Eq` is not implemented for `holochain_core_types::signature::Signature`
   |
   = note: required by `std::cmp::AssertParamIsEq`
```

pjkundert: Unless a type is explicitly not orderable (eg. like `f32`, `f64`, due to NaN), then we should probably ensure we derive both `PartialEq` and `Eq`, because it cannot be "fixed" in user code.

PR #1004, opened 2019-02-13
split tests to decrease wall-time

zippy:


lucksus:
Cool! Why WIP?

lucksus:
This now includes the Circle CI config updated to use more resources which fixes the app-spec tests.

PR #1006, opened 2019-02-13
Add util functions to hdk

willemolding:

Adds a new module hdk::utils which has some functions which make zome development much easier but maybe do not belong in the HDK. This includes the following functions: 

  + get_links_and_load_type - calls try_from for a given type when getting links
  + get_as_type - Similar but for a single entry
  + link_entries_bidir - Same as link_entries but creates link in both directions
  + commit_and_link - Save a line and commit and link in a single function


hdk-rust/specs/hdk-proposal.rs
```diff
@@ -0,0 +1,39 @@
+#![hdk::zome] // this attribute-like macro operates on the whole module. This is equivalent to define_zome! in the old HDK
```
zippy: I don't think this file should make its way into the repo in this PR...

hdk-rust/src/utils.rs
```diff
@@ -0,0 +1,83 @@
...
+                "Could not convert get_links result to requested type".to_string(),
+            )
+        }),
+        _ => Err(ZomeApiError::Internal(
+            "get_links did not return an app entry".to_string(),
+        )),
+    }
+}
+
+pub fn link_entries_bidir<S: Into<String>>(
```
zippy: comments?

hdk-rust/specs/hdk-proposal.rs
lucksus: But it's good. I'd say let's just merge it!

PR #1009, opened 2019-02-14
add Instance class to holochain-nodejs

Connoropolous:

### DIFF
--- a/nodejs_conductor/index.js
+++ b/nodejs_conductor/index.js
@@ -52,54 +52,6 @@ Conductor.prototype.stop = function () {
     return this._stopPromise
 }
 
-Conductor.prototype.call = function (id, zome, fn, params) {
-    const stringInput = JSON.stringify(params)
-    let rawResult
-    let result
-    try {
-        rawResult = this._callRaw(id, zome, fn, stringInput)
-    } catch (e) {
-        console.error("Exception occurred while calling zome function: ", e)
-        throw e
-    }
-    try {
-        result = JSON.parse(rawResult)
-    } catch (e) {
-        console.warn("JSON.parse failed to parse the result. The raw value is: ", rawResult)
-        return rawResult
-    }
-    return result
-}
-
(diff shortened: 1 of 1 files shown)


nodejs_conductor/index.js
```diff
@@ -129,9 +81,59 @@ Conductor.run = function (config, fn) {
...
+class Instance {
+    constructor(instanceId, conductor) {
+        this.id = instanceId
+        this.conductor = conductor
+        this.agentId = this.conductor.agent_id(instanceId)
+        this.dnaAddress = this.conductor.dna_address(instanceId)
+    }
+
+    // internally calls `this.conductor._callRaw`
+    call(zome, fn, params) {
```
maackle: This totally works. I would rather see it as a plain ol' object that gets created from a function, rather than a class, though. I get the desire to make "Instance" a first-class concept, but making it a class doesn't really confer it any more reality as a concept aside from making it more visible in the code.

The rationale is that none of these members (id, conductor, agentId, dnaAddress) should be mutable. It's undefined behavior if anyone were to reassign any of these values. They should know that, but in the spirit of making invalid things impossible, I'd prefer to bake these values into closures directly.

i.e. I'd rather see:

```javascript
function makeInstance(instanceId, conductor) {
  return {
    id: instanceId,
    agentId: conductor.agent_id(instanceId),
    dnaAddress: conductor.dna_address(instanceId),
    call: (...args) => {
      conductor._callRaw(instanceId, ...args)
      // ...etc
    },
    callWithPromise: function(...args) { this.call(..args) /* etc... */ }
  }
}
```

But this is probably just me getting used to Rust where things are tighter. I don't expect you to make this change, just wanted to share the idea. I do think a class is more readable, which may be more important than encapsulating those members.

maackle: Ha, now that I see it I realize this is very much like `makeCaller`...ultimately this is a pretty small point though.

PR #1010, opened 2019-02-14
Signing 3 - Key loading/handling in Conductor

lucksus:
*on top of https://github.com/holochain/holochain-rust/pull/968*
# Context
screenshot 2019-02-14 at 19 34 04
All the pre-work enabling the Conductor to create real signatures from real keys read out of real files.

## Steps
* Conductor gets a new call back `KeyLoader` that is implemented with a default that reads files from the filesystem and prompts for a passphrase
* nodejs Conductor provides its own key loader callback to inject deterministic keys for tests that are created from the agent name's SHA256
* Conductor checks that public keys given in the config match the key load from file


lucksus:
@neonphog, @zo-el, I had to declare `SodiumBuf` *Send* for this to work: https://github.com/holochain/holochain-rust/pull/1010/commits/19cde5994c96c1b24e53c350daf9ae06243a2959
Is it safe to send these across threads? We sure need that and I don't see why not, but rustc can't say much about a `c_void`..

neonphog:
@lucksus - so long as `sodium_init()` is called before any functions are invoked, all libsodium functions are thread safe. The memory allocated in SodiumBuf is not mutexed or anything, so not safe to access from multiple threads, but, as you say, sending should be perfectly safe.

I misread this question at first, and thought we needed a clone for SecBuf... here's that code in case it's useful someday, lol. Keys are generally very small so cloning is cheap: https://github.com/holochain/holochain-rust/compare/clone-secbuf

conductor_api/src/conductor/base.rs
```diff
@@ -472,6 +511,29 @@ impl Conductor {
...
+        notify(format!("Reading agent key from {}", file.display()));
+
+        let mut file = File::open(file)?;
+        let mut contents = String::new();
+        file.read_to_string(&mut contents)?;
+
+        let passphrase = rpassword::read_password_from_tty(Some("Passphrase: "))?;
+
+        let bundle: KeyBundle = serde_json::from_str(&contents)?;
+        let mut passphrase = SecBuf::with_insecure_from_string(passphrase);
```
neonphog: passphrases should use secure memory

Ideally, we'd use a low-level keyboard binding to prevent a class of malware keyloggers and input the data directly into a secure buffer.

For the short-term, we should be sure to zero the memory in "passphrase" returned by the rpassword utility.

conductor_api/src/conductor/base.rs
```diff
@@ -472,6 +511,29 @@ impl Conductor {
...
+        Keypair::from_bundle(
+            &bundle,
+            &mut passphrase,
+            Some(PwHashConfig(
+                pwhash::OPSLIMIT_INTERACTIVE,
+                pwhash::MEMLIMIT_INTERACTIVE,
+                pwhash::ALG_ARGON2ID13,
+            )),
```
neonphog: We should use `None` for the config here. We want to use the default SENSITIVE limits for the real solution.

conductor_api/src/conductor/base.rs
lucksus: I had tried that already in `hc keygen` but `SodiumBuf` doesn't allow arbitrary sizes for secure buffers. Only [multiples of 8](https://github.com/holochain/holochain-rust/blob/develop/sodium/src/secbuf.rs#L70). Also, since the passphrase is coming in as a `String`, it wouldn't make it much safer, right?

But yeah, so you are saying, we should roll our own `read_password_from_tty` eventually?

lucksus:
@neonphog, I've implemented your change requests, except for using a secure `SecBuf` as that enforces sizes % 8. I've tried padding with zeros but I guess that makes it be a different passphrase. Can't unlock keys with that. Any hint how to use `SecBuf` with arbitrary length passphrases?

conductor_api/src/conductor/base.rs
neonphog: : ) right, that totally makes sense. I'll create an issue to do the alignment in the background and deal with appropriately sized slice references in the SecBuf so we don't have to worry about the sizing during implementation.

neonphog: https://github.com/holochain/holochain-rust/issues/1017

conductor_api/src/conductor/base.rs
```diff
@@ -453,6 +465,46 @@ impl Conductor {
...
+
+    /// Get reference to key for given agent ID.
+    /// If the key was not loaded (into secure memory) yet, this will use the KeyLoader
+    /// to do so.
+    fn get_key_for_agent(&mut self, agent_id: &String) -> Result<&Keypair, String> {
+        if !self.agent_keys.contains_key(agent_id) {
+            let agent_config = self
+                .config
+                .agent_by_id(agent_id)
+                .ok_or(format!("Agent '{}' not found", agent_id))?;
```
zippy:
```suggestion
                .ok_or(format!("Agent '{}' not found in config", agent_id))?;
```

PR #1012, opened 2019-02-14
Signing 4 - conductor_api function agent/sign

lucksus:

## Production replacement for the `mock_signer()` using the instance's agent key to sign messages.

screenshot 2019-02-14 at 20 31 06

## Context
With all branches included in this, completes *Entries are signed by the author*:
screenshot 2019-02-14 at 20 34 25


zippy:
hmmm CI process getting killed because of memory too?

lucksus:
@zippy, yeah, the build process needs more resources as well. app-spec is failing already in signing-3. will fix both.

conductor_api/src/interface.rs
```diff
@@ -740,12 +741,30 @@ impl ConductorApiBuilder {
...
-    pub fn with_mock_signing_callback(&mut self) {
+    pub fn with_agent_signature_callback(mut self, keypair: Arc<Mutex<Keypair>>) -> Self {
         self.io.add_method("agent/sign", move |params| {
             let params_map = Self::unwrap_params_map(params)?;
             let payload = Self::get_as_string("payload", &params_map)?;
-            Ok(json!({"payload": payload, "signature": mock_signer(payload)}))
+            // Convert payload string into a SecBuf
+            let mut message = SecBuf::with_insecure_from_string(payload.clone());
+            // Create signature
+            let mut message_signed = SecBuf::with_insecure(SIGNATURESIZE);
```
maackle: I just got to look at this stuff for the first time. Are you concerned that mutation is creeping into higher layers of abstraction? 

Is there any reason why we need an `Arc<Mutex<_>>` for any reason other than that keypair.sign requires does in-place mutation of its arguments? Is there any reason why we can't make keypair::sign create its own empty buffer and return that instead of passing one in from outside? Wouldn't that remove the need for mutable references and maybe this mutex?

maackle: I looked at it more and it does look like there's no way around this, based on how holochain-sodium is implemented. I know this is a little out of scope for this PR but I'm wondering why we need all this shared memory and locking stuff for signing, and if there's any way to keep it from infecting the higher abstractions?

lucksus: We need that because in order to sign we need the agent's private key. `Keypair` is wrapping the key, stored in secure memory, so that we can use keys without having to deal with memory locking etc. and still have a secure way to handle keys. The only drawback is this: we need have a mutable reference to it in order to use the key, because we need to make the memory readable for us. And yes, that is why I switched to `Arc<Mutex<Keypair>>` here..

maackle: I see, it's because of secure memory. If this gets out of hand though, we might want to find a way to contain this shared mutation in an abstraction so it doesn't run amok and infect everything.

neonphog: Indeed there's no reason we can't implement internal mutability to these sodium functions to achieve signing and other crypto on with read-only keys (although we'd probably want to use an unsafe variant so we don't get runtime rust panics) . Just need to have a chance to catch our breaths and prioritize stuff like that...

PR #1016, opened 2019-02-15
Release blocker unblock: cli help fixes

Connoropolous:
I'd also like to get that tiny bit of extra feedback for `hc keygen` in, since its really opaque at the moment.

### DIFF
--- a/cli/src/cli/generate.rs
+++ b/cli/src/cli/generate.rs
@@ -6,7 +6,6 @@ use crate::{
     error::DefaultResult,
     util,
 };
-use colored::*;
 use serde_json;
 use std::{
     fs::{self, File},
@@ -58,14 +57,6 @@ pub fn generate(zome_name: &PathBuf, language: &str) -> DefaultResult<()> {
         // TODO: supply zome name for AssemblyScriptScaffold as well
         _ => bail!("unsupported language: {}", language),
     }
-
-    // CLI feedback
-    println!(
-        "{} new {} Zome at {:?}",
-        "Generated".green().bold(),
-        language,
-        zome_name
-    );
 
     Ok(())
 }
--- a/cli/src/cli/test.rs
+++ b/cli/src/cli/test.rs
@@ -1,6 +1,7 @@
 use crate::{cli::package, error::DefaultResult, util};
 use colored::*;
-use std::{fs, path::PathBuf};
+use failure::Error;
+use std::{fs, io::ErrorKind, path::PathBuf, process::Command};
 
 pub const TEST_DIR_NAME: &str = "test";
 pub const DIST_DIR_NAME: &str = "dist";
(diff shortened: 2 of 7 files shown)


cli/src/main.rs
```diff
@@ -143,7 +143,7 @@ enum Cli {
     #[structopt(
         name = "keygen",
         alias = "k",
-        about = "Creates a new agent key pair, asks for a passphrase and writes an encrypted key bundle to ~/.config/holochain/keys"
+        about = "Creates a new agent key pair, asks for a passphrase and writes an encrypted key bundle to disk in a location specific to Holochain (and dependent on OS platform)"
```
lucksus:
```suggestion
        about = "Creates a new agent key pair, asks for a passphrase and writes an encrypted key bundle to disk in the XDG compliant config directory of Holochain which is dependent on the OS platform (/home/alice/.config/holochain or C:\Users\Alice\AppData\Roaming\holochain\holochain or /Users/Alice/Library/Preferences/com.holochain.holochain)"
```

Connoropolous: the directory you mention is the XDG compliant one, but there is the `keys` subdir where it actually ends up. 
When you actually generate the keys, it shows you the exact location of it, so I thought it should be fine to leave it off/out in the first place, from the help text

lucksus: This way you see what it would be on other platforms..
Here's with /keys
```suggestion
        about = "Creates a new agent key pair, asks for a passphrase and writes an encrypted key bundle to disk in the XDG compliant config directory of Holochain, which is dependent on the OS platform (/home/alice/.config/holochain/keys or C:\Users\Alice\AppData\Roaming\holochain\holochain\keys or /Users/Alice/Library/Preferences/com.holochain.holochain/keys)"
```

Connoropolous: Ok cool. Good with that then

cli/src/cli/generate.rs
```diff
@@ -59,14 +58,6 @@ pub fn generate(zome_name: &PathBuf, language: &str) -> DefaultResult<()> {
         _ => bail!("unsupported language: {}", language),
     }
 
-    // CLI feedback
```
Connoropolous: moved this into the rust zome generator code itself

cli/src/cli/test.rs
```diff
@@ -109,7 +143,10 @@ pub mod tests {
             .exists());
     }
 
+    // TODO: this test is non-deterministic, pivoting around the fact that the
+    // behaviour of the command is different, depending whether nodejs is installed on the system or not
     #[test]
+    #[cfg(not(windows))]
```
thedavidmeister: @Connoropolous for future reference please use `#[cfg(feature = "broken-tests")]` as we might never fix `#[cfg(not(windows))]` flags

PR #1018, opened 2019-02-15
EaviQuery

maackle:
Consolidates EAVI lookup logic into a single datatype as much as possible. This is in preparation for merging develop into #952, abstracting away the current requirement for Attributes to be strings so that we can look them up by prefix, allowing us to move to enum Attributes.

The prefix matching thing went deeper than I thought, and required some significant changes. I also didn't realize we were using glob patterns to match files in EavFileStorage, and by completely abstracting away strings and prefixes, we lose that. If we want it for performance reasons, we should optimize this query language to include a glob-based filter variant.

I'm sorry to add so many changes for something that was already done (Remove Links), but this logic had to be reworked in order for #952 to be usable. I think something like this would have had to be done the next time we wanted to expand our ability to query the EAVI, because the existing query logic was fine-tuned for querying for potentially removed links.

- [ ] ~I have added a summary of my changes to the changelog~


cas_implementations/src/eav/file.rs
```diff
@@ -170,12 +169,12 @@ impl EntityAttributeValueStorage for EavFileStorage {
...
-            Some(eav.attribute()),
-            Some(eav.value()),
-            IndexQuery::default(),
-        )?;
+        let fetched = self.fetch_eavi(&EaviQuery::new(
+            Some(eav.entity()).into(),
+            Some(eav.attribute()).into(),
+            Some(eav.value()).into(),
+            IndexFilter::LatestByAttribute,
+        ))?;
```
maackle: There are a ton of changes just like this in this PR (as many as there are `fetch_eavi` calls). So take comfort, dear reader, that you can skip anything like this pattern once you understand this one.

I added some `impl From<_> for EavFilter` to make these minimally invasive.

cas_implementations/src/eav/file.rs
```diff
@@ -187,49 +186,26 @@ impl EntityAttributeValueStorage for EavFileStorage {
...
-                    .intersection(&entity_set)
-                    .cloned()
-                    .collect();
-                let to_union = set?;
-                Ok(to_union
-                    .union(&entity_attribute_value_inter)
-                    .cloned()
-                    .collect())
-            },
-        );
```
maackle: This is a lot of prefix-specific logic that unfortunately had to go

core_types/src/eav.rs
```diff
@@ -287,87 +357,19 @@ impl EntityAttributeValueStorage for ExampleEntityAttributeValueStorage {
...
-
-        Ok(filtered)
+        let lock = self.storage.read()?;
+        let set = (*lock).clone();
+        let iter = set.iter().cloned();
+        Ok(query.run(iter))
     }
 }
 
-pub fn get_latest(
```
maackle: The logic of `get_latest` was DRY'd up and moved to `EaviQuery::run` under the `LatestByAttribute` variant of `IndexFilter`

core_types/src/eav.rs
```diff
@@ -287,87 +357,19 @@ impl EntityAttributeValueStorage for ExampleEntityAttributeValueStorage {
...
-        "Could not get last value".to_string(),
-    ))
-}
-
 impl PartialEq for EntityAttributeValueStorage {
     fn eq(&self, other: &EntityAttributeValueStorage) -> bool {
-        self.fetch_eavi(None, None, None, IndexQuery::default())
-            == other.fetch_eavi(None, None, None, IndexQuery::default())
+        let query = EaviQuery::default();
+        self.fetch_eavi(&query) == other.fetch_eavi(&query)
```
maackle: I (hopefully) kept this logic the same, which is using the `get_latest` / `LatestByAttribute` method of fetching items, but I wonder if it shouldn't be altered to just return all items? @AshantiMutinta? This is one place where I'm not sure what the original does.

AshantiMutinta: Oh so get latest returns the latest record in the set if the dates are not explicitly provided. I think this is fine but if we ever wanted to do a get latest per record we could always modify the way the eavi works but I think from a functionality perspective a get_latest should be just fine

maackle:
@AshantiMutinta I think that's definitely the way to go. I tried it, and it turns out there's a bit of a rabbit hole to go down because currently `IndexFilter` is `Clone`, and to make it own a `Fn` would make it not so, which would require a bunch of extra code (which should probably be there anyway). But, it's best as a follow-up.

PR #1019, opened 2019-02-15
Release blocker unblock: default js test wasn't up to date

Connoropolous:

### DIFF
--- a/cli/src/cli/js-tests-scaffold/index.js
+++ b/cli/src/cli/js-tests-scaffold/index.js
@@ -1,38 +1,20 @@
 // This test file uses the tape testing framework.
 // To learn more, go here: https://github.com/substack/tape
-const test = require('tape');
-
-const { Config, Conductor } = require("@holochain/holochain-nodejs")
+const { Config, Scenario } = require("@holochain/holochain-nodejs")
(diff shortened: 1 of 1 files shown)


cli/src/cli/js-tests-scaffold/index.js
```diff
@@ -1,38 +1,20 @@
...
-
-const alice = conductor.makeCaller('alice', dnaPath)
-
-test('description of example test', (t) => {
+scenario.runTape("description of example test", (t, instances) => {
   // Make a call to a Zome function
-  // indicating the capability and function, and passing it an input
-    const addr = alice.call("my_zome", "create_my_entry", {"entry" : {"content":"sample content"}})
-
-    const result = alice.call("my_zome", "get_my_entry", {"address": addr.Ok})
```
maackle: @Connoropolous this is likely the pattern all tests will be using since it's part of the scaffolding. Do you prefer this over the destructuring method we use everywhere else? While it shows more clearly what's going on, people might not realize they can write it more succinctly.

Connoropolous: picking which one is low on my list of concerns. I'm up for switching back to destructuring and moving on 👍

PR #1020, opened 2019-02-15
Release blocker unblock: better nodejs logging default

Connoropolous:
This has another nice upside which is that it removes all the crazy amounts of extra logging in app_spec tests, which might actually also help speed it up and maybe eliminate the memory challenge too

### DIFF
--- a/doc/holochain_101/src/scenario_testing_setup.md
+++ b/doc/holochain_101/src/scenario_testing_setup.md
@@ -17,9 +17,7 @@ ___
 
 **Type** `object`
 
-**Description** *conductorOptions.debugLog* `boolean` Which logging to use. There are two options:
-- debugLog = true: Use the "debug" logger. This one has full debug logging and nice, colorful output.
-- debugLog = false: Doesn't output any logs.
+**Description** *conductorOptions.debugLog* `boolean` Enables debug logging. The logger produces nice, colorful output of the internal workings of Holochain.
 
 **Default** `{ debugLog: false }`
 ___
--- a/doc/holochain_101/src/scenario_testing_setup.md
+++ b/doc/holochain_101/src/scenario_testing_setup.md
@@ -17,11 +17,11 @@ ___
 
 **Type** `object`
 
-**Description** *conductorOptions.debugLog* `boolean` Which logger type to use. There are two options:
-- debugLog = true: Use the "debug" logger. This one has nicer, colorful output.
-- debugLog = false: Use the "simple" logger. This one has less interesting output, but can be silenced with the env variable `HC_SIMPLE_LOGGER_MUTE=1`
+**Description** *conductorOptions.debugLog* `boolean` Which logging to use. There are two options:
+- debugLog = true: Use the "debug" logger. This one has full debug logging and nice, colorful output.
+- debugLog = false: Doesn't output any logs.
 
-**Default** `{ debugLog: true }`
+**Default** `{ debugLog: false }`
 ___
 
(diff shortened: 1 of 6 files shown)


doc/holochain_101/src/scenario_testing_setup.md
```diff
@@ -17,11 +17,11 @@ ___
...
 **Type** `object`
 
-**Description** *conductorOptions.debugLog* `boolean` Which logger type to use. There are two options:
-- debugLog = true: Use the "debug" logger. This one has nicer, colorful output.
-- debugLog = false: Use the "simple" logger. This one has less interesting output, but can be silenced with the env variable `HC_SIMPLE_LOGGER_MUTE=1`
+**Description** *conductorOptions.debugLog* `boolean` Which logging to use. There are two options:
+- debugLog = true: Use the "debug" logger. This one has full debug logging and nice, colorful output.
+- debugLog = false: Doesn't output any logs.
 
-**Default** `{ debugLog: true }`
```
maackle: This whole thing could be simplified, people know that the two options for a `boolean` are `true` and `false`

**Description** *conductorOptions.debugLog* `boolean` Enables debug logging. The logger produces nice, colorful output of the internal workings of Holochain.
**Default** `{ debugLog: false }`

doc/holochain_101/src/testing_configuration.md
```diff
@@ -132,11 +132,11 @@ ___
...
 **Type** `object`
 
-**Description** *conductorOptions.debugLog* `boolean` Which logger type to use. There are two options:
-- debugLog = true: Use the "debug" logger. This one has nicer, colorful output.
-- debugLog = false: Use the "simple" logger. This one has less interesting output, but can be silenced with the env variable `HC_SIMPLE_LOGGER_MUTE=1`
+**Description** *conductorOptions.debugLog* `boolean` Which logging to use. There are two options:
+- debugLog = true: Use the "debug" logger. This one has full debug logging and nice, colorful output.
+- debugLog = false: Doesn't output any logs.
 
-**Default** `{ debugLog: true }`
```
maackle: same as above

nodejs_conductor/native/src/config.rs
```diff
@@ -42,9 +42,14 @@ pub fn js_make_config(mut cx: FunctionContext) -> JsResult<JsValue> {
     let logger = if opts.debug_log {
         Default::default()
     } else {
+        // creates a logger which just mutes all
+        let mut rules = LogRules::new();
+        rules
+            .add_rule("^*", true, None)
```
maackle: the `^` is redundant but makes no difference either way

lucksus:
I think we should turn the log on again for app-spec. It does not affect the memory problem much since it's written to a file there and only printed on failure - which does help finding out why CI failed.

Connoropolous:
@lucksus I can do that!

nodejs_conductor/native/src/config.rs
pjkundert: I don't believe that `^*` is a valid Regular Expression, though; the "Kleene star" indicates "match the preceding character 0 or more times", but `^` isn't really a character but an "anchor specifying that the following pattern must begin at the first character of the string". So, a Regular Expression engine could legitimately reject this pattern as invalid, as there is no "character" specified for either the anchor nor the star.  Better to just use `.*` for a universal match.

Connoropolous: thanks Pj. We can update this

PR #1021, opened 2019-02-15
added sleep after init to IPC network initialization

zippy:
fix for 0.0.4 for networking to work!

### DIFF
--- a/net/src/p2p_network.rs
+++ b/net/src/p2p_network.rs
@@ -8,6 +8,7 @@ use holochain_net_connection::{
     protocol::Protocol,
     NetResult,
 };
+use std::{thread::sleep, time::Duration};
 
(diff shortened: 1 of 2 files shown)


net/src/p2p_network.rs
```diff
@@ -49,6 +50,9 @@ impl P2pNetwork {
         };
         // Create NetConnectionThread with appropriate worker factory
         let connection = NetConnectionThread::new(handler, worker_factory, None)?;
+        if let P2pBackendKind::IPC = p2p_config.backend_kind {
+            sleep(Duration::from_millis(1000));
+        }
```
ddd-mtl: I believe it would be nice to add a comment stating that this is a temporary hack until network init is properly handled.

Connoropolous: +1

PR #1022, opened 2019-02-15
Signing 5 - Validate provenances

lucksus:

# Context
This completes the Holochain generic (non-Holo) related part of the signing story:
screenshot 2019-02-19 at 23 30 40

# Content
1. Refactored `validate_entry()`
2. Added signature validation
3. Fixed tests to use real agent IDs backed by real keys so signatures are actually valid

1 and 2 are straightforward. For 3. I had to get creative. Added module `mock_signing` to `test_utils` providing a function `registered_agent_id(String) -> AgentId` that creates deterministic keys from the given agent name and registers them with a static `TEST_AGENT_KEYS: HashMap<Address, Keypair>`. The according `mock_signer` uses this key registry to sign with these deterministic mock keys.


core/src/nucleus/actions/validate.rs
```diff
@@ -3,122 +3,162 @@ use crate::{
...
+            signature_buf
+                .write(0, signature_bytes.as_slice())
+                .expect("SecBuf must be writeable");
+
+            let mut message_buf =
+                SecBuf::with_insecure_from_string(header.entry_address().to_string());
+            let result = Keypair::verify(author.to_string(), &mut signature_buf, &mut message_buf)?;
+
+            (result == 0).ok_or(HolochainError::ValidationFailed(
+                "Signature invalid".to_string(),
```
zippy: Can we add something here about which of the provenences is invalid?  Like:
```suggestion
                format!("Signature for {} invalid",author),
```

test_utils/src/mock_signing.rs
```diff
@@ -0,0 +1,99 @@
...
+    collections::HashMap,
+    sync::{Arc, Mutex},
+};
+
+lazy_static! {
+    pub static ref TEST_AGENT_KEYS: Mutex<HashMap<Address, Arc<Mutex<Keypair>>>>
+        = Mutex::new(HashMap::new());
+}
+
+pub fn registered_test_agent<S: Into<String>>(nick: S) -> AgentId {
```
zippy: nice work!

core/src/context.rs
```diff
@@ -121,6 +61,34 @@ impl Context {
...
+    // There are two different implementations of this function below which get pulled
+    // in depending on if "test" is in the build config, or not.
+    // This allows unit tests of core to not have to deal with a conductor_api.
+    #[cfg(not(test))]
+    fn test_check_conductor_api(
+        conductor_api: Option<Arc<RwLock<IoHandler>>>,
+        _agent_id: AgentId,
+    ) -> Arc<RwLock<IoHandler>> {
+        // If you get here through this panic make sure that the context passed into the instance
+        // gets created with a real container API. In test config it will be populated with mock API
```
maackle:
```suggestion
        // gets created with a real conductor API. In test config it will be populated with mock API
```

core/src/context.rs
```diff
@@ -121,6 +61,34 @@ impl Context {
...
+        conductor_api.expect("Context can't be created without conductor API")
+    }
+
+    #[cfg(test)]
+    fn test_check_conductor_api(
+        conductor_api: Option<Arc<RwLock<IoHandler>>>,
+        agent_id: AgentId,
+    ) -> Arc<RwLock<IoHandler>> {
+        conductor_api
+            .or(Some(Arc::new(RwLock::new(mock_conductor_api(agent_id)))))
```
maackle:
```suggestion
            .unwrap_or_else(|| Arc::new(RwLock::new(mock_conductor_api(agent_id))))
```

core/src/context.rs
```diff
@@ -121,6 +61,34 @@ impl Context {
         100
     }
 
+    // test_check_conductor_api() is used to inject a conductor_api with a working
+    // mock of agent/sign to be used in tests.
+    // There are two different implementations of this function below which get pulled
+    // in depending on if "test" is in the build config, or not.
+    // This allows unit tests of core to not have to deal with a conductor_api.
+    #[cfg(not(test))]
```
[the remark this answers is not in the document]
maackle: Yeah this is a bit of convenience at the expense of explicitness. But it works.

core/src/context.rs
lucksus: Option::or returns an option. so would need to be
```suggestion
            .or(Some(Arc::new(RwLock::new(mock_conductor_api(agent_id))))).unwrap()
```
Maybe a matter of style, but Isn't `unwrap_or_else()` better here?

lucksus: sorry, misread the change request. yes, your suggestion is much better :)

test_utils/src/mock_signing.rs
```diff
@@ -0,0 +1,99 @@
...
+use jsonrpc_ws_server::jsonrpc_core::{self, types::params::Params, IoHandler};
+use std::{
+    collections::HashMap,
+    sync::{Arc, Mutex},
+};
+
+lazy_static! {
+    pub static ref TEST_AGENT_KEYS: Mutex<HashMap<Address, Arc<Mutex<Keypair>>>>
+        = Mutex::new(HashMap::new());
+}
```
maackle: What do you think about making this a `Mutex<Vec<_>>` Instead of `Mutex<HashMap<_>>`, and then `registered_test_agent` doesn't take a nick and instead appends a new agent with name `"test-agent-N"`, to guarantee that we don't have overlap between tests?

lucksus: I don’t think it’s bad to have overlap in the agent IDs. But getting different agent IDs depending on the ordering of simultaneously running tests would be bad because the ID would not be deterministic and we couldn’t have fixtures for signatures and addresses.

PR #1024, opened 2019-02-16
Attribute as enum (take 2)

maackle:
This is based on #952, which is already approved. However, I had to write #1018 in order to make it possible to merge `develop` into #952 due to the addition of #780. So, since this has some significant changes, I decided to open a new PR.

I invited everybody who already approved #952 to have another read over this.

### Request

Could you suggest better names for the Attributes? Particularly:
```
    CrudLink,
    Link,
    LinkRemove,
    LinkTag(String),
    RemovedLink(String),
```

I'd like to have some sense of what they're for without looking into how they're used. I have little understanding of what the difference is between `Link`, `CrudLink`, and `LinkTag`. All I know is `Link` and `LinkRemove` are a pair, and so are `LinkTag` and `RemovedLink`

- [X] ~I have added a summary of my changes to the changelog~


core_types/src/eav/eavi.rs
```diff
@@ -14,15 +14,106 @@ use eav::{
...
+#[serde(rename_all = "snake_case")]
+pub enum Attribute {
+    CrudStatus,
+    CrudLink,
+    EntryHeader,
+    Link,
+    LinkRemove,
+    LinkTag(String),
+    RemovedLink(String),
+    PendingEntry,
```
maackle: a little foreshadowing here

core/src/network/handler/store.rs
```diff
@@ -25,49 +25,46 @@ pub fn handle_store_entry(dht_data: EntryData, context: Arc<Context>) {
...
-        }
-        LINK_NAME => {
-            context.log("debug/net/handle: HandleStoreMeta: got CRUD LINK. processing...");
-            // FIXME: block_on hold crud_link metadata in DHT?
-        }
-        _ => {}
+    let attr = dht_meta_data.attribute;
+    // @TODO: If network crates will switch to using the `Attribute` enum,
+    // we can match on the enum directly
+    if attr == Attribute::Link.to_string() {
```
maackle: Note to reader: no logic changes here other than `match` changing to `if`, and the `if` conditions

cas_implementations/src/eav/file.rs
```diff
@@ -140,7 +141,13 @@ impl EavFileStorage {
                     pathbuf
                         .iter()
                         .last()
-                        .map(|v| eav_filter.check(v.to_string_lossy().to_string().into()))
+                        .and_then(|v| {
+                            v.to_string_lossy()
+                                .to_string()
+                                .try_into()
+                                .ok()
```
AshantiMutinta: I'm a bit skeptical of using ok() because we end up losing any possible error information but this is just a string conversion so I am fine with it

sphinxc0re: Why do we need this *cascade of conversions* ( 😉) anyway? Isn't there a better way of doing this?

maackle: @AshantiMutinta the error case here (coming from `try_into` would be that the string doesn't correspond to a valid attribute. You're right though this would be a hard thing to catch. Maybe it should log a warning message?

@sphinxc0re it's because we have to go from (off the top of my head) `PathBuf` -> `Cow<&str>` -> `String` -> `Result<Attribute, _>`, if you see a shorter path do suggest it!

maackle:
@Connoropolous is core architecture stuff going into the guidebook? Zome developers might be interested to know how things are organized under the hood, but the reason that we could do this in the first place is because zome developers shouldn't need to know anything about EAV, and don't have access to it, it's all internal stuff for storing core and networking metadata.

Connoropolous:
Yeah, so it's more a question of how the design decisions here, bubble up
to there, in terms of their implications. Or would you say this is not an
area where you see direct pathways between?

Connoropolous:
But also, yes, eventually core architecture stuff is also going into the
guidebook because we want to foster open source contribution

PR #1026, opened 2019-02-17
"Pending" validation status

maackle:
Read and comment on the [RFC](https://hackmd.io/8Qs0RSYET1Cd8nXcfgsa2Q?view)!

Based on #1024.

## Changes in this PR

- Add `ValidationResult::MissingValidation` variant
- Remove redundant `Pass` and `Fail` from `CallbackResult`, collapse
  into `ValidationResult` variant
- Use `Result<ValidationResult>` instead of `CallbackResult` when appropriate


maackle:
This isn't really helpful anymore, since we've specced things out in more detail and this deviates from the newest findings

PR #1029, opened 2019-02-18
WASM panic handler

lucksus:

# Reroute `PanicInfo` to `hdk::debug`
```
2019-02-18 23:04:50:alice: debug/dna: '"assertion failed: `(left == right)`\n  left: `1`,\n right: `2`"'
2019-02-18 23:04:50:alice: debug/dna: '"panic occurred in file 'src/blog.rs' at line 34"'
Exception occurred while calling zome function:  unable to call zome "blog" function "show_env": InternalFailure(RibosomeFailed("WASM invocation failed: Trap: Trap { kind: Unreachable }"))
```
Currently, it's not possible to get the stacktrace from WASMi or the WASM Rust runtime, at least not in the general case.


CHANGELOG.md
```diff
@@ -6,7 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 ### Removed
 ### Added
-### Fixed
+- Adds a panic handler to HDK-Rust and that reroutes infos about panics happening inside the WASM Ribosome to the instances logger (https://github.com/holochain/holochain-rust/pull/1029).
```
zippy:
```suggestion
- Adds a panic handler to HDK-Rust and that reroutes infos about panics happening inside the WASM Ribosome to the instances logger [PR#1029](https://github.com/holochain/holochain-rust/pull/1029).
```

hdk-rust/src/panic_handler.rs
```diff
@@ -0,0 +1,24 @@
...
+        //let _ = debug(RawString::from(format!("{}", info.message().unwrap().clone())));
+        let _ = if let Some(location) = info.location() {
+            debug(RawString::from(format!(
+                "panic occurred in file '{}' at line {}",
+                location.file(),
+                location.line()
+            )))
+        } else {
+            debug(RawString::from(format!(
+                "panic occurred but can't get location information..."
```
zippy: If we know this is going to happen every time, we should just comment this out?

lucksus: It's not going to happen every time. With debug symbols we actually see the file and line number. See above output that I got.

PR #1030, opened 2019-02-19
[Do Not Merge] Fix conductor static server windows

willemolding:
In response to holochain/dev-camp-playground#1

It appears that on windows the static file server does not automatically redirect calls to `/` to `index.html`. Mac/linux does not have this issue.

Adds a test for retrieving a static file from `/`. It is expected that this will pass on the linux CI but fail on windows.


willemolding:
I have managed to narrow down the problem to originating in the rust crate we are using for the static file server. The same problem can be replicated by running their example app on windows

See stephank/hyper-staticfile#12

I will try and fix the problem upstream

willemolding:
I have made a PR that fixes the upstream issue stephank/hyper-staticfile#13. The maintainer of the project seems very active on github so hopefully we can get a bugfix release soon.

In the meantime I have updated the conductor_api Cargo.toml to point at the branch with the fix.

Connoropolous:
closing in favor of #1128

PR #1031, opened 2019-02-19
holochain_net: tests update to fit new assumptions

ddd-mtl:
Rewrote parts of the in-memory server & net tests to fit with the new assumptions / mental model (like removing hold_entry_list_test() because it actually didn't make any sense).

Changed test content to be real json and not a string.

Introducing Cell concept:
Cell = Dna + agentId

Much like DnaInstance, but that name is too long and implies that the DnaInstance is currently running, which might not allways be the case. Example: memory_server tracks data related to a cell, even if it is online or not. It would still be the same cell if it unregisters than reregisters. Reregistering doesn't fit the concept of Instance.


net/src/memory_book.rs
```diff
@@ -0,0 +1,153 @@
...
+
+// Return true if data is in book
+pub fn book_has(
+    cell_book: &CellBook,
+    cell_id: CellId,
+    base_address: &Address,
+    data_address: &Address,
+) -> bool {
+    let maybe_entry_book = cell_book.get(&cell_id);
+    if let None = maybe_entry_book {
```
sphinxc0re: Please use `is_none()` here.

net/src/memory_book.rs
```diff
@@ -0,0 +1,153 @@
...
+        return false;
+    }
+    let entry_book = maybe_entry_book.unwrap();
+    let maybe_meta_set = entry_book.get(base_address);
+    if let None = maybe_meta_set {
```
sphinxc0re: Please use `is_none()` here.

net/src/memory_book.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+///
+pub fn book_has_entry(cell_book: &CellBook, cell_id: CellId, entry_address: &Address) -> bool {
+    book_has(&cell_book, cell_id, entry_address, entry_address)
+}
+
+/// Remove an address from a book
+/// Return true if address exists and has been successfully removed.
+pub(crate) fn _unbookkeep_address(
```
sphinxc0re: So, there is absolutely no convention in Rust for functions with underscores as a prefix. We do not implement this convention anywhere else in the codebase. Please remove it.

net/src/memory_book.rs
```diff
@@ -0,0 +1,153 @@
...
+/// return a unique identifier out of an entry_address and attribute
+pub(crate) fn into_meta_id(meta_tuple: &MetaTuple) -> MetaId {
+    HashString::from(format!(
+        "{}||{}||{}",
+        meta_tuple.0, meta_tuple.1, meta_tuple.2
+    ))
+}
+
+/// unmerge meta_id into a tuple
+pub(crate) fn _undo_meta_id(meta_id: &MetaId) -> MetaTuple {
```
sphinxc0re: See the other comment about underscore prefixes.

net/src/memory_book.rs
lucksus:
```suggestion
    if maybe_entry_book.is_none() {
```

net/src/memory_book.rs
lucksus:
```suggestion
    if maybe_meta_set.is_none() {
```

net/src/memory_book.rs
lucksus: Isn't that the convention for unused variables/functions? This function is unused, so the underscore omits the warning, right? So the question is, if it is unused, can we remove it?

sphinxc0re: Yup. If it's unused, we should remove it

sphinxc0re: My point it, that you shouldn't omit the warning but act upon it

ddd-mtl: I'm going to need those function in the future. Not going to remove them.

PR #1036, opened 2019-02-20
Package name cleanup

zippy:

- [x] `bundle.json` & `.hcpkg` unified to `YOUR_DNA_NAME.dna.json`
- [x] `.build` files renamed to `.hcbuild`
- [x] `hc package` now builds to `dist` directory by default, to match how `hc test` works


cli/src/cli/js-tests-scaffold/index.js
```diff
@@ -3,7 +3,7 @@
 const { Config, Scenario } = require("@holochain/holochain-nodejs")
 Scenario.setTape(require("tape"))
 
-const dnaPath = "./dist/bundle.json"
+const dnaPath = "./dist/YOUR_DNA_NAME.dna.json"
```
sphinxc0re: Is that right?

conductor_api/src/conductor/base.rs
```diff
@@ -657,7 +658,7 @@ impl Conductor {
 
     pub fn save_dna(&self, dna: &Dna) -> Result<PathBuf, HolochainError> {
         let mut file_path = self.dna_dir_path().join(dna.address().to_string());
```
sphinxc0re:
```suggestion
        let file_path = self.dna_dir_path().join(dna.address().to_string())
            .with_file_name(DNA_EXTENSION);
```

conductor_api/src/conductor/base.rs
```diff
@@ -657,7 +658,7 @@ impl Conductor {
...
-        file_path.set_extension("hcpkg");
+        file_path.set_extension(DNA_EXTENSION);
```
sphinxc0re:
```suggestion
```

cli/src/cli/js-tests-scaffold/index.js
zippy: Well, actually we need some automation to edit that file on the fly during `init`.  I'll work on that

conductor_api/src/conductor/base.rs
zippy: Will that really work?  I think not because `set_extension` will add the "." between the file name and the path, but `with_file_name` I don't think will.

sphinxc0re: 🙈 Oh, sorry,... I meant:
```suggestion
        let mut file_path = self.dna_dir_path().join(dna.address().to_string())
            .with_extension(DNA_EXTENSION);
```

cli/src/cli/init.rs
```diff
@@ -75,7 +102,7 @@ pub fn init(path: &PathBuf) -> DefaultResult<()> {
     let ignores = [
         &DIST_DIR_NAME,
         &TEST_DIR_NAME,
-        &DEFAULT_BUNDLE_FILE_NAME,
+        format!("*.{}", DNA_EXTENSION).as_str(),
```
Connoropolous: you'll only need this here, if it's still going to be building the file in the root of the app, instead of in the DIST folder

cli/src/cli/js-tests-scaffold/index.js
zippy: done

zippy:
I'm not worried about the .build rename because we are in alpha, not backward compatibility!!

cli/src/cli/init.rs
zippy: fixed.

Connoropolous:
fair enough

PR #1038, opened 2019-02-20
Make p2p bindings available in conductor API

ddd-mtl:
Added p2p_bindings() getter in conductor API


cli/src/cli/run.rs
```diff
@@ -32,6 +32,11 @@ pub fn run(
         .load_config()
         .map_err(|err| format_err!("{}", err))?;
 
+    println!(
```
Connoropolous: Won't this be a weird message if p2p_bindings is None?

ddd-mtl: Yeah, kind of. Us lazy devs are used to weird messages like that :)

Connoropolous: Could you make it conditional?

Connoropolous: Us obsessive documenters go nuts over messages like that :p

Connoropolous: You can use [ci skip] and [skip travis] in the commit message to prevent rebuilding

ddd-mtl: i'll just erase it

PR #1039, opened 2019-02-21
Signals 1 additions

lucksus:

@maackle, these are my changes to #982. Please merge into your branch at will. This is not 100% finished. We need to shutdown and respawn the signal thread if the config gets changed..


conductor_api/src/conductor/base.rs
```diff
@@ -438,9 +449,12 @@ impl Conductor {
                 context_builder = context_builder.with_p2p_config(self.instance_p2p_config());
 
                 // Signal config:
-                self.setup_signals(signal_tx.clone());
-                context_builder = context_builder
-                    .with_signals(self.signal_tx.clone().expect("Signal sender not set up"));
+                let (sender, receiver) = unbounded();
```
maackle: Why the switch to crossbeam channels?

conductor_api/src/conductor/base.rs
```diff
@@ -131,48 +128,66 @@ impl Conductor {
...
+                                let broadcasters = broadcasters.read().unwrap();
+                                let interfaces_with_instance: Vec<&InterfaceConfiguration> = config
+                                    .interfaces
+                                    .iter()
+                                    .filter(|interface_config| {
+                                        interface_config
+                                            .instances
+                                            .iter()
+                                            .find(|instance| instance.id == *instance_id)
+                                            .is_some()
```
maackle: yes, this is the key piece

conductor_api/src/conductor/base.rs
```diff
@@ -897,16 +911,17 @@ pub mod tests {
...
     }
 
     fn test_conductor_with_signals(signal_tx: SignalSender) -> Conductor {
         let config = load_configuration::<Configuration>(&test_toml()).unwrap();
         let mut conductor = Conductor::from_config(config.clone());
+        conductor.set_signal_sender(signal_tx);
         conductor.dna_loader = test_dna_loader();
         conductor.key_loader = test_key_loader();
-        conductor.load_config(Some(signal_tx)).unwrap();
+        conductor.load_config().unwrap();
```
maackle: I kinda liked it having load_config take the Option<Sender>, so that way you have to declare whether you want it or not, instead of maybe forgetting it. What do you think?

conductor_api/src/conductor/base.rs
lucksus: Because mpsc Receivers are mit Send, which would turn the whole Conductor struct not Send but is needed for admin interfaces and the lazy_static.

conductor_api/src/conductor/base.rs
lucksus: But most places don’t need to extract all signals from all instances. It’s only the waiter that does. And it’s completely safe to not set it when it’s not needed.

lucksus:
Hey @maackle, can we merge this or do you want to see changes first? I would basically merge this one into signals-1 and would deem signals-1 then ready for merge into develop. Would you agree? (Want to clean up our PR list a bit..)

conductor_api/src/conductor/base.rs
maackle: That was confusing at first because I read "mit" as "with" and not "not" ;) Gotcha though

PR #1041, opened 2019-02-21
Small improvements

sphinxc0re:

The only place where tests fail because if this is in the tests of the c bindings. Which brings up the question: Do we still need the bindings or can we drop them for the WebSocket API?


sphinxc0re:
I think the hashing itself is a problem. How should I do this? The hash of a DNA is changing when commiting a change

sphinxc0re:
I could ignore the git_hash in the hash generation process

PR #1043, opened 2019-02-21
Holo client-side signing

lucksus:
*-> No, this feature is meant only for Holo*

# Context
Enabling outsourced signing so agent keys can live client side while the conductor is running on the HoloPort. (Bootsrapping interims hack until we have a WASM build of core and conductor that can run in the browser)
screenshot 2019-02-21 at 17 22 51

## Usage
The conductor has two new config options:
* `signing_service_uri = "http://localhost:8888"` on root level
* `holo_remote_key = true` in the agent config

If both are set, the conductor will ignore `key_file` for the agent for which `holo_remote_key` is set and instead route signing request to the signing service by HTTP posting a JSON object like this:
```json
{
  "agent_id": "HoloTester1-----------------------------------------------------------------------AAACZp4xHB",
  "payload: "xxxx"
}
```
and expecting the base64 encoded signature string as plain text response.

There is an example config in conductor/example-config/holo.toml that can be run with:
```
cd conductor
cargo run -- -c example-config/holo.toml
```
(make sure app_spec was build before since the example is trying to load that DNA from `../app_spec/dist/app_spec.dna.json`)

## Follow up
If you try running this without anything listening on localhost:8888 you'll see core panic as it's currently still expecting signing to work.


conductor/example-config/holo.toml
```diff
@@ -0,0 +1,65 @@
+bridges = []
+signing_service_uri = "http://localhost:8888"
+
+[[agents]]
+id = "test agent 1"
+key_file = "holo_tester.key"
+name = "Holo Tester 1"
+public_address = "HoloTester1-----------------------------------------------------------------------AAACZp4xHB"
+holo_remote_key = true
```
maackle: Remote keys could potentially be a use case outside of Holo. Does it make sense to start thinking about it as such, and not use 'holo' in the name? Also isn't key_file irrelevant now?

conductor_api/src/config.rs
```diff
@@ -304,6 +310,9 @@ pub struct AgentConfiguration {
     pub name: String,
     pub public_address: String,
     pub key_file: String,
+    /// If set to true conductor will ignore key_file and instead use the remote signer
+    /// accessible through signing_service_uri to request signatures.
+    pub holo_remote_key: Option<bool>,
```
maackle: This is a perfect job for an enum!

```rust

    // ...
    pub key: KeySource
    // ...

enum KeySource = { 
    KeyFile(PathBuf), 
    RemoteKey(String), // then we can put the signing URL directly into the enum 
}
```

conductor_api/src/interface.rs
```diff
@@ -767,6 +771,29 @@ impl ConductorApiBuilder {
...
+        let agent_id = agent_id.clone();
+        let signing_service_uri = signing_service_uri.clone();
+
+        self.io.add_method("agent/sign", move |params| {
+            let params_map = Self::unwrap_params_map(params)?;
+            let payload = Self::get_as_string("payload", &params_map)?;
+
+            let signature = request_signing_service(&agent_id, &payload, &signing_service_uri)
+                .map_err(|holochain_error| {
+                    println!("Error in signing hack: {:?}", holochain_error);
```
maackle: Maybe the actual output shouldn't mention this is a hack ;)

conductor_api/src/config.rs
maackle: That is, only if we're going to keep the remote key functionality after the holo hack goes away.

lucksus:
@maackle, I see your requests to change to an enum and not mention "hack" in the output be connected to the high-level question of do we regard this as a temporary hack or something long-term.

I see it as a hack and would rather name it so and don't have the code (nor the error output) mask that fact.

Even Holo won't need this anymore as soon as we have a running WASM build of Holochain so that the whole creation of entries runs client-side. So I would like to get rid of this code then.

maackle:
Yeah @lucksus OK, I agree on both points. It's a hack, and let's not pretend it's not. The way you wrote it will make it easy to rip out later.

neonphog:
+1 treating it as a hack, ultimately we want to limit things that request signatures, because they are attack vectors for fraud

conductor_api/src/holo_signing_service.rs
```diff
@@ -0,0 +1,33 @@
+use boolinator::Boolinator;
+use holochain_core_types::{
+    agent::AgentId, cas::content::AddressableContent, error::HolochainError,
+};
+
+pub fn request_signing_service(
+    agent_id: &AgentId,
+    payload: &String,
```
neonphog: do we want this to be a string? I feel like it's more appropriate to sign a byte array - or rather, it *has* to be, so do we do a utf8 conversion on the other side?

lucksus: I was using Strings for the payload elsewhere before. And currently we only sign entry addresses which are represented as Strings in our code case as well. So I wouldn’t like to introduce a change just here...

PR #1045, opened 2019-02-21
hc run dna path was wrong, and the function untestable

Connoropolous:

### DIFF
--- a/cli/src/cli/run.rs
+++ b/cli/src/cli/run.rs
@@ -11,13 +11,6 @@ use holochain_conductor_api::{
 use holochain_core_types::agent::{AgentId, KeyBuffer};
 use std::{fs, path::PathBuf};
 
-const LOCAL_STORAGE_PATH: &str = ".hc";
-
-const AGENT_CONFIG_ID: &str = "hc-run-agent";
-const DNA_CONFIG_ID: &str = "hc-run-dna";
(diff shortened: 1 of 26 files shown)


Connoropolous:
@zippy was it you who addded "fish" into the expects, what should be there?

zippy:
Oh shoot.  I forgot to clean that up!

Connoropolous:
can you push a better error message, that could totally end up in a devs command line :p ?

zippy:
ok passing the error through now instead...

Connoropolous:
oh, hehe, I think you read that late... I meant in response to "fish"

Connoropolous:
@lucksus can you review?

cli/src/cli/run.rs
```diff
@@ -11,95 +11,181 @@ use holochain_conductor_api::{
...
+            .expect("Expected DNA path to be valid unicode")
+            .to_string(),
         hash: None,
-    };
+    }
+}
 
-    let storage = if persist {
+// STORAGE
+const LOCAL_STORAGE_PATH: &str = ".hc";
```
lucksus: Isn't this something that should either go to `holochain_common` or just use the path from there?

Connoropolous: I think if anything should change, it’s the folder name itself... it’s a bit different than the others because this directory sits inside the NDA folder itself

PR #1048, opened 2019-02-22
Crud status network

AshantiMutinta:
- [x] Changed CrudStatus Action creators to use author entry
- [x] Handled CrudStatus and CrudLink data sent over network


core/src/workflows/crud_status.rs
```diff
@@ -0,0 +1,163 @@
+use crate::context::Context;
```
lucksus: The functions in this file are not workflows in the sense of the other workflows in this directory. They don’t aggregate async action creators to a high-level process. 

Actually these functions are misplaced reducers.

core/src/workflows/author_update_entry.rs
```diff
@@ -0,0 +1,81 @@
...
+    cas::content::{Address, AddressableContent},
+    entry::Entry,
+    error::HolochainError,
+    validation::{EntryAction, EntryLifecycle, ValidationData},
+};
+use std::sync::Arc;
+
+pub async fn author_update_entry<'a>(
+    entry: &'a Entry,
+    maybe_link_update_delete: Option<Address>,
```
zippy: I think for this function, the `link_update_delete` shouldn't be an Option.  I don't think you should be able to call author_update_entry without actually specifying which other entry is being updated.  It does mean you'll have to add the `Some()` around the link in the call to `commit` in step 3

hdk-rust/wasm-test/src/handle_crud.rs
```diff
@@ -6,11 +6,16 @@ use holochain_wasm_utils::{
...
+use std::{
+    time::Duration
+};
+
 //
 pub(crate) fn handle_update_entry_ok() -> JsonString {
     // Commit v1 entry
     hdk::debug("**** Commit v1 entry").ok();
     let res = hdk::commit_entry(&hdk_test_entry());
+    hdk::sleep(Duration::from_millis(35000)).unwrap();
```
zippy: whoa!!  Do you really need to wait 35 seconds for this test to work?  All these sleeps will dramatically increase the wall time on CI.  Can we decrease this?

core/src/dht/dht_reducers.rs
```diff
@@ -295,6 +297,7 @@ fn reduce_remove_entry_inner(
         return Err(err);
     }
     let status_eavs = maybe_status_eav.unwrap();
+
     assert!(!status_eavs.is_empty(), "Entry should have a Status");
```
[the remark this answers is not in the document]
lucksus: Hm, yeah. I guess this should be part of the sub-conscious validation of removals, right?

core/src/network/handler/store.rs
```diff
@@ -60,11 +61,43 @@ pub fn handle_store_meta(dht_meta_data: DhtMetaData, context: Arc<Context>) {
...
-    } else if attr == Attribute::CrudStatus.to_string() {
-        context.log("debug/net/handle: HandleStoreMeta: got CRUD status. processing...");
-    // FIXME: block_on hold crud_status metadata in DHT?
-    } else if attr == Attribute::CrudLink.to_string() {
+    } else if CrudStatus::from_str(&attr)
+        .expect("Could not convert deleted attribute to CrudStatus")
+        == CrudStatus::Deleted
+    {
+        context.log("debug/net/handle: HandleStoreMeta: got CRUD STATUS. processing...");
+        // FIXME: block_on hold crud_status metadata in DHT?
```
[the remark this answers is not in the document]
lucksus: yes.

core/src/network/handler/store.rs
```diff
@@ -60,11 +61,43 @@ pub fn handle_store_meta(dht_meta_data: DhtMetaData, context: Arc<Context>) {
...
+            {
+                context.log(format!("err/net/dht: {}", error))
+            }
+        });
+    } else if CrudStatus::from_str(&attr)
+        .expect("Could not convert modified attribute to CrudStatus")
+        == CrudStatus::Modified
+    {
         context.log("debug/net/handle: HandleStoreMeta: got CRUD LINK. processing...");
         // FIXME: block_on hold crud_link metadata in DHT?
```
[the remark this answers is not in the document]
lucksus: +1

core/src/workflows/author_update_entry.rs
lucksus: Yes, I agree. I am also wondering why we need a separate workflow. Can we not reuse `author_entry`?

Looking at this I notice that the only thing different is `EntryAction::Modify` in the validation data, which is correct. But since the `remove_entry` action also calls `author_entry` the action in the validation data will be wrong in the deletion case..

core/src/workflows/hold_entry_remove.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    // 1. Get validation package from source
+    let maybe_validation_package = await!(get_validation_package(header.clone(), &context))?;
+    let validation_package = maybe_validation_package
+        .ok_or("Could not get validation package from source".to_string())?;
+
+    // 2. Create validation data struct
+    let validation_data = ValidationData {
+        package: validation_package,
+        lifecycle: EntryLifecycle::Dht,
```
lucksus: This has to be meta:
```suggestion
        lifecycle: EntryLifecycle::Meta,
```

...which reminds me that we do need set the action right when the lifecycle actually is `Dht`, which is the case for storing the deletion entry, which happens in the regular `hold_entry_workflow`. I guess we need to test if the entry is a deletion entry, or if it is an update entry, and set the action there accordingly.

core/src/workflows/hold_entry_update.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    // 1. Get validation package from source
+    let maybe_validation_package = await!(get_validation_package(header.clone(), &context))?;
+    let validation_package = maybe_validation_package
+        .ok_or("Could not get validation package from source".to_string())?;
+
+    // 2. Create validation data struct
+    let validation_data = ValidationData {
+        package: validation_package,
+        lifecycle: EntryLifecycle::Dht,
```
lucksus: Same here:
```suggestion
        lifecycle: EntryLifecycle::Meta,
```

hdk-rust/wasm-test/src/handle_crud.rs
lucksus: Yeah, we need too find a different solution here, I think...

core/src/dht/dht_reducers.rs
```diff
@@ -316,14 +314,17 @@ fn reduce_remove_entry_inner(
     }
     let new_status_eav = result.expect("should unwrap eav");
     let meta_storage = &new_store.meta_storage().clone();
+
     let res = (*meta_storage.write().unwrap()).add_eavi(&new_status_eav);
```
sphinxc0re:
```suggestion
     let res = (*meta_storage.write().unwrap()).add_eavi(&new_status_eav)?;
```

core/src/dht/dht_reducers.rs
```diff
@@ -316,14 +314,17 @@ fn reduce_remove_entry_inner(
...
     if let Err(err) = res {
```
sphinxc0re:
```suggestion
```

core/src/dht/dht_reducers.rs
```diff
@@ -316,14 +314,17 @@ fn reduce_remove_entry_inner(
...
         return Err(err);
```
sphinxc0re:
```suggestion
```

core/src/dht/dht_reducers.rs
```diff
@@ -316,14 +314,17 @@ fn reduce_remove_entry_inner(
...
     }
```
sphinxc0re:
```suggestion
```

core_types/src/entry/entry_type.rs
```diff
@@ -85,6 +85,7 @@ impl EntryType {
     pub fn is_app(&self) -> bool {
         match self {
             EntryType::App(_) => true,
+            EntryType::Deletion => true,
```
lucksus: ?
Why this? A deletion entry is a "system" entry type, not an app entry..

AshantiMutinta: Woops, This stayed form me debugging. Will change that asap

PR #1049, opened 2019-02-24
holochain_dpki: refactor for using HCID

ddd-mtl:
## context reading

https://medium.com/holochain/part-1-holochain-holo-accounts-and-cryptographic-key-management-8b6306555e09

## core_types: AgentId
Removed KeyBuffer and use HCID crate instead.

**Base32**: new type definition of a String to represent a Base32 encoded public key with HCID.
AgentId now stores the hcid encoded key.
Added method  `AgentId.decoded_key()`

#### KeyPair
**KeyPair** is a Trait that represents an asymetric cryptographic key pair.
Trait is expressed with:
 - **SigningKeyPair**
 - **EncryptingKeyPair**

#### KeyBundle
Struct holding both the signing keys and encrypting keys generated from a seed. It also holds the seed type.

#### SeedType
An enum for specifying the type of seed used for creating the KeyBundle (Root, Revocation, Device, etc).

#### KeyBlob
A passphrase encrypted KeyBundle with some metadata.
Intended to be stored for persistence.

### Other
Removed DpkiError since it was not used.

## Sodium crate
OutOfMemory error according to libSodium api.
Enhanced a bit SecBuf with methods:
 - from_array() 
 - randomize()
 - is_zero()
 - compare()

## Misc.

Added unit tests here and there.


sodium/src/secbuf.rs
```diff
@@ -232,16 +232,48 @@ impl SecBuf {
...
         }
+        Ok(())
+    }
+
+    // -- Helpers -- //
+
+    /// Check if the buffer is empty i.e. [0,0,0,0,0,0,0,0]
+    pub fn is_empty(&mut self) -> bool {
+        let buf = self.read_lock();
+        for i in 0..buf.len() {
```
neonphog: If we're going to implement utility functions like this, I feel we should defer to the sodium utilities : https://download.libsodium.org/doc/helpers#testing-for-all-zeros

sodium/src/secbuf.rs
```diff
@@ -232,16 +232,48 @@ impl SecBuf {
...
+            buf[x] = data[x];
+        }
         Ok(())
     }
+
+    /// Return true if both buffers hold the same data
+    pub fn is_same(&mut self, other: &mut Self) -> bool {
+        let a = self.read_lock();
+        let b = other.read_lock();
+        **a == **b
```
neonphog: Feel like this should be implemented using PartialEq

ddd-mtl: Tried but PartialEq doesn't work with `&mut` args :/

sodium/src/secbuf.rs
```diff
@@ -232,16 +232,48 @@ impl SecBuf {
...
+    /// Load the [u8] into the SecBuf
+    pub fn from_array(&mut self, data: &[u8]) -> Result<(), SodiumError> {
+        if (data.len() != self.len()) {
+            return Err(SodiumError::Generic(
+                "Input does not have same size as SecBuf".to_string(),
+            ));
+        }
+        let mut buf = self.write_lock();
+        for x in 0..data.len() {
+            buf[x] = data[x];
```
neonphog: should not do this with iteration... please see the `write` function on secbuf. In fact, I don't think we need this function... just use `buf.write(0, &[ /* ... */])`

sodium/src/secbuf.rs
neonphog: huh... in that case, maybe we could make use of https://download.libsodium.org/doc/helpers#comparing-large-numbers ? because we have need of that as well.

sodium/src/pwhash.rs
```diff
@@ -54,23 +54,25 @@ pub fn hash(
             ops_limit as libc::c_ulonglong,
             mem_limit,
             alg as libc::c_int,
-        );
+        )
```
neonphog: from the docs
> The function returns 0 on success, and -1 if the computation didn't complete, usually because the operating system refused to allocate the amount of requested memory.

and the [sys bindings](https://docs.rs/rust_sodium-sys/0.10.2/rust_sodium_sys/fn.crypto_pwhash.html) look like they return that integer... maybe we should be grabbing it to make sure

hdk-rust/tests/integration_test.rs
```diff
@@ -306,7 +306,7 @@ fn can_use_globals() {
     assert_eq!(
         result.clone(),
         Ok(JsonString::from(HashString::from(
-            "zDTy3nif1_9khZUc9fP304PFdV-53yLTI_bVinjvZhwFkd3EMb7P4nXu8LRtMun5g3uR2ZjY_p5h54P_T_8KFog0SnQ2"
+            "HcSCJUBV8mqhsh8y97TIMFi68Y39qv6dzw4W9pP9Emjth7xwsj6P83R6RkBXqsa"
```
neonphog: bwahaha - yay real keys :+1:

hc_dpki/src/utils.rs
```diff
@@ -0,0 +1,93 @@
...
+/// @return {SecBuf} Resulting decoded key
+pub(crate) fn decode_pub_key(pub_key_b32: Base32, codec: &HcidEncoding) -> HcResult<SecBuf> {
+    // Decode Base32 public key
+    let pub_key = codec.decode(&pub_key_b32)?;
+    // convert to SecBuf
+    let mut pub_key_sec = SecBuf::with_insecure(sign::PUBLICKEYBYTES);
+    {
+        let mut pub_key_lock = pub_key_sec.write_lock();
+        for x in 0..pub_key.len() {
+            pub_key_lock[x] = pub_key[x];
```
neonphog: We shouldn't iterate on bytes... it's slow.

```rust
let mut pub_key_sec = SecBuf::with_insecure(sign::PUBLICKEYBYTES)
pub_key_sec.write(&codec.decode(&pub_key_b32)?);
```

hc_dpki/src/utils.rs
```diff
@@ -0,0 +1,93 @@
...
+
+/// Encode with HCID a public key given as a SecBuf
+/// @param {SecBuf} pub_key_sec - Public signing key to encode
+/// @param {HcidEncoding} codec - The configured HCID encoder to use
+/// @return {Base32} Resulting HCID encoded key
+pub(crate) fn encode_pub_key(pub_key_sec: &mut SecBuf, codec: &HcidEncoding) -> HcResult<Base32> {
+    let locker = pub_key_sec.read_lock();
+    let pub_buf = &*locker;
+    let pub_key = array_ref![pub_buf, 0, SEED_SIZE];
+    Ok(codec.encode(pub_key)?)
```
neonphog: Curious why the `array_ref!`? a direct reference, should work, or if not, a slice
```rust
Ok(codec.encode(&locker[0..SEED_SIZE]))
```

hc_dpki/src/utils.rs
```diff
@@ -0,0 +1,93 @@
...
+/// @param {SecBuf} signature - Candidate signature for that data buffer
+/// @return true if verification succeeded
+pub fn verify(
+    pub_sign_key_b32: Base32,
+    data: &mut SecBuf,
+    signature: &mut SecBuf,
+) -> HcResult<bool> {
+    let mut pub_key = decode_pub_key(
+        pub_sign_key_b32,
+        &with_hcs0().expect("HCID failed miserably with_hcs0."),
```
neonphog: lol.

there is some overhead to creating the encodings... I'd recommend putting it in a lazy_static!

hc_dpki/src/keypair.rs
```diff
@@ -1,610 +1,242 @@
...
-        let bundle_data_serialized = json::encode(&password_encrypted).unwrap();
+impl KeyPair for SigningKeyPair {
+    fn public(&self) -> String {
+        self.public.clone()
+    }
+    fn private(&mut self) -> &mut SecBuf {
+        &mut self.private
+    }
+    fn codec() -> HcidEncoding {
+        with_hcs0().expect("HCID failed miserably with_hcs0.")
```
neonphog: lazy_static

hc_dpki/src/keypair.rs
```diff
@@ -1,610 +1,242 @@
...
-
-        util::decode_id(pub_keys, &mut sign_pub, &mut enc_pub)?;
-        let verified: i32 = sign::verify(signature, data, &mut sign_pub);
-        Ok(verified)
+    /// @param {SecBuf} signature
+    /// @return true if verification succeeded
+    pub fn verify(&mut self, data: &mut SecBuf, signature: &mut SecBuf) -> bool {
+        let mut pub_key = self.decode_pub_key_into_secbuf();
+        let res = holochain_sodium::sign::verify(signature, data, &mut pub_key);
+        res == 0
```
neonphog: can we make the base `holochain_sodium::sign::verify` return the boolean?

hc_dpki/src/keypair.rs
```diff
@@ -1,610 +1,242 @@
...
-        let mut seed = SecBuf::with_insecure(SEEDSIZE);
-        random_secbuf(&mut seed);
-        let mut keypair = Keypair::new_from_seed(&mut seed).unwrap();
-
-        let mut message = SecBuf::with_insecure(16);
-        random_secbuf(&mut message);
-
-        let mut message_signed = SecBuf::with_insecure(SIGNATURESIZE);
+    fn codec() -> HcidEncoding {
+        with_hck0().expect("HCID failed miserably with_hck0.")
```
neonphog: lazy_static

hc_dpki/src/key_blob.rs
```diff
@@ -0,0 +1,178 @@
...
+        let blob_decoded = base64::decode(&blob.data)?;
+
+        // Deserialize
+        let blob_string = str::from_utf8(&blob_decoded).unwrap();
+        let data: EncryptedData = json::decode(&blob_string).unwrap();
+        // Decrypt
+        let mut decrypted_data = SecBuf::with_secure(BLOB_DATA_LEN);
+        password_encryption::pw_dec(&data, passphrase, &mut decrypted_data, config)?;
+
+        let mut pub_sign = SecBuf::with_secure(sign::PUBLICKEYBYTES);
```
neonphog: use insecure for pubkey

hc_dpki/src/key_blob.rs
```diff
@@ -0,0 +1,178 @@
...
+        let mut pub_enc = SecBuf::with_secure(kx::PUBLICKEYBYTES);
```
neonphog: insecure for pubkey

core_types/src/agent.rs
```diff
@@ -1,129 +1,64 @@
...
+/// This data struct is meant be stored in the CAS and source-chain.
+/// Its key is the public signing key, and is also used as its address.
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, DefaultJson)]
 pub struct AgentId {
     /// a nickname for referencing this agent
     pub nick: String,
-    /// the base64url encoded public identity string for this agent
-    pub key: String,
+    /// the encoded public signing key of this agent (the magnifier)
+    pub pub_sign_key: Base32,
```
neonphog: can we go ahead and add the `pub_enc_key` here too? or would you rather do that at a later time?

core_types/src/agent.rs
```diff
@@ -1,129 +1,64 @@
...
-        while buf.len() < 82 {
-            buf.push_str("+");
-        }
-        buf.push_str("AAAA");
-        let buf = base64::decode(&buf)
-            .expect("could not decode the generated fake base64 string - use the base64 alphabet");
-        let buf = KeyBuffer::with_raw(array_ref![buf, 0, KeyBuffer::KEY_LEN]);
-        AgentId::new(nick, &buf)
+        let mut key: [u8; 32] = [0; 32];
+        key[0] = 42;
```
neonphog: Would it make sense to have a well-known testing keypair (i.e. the priv-key is published in the sourcecode with warnings not to really use it?)

sodium/src/secbuf.rs
ddd-mtl: Done in secbuf_util.rs

sodium/src/secbuf.rs
ddd-mtl: Just discovered that sodium_compare() was already implemented in util.rs
I reworked util.rs to secbuf_util.rs and maybe methods instead.
Moved all helpers there.

hc_dpki/src/utils.rs
ddd-mtl: Roger that.

hc_dpki/src/keypair.rs
ddd-mtl: Ok, but shouldn't HCID crate provide those?

hc_dpki/src/utils.rs
ddd-mtl: No reason, just copy-paste ^^'''

hc_dpki/src/keypair.rs
ddd-mtl: Done

core_types/src/agent.rs
ddd-mtl: Better to do that later, since enc doesn't work yet.
I'll add a TODO though.

core_types/src/agent.rs
ddd-mtl: Probably. Changed it to a const within AgentId.

sodium/src/pwhash.rs
ddd-mtl: Ok. I'll add an unreachable!

ddd-mtl:
@neonphog Thanks for the review. I made all the changes.
There is still some `with_hcs0()` used in `agent.rs` because I put the lazy statics in `hc_dpki` and I don't want a circular dependency, and I also don't want `holochain_core_types` to be dependant to `holochain_sodium`.
If its an issue, we could move the decoding methods in a different crate in a later PR.

hc_dpki/src/key_blob.rs
```diff
@@ -0,0 +1,178 @@
...
+        blob: &KeyBlob,
+        passphrase: &mut SecBuf,
+        config: Option<PwHashConfig>,
+    ) -> HcResult<KeyBundle> {
+        // decoding the blob.data of type EncryptedData
+        let blob_decoded = base64::decode(&blob.data)?;
+
+        // Deserialize
+        let blob_string = str::from_utf8(&blob_decoded).unwrap();
+        let data: EncryptedData = json::decode(&blob_string).unwrap();
```
thedavidmeister: lets get this deserialize logic behind a `?` in case bad input

hc_dpki/src/key_blob.rs
```diff
@@ -0,0 +1,178 @@
...
+        data_buf
+            .write(offset, &**self.enc_keys.private.read_lock())
+            .expect("Failed blobbing private encoding key");
+        offset += kx::SECRETKEYBYTES;
+        assert_eq!(offset, BLOB_DATA_LEN_MISALIGN);
+
+        // encrypt buffer
+        let encrypted_blob = password_encryption::pw_enc(&mut data_buf, passphrase, config)?;
+        let serialized_blob = json::encode(&encrypted_blob).expect("Failed to serialize KeyBundle");
+        // conver to base64
```
thedavidmeister: convert typo

hc_dpki/src/key_blob.rs
ddd-mtl: Fixed and changed to serde_json

sodium/src/secbuf_util.rs
```diff
@@ -0,0 +1,141 @@
...
+    /// Compares two SecBuf
+    /// Return :
+    /// | if a > b; return 1
+    /// | if a < b; return -1
+    /// | if a == b; return 0
+    pub fn compare(&mut self, b: &mut SecBuf) -> i32 {
+        check_init();
+        let mut a = self.write_lock();
+        let mut b = b.write_lock();
+        unsafe { rust_sodium_sys::sodium_compare(raw_ptr_char!(a), raw_ptr_char!(b), a.len()) }
```
neonphog: :+1:

hc_dpki/src/keypair.rs
neonphog: indeed, that might make the most sense. I have some vague unfounded worry that lazy_static and wasm compilation don't work well together... I wonder where that came from... but it would totally be worth trying.

PR #1050, opened 2019-02-25
Added zome integrity check

sphinxc0re:

### DIFF
--- a/conductor_api/src/holochain.rs
+++ b/conductor_api/src/holochain.rs
@@ -122,7 +122,9 @@ impl Holochain {
 
             if let Ok(json_string) = maybe_json_string {
                 if json_string != holochain_core_types::GIT_HASH.into() {
-                    return Err(HolochainError::ErrorGeneric("HALP!".into()).into());
+                    eprintln!("WARNING! The git-hash of the runtime and the zome don't match.");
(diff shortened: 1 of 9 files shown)


conductor_api/src/holochain.rs
```diff
@@ -108,6 +112,21 @@ impl Holochain {
     /// create a new Holochain instance
     pub fn new(dna: Dna, context: Arc<Context>) -> HolochainResult<Self> {
         let instance = Instance::new(context.clone());
+
+        for zome in dna.zomes.values() {
```
zippy: I think this needs to be broken out into it's own function for use by conductors, not as part of `new()` function.

This way makes it impossible to instantiate a instance unless the git hashes match.  I think this actually has to be configuration call, i.e. maybe there's a "strict" mode.  For now.  I think we should simply have our conductors spit out a bit warning if the hashes don't match.

sphinxc0re: To get an understanding of what is needed here: How and where would such a function be called? Where should it go? Should I define it on `Zome`?

lucksus: Yes. Or: instead of return an error below we just `println!` the version mismatch for now.

sphinxc0re:
@lucksus sure, I can just `eprintln!` an error message, which prints it to `stderr`

PR #1054, opened 2019-02-26
Pending Link Validation

lucksus:

screenshot from 2019-02-27 00-32-34

## Preparation
* Added `clokwerk` and created a module for scheduled jobs
* Big refactor of the validation code. Validation lives now under `core/nucleus/validation` and uses only one new result type (no `CallbackResult` anymore):
```rust
pub enum ValidationError {
    Fail(String),
    UnresolvedDependencies(Vec<Address>),
    NotImplemented,
    Error(String),
}
pub type ValidationResult = Result<(), ValidationError>;
```

## Solution
`hold_link_workflow` makes use of this new return value and signals a pending validation by dispatching an action:
```rust
    // 3. Validate the entry
    context.log(format!("debug/workflow/hold_link: validate..."));
    await!(validate_entry(entry.clone(), validation_data, &context)).map_err(|err| {
        context.log(format!("debug/workflow/hold_link: invalid! {:?}", err));
        if let ValidationError::UnresolvedDependencies(dependencies) = &err {
            dispatch_action(
                context.action_channel(),
                ActionWrapper::new(Action::AddPendingValidation(Box::new((
                    entry_with_header.to_owned(),
                    dependencies.clone(),
                )))),
            );
        }
        err
    })?;
```
which results in the `NucleusState` adding the boxed entry to `pending_validations`.

`run_pending_validations` (in core/scheduled_jobs/pending_validations.rs) picks this up and re-runs the `hold_link_workflow`:

```rust
fn retry_validation(pending: PendingValidation, context: Arc<Context>) {
    thread::spawn(move || context.block_on(hold_link_workflow(&pending.0, &context)));
}

pub fn run_pending_validations(context: Arc<Context>) {
    context
        .state()
        .unwrap()
        .nucleus()
        .pending_validations
        .iter()
        .for_each(|(_, boxed)| match boxed.0.entry.entry_type() {
            EntryType::LinkAdd => retry_validation(boxed.clone(), context.clone()),
            EntryType::LinkRemove => retry_validation(boxed.clone(), context.clone()),
            _ => panic!("Pending validations are (currently) only implemented for links"),
        });
}
```

## Next Steps and Follow-ups
* **Tests!** (or not follow-up but added to this PR)
* **Persisting** `NucleusState` in the same way we persist `AgentState` so the pending_list survives reboots
* Exponential back-off for retries
* Garbage collection
* Storing validation packages
* Same for CRUD validations
* Custom/app defined dependencies


core/src/instance.rs
```diff
@@ -48,6 +51,11 @@ impl Instance {
         let (rx_action, rx_observer) = self.initialize_channels();
         let context = self.initialize_context(context);
         self.start_action_loop(context.clone(), rx_action, rx_observer);
+        let mut scheduler = Scheduler::new();
+        scheduler
+            .every(10.seconds())
+            .run(scheduled_jobs::create_callback(context.clone()));
+        self.scheduler_handle = Some(Arc::new(scheduler.watch_thread(Duration::from_millis(5))));
```
maackle: 5ms seems pretty short to be checking if 10 minutes have passed yet, am I reading this right? The interval isn't so important, this could easily be 1 sec to give us retries every 9-11 seconds, right?

core/src/nucleus/validation/link_entry.rs
```diff
@@ -0,0 +1,60 @@
...
+            return Err(ValidationError::Error(
+                "Could not extract link_add from entry".into(),
+            ));
+        }
+    };
+    let link = link.link().clone();
+    let (base, target) = links_utils::get_link_entries(&link, context).map_err(|_| {
+        ValidationError::UnresolvedDependencies(
+            [link.base().clone(), link.target().clone()].to_vec(),
+        )
```
maackle: I like that this hooks into the general concept of unresolved dependencies which we can easily use to drive the userland validation case as well :+1:

core/src/nucleus/validation/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+            entry.clone(),
+            validation_data,
+            context
+        )),
+
+        EntryType::LinkRemove => await!(link_entry::validate_link_entry(
+            entry.clone(),
+            validation_data,
+            context
+        )),
```
maackle: Could DRY these matches into

```rust
EntryType::LinkAdd | EntryType::LinkRemove => // the identical code for each
```

core/src/scheduled_jobs/mod.rs
```diff
@@ -0,0 +1,10 @@
+pub mod pending_validations;
+
+use crate::context::Context;
+use std::sync::Arc;
+
+pub fn create_callback(context: Arc<Context>) -> impl 'static + FnMut() + Sync + Send {
+    move || {
+        pending_validations::run_pending_validations(context.clone());
+    }
+}
```
maackle: Nice! I guess closures have come a long way.

core/src/scheduled_jobs/pending_validations.rs
```diff
@@ -0,0 +1,26 @@
...
+pub fn run_pending_validations(context: Arc<Context>) {
+    context
+        .state()
+        .unwrap()
+        .nucleus()
+        .pending_validations
+        .iter()
+        .for_each(|(_, boxed)| match boxed.0.entry.entry_type() {
+            EntryType::LinkAdd => retry_validation(boxed.clone(), context.clone()),
+            EntryType::LinkRemove => retry_validation(boxed.clone(), context.clone()),
```
maackle: Could also collapse these into one match, `EntryType::LinkAdd | EntryType::LinkRemove`

core/src/scheduled_jobs/pending_validations.rs
```diff
@@ -0,0 +1,26 @@
+use crate::{
+    context::Context, network::entry_with_header::EntryWithHeader,
+    workflows::hold_link::hold_link_workflow,
+};
+use holochain_core_types::{cas::content::Address, entry::entry_type::EntryType};
+use std::{sync::Arc, thread};
+
+pub type PendingValidation = Box<(EntryWithHeader, Vec<Address>)>;
```
maackle: Is the `Vec<Address>` even used anywhere? Might be able to take it out. Especially once we have a function to calculate dependencies, we can just generate them. Also this would get stale if we have to retry validation multiple times, some of the deps might be resolvable, not all, and then we should be updating this list of dependencies.

core/src/instance.rs
thedavidmeister:
```suggestion
        self.scheduler_handle = Some(Arc::new(scheduler.watch_thread(Duration::from_millis(1000))));
```

thedavidmeister: it's checking if 10 seconds have passed not 10 minutes, but yeah...

core/src/nucleus/actions/mod.rs
```diff
@@ -2,7 +2,8 @@ pub mod build_validation_package;
 pub mod call_zome_function;
 pub mod get_entry;
 pub mod initialize;
-pub mod validate;
+pub mod run_validation_callback;
+//pub mod validate;
```
thedavidmeister:
```suggestion
```

core/src/scheduled_jobs/pending_validations.rs
lucksus: Yeah, this is not used yet. I added it to show how this can be used for the general case as well. I thought we would then run the callback once during validation and then check from the scheduler if those entries are there, before retrying validation. I was also thinking about storing the validation packagage which we now re-request from the source every time.. so lots of room for improvement. I just wanted to get the basic fix in first.

thedavidmeister: not sure about what @maackle is saying here because:

- calculating deps could be a heavy operation so we don't want to redo it and we don't have to because it is pure
- yes the deps might be resolvable but that doesn't mean we get a different array of deps

let's have this conversation in a followup and not derail the MVP

maackle: Viewing it as a cache makes sense. I was viewing it as a list of dependencies we're still waiting for, and that wouldn't be correct unless we keep this list updated. Makes sense as a cache.

PR #1058, opened 2019-02-27
Expose holo_remote_key to admin interface method

maackle:
Needed for Holo Interceptr.

Also takes `holo_remote_key` out of the Option

### DIFF
--- a/conductor_api/src/interface.rs
+++ b/conductor_api/src/interface.rs
@@ -552,13 +552,14 @@ impl ConductorApiBuilder {
             let name = Self::get_as_string("name", &params_map)?;
             let public_address = Self::get_as_string("public_address", &params_map)?;
             let key_file = Self::get_as_string("key_file", &params_map)?;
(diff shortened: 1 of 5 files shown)


conductor_api/src/interface.rs
```diff
@@ -552,13 +552,14 @@ impl ConductorApiBuilder {
             let name = Self::get_as_string("name", &params_map)?;
             let public_address = Self::get_as_string("public_address", &params_map)?;
             let key_file = Self::get_as_string("key_file", &params_map)?;
+            let holo_remote_key = Self::get_as_bool("key_file", &params_map).unwrap_or_default();
```
maackle: This is the only significant change, the rest is refactoring to remove the Option

lucksus:
Those failing tests are the rationale for `Option` I'd say. If this wouldn't be something we plan to remove again soon, I'd agree to change to a plain bool. But in this case I'd suggest to keep the option so we don't have to change those tests now and back again later.

maackle:
Ugh, yeah, I agree. Closing this and starting a new PR with a simpler commit

PR #1059, opened 2019-02-27
More Pending Validation

lucksus:
# Immediate follow-ups of the pending link validation merge
Builds on the work of https://github.com/holochain/holochain-rust/pull/1054
screenshot from 2019-02-27 18-59-31

## Pending because of validation package

I couldn't easily write a test for pending link validations because having a node fake a link entry without actually validating it will also break the generation of validation packages.

That pointed me to the fact that validation can also fail early because we couldn't get the validation package from the source, usually because the source is offline.

I think we should store validation packages in the DHT to circumvent this long-term. For now, retrying validation of any entry in this case is the bare minimum we need to not break the DHT's consistency (nodes online during authoring can validate, future ones won't if source is offline).

## Removing pending entries from pending list
We have to make sure that entries are removed from the pending list not only when validation succeeds, but also when it finally fails. Therefore added `HolochainError::ValidationPending` to tell pending cases apart from other failures. The scheduled background task now removes a processed item from the list unless it returns `HolochainError::ValidationPending`.

## Memory management
Entries to be validated are now `Arc`ed. The map of these Arcs gets copied out of the state so we can drop the state lock immediately to not block the instance while running pending validations.


core/src/scheduled_jobs/pending_validations.rs
```diff
@@ -1,26 +1,47 @@
...
         .unwrap()
         .nucleus()
         .pending_validations
-        .iter()
-        .for_each(|(_, boxed)| match boxed.0.entry.entry_type() {
-            EntryType::LinkAdd => retry_validation(boxed.clone(), context.clone()),
-            EntryType::LinkRemove => retry_validation(boxed.clone(), context.clone()),
-            _ => panic!("Pending validations are (currently) only implemented for links"),
-        });
+        .clone();
```
lucksus: We get this clone of a `HashMap<Arc<...>>` so we can drop the state lock immediately.

core/src/nucleus/reducers/remove_pending_validation.rs
```diff
@@ -0,0 +1,63 @@
...
+        ))));
+
+        reduce_add_pending_validation(context.clone(), &mut state, &action_wrapper);
+
+        assert!(state.pending_validations.contains_key(&entry.address()));
+
+        let action_wrapper = ActionWrapper::new(Action::RemovePendingValidation(entry.address()));
+
+        reduce_remove_pending_validation(context, &mut state, &action_wrapper)
+    }
```
zippy: I think you need an assert after this to prove that it doesn't contain the key any more!

core/src/workflows/hold_entry.rs
```diff
@@ -4,26 +4,38 @@ use crate::{
...
+            let message = "Could not get validation package from source! -> Add to pending...";
+            context.log(format!("debug/workflow/hold_entry: {}", message));
+            context.log(format!("debug/workflow/hold_entry: Error was: {:?}", err));
+            add_pending_validation(entry_with_header.to_owned(), Vec::new(), &context);
+            HolochainError::ValidationPending
+        })?;
+    let validation_package = maybe_validation_package.ok_or({
+        let message = "Source did respond to request but did not deliver validation package! This is weird! Entry is not valid!";
+        context.log(format!("debug/workflow/hold_entry: {}", message));
+        HolochainError::ValidationPending
```
zippy: Why would this be peding, shouldn't it be failed?

core/src/workflows/hold_link.rs
```diff
@@ -35,10 +36,20 @@ pub async fn hold_link_workflow<'a>(
...
+            let message = "Could not get validation package from source! -> Add to pending...";
+            context.log(format!("debug/workflow/hold_link: {}", message));
+            context.log(format!("debug/workflow/hold_link: Error was: {:?}", err));
+            add_pending_validation(entry_with_header.to_owned(), Vec::new(), context);
+            HolochainError::ValidationPending
+        })?;
+    let validation_package = maybe_validation_package.ok_or({
+        let message = "Source did respond to request but did not deliver validation package! This is weird! Entry is not valid!";
+        context.log(format!("debug/workflow/hold_link: {}", message));
+        HolochainError::ValidationPending
```
zippy: same...

core/src/workflows/hold_entry.rs
lucksus: Yeah, I've noticed that just after releasing for review.. ;) You were a bit too fast! ;D

core/src/nucleus/reducers/remove_pending_validation.rs
lucksus: :see_no_evil:

lucksus:
All addressed, @zippy.

core/src/nucleus/reducers/remove_pending_validation.rs
```diff
@@ -0,0 +1,65 @@
+use crate::{
+    action::{Action, ActionWrapper},
+    context::Context,
+    nucleus::state::NucleusState,
+};
+use std::sync::Arc;
+
+/// Reduce AddPendingValidation Action.
```
thedavidmeister: typo

PR #1060, opened 2019-02-27
Persistence Fix + Persist Nucleus State (for pending validations)

lucksus:
*on top of #1059*
 
# Context: persistence of pending validations
screenshot from 2019-02-27 23-16-59

## How?
Same as with `AgentState` already, a snapshot of `NucleusState` gets saved in the CAS.

## BUT 
We are currently not using `Holochain::load()` anywhere and the test for this function is definitely broken. -> Follow-up.


Connoropolous:
@thedavidmeister your comments resonate with something I just shared with Nico where I was considering whether a scheduler like clokwerk can actually compile to WASM and work

PR #1069, opened 2019-03-01
hc keygen --silent

lucksus:

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -139,11 +139,7 @@ enum Cli {
         about = "Creates a new agent key pair, asks for a passphrase and writes an encrypted key bundle to disk in the XDG compliant config directory of Holochain, which is dependent on the OS platform (/home/alice/.config/holochain/keys or C:\\Users\\Alice\\AppData\\Roaming\\holochain\\holochain\\keys or /Users/Alice/Library/Preferences/com.holochain.holochain/keys)"
     )]
     KeyGen {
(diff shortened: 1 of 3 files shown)


common/src/lib.rs
```diff
@@ -1,2 +1,4 @@
 pub mod env_vars;
 pub mod paths;
+
+pub const DEFAULT_PASSPHRASE: &str = "convenient and insecure keystore";
```
zippy: Please add a TODO comment here that this is going away as soon as we can have a single passphrase for keystore bundle.

cli/src/main.rs
```diff
@@ -195,8 +198,14 @@ fn run() -> HolochainResult<()> {
             cli::test(&current_path, &dir, &testfile, skip_build)
         }
         .map_err(HolochainError::Default)?,
-        Cli::KeyGen => {
-            cli::keygen(None, None).map_err(|e| HolochainError::Default(format_err!("{}", e)))?
+        Cli::KeyGen { silent } => {
+            let passphrase = if silent {
```
sphinxc0re: You can use the `filter` method on `Option<T>`:

```rust
let passphrase = Some(holochain_common::DEFAULT_PASSPHRASE.to_string()).filter(silent);
```

zippy: Nope, that actually doesn't work.  filter takes a function not a boolean.  So I'm just going to merge as is.

PR #1071, opened 2019-03-01
When creating a CAS storage_path, create parent dirs, log path on fail

pjkundert:
If I'm starting up a new instance, and I've provided a path (eg. `tmp-storage`) that doesn't (yet) exist, the `holochain -c <config-file>` will fail; it will current neither log the name of the directory it couldn't create (eg. `tmp-storage/cas`, nor will it attempt to create the parent directory I've named (`tmp-storage`):

    [[instance]]
    ....
    [instance.storage]
    type = "file"
    path = "tmp-storage"

This corrects both deficiencies; it attempts to create the specified parent directories, and it logs the name of the path it couldn't create, if unsuccessful.


pjkundert:
@Connoropolous No automated tests, just manual testing trying to start up an instance.  I couldn't start the instance, and the failure message was unhelpful -- I had to `strace holochain -c ...` the startup to find out where it was failing.  So, I improved the message.

The `.recursive(true)` on the `DirBuilder` to allow it to create leading directory path entries was also tested manually. I can't imagine it being a security issue, unless of course `holochain` is run as root, but not doing this will cause a lot of issues; configuration of any CAS storage path will fail unless you first create the directory, because it tries to create the `cas` sub-directory in there without first creating the target storage directory.  That'll be a surprising failure in most cases.

pjkundert:
Sorry, the tests are failing; it successfully now creates `/*?abc/bar` in the `test_create_path_if_not_exists` test, which is somewhat odd.  Are these tests running in a sandbox where `/...` is writable to the test?

pjkundert:
Ah; it was failing; just testing against the new (changed) failure message.  Fixed test case.

cas_implementations/src/path.rs
```diff
@@ -13,8 +13,11 @@ pub fn storage_path(path: &Path, folder_name: &str) -> HcResult<PathBuf> {
 pub fn create_path_if_not_exists(path: &Path) -> HcResult<()> {
     if !path.exists() {
         return DirBuilder::new()
+            .recursive(true) // create parent dirs if necessary
             .create(path)
-            .map_err(|_| HolochainError::IoError("Could not create directory".to_string()));
+            .map_err(|_| {
+                HolochainError::IoError(format!("Could not create directory: {:?}", path))
+            });
```
sphinxc0re: Okay, this is ridiculous... this whole function already exists in the `stdlib`. It's called `std::fs::create_dir_all`

cas_implementations/src/path.rs
```diff
@@ -36,14 +41,38 @@ pub mod tests {
...
+            Err(err) => {
+                assert!(err.to_string() == "Could not create directory: \"/*?abc/.hc/storage/bar\"")
+            }
+        };
+        let dir = tempdir().unwrap();
+        let result = create_path_if_not_exists(dir.path());
+        match result {
+            Ok(val) => assert_eq!(val, ()),
+            Err(_) => unreachable!(),
+        };
```
sphinxc0re: This whole match statement can be omitted by a simple 
```rust
std::fs::create_dir_all(dir.path()).expect("failed to create directory");
```

cas_implementations/src/path.rs
zippy: Hey @sphinxc0re, you definitely have a deeper knowledge of the stdlib than all of, so thanks for that.  It would be more helpful if you could just provide a suggested edit using github's UI, or even just check out the branch and make the fix.  Thanks.

pjkundert: Ya, I saw that `std:fs::create_dir_all`, but didn't look into it too deeply -- I didn't want to get into changing crates, etc.  I'll take a look at it.  Thanks! ...  Ah; it takes a Path, not an &Path:
```
pub fn create_dir_all<P: AsRef<Path>>(path: P) -> io::Result<()> {
    DirBuilder::new().recursive(true).create(path.as_ref())
}
```
so we'd have to `.to_owned()` the path, and then call it.  I don't know if its worth it to make the implementation of the function more opaque.  Its just one line of method calls, and I think it can stay like it is...

pjkundert: I made the change, copying the `path` (no biggie, we only use this once right now), but I don't think it improves the code.  I like it explicit and clear, the way it is. And, no copy.  Changing the API to a plain `Path` also was not better, because when you want to create a path, usually you want to do something with it later, requiring the (now moved) Path.  And, it launched a bunch of complaints by the compiler that you really should borrow the `Path`, and `[u8]` arrays and sizes and...  I don't want to waste time on investigating that right now.  Basically, the `std::fs::create_dirs_all` API is worse! :)

pjkundert:
Thanks, @sphinxc0re -- ﻿this solves the problem appropriately, without needing to emit an error message to help diagnosis.  The original problem -- needing to run `strace holochain -c ...` to try to find out why an instance wouldn't start -- is no longer necessary for this failure case, unless you specify a target `tmp-storage` path in a non-writable area of the filesystem.  Then, you'll get an opaque failure to start, and need to run `strace` to find it.  But, this should be rare.

PR #1072, opened 2019-03-02
Make nodejs_waiter aware of pending validation

maackle:
Let's see if this helps with the hanging tests.
- Adds a wait condition: Every AddPendingValidation must be matched with a RemovePendingValidation
- DRY's waiter logic a bit 


nodejs_waiter/src/waiter.rs
```diff
@@ -4,7 +4,7 @@ use holochain_core::{
     nucleus::ZomeFnCall,
     signal::{Signal, SignalReceiver},
 };
-use holochain_core_types::entry::Entry;
+use holochain_core_types::{cas::content::AddressableContent, entry::Entry};
```
Connoropolous: why does this come in?

zippy: If you use `.address()` on something that implements AddressableContent, you also have to include the trait in scope.

Connoropolous: I see

zippy:
@pjkundert can you see if this fixes some of the hanging problems you were having in your tests?

nodejs_waiter/src/waiter.rs
maackle: Yeah here https://github.com/holochain/holochain-rust/pull/1072/files/8ddb359fbe20ed4521311e00b3d51376c835fbcc#diff-f6e488476c3a39ed86c9080bee61ef9aR209

maackle:
@Connoropolous there are tests: https://github.com/holochain/holochain-rust/blob/develop/nodejs_waiter/src/waiter.rs#L433

pjkundert:
Sorry, yes -- I tested this.  No effect.

nodejs_conductor/publish.js
```diff
@@ -58,7 +64,7 @@ npmPackageJson.scripts.fallback = "npm run compile && mkdir -p bin-package && cp
 
 fs.writeFileSync("./dist/package.json", JSON.stringify(npmPackageJson, null, 2));
 
-shell.mkdir("./bin-package");
+shell.mkdir('-p', "./bin-package");
```
Connoropolous: what's -p?

Connoropolous:
Cool, good to know there's some tests, this feels like a pretty complex area in terms of functionality, test coverage here is a big deal I think

nodejs_conductor/publish.js
maackle: recursive -- the easiest way to make it not complain if the dir already exists, but still create it if it doesn't

zippy:
Even though this doesn't fix @pjkundert blocking issue, I think this is ready to merge no?  We still need it.

PR #1074, opened 2019-03-02
0.0.6 merge: Fix peer URLs for networking

Connoropolous:
Networking URLs changed, but the docs and some references didn't. Fixing that.

### DIFF
--- a/conductor_api/src/config.rs
+++ b/conductor_api/src/config.rs
@@ -643,7 +643,7 @@ pub mod tests {
     id = "app spec instance"
 
     [network]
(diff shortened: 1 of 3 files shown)


Connoropolous:
@ddd-mtl would it be satisfactory just to change it so the URL format is overall right, but it says ?a=blabla ?

ddd-mtl:
I wouldn't say "satisfactory" but I'd be okay with it.

Connoropolous:
Sorry maybe I’m not understanding your reasoning for wanting to leave in a
URL format that’s not relevant in the codebase anymore. If you can
elaborate...

PR #1077, opened 2019-03-03
use real signing for capabilities

zippy:
- [x] updates to use real signing instead of mock
- [x] updates to use new `call` rpc to get token and provenance for generating capability request on zome calls if provided.


conductor_api/src/interface.rs
```diff
@@ -1055,7 +1072,58 @@ pub mod tests {
...
+                "call",
+                Some(json!({
+                    "instance_id" : "test-instance-1",
+                    "zome" : "greeter",
+                    "function" : "hello",
+                    "token" : "bogus token",
+                    "provenance" : {
+                        "source" : "some_source",
+                        "signature" : "some_signature",
+                    },
```
lucksus: I was wondering why we see the "instance not active yet" error below, for this provenance. The answer is because the provenance check happens within the instance but in this test we don't have an instance, right?

zippy: yep.  I wish we did, and then it could be a better test, but this just tests that the routes are in place.

core_types/src/signature.rs
```diff
@@ -2,17 +2,28 @@
...
-/// Provenance is a tuple of initiating agent and signature of some item being signed
+/// Provenance is a tuple of initiating agent public key and signature of some item being signed
 /// this type is used in headers and in capability requests where the item being signed
 /// is implicitly known by context
-pub type Provenance = (Address, Signature);
-/*impl Provenance {
-    fn source(&self) -> Address {
-        self.1.clone()
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Hash, Eq, DefaultJson)]
+pub struct Provenance {
```
thedavidmeister: this could add "a lot" of redundant data if we're serialising the same `source` and `signature` keys and sending them around

use the new ergonomics methods and leave the tuple in place?

conductor_api/src/interface.rs
```diff
@@ -138,23 +144,36 @@ impl ConductorApiBuilder {
...
+                match maybe_provenance {
+                    None => make_cap_request_for_call(
+                        context.clone(),
+                        token,
+                        &func_name,
+                        params_string.clone(),
+                    ),
+                    Some(json_provenance) => {
+                        let provenance =
+                            Provenance::try_from(JsonString::from(json_provenance.to_string()))
```
thedavidmeister: does `JsonString::from(json_provenance)` work?

why do we have to `to_string()` it here? that runs the risk of the output not being valid JSON, depending on the `to_string` implementation...

zippy: so that doesn't work, because there dosen't seem to be an implementation of From for serde::json_values... !!  So ti think we are safe here becasue it is valid json.

zippy: Should there be a followup to covnert a serdejsonvalue to a JsonString?

zippy: Ok, actually this got resolved by removing JsonString and just using `serde_json::from_value`

core_types/src/signature.rs
zippy: good catch, fixed.

PR #1075, opened 2019-03-02
improve startup error message

Connoropolous:

PR #1078, opened 2019-03-04
Make AgentStateSnapshot chain header optional

maackle:

### DIFF
--- a/core/src/agent/state.rs
+++ b/core/src/agent/state.rs
@@ -43,12 +43,12 @@ impl AgentState {
 
     pub fn new_with_top_chain_header(
         chain_store: ChainStore,
-        chain_header: ChainHeader,
+        chain_header: Option<ChainHeader>,
     ) -> AgentState {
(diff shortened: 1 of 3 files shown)


thedavidmeister:
@lucksus isn't the purpose of the `AgentStateSnapshot` to save the `AgentState`?

it is possible for the agent state to not have a top chain header

i think this is breaking all our cold ubuntu builds with the error:

`Could not save instance! Error: ErrorGeneric("Could not serialize")`

which is blocking releases right now

is that right @maackle ?

maackle:
I thought this might help but it seems related to https://github.com/holochain/holochain-rust/pull/1075 which was just a matter of fixing misleading log messages. So, I'm not sure and will leave it up to you two to decide whether this is significant.

thedavidmeister:
even if it doesn't fix the tests it makes perfect sense to me :man_shrugging: 

i still approve

@lucksus what do you think?

PR #1079, opened 2019-03-04
Release 0.0.5 alpha

thedavidmeister:


conductor/README.md
```diff
@@ -41,7 +41,7 @@ These dependencies need to be installed in order to compile, and use `holochain`
 
 To install the latest version of the Holochain conductor, run the following command in a terminal
 ```shell
-$ cargo install holochain --force --git https://github.com/holochain/holochain-rust.git --branch develop
+$ cargo install holochain --force --git https://github.com/holochain/holochain-rust.git --tag v0.0.5-alpha
```
zippy: I don't think we want to change this instance in the readme...  because it also says "latest".

zmq-sys/Cargo.lock
```diff
@@ -0,0 +1,44 @@
+[[package]]
```
zippy: Shouldn't this file be gone?  I thought we removed zmq?

thedavidmeister: @zippy that would be on 0.0.6

conductor/README.md
Connoropolous: Yeah, we either change the above text, or change this back

PR #1081, opened 2019-03-04
holochain_net: fixing IPC mock mode

ddd-mtl:
Misc stuff while fixing n3h-mock mode:
 -  Added test `setup_one_node()`
- P2pNode now handles incoming  `Protocol::P2pReady` sets a is_network_ready field.
 - Tests now waits for `is_network_ready()`
 - Moved all `impl TransportWss<std::net::TcpStream>` into its own file
 - changed `TransportWss::priv_close_one()` to `TransportInfo::close()`
- Renamed `TransportInfo.socket` to `TransportInfo.stateful_socket`
- Changed `wait_connect` from a function to a method.
- Changed `wait_connect()` to return TransportId on success, and Vec<TransportEvent> on failure.
- added transport_id field to IpcNetWorker 
- Removed an indentation level in IpcNetWorker::new()
- Removed unused field `endpoint` in IpcNetWorker

### DIFF
--- a/net/src/ipc/transport_wss/tcp.rs
+++ b/net/src/ipc/transport_wss/tcp.rs
@@ -4,7 +4,7 @@
 
 use crate::ipc::{
     transport::{
-        DidWork, Transport, TransportError, TransportEvent, TransportId, TransportIdRef,
+        Transport, TransportError, TransportEvent, TransportId,
         TransportResult,
     },
     transport_wss::{
--- a/net/src/ipc/transport_wss/tcp.rs
+++ b/net/src/ipc/transport_wss/tcp.rs
@@ -30,7 +30,6 @@ impl TransportWss<std::net::TcpStream> {
     ) -> TransportResult<TransportId> {
         // Launch connection attempt
         let transport_id = self.connect(&uri)?;
-
         // Wait for a successful response
         let mut out = Vec::new();
         let start = std::time::Instant::now();
--- a/net/src/ipc/transport_wss/tcp.rs
+++ b/net/src/ipc/transport_wss/tcp.rs
@@ -1,10 +1,7 @@
 //! abstraction for working with Websocket connections
 //! TcpStream specific functions
 
-use crate::ipc::{
-    transport::{Transport, TransportError, TransportEvent, TransportId, TransportResult},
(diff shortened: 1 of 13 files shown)


net/src/ipc/transport_wss/tcp.rs
```diff
@@ -0,0 +1,47 @@
...
+    pub fn with_std_tcp_stream() -> Self {
+        TransportWss::new(|uri| {
+            let socket = std::net::TcpStream::connect(uri)?;
+            socket.set_nonblocking(true)?;
+            Ok(socket)
+        })
+    }
+
+    /// connect and wait for a Connect event response
+    pub fn wait_connect(&mut self, uri: &str) -> TransportResult<TransportId> {
```
neonphog: I don't think anything in here is specific to a TCP stream... should this move to the generic?

net/src/ipc_net_worker.rs
```diff
@@ -126,60 +119,34 @@ impl IpcNetWorker {
...
-    }
-
-    bail!("ipc connection timeout");
-}
-
 impl NetWorker for IpcNetWorker {
     /// stop the net worker
     fn stop(mut self: Box<Self>) -> NetResult<()> {
-        self.socket.close_all()?;
+        self.wss_socket.close(self.transport_id)?;
```
neonphog: while there isn't currently a path that creates multiple connections, shouldn't we still close them all just in case?

net/src/ipc/transport_wss/tcp.rs
ddd-mtl: Totally. Wasn't sure since it was only used for a TcpStream...

net/src/ipc_net_worker.rs
ddd-mtl: Yeah. There is currently an issue with closing with n3h-mock and I thought this would fix it but it actually doesn't so i can revert.

net/src/ipc/transport_wss/tcp.rs
ddd-mtl: Moved.

net/src/ipc_net_worker.rs
ddd-mtl: reverted

ddd-mtl:
@zippy Nope. Just refactoring and having the tests work when in ipc mock.

PR #1082, opened 2019-03-05
Fix Pending Validations and thereby HoloFuel tests

lucksus:
 *-> No, fix for feature that is introduced in next version.*

# Fix for blocking dead-locks in HoloFuel integration tests
The reason the tests were blocking was that the Pending Validations feature was not working correctly in all cases, but the waiter correctly waits for all nodes in the test scenario to have received and validated all entries. 

## Problem in Detail
There was no differentiation between:
* storing an `Entry::LinkAdd` as an entry itself in the DHT, and
* updating the meta information of the link's base entry.

Both cases resulted in calling `hold_link` which only works in the second case.

## Solution
With these changes the pending validation `HashMap` includes the workflow that added the entry so the worker thread now knows which kind of workflow to run when retrying validation.

----

Furthermore, receiving an empty response when requesting the validation package from source was treated as validation failure. This lead to another reason for blocking. Now, receiving an empty response also adds the entry to the pending list which fixes the last remaining waiter dead-locks.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@ jobs:
  app-spec-tests:
   docker:
    - image: holochain/holochain-rust:circle.zz.end
-  resource_class: large
+  resource_class: xlarge
   steps:
    - checkout
 
--- a/core/src/workflows/hold_entry.rs
+++ b/core/src/workflows/hold_entry.rs
@@ -36,7 +36,7 @@ pub async fn hold_entry_workflow<'a>(
             );
             HolochainError::ValidationPending
         })?;
-    let validation_package = maybe_validation_package.ok_or_else(|| {
+    let validation_package = maybe_validation_package.ok_or({
         let message = "Source did respond to request but did not deliver validation package! (Empty response) This is weird! Let's try this again later -> Add to pending";
         context.log(format!("debug/workflow/hold_entry: {}", message));
         add_pending_validation(
--- a/core/src/workflows/hold_entry.rs
+++ b/core/src/workflows/hold_entry.rs
@@ -9,8 +9,11 @@ use crate::{
     },
 };
 
-use crate::scheduled_jobs::pending_validations::ValidatingWorkflow;
+use crate::{
+    nucleus::validation::ValidationError, scheduled_jobs::pending_validations::ValidatingWorkflow,
+};
 use holochain_core_types::{
+    cas::content::AddressableContent,
     error::HolochainError,
     validation::{EntryAction, EntryLifecycle, ValidationData},
 };
@@ -57,10 +60,35 @@ pub async fn hold_entry_workflow<'a>(
     };
 
     // 3. Validate the entry
-    await!(validate_entry(entry.clone(), validation_data, &context))?;
+    await!(validate_entry(entry.clone(), validation_data, &context)).map_err(|err| {
+        context.log(format!(
+            "err/workflow/hold_entry: {} is NOT valid! {:?}",
+            entry_with_header.entry.address(),
+            err,
+        ));
+        if let ValidationError::UnresolvedDependencies(dependencies) = &err {
+            add_pending_validation(
+                entry_with_header.to_owned(),
+                dependencies.clone(),
+                ValidatingWorkflow::HoldEntry,
+                &context,
+            );
+        }
+        HolochainError::ValidationPending
+    })?;
+
+    context.log(format!(
+        "debug/workflow/hold_entry: is valid! {}",
+        entry_with_header.entry.address()
+    ));
 
     // 3. If valid store the entry in the local DHT shard
-    await!(hold_entry(entry_with_header, context))?;
+    await!(hold_entry(entry_with_header, context.clone()))?;
+
+    context.log(format!(
+        "debug/workflow/hold_entry: HOLDING: {}",
+        entry_with_header.entry.address()
+    ));
 
     Ok(())
 }
--- a/core/src/workflows/hold_entry.rs
+++ b/core/src/workflows/hold_entry.rs
@@ -9,6 +9,7 @@ use crate::{
     },
 };
 
+use crate::scheduled_jobs::pending_validations::ValidatingWorkflow;
 use holochain_core_types::{
     error::HolochainError,
     validation::{EntryAction, EntryLifecycle, ValidationData},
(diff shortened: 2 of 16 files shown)


core_types_derive/src/lib.rs
```diff
@@ -17,7 +17,10 @@ fn impl_default_json_macro(ast: &syn::DeriveInput) -> TokenStream {
             fn from(v: &#name) -> JsonString {
                 match ::serde_json::to_string(v) {
                     Ok(s) => Ok(JsonString::from(s)),
-                    Err(e) => Err(HolochainError::SerializationError(e.to_string())),
+                    Err(e) => {
+                        println!("Error serializing to JSON: {:?}", e);
+                        Err(HolochainError::SerializationError(e.to_string()))
+                    },
```
lucksus: @thedavidmeister, I've added this `println!` because I spent several hours debugging until reaching this place. For some strange reason the expect below did not lead to an output showing up, but it killed the thread this was happening in. Since it was the main redux thread that was trying to save a snapshot of the state including `pending_validations` which I had with a key that could not be serialized into a string (but that's what JSON requires) it lead to strange blocking..

Since this is hidden behind a derive, I'd rather have an error output twice rather than none, as in my case.. So I'm suggesting to add this `println` unless you have a better idea.

pjkundert:
Awesome, @lucksus!  I'll check this out right away.

core_types_derive/src/lib.rs
ddd-mtl: `eprintln` ?

.circleci/config.yml
```diff
@@ -21,7 +21,7 @@ jobs:
  app-spec-tests:
   docker:
    - image: holochain/holochain-rust:circle.zz.end
-  resource_class: large
+  resource_class: xlarge
```
maackle: are we keeping this xlarge?

core/src/workflows/hold_entry.rs
```diff
@@ -27,13 +28,24 @@ pub async fn hold_entry_workflow<'a>(
...
-            add_pending_validation(entry_with_header.to_owned(), Vec::new(), &context);
+            add_pending_validation(
+                entry_with_header.to_owned(),
+                Vec::new(),
+                ValidatingWorkflow::HoldEntry,
+                &context,
+            );
             HolochainError::ValidationPending
         })?;
     let validation_package = maybe_validation_package.ok_or({
```
maackle:
```suggestion
    let validation_package = maybe_validation_package.ok_or_else(|| {
```

core/src/workflows/hold_entry.rs
```diff
@@ -27,13 +28,24 @@ pub async fn hold_entry_workflow<'a>(
...
-        let message = "Source did respond to request but did not deliver validation package! This is weird! Entry is not valid!";
+        let message = "Source did respond to request but did not deliver validation package! (Empty response) This is weird! Let's try this again later -> Add to pending";
         context.log(format!("debug/workflow/hold_entry: {}", message));
-        HolochainError::ValidationFailed("Entry not backed by source".to_string())
+        add_pending_validation(
+            entry_with_header.to_owned(),
+            Vec::new(),
+            ValidatingWorkflow::HoldEntry,
+            &context,
+        );
```
maackle: If you use `ok_or`, it's going to run this every time. You probably want `ok_or_else` to only run this in the Err case, right?

core/src/workflows/hold_link.rs
```diff
@@ -41,13 +44,24 @@ pub async fn hold_link_workflow<'a>(
...
-            add_pending_validation(entry_with_header.to_owned(), Vec::new(), context);
+            add_pending_validation(
+                entry_with_header.to_owned(),
+                Vec::new(),
+                ValidatingWorkflow::HoldLink,
+                context,
+            );
             HolochainError::ValidationPending
         })?;
     let validation_package = maybe_validation_package.ok_or({
```
maackle: Same here

```suggestion
    let validation_package = maybe_validation_package.ok_or_else(|| {
```

pjkundert:
Even the code in my latest development branch, which used to fail even with very long delays between the commit/Link now passes!  Thanks.

.circleci/config.yml
lucksus: Yes, app-spec failed with *only* `large`.

core/src/workflows/hold_entry.rs
lucksus: Jeez, yes! :see_no_evil: 
Thanks, @maackle :pray:

core_types_derive/src/lib.rs
lucksus: Yes, thanks, @ddd-mtl.

pjkundert:
Unfortunately -- the lockup is back.  I haven't tried to see if the `hdk::sleep` mitigates it, but I wanted to let y'all know.

I am uncertain why this has occurred.  But, it appears to be reliably locking up in my tests which (until recently) were working.

pjkundert:
I'm going to pull this morning's commit (which seemed to work), and see if I can duplicate my success, and then maybe try to do a `git bisect`...

pjkundert:
Rolling back to this morning's https://github.com/holochain/holochain-rust/pull/1082/commits/b24726ef6e50d48788c8bde6ba6d240a950e7496 allows my tests to again pass reliably...

pjkundert:
The problem is somewhere in here: `git diff b24726e..47bf445`.  It is the changes from `.ok_or` to `.ok_or_else` -- something that was being done all the time (probably the the `add_pending_validation` calls) via the `.ok_or` calls, changed to only being done in !Ok cases.

pjkundert:
Hmmm.  It seems like if we restore the old behavior, which works, that we are *always* calling `add_pending_validation`, whether or not `maybe_validation_package` or Ok or not -- and this was necessary to make it not lock up:
```
    let validation_package = maybe_validation_package.ok_or({ // _else(|| {
        let message = "Source did respond to request but did not deliver validation package! (Empty response) This is weird! Let's try this again later -> Add to pending";
        context.log(format!("debug/workflow/hold_link: {}", message));
        add_pending_validation(
            entry_with_header.to_owned(),
            Vec::new(),
            ValidatingWorkflow::HoldLink,
            &context,
        );
        HolochainError::ValidationPending
    })?;
```

Does this make sense?  Why should be *always* need to `add_pending_validation`, even when we have a `validation_package`, or suffer a lock-up?

lucksus:
Oh, wow. That is interesting. But when I was running the HoloFuel tests I did not see pending validations triggered in each tests. Really just a few..

I will revert those changes for now so that we have something we can merge now and then investigate this more tomorrow. @maackle would you have 15 minutes some point tomorrow to help me get decent debug output from the waiter?

pjkundert:
I tried changing just `hold_entry.rs` to use `.ok_or(...` (thinking that perhaps it was just the hold_link that might need to *always* be added to pending because of possibly missing dependencies), but it locked up...

core_types_derive/src/lib.rs
thedavidmeister: @lucksus i think that is fine, the only time we get here anyway is due to runtime data or other weird edge cases

lucksus:
So we also see the dead-lock in app-spec tests, both without @maackle's fixes and even in develop. I think it came in through https://github.com/holochain/holochain-rust/pull/1048.

PR #1052, opened 2019-02-26
WIP: remove nodejs conductor

thedavidmeister:
- [ ] I have added a summary of my changes to the changelog

## todo

- [x] remove nodejs conductor
- [ ] reinstate nodejs conductor repo with correct version
- [ ] loosen node deps in install/ci
- [ ] tests passing with native conductor
- [x] app spec tests can hit websockets from conductor
- [ ] app spec tests use the web client to interact with websockets

PR #1084, opened 2019-03-05
Ensure build of libsodium/rust_sodium-sys is documented in Makefile and works

pjkundert:
Presently, building libsodium and linking against the result fails on important linux platforms (eg. Debian), and it is not clear in the Makefile or elsewhere how to fix the issues.  I've added some documentation to the Makefile to assist the developer in overcoming this obstacle.

* Detail how to select system/pkg_config/download build approach for libsodium in the Makefile
* Logs the selected libsodium version and/or build approach and builds rust_sodium-sys
  as a separate target, to isolate failures.
* Remove an unnecessary logging of CFLAGS used to build libsodium, but add
  logging of selected libsodium build choice.

By default, we'll download/build our pinned libsodium version, and try to log alternatives for the user if it fails (but this failure is not always detectable, because it builds fine but fails in the later linking stage).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -279,7 +279,7 @@ clean: ${C_BINDING_CLEAN}
 	@$(RM) -rf app_spec/dist
 	@for cargo in $$( find . -name 'Cargo.toml' ); do \
 	    echo -e "\033[0;93m## 'cargo clean' in $${cargo%/*} ##\033[0m"; \
-	    ( cd $${cargo%/*} && cargo clean ); \
+	    ( cd $${cargo%/*} && $(CARGO) clean ); \
 	done
 
 # clean up the extraneous "C" binding test files
--- a/Makefile
+++ b/Makefile
@@ -221,13 +221,13 @@ wasm_build: ensure_wasm_target
 
 .PHONY: build_holochain libsodium_version
 libsodium_version:
-	@[ ! -z "$$RUST_SODIUM_LIB_DIR" ] \
-	    && echo -e "\033[0;93m## Building rust_sodium-sys -- with system libsodium: $(RUST_SODIUM_LIB) ##\033[0m" \
-	    || ( [ ! -z "$$RUST_SODIUM_USE_PKG_CONFIG" ] \
+	@[ ! -z "$${RUST_SODIUM_LIB_DIR+defined}" ] \
+	    && echo -e "\033[0;93m## Building rust_sodium-sys -- with system libsodium \"$(RUST_SODIUM_LIB)\" in: $${RUST_SODIUM_LIB_DIR:-(unknown)}) ##\033[0m" \
+	    || ( [ ! -z "$${RUST_SODIUM_USE_PKG_CONFIG+defined}" ] \
 		&& echo -e "\033[0;93m## Building rust_sodium-sys -- with pkg_config libsodium ##\033[0m" \
 		|| ( echo -e "\033[0;93m## Building rust_sodium-sys -- with download/build libsodium ##\033[0m" ))
 	@$(CARGO) build --manifest-path rust_sodium-sys/Cargo.toml \
-	    || echo -e "\031[0;93m##  *** Building rust_sodium-sys failed: Ensure libsodium version 1.0.12+ is installed ##\031[0m" \
+	    || echo -e "\033[0;91m##  *** Building rust_sodium-sys failed: Ensure libsodium version 1.0.12+ is installed ##\033[0m" \
 
 build_holochain: core_toolchain wasm_build libsodium_version
 	@echo -e "\033[0;93m## Building holochain... ##\033[0m"
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,12 @@ export RUST_BACKTRACE=1
 # There are 3 methods to obtain the libsodium encryption library required by holochain-rust,
 # selectable by setting/clearing various RUST_SODIUM_...  environment variables (see:
 # https://github.com/maidsafe/rust_sodium).  These selections are implemented and enforced in
-# rust_sodium-sys/build.rs
+# rust_sodium-sys/build.rs.  The default is to download/compile libsodium 1.0.17 (06-Jan-2019).
+
+# 0) DEFAULT: Downloaded/compiled by holochain-rust build: select by clearing RUST_SODIUM_LIB_DIR
+# and RUST_SODIUM_USE_PKG_CONFIG.  Some systems require libsodium to be configured and built with
+# `--disable-pie`; select this here by setting RUST_SODIUM_DISABLE_PIE.
+#export RUST_SODIUM_DISABLE_PIE=1
 
 # 1) System installed: select by setting RUST_SODIUM_LIB_DIR. We need to find the location of the
 # system's libsodium dynamic library: at least version 1.0.12 is required.  On Mac, `brew install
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -20,12 +20,43 @@ help:
...
+
+# 1) System installed: select by setting RUST_SODIUM_LIB_DIR. We need to find the location of the
+# system's libsodium dynamic library: at least version 1.0.12 is required.  On Mac, `brew install
+# libsodium`.  On Ubuntu Bionic (libsodium 1.0.16), Consmic (libsodium 1.0.16) and Disco (libsodium
+# 1.0.17): `apt-get install libsodium-dev`.  On Debian Stretch (stable, libsodium 1.0.11 *to old*,
+# add `buster` to /etc/apt/sources...), or Debian Buster (testing, libsodium 1.0.17): `apt-get -t
+# buster -u install libsodium-dev`.  On other Linux distros, ensure at least libsodium 1.0.12+ is
+# installed.
+RUST_SODIUM_LIB=$(shell find /usr/local/lib /usr/lib /lib -name 'libsodium.so' -o -name 'libsodium.dylib' | head -1)
+export RUST_SODIUM_LIB_DIR=$(dir $(RUST_SODIUM_LIB))
```
neonphog: I don't have the context for the node-js neon binding problems... but part of the reason for choosing rust_sodium-sys in the first place was it defaulted to downloading / building a pinned version of libsodium on all target os-s (linux, windows, macOs)... so we wouldn't have to deal with lame stale os versions...

Not saying we shouldn't do this... just wish we could get it to work the other way.

pjkundert: I'll take another run at getting the download/build option working on Debian. I'm not sure why it doesn't work properly -- it should. If something can't build on Debian, there's something seriously wrong...

zippy: Please sync with @philipbeadle and @thedavidmeister about this. They have spent MANY hours working on the lib-sodium issue, lets' not duplicate work.

thedavidmeister:
soz all i think i need to close this one out (@pjkundert @zippy @neonphog )

i appreciate the initiative/work but

- i'd like to not perpetuate the exports of the env that are in nix-shell, that was a bit of a hack because of time, ideally we'd manage env on a per command basis OR install them with the install script, not set the current shell environment state silently
- we don't want to hit the shared lib unless we absolutely need to - we enforce v `1.0.16` as much as possible as it is the only version we've tested, which e.g. isn't shipped with xenial (bionic is needed for ubuntu), as per @neonphog 's comments
- the shared lib is only needed for the node conductor because it builds as `dylib`, see https://github.com/holochain/holochain-rust/blob/develop/nodejs_conductor/native/Cargo.toml#L11 and cargo can't output dynamic libs unless the inputs are dynamic
- the node conductor is in an active process of being removed from this repo, this PR would increase the extraction process https://github.com/holochain/holochain-rust/pull/1052

what we really need to do is finish removing the node conductor as a priority as soon as 0.0.6 goes out the door >.<

at that point everything can go back to using the static/pinned setup in `rust_sodium-sys`

if there is some _urgent_ need to get make commands to use these environment variables feel free to reopen with additional context in the issue description :)

thedavidmeister:
also note that the nodejs conductor repo ships the correct version of libsodium directly in the repo and make configures, builds and installs it from there

https://github.com/holochain/holochain-nodejs/pull/16/files#diff-1666b50e69e9393a71f6edeab6bf7a70

even though it is shared over there, we aren't using a version injected by the OS, we're hitting the version that we test/provide

pjkundert:
OK...  I'm not sure what to do; holochain-rust won't build on Debian, which I don't think is going to be acceptable to the community.  Debian is the *granddaddy*; if it won't build on Debian, then there is something seriously wrong!  I really believe that we have to have a working Makefile, and Debian is absolutely going to be an early and common target for development in the Alpha, and later.

Documenting the build requirements for rust_sodium-sys on various systems is necessary, and (as far as I can tell), it is only currently documented in `shell.nix` and `rust_libsodium-sys/build.rs`.  Documenting it in the Makefile is perhaps more accessible?  It has to be documented somewhere, because this is (presently, I think), the most fragile part of the holochain-rust build, and is likely to catch people. And when the build fails, people typically start with the Makefile.

I'll make sure it defaults to download/build our pinned version, and I'll try to continue diagnosing why it fails to link on Debian, now even with RUST_SODIUM_DISABLE_PIE=1 set.

pjkundert:
Sorry, not trying to make trouble -- I'm just trying to get holochain-rust to build, so I can continue working on holofuel, and its really causing me a serious time drain.  If it was only going to be me running into this, I'd just switch over to Nix, but I don't think requiring every developer to abandon Debian as a target platform is going to be palatable (at least to any old greybeards like me out there who deploy to Debian at scale!)

pjkundert:
OK, updating to libsodium 1.0.17 (03-Jan-2019) solves the build problems.  1.0.16 is from 2017, so upgrading is probably starting to look like a good idea, if it doesn't cause anyone problems.

I was unable to determine *why* the link failures on Debian occurred; updating the C compiler version, setting RUST_SODIUM_DISABLE_PIE=1, etc. didn't change the failure.  Updating to 1.0.17 was smooth and seemed to fix the build issues, and seems to work fine across Linux, Mac and Windows.

pjkundert:
All platforms (Linux, Mac, Windows, and now NixOS) download and build a pinned version of libsodium, as defined in `rust_sodium-sys/build.rs`.  Previously, we were using the pre-built Nix libsodium library (which happened to be 1.0.16, but we were not specifying that anywhere).

How that we've upgrade to libsodium 1.0.17 03-Jan-2019, it seems to build reliably on every platform, and we're all using the same pinned version.

pjkundert:
Actually, it turns out that this doesn't fix the issue.  The only reason nix-shell *was* working, is because it *wasn't* downloading/building libsodium.  It was just using the pre-built nix libsodium library, which happened to be 1.0.16.  Trying to use the download/build on nix runs into the exact same failures as I am seeing on Debian.

I'm going to try to dive into the download/build process in the `rust_sodium-sys/build.rs` and fix it -- assuming that is what we want to standardize on going forward.  Basically, we were only using it in a few places... maybe on Mac and Windows?  It wasn't being used under NixOS, for sure...

pjkundert:
OK, I think I'm starting to see what's going on.

1) nix-shell is just a reproducible package management shell (like apt or brew).  Not a VM.  Doesn't require people to run a different OS.  Much lower threshold of pain than I previously understood.

2) I'll dump the current Makefile, and make a much simpler `./configure; make; make install` process that just makes sure the nix-shell is installed, and then a Makefile that fires the nix-shell for various targets.

3) This lowers the developer threshold *dramatically*, on all Linux and Mac platforms!  Basically, no software has to be installed to begin developing and generating Holochain apps, and compiling Zomes into *.dna.json targets!  No nodejs, nothing.  Just nix-shell -- because everything we need (gcc, nodejs, etc.) is all installed by the nix-shell.

Perhaps everyone else understood this.  I did not.

thedavidmeister:
@pjkundert yup :+1: 

and even more than this, everything you just said is only for the nodejs conductor, which is being split out

libsodium used by rust is handled by the build script - if _that_ is not working for the rust conductor that is definitely an issue to fix

thedavidmeister:
nix comes in many flavours (and probably some i don't know about)

- nixos: an operating system
- nix alpine docker: an alpine docker box with nix pre-installed
- nix build: manages deps/environment for building/compiling things
- nix shell: manages deps/environment for a cli shell
- nix container: lightweight docker compatible containers with a nixos inside
- nix repl: a repl for the nix functional language

thedavidmeister:
@pjkundert 

OK, so let's see if i'm understanding you right here

the _urgent_ part of this is getting you unblocked for holofuel, is that satisfied by you personally using nix-shell in the immediate term?

the _important_ part of this is debian support, which we don't explicitly build against right now. i agree that it's something we need to do. @zippy and i have added something to the SOA tree for the short term which is a debian "cold build" (we have ubuntu atm) that would need to pass on develop before we cut a new release. the cold build does a full installation and test process from a fresh OS (which is very slow so we don't do it on every PR, but is OK for a final release QA process)

thedavidmeister:
fwiw i just tested 1.0.17 using the hashes in this branch in build.rs on a cold ubuntu and it doesn't help the node conductor

static version downloaded by build.rs = all rust compiling
shared version = node/neon compiling

thedavidmeister:
@pjkundert 

so we only officially support ubuntu bionic, not all ubuntu atm and we don't support debian
ubuntu bionic installs the correct libsodium with `apt-get install libsodium23`

we "sort of" support xenial because of a quirk of travis that we noticed recently, where it uses xenial when asked for bionic :roll_eyes: but that's entirely an accident, and moving to circle fixes this bug and removes our xenial coverage until we setup a cold build for it

i get the desire/need to support more linux distros and older releases (e.g. xenial), it's just that attempting to support them before we have infra/builds to show what works/breaks is putting the cart before the horse imo

e.g. i have no way of knowing whether some PR breaks/fixes some debian issue that you're reporting, so what would i really be reviewing? :man_shrugging: 

the difference between fixed/broken can be very subtle, especially when working locally (e.g. the 1.0.17 apparent fix that turned out to be an illusion), so yeah... need some builds to be spun up

_luckily_ i don't think it is that hard to get said builds happening against develop, thanks to docker! i might even get some time to tackle this over the weekend :)

thedavidmeister:
ugh, and to make things more annoying, looks like a cold ubuntu build is running into this too on develop

i just ran through it in bionic docker on a hunch >.<

more environment swizzling...

thedavidmeister:
... i'm gonna work on a PR that brings sodium into the repo i think ...

pjkundert:
Well, I was actually unblocked yesterday, when I figured out the minimum libsodium API versioning (1.0.12+), and environment variables (RUST_SODIUM_LIB_DIR) required to get nodejs_container to build using the system libraries.  However, I was under the impression that this was incorrect -- that it was architected to use the download/build version.  Now that I know that every build of nodejs_container has been just using the OS version of libsodium, I can carry on doing that!

thedavidmeister:
@pjkundert awesome

i'm still going to keep digging tho, but glad to hear it can wait a few days at least :)

pjkundert:
Ya, this pull request is probably not the approach we want to take; I'll go ahead and close it.

PR #1088, opened 2019-03-06
Added temp hotfix for tests on macOS

sphinxc0re:


AshantiMutinta:
Additional : We have changed the timeout sleep to 40000 and it passes on windows but something weird is going on there. They update and remove tests work quickly on app_spec but are very time variant on windows. The issue is with the sleep, it is best to move out all of integration tests into app_spec as it has the waiter but for a hotfix this is fine.

PR #1090, opened 2019-03-06
added new format for json rpc call into conductor

zippy:

Adds a `call` route to the json rpc for the conductor for making zome calls.

### DIFF
--- a/conductor_api/src/interface.rs
+++ b/conductor_api/src/interface.rs
@@ -28,6 +28,20 @@ use serde_json::map::Map;
 pub type InterfaceError = String;
 pub type InstanceMap = HashMap<String, Arc<RwLock<Holochain>>>;
 
+/// An identifier for an instance that is usable by UI in making calls to the conductor
+/// this type allows us to implement this identifier differently, i.e. as a DNA/agent ID pair, etc
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Hash, Eq)]
+pub struct PublicInstanceIdentifier(String);
+
+/// A mapper type between the public identifier and the conductor config identifier string
+pub type PublicInstanceMap = HashMap<PublicInstanceIdentifier, String>;
+
+impl From<String> for PublicInstanceIdentifier {
+    fn from(s: String) -> PublicInstanceIdentifier {
+        PublicInstanceIdentifier(s)
+    }
+}
+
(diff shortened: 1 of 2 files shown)


conductor_api/src/interface.rs
```diff
@@ -93,9 +109,61 @@ impl ConductorApiBuilder {
...
         *self.io
     }
 
+    /// Adds a "call" method for making zome function calls
+    fn setup_call_api(&mut self) {
+        let instances = self.instances.clone();
+        let instances_map = self.instances_map.clone();
+        self.io.add_method("call", move |params| {
+            let params_map = Self::unwrap_params_map(params)?;
+            let public_id_str = Self::get_as_string("id", &params_map)?;
```
willemolding: A little bit unsure about using `"id"` to identify the instance. Only because it might be confused with the JSON-RPC field also called `"id"` which is a unique identifier for each call.

Maybe call it `"instanceId"` instead but this might also be misleading

zippy: interesting. all the admin routes use "id" except the one that have to distinguish between interface_id and instance_id.  So I think for now I'll leave it, until will decide to make all of those also be "instance_id".

maackle: I'm also a little unsure about this. @zippy the difference here is that the admin routes are all REST-like, e.g. `admin/instance/add` where it's clear what thing you're dealing with. `call` doesn't suggest a particular class of things that the ID would be associated with. That's why `id` feels a little ambiguous here.

conductor_api/src/interface.rs
```diff
@@ -77,6 +91,7 @@ macro_rules! conductor_call {
 /// with spawn() to retrieve the IoHandler.
 pub struct ConductorApiBuilder {
     instances: InstanceMap,
+    instances_map: PublicInstanceMap,
```
maackle: Could you rename this to something else, or rename `instances` above? It's confusing that `instances` has type `InstanceMap`, and then there's something else called `instances_map`. Maybe `instance_id_map`?

conductor_api/src/interface.rs
zippy: Ok.  That makes sense.

conductor_api/src/interface.rs
zippy: will do.

PR #1100, opened 2019-03-07
Added basic hdk signature verification

sphinxc0re:


lucksus:
This is a WIP, right?

wasm_utils/src/api_serialization/verify_signature.rs
```diff
@@ -0,0 +1,8 @@
+use holochain_core_types::{error::HolochainError, json::*};
+
+#[derive(Deserialize, Default, Clone, PartialEq, Eq, Hash, Debug, Serialize, DefaultJson)]
+pub struct VerifySignatureArgs {
+    pub pub_key: String,
```
ddd-mtl: There is a `Base32` type in `hc_dpki` that you can use for public keys.

core/src/context.rs
```diff
@@ -276,6 +274,20 @@ impl Context {
         }
     }
 
+    pub fn verify_signature(
```
lucksus: Why is this function in `Context`? It doesn't even use `self`. I think these three lines should go straight into `invoke_verify_signatureinvoke_verify_signature` instead of calling this function on `context`.

For context ;), `Context::sign` is in here because it wraps the conductor API callback in way that code deeper in core does not need to handle JSONRPC. But verification can happen locally, so I don't see why we would want to put it in `Context`.

wasm_utils/src/api_serialization/verify_signature.rs
```diff
@@ -0,0 +1,8 @@
+use holochain_core_types::{agent::Base32, error::HolochainError, json::*};
+
+#[derive(Deserialize, Default, Clone, PartialEq, Eq, Hash, Debug, Serialize, DefaultJson)]
+pub struct VerifySignatureArgs {
```
zippy: Lets make the arguments have a `Provenance` and a payload, not the pub_key and the signature.  I think this is semantically more accurate, and also, that way when we merge `crypto-conductor-1` then the verify function will just be super easy: `provenance.verify(payload)`  Because that's all ready [been added there](https://github.com/holochain/holochain-rust/blob/912b56dbae93b1a49b0c7f00cebd690e1afcf174/dpki/src/utils.rs#L17).

core/src/context.rs
zippy: I agree with this in principle, but I think it will just get handled if we do what I asked below of having the arg be a provenance, because verify is defined on a provenance.

wasm_utils/src/api_serialization/verify_signature.rs
sphinxc0re: Good idea. I would like to get the other PR merged before though. I would rather do this as a follow-up

core/src/context.rs
```diff
@@ -286,7 +282,7 @@ impl Context {
                 .agent()
                 .chain_store()
                 .iter_type(&Some(top), &EntryType::CapTokenGrant)
-                .nth(0)?
+                .next()?
```
lucksus: What is the difference between `next()` and `nth(0)`? Why did you change this? Does `nth(0)` break anything wrt signature verification?

PR #1101, opened 2019-03-07
Small fixes

sphinxc0re:

### DIFF
--- a/net/src/ipc/transport_wss/mod.rs
+++ b/net/src/ipc/transport_wss/mod.rs
@@ -292,6 +292,7 @@ impl<T: Read + Write + std::fmt::Debug> TransportWss<T> {
                 // This seems to be wrong. Messages shouldn't be drained.
                 let msgs: Vec<Vec<u8>> = info.send_queue.drain(..).collect();
                 for msg in msgs {
+                    // TODO: fix this line! if there is an error, all the remaining messages will be lost!
                     socket.write_message(tungstenite::Message::Binary(msg))?;
                 }
 
--- a/net/src/ipc/transport_wss/mod.rs
+++ b/net/src/ipc/transport_wss/mod.rs
@@ -289,10 +289,12 @@ impl<T: Read + Write + std::fmt::Debug> TransportWss<T> {
                 Ok(())
             }
             WssStreamState::Ready(mut socket) => {
+                // This seems to be wrong. Messages shouldn't be drained.
                 let msgs: Vec<Vec<u8>> = info.send_queue.drain(..).collect();
                 for msg in msgs {
                     socket.write_message(tungstenite::Message::Binary(msg))?;
                 }
+
                 match socket.read_message() {
                     Err(tungstenite::error::Error::Io(e)) => {
                         if e.kind() == std::io::ErrorKind::WouldBlock {
--- a/net/src/ipc/transport_wss/mod.rs
+++ b/net/src/ipc/transport_wss/mod.rs
@@ -85,9 +85,9 @@ impl<T: Read + Write + std::fmt::Debug> Transport for TransportWss<T> {
         let host_port = format!(
             "{}:{}",
             uri.host_str()
-                .ok_or(TransportError("bad connect host".into()))?,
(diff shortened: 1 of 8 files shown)


net/src/ipc/transport_wss/mod.rs
```diff
@@ -85,9 +85,9 @@ impl<T: Read + Write + std::fmt::Debug> Transport for TransportWss<T> {
         let host_port = format!(
             "{}:{}",
             uri.host_str()
-                .ok_or(TransportError("bad connect host".into()))?,
+                .ok_or_else(|| TransportError("bad connect host".into()))?,
             uri.port()
-                .ok_or(TransportError("bad connect port".into()))?,
+                .ok_or_else(|| TransportError("bad connect port".into()))?,
```
lucksus: Why do you consider this change an improvement?

sphinxc0re: Because the error creation and string conversion is done lazily this way

ddd-mtl:
Make sure to run `cargo run -p holochain_test_bin --bin holochain_test_bin`
to test that your changes in `net` don't break anything.

zippy:
Hey @ddd-mtl is there any reason why we can't get that into CI, at least into the CI that runs when we merge to develop?

ddd-mtl:
@zippy Yes, its because it requires a local install of n3h. I believe @neonphog is looking at a work around. A solution could be to to have n3h as git submodule.
For now the best way to not worry about it is to not touch anything in `net`

net/src/ipc/transport_wss/mod.rs
lucksus: But that doesn't make any difference in this line, or does it? So in what case would then use `.ok_or`? I would use it when the creation of the alternative has no side-effects and is basically just calling a constructor, like in this example.

net/src/ipc/transport_wss/mod.rs
```diff
@@ -271,60 +271,56 @@ impl<T: Read + Write + std::fmt::Debug> TransportWss<T> {
...
-                return Ok(());
+                Ok(())
             }
             WssStreamState::Ready(mut socket) => {
                 let msgs: Vec<Vec<u8>> = info.send_queue.drain(..).collect();
                 for msg in msgs {
-                    if let Err(e) = socket.write_message(tungstenite::Message::Binary(msg)) {
-                        return Err(e.into());
-                    }
+                    socket.write_message(tungstenite::Message::Binary(msg))?;
```
neonphog: uhh... this logic is wrong (it was wrong before your change too).

Can you add a comment or something? since it drains them, if an error happens, all the remaining messages will be lost.

sphinxc0re:
Okay, the test binary can't seem to find the `logs` directory in a seemingly temporary directory (`/var/folders/bc/mzqtfdmd69zb9vhx2gfyw6nh0000gn/T/.tmppnmUUK/` on macOS).

Other than that, it seems to work

net/src/ipc/transport_wss/mod.rs
sphinxc0re: You can use `ok_or` if you have a static value that doesn't need evaluation at runtime

net/src/ipc/transport_wss/mod.rs
```diff
@@ -271,60 +271,58 @@ impl<T: Read + Write + std::fmt::Debug> TransportWss<T> {
...
+                Ok(())
             }
             WssStreamState::Ready(mut socket) => {
+                // This seems to be wrong. Messages shouldn't be drained.
                 let msgs: Vec<Vec<u8>> = info.send_queue.drain(..).collect();
                 for msg in msgs {
-                    if let Err(e) = socket.write_message(tungstenite::Message::Binary(msg)) {
-                        return Err(e.into());
-                    }
+                    socket.write_message(tungstenite::Message::Binary(msg))?;
```
lucksus:
```suggestion
                    // TODO: fix this line! if there is an error, all the remaining messages will be lost!
                    socket.write_message(tungstenite::Message::Binary(msg))?;
```

PR #1102, opened 2019-03-07
skip running tests during the build jobs

Connoropolous:


Connoropolous:
by that do you mean all the circle AND travis jobs, before deploying?

I feel like all that happens in the stages before we tag... isn't that in the release process docs itself?

And right now, all those jobs just happen concurrently to the builds, not as a pre-step anyways. Not oppose to what you're saying, just not sure how it would work

thedavidmeister:
@Connoropolous yeah, so realistically shifting as much as possible to circle and having the deploy triggered automatically as the builds pass

we definitely have builds that run only on develop atm, and rely on manual review/tagging to kick off each step

i dunno the exact workflow at that point, but it seems like we would want to use proper workflows that our CI gives us like https://circleci.com/docs/2.0/workflows/ to automate a lot of what we're doing by hand right now

i'm not really opposed to this change, it feels like a bit of a bandaid as long as we have like 20 manual steps every release :/

PR #1104, opened 2019-03-09
crypto and key-management api for DeepKey: Keystore & PassphraseManager

zippy:
This PR 

1. creates a `Keystore` implementation in `dpki` crate which has the functions necessary for crypto to be usable both by DPKI and other apps that want to do secure crypto functions.  The work is based on this spec:
https://hackmd.io/tVy22HhJQPuHO28gtD_UTQ?both
and this diagram:
https://realtimeboard.com/app/board/o9J_kycnaks=/
2. creats a PassphraseManager service/trait for getting passphrases from somewhere (either testing environments, or UIs) 
3. Makes Keystores persistable (with encryption using the passphrase)

What it does not do, and needs to be done in future prs is:
1. Actually implement a UI passphrase service
2. Wrap the keystore functions in hdk functions available to running in zomes.
3. Follow the designed bootup sequence, using an instance of DeepKey to actually generate agent keys instead of using our current `hc keygen` solution


dpki/src/seed.rs
```diff
@@ -26,10 +28,10 @@ pub enum SeedType {
...
     /// Revocation seed
     Revocation,
-    /// Device specific seed
-    Device,
-    /// Derivative of a Device seed with a PIN
-    DevicePin,
+    /// Indexed seed
+    Indexed,
+    /// Derivative of a Indexed seed with a PIN
+    IndexedPin,
```
neonphog: mmm... in the whole flow, there are several steps that include indexing... so this change might be confusing.

i.e. root -> device(indexed) -> devicePin(pin hashed) -> application(indexed)

dpki/src/key_blob.rs
```diff
@@ -287,13 +287,13 @@ impl Blobbable for KeyBundle {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::key_bundle::tests::*;
+    use crate::{key_bundle::tests::*, seed::generate_random_seed_buf, SEED_SIZE};
```
neonphog: this file got marked executable `100644 → 100755`

dpki/src/key_bundle.rs
```diff
@@ -73,7 +73,7 @@ impl KeyBundle {
 #[cfg(test)]
 pub(crate) mod tests {
     use super::*;
-    use crate::SIGNATURE_SIZE;
+    use crate::{seed::generate_random_seed_buf, SIGNATURE_SIZE};
```
neonphog: file marked executable `100644 → 100755`

dpki/src/keypair.rs
```diff
@@ -3,6 +3,7 @@
 use crate::{
     key_bundle,
     password_encryption::{self, PwHashConfig},
+    seed::generate_random_seed_buf,
```
neonphog: file marked executable `100644 → 100755`

dpki/src/seed.rs
```diff
@@ -38,11 +40,33 @@ pub enum SeedType {
     Mock,
 }
 
+pub struct SeedContext {
+    inner: String,
```
neonphog: If this is non-public anyways, it might be more efficient to store this as the bytes... and use `b""` strings, a'la:

```rust
let inner: [u8; 8] = *b"HCDEVICE";
```

dpki/src/keystore.rs
```diff
@@ -0,0 +1,330 @@
...
+    /// adds a random root seed into the keystore
+    #[allow(dead_code)]
+    pub fn add_random_seed(&mut self, id_str: &str, size: usize) -> HcResult<()> {
+        let id = id_str.to_string();
+        if self.keys.contains_key(&id) {
+            return Err(HolochainError::ErrorGeneric(
+                "identifier already exists".to_string(),
+            ));
+        }
+        let seed_buf = generate_random_seed_buf(size);
```
neonphog: the other functions use `SEED_SIZE` should we simplify this api?

dpki/src/keystore.rs
```diff
@@ -0,0 +1,330 @@
...
+        let secret = Arc::new(Mutex::new(Secret::RootSeed(RootSeed::new(seed_buf))));
```
neonphog: How do we know this was intended to be a root seed? Perhaps the user just wishes to create a one-off application that doesn't follow the seed derivation (not sure why they'd do this...) but, I was imagining this store just being a map of `[u8; 32]`s that could have arbitrary meaning.

dpki/src/keystore.rs
```diff
@@ -0,0 +1,330 @@
...
+                    Arc::new(Mutex::new(Secret::IndexedSeed(seed)))
+                }
+                _ => {
+                    return Err(HolochainError::ErrorGeneric(
+                        "source secret is not a root seed".to_string(),
+                    ));
+                }
+            }
+        };
+        let _ = self.keys.insert(dst_id, secret);
```
neonphog: what's the unused binding for?

dpki/src/keystore.rs
```diff
@@ -0,0 +1,330 @@
...
+
+    /// adds a keypair into the keystore based on a seed already in the keystore
+    /// returns the public key
+    #[allow(dead_code)]
+    pub fn add_key_from_seed(
+        &mut self,
+        src_id_str: &str,
+        dst_id_str: &str,
+        context: &SeedContext,
+        index: u64,
```
neonphog: I feel like we shouldn't dual purpose this. If we provide access to discreet functions, then the app dev can make use of them how they want:
- derivation (seed + context + index = new_seed)
- key generation (seed = keypair)

then, the base32 key id (which for an application key is the agentId) could be the key in the keystore that stores the private side)

dpki/src/keystore.rs
```diff
@@ -0,0 +1,330 @@
...
+
+pub enum Secret {
+    Key(KeyBundle),
+    RootSeed(RootSeed),
+    IndexedSeed(IndexedSeed),
+}
+
+#[allow(dead_code)]
+struct Keystore {
+    keys: HashMap<String, Arc<Mutex<Secret>>>,
```
neonphog: Do you think it would be worth adding in a stub for requesting the passphrase (it can return a test string or something), then either stubs or real implementations of the encryption / decryption / disk persistence?

I don't know how realistic interacting with a `HashMap<String, Arc<Mutex<Secret>>>` is for the target solution.

dpki/src/keystore.rs
zippy: well actually, the reason I did this is because in https://hackmd.io/tVy22HhJQPuHO28gtD_UTQ?both you have the "size" param for `keystore_new_random` so I made the generalized function in keystore to use that.  If we don't need to give the end-user access to the size of the random seed, then I'll take it out.

dpki/src/keystore.rs
zippy: Well as I said I had assumed that most of these functions were supposed to be wrappers of what had been written in the dpki crate, so I didn't know which else to use and just made the assumption.  Are you thinking that I should simplify this and not even use the seed types that were created in the crate?  And if so then move the derivations directly into the functions I implemented and not use those Seed type implementations in the Secret variants?

 (And if they aren't supposed to be wrappers of those functions, then who is actually using those functions?)

dpki/src/seed.rs
zippy: I think there's something that I'm still fundamentally confused about regarding what was implemented in the crate and how to use it to implement the functions that are supposed to be exposed as per the spec.  I know we are using derived indexed seeds for device seeds.  But the API is allowing you to do derived indexed seeds for any purpose.  So that's why I renamed them to indexed.   And since (as far as I can tell) these functions are the only clients of the crate's functions, I thought it made sense.  But I think I may have this wrong.

dpki/src/seed.rs
zippy: done

dpki/src/keystore.rs
zippy: for some reason I thought I had to have that, I've cleaned it up...

dpki/src/keystore.rs
zippy: I'm sorry I really don't understand this comment.  Where does this function dual purpose anything?  Is it because I was allowing creating a key from the RootSeed?  That was my attempt at allowing people access to creating keys directly from random_seeds and not having to do derivation if they didn't want to.  But I could change that.

dpki/src/keystore.rs
zippy: The passphrase doesn't happen at the keystore level.  There is a single keystore is per-instance.  The keystore manager does the passphrase.

dpki/src/seed.rs
zippy: Renamed back to the old way and pulled out a common function that can be used both by the HCadmin used of the crate and Keystore.

dpki/src/keystore.rs
zippy: removed now and there is just a direct buf stored as the seed.

conductor_api/src/keystore.rs
```diff
@@ -0,0 +1,763 @@
...
+
+    /// adds a keypair into the keystore based on a seed already in the keystore
+    /// returns the public key
+    #[allow(dead_code)]
+    fn add_key_from_seed(
+        &mut self,
+        src_id_str: &str,
+        dst_id_str: &str,
+        context: &SeedContext,
+        index: u64,
```
neonphog: I still feel like we need to delete `context` and `index` from this function to remove duplication.

Implementors can make two calls:

`add_seed_from_seed("device_pin", "app_seed_42", b"HCAPPLIC", 42);`

then

`add_key_from_seed("app_seed_42", "app_sign_42", KeyType::Signing);`

conductor_api/src/keystore.rs
```diff
@@ -0,0 +1,763 @@
...
+    Encrypting,
+}
+
+#[allow(dead_code)]
+#[derive(Serialize, Deserialize)]
+pub struct Keystore {
+    passphrase_check: String,
+    secrets: BTreeMap<String, KeyBlob>,
+    #[serde(skip_serializing, skip_deserializing)]
+    cache: HashMap<String, Arc<Mutex<Secret>>>,
```
neonphog: curious `secrets` and `cache` are both indexed by the same id strings, why is one a btree and the other a hash map?

conductor_api/src/keystore.rs
```diff
@@ -0,0 +1,763 @@
...
+            )?),
+            _ => {
+                return Err(HolochainError::ErrorGeneric(format!(
+                    "Tried to decrypt unsupported BlobType in Keystore: {}",
+                    id_str
+                )));
+            }
+        };
+        self.cache
+            .insert(id_str.clone(), Arc::new(Mutex::new(secret)));
```
neonphog: I'm uncertain about having a cache here. Part of the point of the passphrase manager is to give users control over "locking" their authorizor capabilities after an expiration timer, but if they authorize something and their authorizor key gets cached here, then we have to coordinate expiring the cache when the passphrase expires?

conductor_api/src/keystore.rs
```diff
@@ -0,0 +1,763 @@
...
+        let mut decrypted_buf = decrypt_with_passphrase_buf(
+            &self.passphrase_check,
+            &mut passphrase,
+            self.hash_config.clone(),
+            PCHECK_SIZE,
+        )?;
+        let mut decrypted_header = SecBuf::with_insecure(PCHECK_HEADER_SIZE);
+        let decrypted_buf = decrypted_buf.read_lock();
+        decrypted_header.write(0, &decrypted_buf[0..PCHECK_HEADER_SIZE])?;
+        let mut expected_header = SecBuf::with_secure(PCHECK_HEADER_SIZE);
```
neonphog: I'm not sure I'm following this passphrase header bit. Perhaps some comments? also, one is in a secure buffer and one is in an insecure buffer. If they are expected to be the same data, probably should have the same security.

conductor_api/src/keystore.rs
lucksus: `cache` is only held in memory -> sorting is irrelevant. `secrets` is persisted to files. `BTreeMap` makes sure that the sorting and hence the exact serialized representation is deterministic.

@zippy wrote the `HashMap` with unencrypted `Secret`s and I left it in and just renamed it to `cache` when adding the `BTreeMap`. I default to `BTreeMap` when it gets persisted after wasting hours in trying to figure out why DNA hashes were different, non-deterministically. It was the non-determinism of `HashMap`.

conductor_api/src/keystore.rs
lucksus: Hm, I get that.

But we do need to hold some secrets unencrypted indefinitely: agent keys in particular.

I would say we might need to add more complexity around this cache so that agent keys stay in, even after the PassphraseManager locks itself up. But everything else gets cleared from the cache at some point, as a default that applies in particular to seeds.. Maybe dependent on the secret type? Or everything gets cleared except for `PRIMARY_KEYBUNDLE_ID`? What do you think?

conductor_api/src/keystore.rs
lucksus: So, we are encrypting `PCHECK_HEADER` followed by 32 random bytes with the passphrase and are storing this cipher in `Keystore::passphrase_check`. Everytime the keystore gets modified, we have to encrypt the new secret but need to make sure it is encrypted with the same passphrase. Before using the provided passphrase to encrypt the new secret, we try to decrypt `passphrase_check` and expect to get back our `PCHECK_HEADER` followed by some random bytes. We only compare `PCHECK_HEADER` and ignore the random bytes during check.

Yeah, I don't think this second `SecBuf` needs to be secure..

Will add comments.

conductor_api/src/keystore.rs
zippy: Hmmm. So are you saying that I've added an extra derivation call to `kdf::derive` (via the call to `generate_derived_seed_buf` at line 308)  which is unnecessary, and that all we need to do is used the key returned from the src seed?  I guess it would be a very easy fix.

zippy: Ok.  Removed the extra derive.

conductor_api/src/keystore.rs
neonphog: :+1:

conductor_api/src/keystore.rs
neonphog: I guess I was imagining anything that needed persistence like for the app keys handling it on that end... i.e. keeping a copy of the secret after requesting it. But definitely something we can iterate on to see what works.

conductor_api/src/keystore.rs
neonphog: :+1:

PR #1105, opened 2019-03-10
use upstream rust_sodium-sys (0.0.6 blocker)

thedavidmeister:

## context

https://hackmd.io/8QrSSc3IRnOsgO5M-h83lA#

## proposed fix

our bespoke fork of `rust_sodium-sys` injects compiler flags that don't help and might hurt, and maybe other things?

it's hard to track upstream with our fork, or what exactly we changed or why...

roll back to the upstream `rust_sodium-sys` from maidsafe and configure it to work across all the platforms we support

if we disable pie then we can support older ubuntu/debian/andriod/etc: 
- https://github.com/wireapp/libsodium-neon/issues/15 
- https://github.com/maidsafe/rust_sodium/pull/40 
- https://askubuntu.com/questions/911538/disable-pie-and-pic-defaults-in-gcc-on-ubuntu-17-04

this lovely error:

```
= note: /usr/bin/ld: /tmp/rustc.ZbjniGJ7D9Ug/librust_sodium_sys-24ee5a8b5f6825c1.rlib(libsodium_la-aead_chacha20poly1305.o): relocation R_X86_64_PC32 against symbol `cr
ypto_aead_chacha20poly1305_encrypt_detached' can not be used when making a shared object; recompile with -fPIC
```
appears on some linux OS even if we disable pie and add `-fPIC` until we also `apt-get install build-essential`, so include that in the auto install script

we know adding the `-fPIC` flag does not help without `build-essential`, because our fork hardcoded this flag but still shows these errors :(

we haven't been testing debian or older ubuntu (except accidentally on travis), and they were breaking as reported by @pjkundert in https://github.com/holochain/holochain-rust/pull/1084, so add cold builds for those

the end result is:

- some small additional `apt-get` overhead
- reliable (mostly, see below) libsodium builds
- no more bespoke forks to manage, clear upgrade path from upstream, etc.
- better/easier cross platform support than the old approach of shivving in shared libs with env variables
- no more reliance on shared versions of libsodium
- unblocks linking issues blocking `0.0.6` release due to failing ubuntu cold builds on `develop` https://circleci.com/gh/holochain/holochain-rust/10507

## some flakiness still exists

upstream `rust_sodium-sys` still downloads libsodium `1.0.16`, checksums it then configures/builds that

it is a thin wrapper around https://libsodium.gitbook.io/doc/installation#compilation-on-unix-like-systems

this means that __the download can still fail and give a bad checksum__ leading to flakiness in the build task

this pain could be mitigated in CI at least by triggering this build at the start of CI, instead of ~20-30 minutes into cold builds, so a manual rebuild can be noticed, kicked off asap and easily monitored

## changes

- removes our fork of `rust_sodium-sys` and uses upstream
- sets `RUST_SODIUM_DISABLE_PIE` where it is needed in `build_nodejs_conductor.sh`
- adds `apt-get install build-essential gnupg apt-transport-https` for ubuntu/debian support
- adds debian stable support and cold builds
- tests both ubuntu `xenial` and `bionic` cold builds (both are LTS)
- adds dry run of node build for windows
- add section in README that strongly recommends rerunning the install script when upgrading to a new tagged holochain release
- removes ad-hoc "release install" scripts and uses the auto script for everything

works with:

- [x] ubuntu bionic https://circleci.com/gh/holochain/holochain-rust/10625
- [x] ubuntu xenial https://circleci.com/gh/holochain/holochain-rust/10623
- [x] nixos https://circleci.com/gh/holochain/holochain-rust/10621
- [x] mac https://circleci.com/gh/holochain/holochain-rust/10626
- [x] windows https://travis-ci.com/holochain/holochain-rust/jobs/183642791
- [x] debian https://circleci.com/gh/holochain/holochain-rust/10629

## manual deploy step

__DON'T DEPLOY UNTIL THE COLD BUILDS ARE BACK ON DEVELOP ONLY__


sphinxc0re:
We might have to ship the `libsodium.dll` on Windows. But regardless, this sounds like a good idea

sphinxc0re:
Which means, the release process could include obtaining the correct version of `libsodium` and packaging it with the binary.

thedavidmeister:
> We might have to ship the `libsodium.dll` on Windows. But regardless, this sounds like a good idea

@sphinxc0re not necessary, upstream covers windows and other platforms we haven't even touched yet like iOS - see https://github.com/maidsafe/rust_sodium/blob/master/rust_sodium-sys/build.rs

i added a dry run to show it working on windows under neon - https://travis-ci.com/holochain/holochain-rust/jobs/183642791

scripts/install/linux.sh
```diff
@@ -0,0 +1,39 @@
...
+# https://stackoverflow.com/questions/50757647/e-gnupg-gnupg2-and-gnupg1-do-not-seem-to-be-installed-but-one-of-them-is-requ
+sudo apt-get install -y gnupg
+
+sudo apt-get install -y \
+  libssl-dev \
+  pkg-config \
+  python2.7
+
+# hc deps
+apt-get install -y qt5-default;
```
Connoropolous: do we still need this? maybe it could be removed, but in a separate PR if not

thedavidmeister:
oh noes, @zippy you didn't see my warning

> DON'T DEPLOY UNTIL THE COLD BUILDS ARE BACK ON DEVELOP ONLY

thedavidmeister:
@neonphog upstream PRs:

libsodium v1.0.17 - https://github.com/maidsafe/rust_sodium/pull/101
better docs - https://github.com/maidsafe/rust_sodium/pull/100

@zippy cold builds back on develop https://github.com/holochain/holochain-rust/pull/1106

thedavidmeister:
upstream issue to discuss putting libsodium directly in the crate rather than relying on a flaky download step https://github.com/maidsafe/rust_sodium/issues/102

PR #1110, opened 2019-03-12
holochain_net: Update tests for the gossip rework

ddd-mtl:


net/src/ipc_net_worker.rs
```diff
@@ -78,7 +78,7 @@ impl IpcNetWorker {
             .map(|(k, v)| (k.to_string(), v.as_str().unwrap().to_string()))
             .collect();
         // create a new IpcNetWorker witch spawns the n3h process
-        IpcNetWorker::priv_new_with_spawn(
+        return IpcNetWorker::priv_new_with_spawn(
```
lucksus: Unnecessary `return`, this just got fixed by @sphinxc0re. We should keep it this way to have less work when turning on clippy again.

ddd-mtl: How many times do you want to have this conversation?
Its not a 'fix', its worse.
Newlines are also unnecessary, should we remove them also?

maackle: Doesn't look like much of a "conversation" to me

lucksus:
Ehm, you just merged this without having two approvals and despite my change request, @ddd-mtl. And by overriding my change request you inserted a style that is not backed by the team but rather your own personal preference which keeps us away from using community tooling which, if it hadn't been switched off earlier, would have saved us many ours of debugging.

net/src/ipc_net_worker.rs
ddd-mtl: @maackle That's because you were not part of core dev when those conversations happened. I explained this at least 3 times, and we also had lots of conversations on clippy.
During mtl hodlpodl we even spent an hour collectively deciding on clippy rules. Result is here:
https://trello.com/c/Pk9f9j7F/18-the-clippy-linter-enforces-requirements-that-not-all-buy-into

As you can see, we decided to disable style lints. 'needless return' is part of them.

ddd-mtl:
@lucksus 
Since I know and trust you Nico, I know you really believe in the assertions you just stated.
I invite you to take a step back, read your statements and ask yourself if what you just said is really the truth?

Here are some facts:
1. My code in this PR respects the 'team' style, since it is enforced by rustfmt in CI.
2. There has never been a team agreement on 'team' style. You enforced rustfmt defaults.
3. I have accepted those defaults despite not being my preference. [Proof](https://github.com/holochain/holochain-rust/pull/211)
4. There has been a team agreement on clippy configuration during mtl hodlpodl, which has not been enforced ([Proof](https://trello.com/c/Pk9f9j7F/18-the-clippy-linter-enforces-requirements-that-not-all-buy-into)).
5. You and Julian are manually enforcing clippy lint rules that are contrary to that team agreement (proof: your change request in this PR; Julian's 'small fixes' PR)
6. There are plenty of PRs that have been merged without two approvals.

 `net` and `test_bin` crates are not core dev team's responsiblity so there is no reason for you to enforce  core dev team's rules on those crates. I'm pretty sure you have more urgent and important things to do.

net/src/ipc_net_worker.rs
maackle: @ddd-mtl at the last core meeting in Dortmund, it seemed that everybody on that team without exception was in favor of using clippy (someone correct me if I'm wrong about that), and the only thing halting the use of clippy now is the fact that it would take a long time to address all of the existing style discrepancies that have accrued -- that and the fact that net code is part of the holochain-rust repo.

So, since months have passed since those initial conversations, and since the makeup of the core team has changed, it does seem like this issue is ready to be revisited, since it has naturally arisen here. I don't think we can pretend that a Trello card from 5 months ago that starts "there are differences on the team" can be construed as a firm decision.

I'm focused on Holo at the moment so I'm not even working closely with core team now, but am commenting as a bystander now because I have spent months on core and am familiar with the issue.

ddd-mtl: "I don't think we can pretend that a Trello card from 5 months ago that starts "there are differences on the team" can be construed as a firm decision."

@maackle 
I wrote in that card the agreement the team concluded after one hour of in-person discussions.
This apparently has no value to you, and actually proves my point that having conversations to solve issues with this team is pointless.
Hence my conclusion that since we are not going to agree on the issue at hand, the only solution I see is to consider `net` and `test_bin` as not part of devcore's responsility, and so, not under devcore's law.
Clippy can be deactivated for those crates, and if its not enough, I don't mind moving those crates in a separate repo.

PR #1121, opened 2019-03-14
SPIKE: testing windows docker

duffybelfield:


ddd-mtl:
If this is not conclusive, there is still appVeyor.

thedavidmeister:
@ddd-mtl yeah if we get stuck here we have options, appveyor and also raw AWS builds :)

thedavidmeister:
see https://github.com/holochain/holochain-rust/pull/1131

Issue #1017, opened 2019-02-15
Hide SecBuf secure alignment complexity


neonphog:
Secure sodium buffers can only be sized % 8 :

https://github.com/holochain/holochain-rust/blob/85e6bfff0b814922fd73968a47e7b3cd9ece2462/sodium/src/secbuf.rs#L70

We should hide this in the background, track the "used" size, and return slice references that are limited to that length, to make it usable for things like passphrases.

PR #1122, opened 2019-03-15
Passphrase manager

lucksus:
Screenshot from 2019-03-15 18-05-08


maackle:
Will this eventually address the event of a power cycle, where the machine can't join the network again without human input?

lucksus:
Hey @neonphog, I would love to have you have a quick look over this. Note that I'm currently using an insecure `SecBuf` because of the size constraint currently still present for secure ones. I think we need to change this before beta.

thedavidmeister:
> Note that I'm currently using an insecure SecBuf

mm i wondered about that

conductor_api/src/conductor/passphrase_manager.rs
```diff
@@ -0,0 +1,110 @@
...
+        let pm = PassphraseManager {
+            passphrase_cache: Arc::new(Mutex::new(None)),
+            passphrase_service,
+            last_read: Arc::new(Mutex::new(Instant::now())),
+            timeout_kill_switch: kill_switch_tx,
+        };
+
+        let pm_clone = pm.clone();
+
+        thread::spawn(move || loop {
```
neonphog: I think we're going to need to do an audit of system threads at some point... they should probably mostly be moved into some kind of pool executor scheme.

neonphog:
> Note that I'm currently using an insecure SecBuf

https://github.com/holochain/holochain-rust/issues/1017 strikes again!

conductor_api/src/conductor/passphrase_manager.rs
lucksus: :+1:

PR #1125, opened 2019-03-18
More helpful error message on missing public token

maackle:


zippy:
Interesting.  This last addition breaks a few tests.  Apparently we have tests on DNAs with no zomes!

zippy:
Looks like there's still an "Attempting to initialize DNA with zero zomes!"

core_types/src/error/error.rs
```diff
@@ -0,0 +1,369 @@
+use self::HolochainError::*;
```
maackle: Uh oh, did this file renamed and then re-added?

maackle: Removed it again in https://github.com/holochain/holochain-rust/pull/1240

PR #1136, opened 2019-03-19
update guidebook links to 0.0.7-alpha

Connoropolous:


Connoropolous:
yes and no... I have tried to depart from using 'latest' in links to documentation, since its kind of like pointing a link to develop branch on github with a code snippet, then that code snippet or whole file can just disappear. I would prefer to find an automated and easy way to swap values for versions in, during build time, cause even docs have a build step

PR #1137, opened 2019-03-19
Download / Run Pinned N3H Version

neonphog:

Changes: 
- Pins n3h to a specific version.
- If n3h is found in the path, and `n3h --version` responds with the correct version, it will be used.
- Otherwise a release binary will be downloaded to $HOME/.hc/n3h-binaries/$VERSION/` and run from there.
- The sha256 hashes of these download binaries are built into the net crate for security via the `pin_n3h` script.

For n3h development, this means:
- export PATH=your/n3h/path/bin:$PATH
- make sure the version in net/src/ipc/n3h_pin.json matches what is reported by your n3h code


net/src/ipc/n3h.rs
```diff
@@ -0,0 +1,294 @@
...
+    let hash = format!("{:x}", hash.result());
+
+    if &hash != sha256 {
+        tlog_e!("bad hash, expected {}, got {}", sha256, &hash);
+    }
+
+    &hash == sha256
+}
+
+/// hash a file, if it doesn't exist or fails, download && hash
```
ddd-mtl: `Download a missing file from disk. Check is done by hashing the file's content.`

net/src/ipc/n3h.rs
```diff
@@ -0,0 +1,294 @@
...
+        return false;
+    }
+
+    let hash = format!("{:x}", hash.result());
+
+    if &hash != sha256 {
+        tlog_e!("bad hash, expected {}, got {}", sha256, &hash);
+    }
+
+    &hash == sha256
```
ddd-mtl: Suggestion:
```Rust
    if &hash != sha256 {
        tlog_e!("bad hash, expected {}, got {}", sha256, &hash);
        return false
    }
    true
```

net/src/ipc/n3h.rs
```diff
@@ -0,0 +1,294 @@
...
+        );
+        exec_output(
+            "cp",
+            vec!["-a", "./dmg-mount/n3h.app", &dest.to_string_lossy()],
+            true,
+        )?;
+        tlog_d!(
+            "{}",
+            exec_output("hdiutil", vec!["detach", "./dmg-mount"], false)?
+        );
```
ddd-mtl: I believe it would better to separate cmd execution from logging. e.g.
```Rust
    let output = exec_output("hdiutil", vec!["detach", "./dmg-mount"], false)?;
    tlog_d!("{}", output);
```
That way you are resilient to future changes of logging doctrine.

conductor_api/src/conductor/base.rs
```diff
@@ -267,11 +267,6 @@ impl Conductor {
             network_config.n3h_persistence_path
         );
         let spawn_result = ipc_spawn(
-            "node".to_string(),
-            vec![format!(
-                "{}/packages/n3h/bin/n3h",
-                network_config.n3h_path.clone()
-            )],
```
ddd-mtl: \o/

net/src/ipc/n3h.rs
```diff
@@ -0,0 +1,294 @@
...
+}
+
+macro_rules! tlog_e {
+    ($($arg:tt)+) => {
+        log_ee!("get_verify_n3h", $($arg)+);
+    }
+}
+
+#[cfg(not(target_os = "windows"))]
+use std::os::unix::fs::OpenOptionsExt;
```
ddd-mtl: This doesn't seem to be used.

neonphog: one of those weird rust things... this makes the

```rust
open_opts.mode(0o755);
```

below be available on non-windows systems.

net/src/ipc/n3h.rs
neonphog: updating comment :+1:

net/src/ipc/n3h.rs
neonphog: :+1:

net/src/ipc/n3h.rs
neonphog: :+1:

conductor_api/src/conductor/base.rs
neonphog: :tada:

net/src/ipc/n3h.rs
```diff
@@ -0,0 +1,300 @@
...
+fn download(dest: &std::ffi::OsStr, url: &str, sha256: &str) -> NetResult<()> {
+    if check_hash(dest, sha256) {
+        return Ok(());
+    }
+    {
+        tlog_d!("downloading {}...", url);
+        let mut open_opts = std::fs::OpenOptions::new();
+        open_opts.create(true).write(true);
+        if cfg!(not(target_os = "windows")) {
+            open_opts.mode(0o755);
```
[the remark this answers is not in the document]
neonphog: Marks the file executable. Adding comment.

net/src/ipc/n3h.rs
ddd-mtl: 👍

neonphog:
> Before merging, please give me a short window to confirm that this is the right path to do this.
We have path mixups happening, like there is a .hc and a .holochain, and even a third path that is a default.

:+1:

pin_n3h
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env node
+
+/* Helper script for pinning downloadable version of n3h binaries.
+ * Bakes the sha256 hashes into the code for security.
+ *
+ * Run with the target n3h release name (tag name) e.g.
+ * `./pin_n3h adhoc-2019-03-19`
```
neonphog: @Connoropolous

> There is so far no documentation added to the repo itself on how to utilize the added script to update the pinned version.

Here's the documentation. Happy to put it somewhere else, too, if you have a suggestion where.

Connoropolous: Ok cool. I will have a quick think about where to put it, but in addition, I have the question:

who should update this version, and when? do you believe?

neonphog: Anyone iterating on n3h... I think the flow should be:

- bump n3h version in n3h feature branch
- manually bump pinned version in holochain-rust feature branch
- iterate
- code review
- code merged to n3h master
- adhoc n3h release / tag
- run n3h_pin script in holochain-rust feature branch
- code merged to holochain-rust develop

Then we no longer need to cut separate "releases" for n3h whenever holochain-rust releases are cut.

Connoropolous: thanks for laying this out. I think I'm picking up what you're putting down

net/src/ipc/n3h.rs
```diff
@@ -0,0 +1,294 @@
+//! check that n3h is in the path, or download it
```
Connoropolous: Actually, I believe it would be much better to do some logging in here, not just in the error cases, but the success case, to let the user know what is going on here

neonphog: there is indeed some logging... like when [downloading](https://github.com/holochain/holochain-rust/pull/1137/files#diff-487ce26ffed9d1889698c2d916ad0162R223) or [executing commands](https://github.com/holochain/holochain-rust/pull/1137/files#diff-487ce26ffed9d1889698c2d916ad0162R135).

What else would you like to see?

Connoropolous: Ah cool yeah sorry my eyes were looking for println s

neonphog:
@thedavidmeister - made some progress on this. It runs now on my copy of @samrose `s holoportOs nix image... but it doesn't run on circleci still, due to some funky docker permission I don't know how to fix:

```
# ./n3h-nix-runner.bash 

** (process:16772): ERROR **: 22:04:44.443: bind_mount: mount(source, target, "bind", MS_BIND | MS_REC, NULL): Permission denied
Trace/breakpoint trap
```

shell.nix
```diff
@@ -167,6 +167,8 @@ stdenv.mkDerivation rec {
   name = "holochain-rust-environment";
 
   buildInputs = [
+    # needed for appimage running
+    appimage-run which bash
```
neonphog: @thedavidmeister - fyi, needed to add this ^^

net/src/ipc/n3h.rs
```diff
@@ -0,0 +1,321 @@
...
+    download(path.as_os_str(), &artifact.url, &artifact.hash)?;
+
+    let path = if os == "mac" {
+        // we need to extract the dmg into n3h.app
+        extract_dmg(path.as_os_str(), &bin_dir)?
+    } else if os == "linux" && std::env::var("NIX_STORE").is_ok() {
+        // on nix, we need some extra appimage-run magic
+        exec_output("bash", vec!["-c", NIX_MKRUNNER], &bin_dir, false)?;
+        let mut path = bin_dir.clone();
+        path.push("n3h-nix-runner.bash");
```
neonphog: @thedavidmeister - fyi needed to add this ^^

thedavidmeister:
@neonphog ok, just working on reproducing this locally

works:

```
nix-shell
hc-test-all
```

works:

```
docker pull holochain/holochain-rust
docker run --rm -it holochain/holochain-rust bash
nix-shell
hc-test-all
```

doesn't work:

```
docker pull holochain/holochain-rust
nix-shell
circleci-cli local execute
```

but the last command is very slow :(

will do some more debugging tomoz, probably will need to keep a ssh session to circle open...

thedavidmeister:
it's possibly just some weirdness with setting the `$PATH` in CI (fairly common issue)...

neonphog:
super appreciate you taking a look @thedavidmeister ! let me know if I can help with anything.

thedavidmeister:
@neonphog got some more debug info

```
---- ipc::n3h::tests::can_dl_n3h stdout ----
[d] EXEC: "n3h" "--version"

response: Err(Os { code: 2, kind: NotFound, message: "No such file or directory" })
[e] Err(ErrorMessage { msg: "n3h not found in path: \"n3h\"" })

[d] downloading https://github.com/holochain/n3h/releases/download/adhoc-2019-03-21/n3h-0.0.7-alpha2-linux-x64.AppImage...

[d] EXEC: "bash" "-c" "cat > n3h-nix-runner.bash <<EOF\n#! $(which bash)\nexec $(which appimage-run) $(ls $(pwd)/n3h*.AppImage) \"\\$@\"\nEOF\nchmod a+x n3h-nix-runner.bash\n"

[d] EXEC: "/root/.local/share/holochain/n3h-binaries/v0.0.7-alpha2/n3h-nix-runner.bash" "--version"

response: Err(ErrorMessage { msg: "bad exit None \"\\n** (process:11347): ERROR **: 12:26:25.262: bind_mount: mount(source, target, \\\"bind\\\", MS_BIND | MS_REC, NULL): Permission denied\\n\"" })
thread 'ipc::n3h::tests::can_dl_n3h' panicked at 'called `Result::unwrap()` on an `Err` value: ErrorMessage { msg: "n3h not found in path: \"/root/.local/share/holochain/n3h-binaries/v0.0.7-alpha2/n3h-nix-runner.bash\"" }', src/libcore/result.rs:997:5
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

i'm drilling down on just the `can_dl_n3h` test in `n3h_debug` and adding prints

thedavidmeister:
- https://github.com/AppImage/AppImageKit/issues/828
- https://github.com/AppImage/AppImageKit/issues/841

samrose:
for what it's worth, it looks like you can do VM as executor type, in addition to Docker container

https://circleci.com/docs/2.0/executor-types/

thedavidmeister:
yes there is machine, but then how do we test our official docker boxes?

thedavidmeister:
as per discussions in chat, we should look at using `appimagekit` to build the app images so that we get access to the handy `--appimage-extract-and-run` flag

IIRC that can be used to workaround FUSE/mount issues in restricted environments

somehow detecting issues in the environment in the script then falling back to the extract option sounds like the most reliable/portable option if we can get it working :)

neonphog:
> as per discussions in chat, we should look at using appimagekit to build the app images so that we get access to the handy --appimage-extract-and-run flag

for clarity, we were already building the images with appimagekit... not sure there is an alternative.

Confirmed that the n3h release artifacts already include and can use the `--appimage-extract-and-run` flag. I'll update the code here to include it when running in a non-fuse environment.

PR #1141, opened 2019-03-20
Get headers network

AshantiMutinta:
- [x] changed to use tuples rather than just EntryWithMeta
- [x] use get_enty_with_workflow_meta for EntryValidationData conversions


Connoropolous:
if you're touching JsonPorotocol I believe this will require an accompanying change in n3h
cc @ddd-mtl

net/src/connection/json_protocol.rs
```diff
@@ -177,6 +179,9 @@ pub struct FetchEntryResultData {
     pub entry_address: Address,
     #[serde(rename = "content")]
     pub entry_content: serde_json::Value,
+
+    #[serde(rename = "chainHeader")]
+    pub headers: Vec<ChainHeader>,
```
lucksus: As @Connoropolous stated, this change would need a change to n3h as well (@ddd-mtl, @neonphog).

But I don't think we need to add this here. `entry_content` already includes the entry with meta (its a serialized `EntryWithMeta` struct). It could be an `EntryWithMetaAndHeader` ;)

core/src/nucleus/actions/get_entry.rs
```diff
@@ -66,15 +67,17 @@ pub(crate) fn get_entry_from_dht(
     context: &Arc<Context>,
     address: &Address,
 ) -> Result<Option<Entry>, HolochainError> {
-    let cas = context.state().unwrap().dht().content_storage();
+    let cas = context.state().unwrap().dht().content_storage().clone();
     get_entry_from_cas(&cas.clone(), address)
```
lucksus: Double clone?

core/src/workflows/get_entry_result.rs
```diff
@@ -17,20 +17,38 @@ pub async fn get_entry_with_meta_workflow<'a>(
...
     timeout: &'a Timeout,
-) -> Result<Option<EntryWithMeta>, HolochainError> {
+) -> Result<Option<EntryWithMetaAndHeader>, HolochainError> {
     // 1. Try to get the entry locally (i.e. local DHT shard)
-
     let maybe_entry_with_meta =
         nucleus::actions::get_entry::get_entry_with_meta(context, address.clone())?;
-    if maybe_entry_with_meta.is_some() {
-        return Ok(maybe_entry_with_meta);
+    if let None = maybe_entry_with_meta {
```
lucksus:
```suggestion
    // 2. No result, so try on the network
    if let None = maybe_entry_with_meta {
```

core/src/workflows/get_entry_result.rs
```diff
@@ -17,20 +17,38 @@ pub async fn get_entry_with_meta_workflow<'a>(
...
+        await!(network::actions::get_entry::get_entry(
+            context.clone(),
+            address.clone(),
+            timeout.clone(),
+        ))
+    } else {
+        let entry = maybe_entry_with_meta.ok_or(HolochainError::ErrorGeneric(
```
lucksus:
```suggestion
        // 3. If we've found the entry locally we also need to get the header from the local state:
        let entry = maybe_entry_with_meta.ok_or(HolochainError::ErrorGeneric(
```

core/src/workflows/get_entry_result.rs
```diff
@@ -68,22 +86,30 @@ pub async fn get_entry_result_workflow<'a>(
                 context
                     .state()
                     .expect("state uninitialized! :)")
-                    .get_headers(entry_with_meta.entry.address().clone())?
+                    .get_headers(entry_with_meta.entry_with_meta.entry.address().clone())?
```
lucksus: Why is this still here? Shouldn't this be replaced by the code that was added?

lucksus: In line 68 above, `get_entry_with_meta_workflow` now returns an `EntryWithMetaAndHeader`, so the header is already there..

Maybe that function and the local variable `maybe_entry_with_meta` should be renamed as well to reflect that the header is already included?

lucksus: So I think this would even overwrite the right header that was returned by the new code with what is found locally, right? So I think this is a bug and needs changing.

AshantiMutinta: Yes,  I think I missed this during my changes. This is a bug

AshantiMutinta: I think this might have been remnants of the old code

core/src/nucleus/actions/get_entry.rs
AshantiMutinta: Good catch

PR #1142, opened 2019-03-20
Some minor cleanups for v0.0.8

pjkundert:
o Add an `hc run --logging` option
o Comment out some debugging messages
o Remove a period after the key bundle name (may be included if copying)
o Upgrade Makefile to support Nix targets (recommended), as well as local toolchain/library builds
   - This shouldn't be a regression; it is intended to support the host-native builds using native libraries, etc., while still maintaining access to the Nix builds via the `nix-shell --pure --run ...` targets.


thedavidmeister:
see review here and comment https://github.com/holochain/holochain-rust/pull/1152#issuecomment-476407954

let's split this into items that can be reviewed separately as some are minor and some are major

PR #1143, opened 2019-03-20
Reinstate entry lifecycle

AshantiMutinta:
- [x] EntryValidation and LinkValidation data types both have lifecycle

### DIFF
--- a/core/src/nucleus/validation/link_entry.rs
+++ b/core/src/nucleus/validation/link_entry.rs
@@ -49,10 +49,12 @@ pub async fn validate_link_entry(
         Entry::LinkAdd(link) => Ok(LinkValidationData::LinkAdd {
             link,
             validation_package: validation_data.package.clone(),
+            life_cycle : validation_data.lifecycle.clone()
         }),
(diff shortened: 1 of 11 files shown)


core/src/nucleus/validation/link_entry.rs
```diff
@@ -49,10 +49,12 @@ pub async fn validate_link_entry(
         Entry::LinkAdd(link) => Ok(LinkValidationData::LinkAdd {
             link,
             validation_package: validation_data.package.clone(),
+            lifecycle: validation_data.lifecycle.clone(),
         }),
         Entry::LinkRemove(link) => Ok(LinkValidationData::LinkRemove {
             link,
-            validation_package: validation_data.package.clone(),
```
pjkundert: Why don't we go back to passing the `validation_data` (which includes the `.lifecycle`), instead of breaking it apart and passing the validation_package and lifecycle separately?

AshantiMutinta: I guess I looked it as in the package is paired with the entry for the sources? but that would make more sense. i will make changes.

PR #1146, opened 2019-03-21
Hdk crypto app spec test

sphinxc0re:
Screenshot 2019-03-21 at 12 15 19


sphinxc0re:
> Looks good, but app-spec tests are failing.

Jup... the verification is failing. Could you take a look at my code? I think I am missing something

zippy:
Ok.  After a bunch of research I discovered that the problem was that the verify code in your original branch wasn't decoding the base64 signature before passing to sodum::verify.  But as soon as I merged with develp which now also has the crypto stuff in it and uses `provenance.verify`  this was fixed.  In the mean time I also added in an api level test for the verify function.

PR #1147, opened 2019-03-21
remove deprecated zome calling route

zippy:

This PR removes the deprecated zome calling route `/instance/zome/function`  leaving just the `/call` route.


zippy:
The timing is up for grabs.  I just assume that with breaking changes like this, the sooner the better because the longer we leave things in, the more people will have to make the changes.

PR #1149, opened 2019-03-21
Updated ipc_spawn to wait for p2p-ready

ddd-mtl:
Allow ipc_spawn() to block for p2p_bindings so they can be available for the conductor, i.e. fixes problem mentionned by connor:
https://chat.holochain.org/holo/pl/nneqez1147ruuqebsgcpjmzioa


net/src/ipc/spawn.rs
```diff
@@ -57,39 +60,54 @@ pub fn ipc_spawn(
...
+                        has_ipc = true
                     }
-                    for m in re_p2p.captures_iter(&tmp) {
-                        out.p2p_bindings.push(m[1].to_string());
+                }
+                // look for P2P-READY
+                if !has_p2p.clone() {
+                    if re_p2p_ready.is_match(&tmp) {
+                        for m in re_p2p.captures_iter(&tmp) {
+                            out.p2p_bindings.push(m[1].to_string());
```
Connoropolous: The first loop does ‘break’ but this one doesn’t, is there a reason?

conductor_api/src/conductor/base.rs
```diff
@@ -280,6 +280,7 @@ impl Conductor {
                 String::from("N3H_IPC_SOCKET") => String::from("tcp://127.0.0.1:*"),
                 String::from("N3H_LOG_LEVEL") => network_config.n3h_log_level.clone(),
             },
+            2000,
```
Connoropolous: Is 2000 long enough? Will people encounter timeout issues and not know how, or have a way, to set the timeout value?

net/src/ipc/spawn.rs
```diff
@@ -57,39 +60,54 @@ pub fn ipc_spawn(
...
+    // the child process is ready for ipc connections
+    let re_ipc_ready = regex::Regex::new("#IPC-READY#")?;
+
+    // the child process is ready for p2p connections
+    let re_p2p_ready = regex::Regex::new("#P2P-READY#")?;
 
     // we need to know when our child process is ready for IPC connections
     // it will run some startup algorithms, and then output some binding
     // info on stdout and finally a `#IPC-READY#` message.
     // collect the binding info, and proceed when `#IPC-READY#`
```
Connoropolous: This comment is just slightly incorrect but should be corrected I thin

net/src/ipc/spawn.rs
ddd-mtl: Yes. There is only one ipc binding but you can have many p2p binding*s*

conductor_api/src/conductor/base.rs
ddd-mtl: Thats a conductor level decision to make. You don't want to wait too long either, otherwise it will appear 'frozen'.

net/src/ipc/spawn.rs
ddd-mtl: fixed

net/src/ipc/spawn.rs
```diff
@@ -57,39 +60,56 @@ pub fn ipc_spawn(
...
-                    break;
                 }
-            } else {
-                std::thread::sleep(std::time::Duration::from_millis(10));
             }
-
-            if !block_connect {
-                break;
+            if wait_ms >= timeout_ms.clone() {
+                bail!("ipc_spawn timed out");
```
Connoropolous: Just in case we haven't quite got the timeout tuned right, could you add a little bit more supportive messaging in this log? Like, what does that mean to them? Is there anything they can do about it?

ddd-mtl: fixed

Connoropolous:
Note to self, this change should result in some minor updates to documentation

PR #1152, opened 2019-03-22
Makefile improvements

zippy:
- [ ] ~~I have added a summary of my changes to the changelog~~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -243,12 +243,10 @@ wasm_build: ensure_wasm_target
 .PHONY: build_holochain libsodium_version
 libsodium_version:
 	@[ ! -z "$${RUST_SODIUM_LIB_DIR+defined}" ] \
-	    && echo -e "\033[0;93m## Building rust_sodium-sys -- with system libsodium \"$(RUST_SODIUM_LIB)\" in: $${RUST_SODIUM_LIB_DIR:-(unknown)}) ##\033[0m" \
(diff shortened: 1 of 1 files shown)


thedavidmeister:
@zippy @lucksus 

i'm closing this because (i think) it has (almost) all been addressed elsewhere

for context:

comments in https://github.com/holochain/holochain-rust/pull/1084 re: libsodium and nix handling, i.e. libsodium is handled upstream https://github.com/holochain/holochain-rust/pull/1105 and so shouldn't use a (potentially incorrect/compromised) shared lib from the host OS and nix/make are separated by design, this approach works on all supported platforms with working cold installs including debian https://circleci.com/gh/holochain/holochain-rust/10629 and the addition of support for nix shell on mac https://github.com/holochain/holochain-rust/pull/1132 and the nix pod should be used to provide any bug reports https://github.com/holochain/holochain-rust/pull/1139 which is in the readme and the recent debian/xenial additions to the auto install script bring local deps to where they need to be for make support https://github.com/holochain/holochain-rust/blob/develop/scripts/install/linux.sh#L16

if you're seeing issues with nix please send me the the output of the pod script for the command(s) not working
if you're seeing issues with any linux/mac sans nix could you rerun the install script and send me debug info from that?

if you are a downstream consumer of hc, make sure that you're on a tagged release as cold builds aren't run against every PR/merge but tests pass against a fresh install on every supported OS for release commits :)

the `install_conductor` looks like it could be convenient in the makefile, maybe we could get that split out as a separate PR?

PR #1168, opened 2019-03-26
Add make targets to allow simple building of holochain-rust Nix targets

pjkundert:
Further support for Nix/Native build.  Adds support for building the main hc, holochain and WASM API targets via Nix Environment, as well as for selecting the Native `libsodium` library options to use for native builds.

Perhaps there is somewhere else where the available nix-shell targets is summarized, and how to select/build them is detailed, but putting them in the Makefile for easy documentation, selection and execution seems useful. 

Likewise, the selection of system libsodium library is foundational to the success of building/running the NodeJS tests.  The options for selecting/building that are detailed here.


thedavidmeister:
looks like a dupe of https://github.com/holochain/holochain-rust/pull/1152

`RUST_SODIUM_SHARED=1` is a regression, for example

see https://github.com/maidsafe/rust_sodium/issues/98 if you're experiencing linking errors

PR #1180, opened 2019-03-28
Corrected nix installation command.

btakita:


thedavidmeister:
@btakita hey, could i get some context around this change? i'm not sure why it helps?

thedavidmeister:
@btakita i'm not sure the problem here, feel free to provide context and reopen

looking at the nix install scripts it seems like `--no-daemon` is the default:

https://github.com/NixOS/nix/blob/master/scripts/install-nix-from-closure.sh#L58

the command we are running currently is copied from the official docs:

https://nixos.org/nix/download.html

PR #1182, opened 2019-03-28
DPKI bootstrap

zippy:
This PR adds conductor DPKI integration (for bootstraping and adding agents):
- Conductor config now has an optional dpki section to specify a dpki instance that will handle creating and validating new instances when added..
- Keystore now has a `new_standalone` method for when NOT using dpki. This method is now used to create the keystores generated by `hc keygen`
- Added a rust trait to mirror the dpki DNA "trait" for use by the conductor in bootstrapping and and adding to agent_keys.
- Barebones documentation added to MDbook describing DPKI.


conductor_api/src/conductor/admin.rs
```diff
@@ -381,22 +388,70 @@ impl ConductorAdmin for Conductor {
...
+            {
+                let instance = self.instances.get(&dpki_instance_id)?;
+                let hc_lock = instance.clone();
+                let hc_lock_inner = hc_lock.clone();
+                let mut hc = hc_lock_inner.write().unwrap();
+                hc.dpki_create_agent_key(name.clone())?;
+            }
+            // TODO: how do we clean-up now if this fails? i.e. the dpki dna will have registered
+            // the identity to its DHT, but we failed, for what ever reason, to set up
+            // the agent in the conductor, so we should do something...
```
lucksus: Maybe we should not exit this function if we fail to access the DPKI keystore and regard it as a transient problem - and instead continue with adding the agent to the conductor config?

conductor_api/src/dpki_instance.rs
```diff
@@ -0,0 +1,83 @@
...
+
+pub const DPKI_ZOME_NAME: &str = "dpki";
+pub const DPKI_TRAIT_FN_ADD_AGENT: &str = "create_agent_key";
+pub const DPKI_TRAIT_FN_INIT: &str = "init";
+pub const DPKI_TRAIT_FN_IS_INITIALIZED: &str = "is_initialized";
+
+use std::convert::TryInto;
+
+pub trait DpkiInstance {
+    fn dpki_cap_request(
```
lucksus: Not sure if this function actually needs to be in the trait?
It is only used internally to the implementations of the other functions right? Would rather have it be not part of the trait then since it is also not part of the zome trait.

doc/holochain_101/src/dpki.md
```diff
@@ -0,0 +1,54 @@
...
+- https://hackmd.io/oobu0sKMSMadLXza4rHY_g
+
+## DPKI Implementation Technical Details
+
+### Keystore
+For each Holochain DNA instance, the Conductor maintains a Keystore, which holds "secrets" (seeds and keys) needed for cryptographic signing and encrypting. Each of the secrets in the Keystore is associated with a string which is a handle needed when using that secret for some cryptographic operation.  Our cryptographic implementation is based on libsodium, and the seeds use their notions of context and index for key derivation paths.  This implementation allows DNA developers to securely call cryptographic functions from WASM which will be executed in the conductor's secure memory space when actually doing the cryptographic processing.
+
+Decrypting a secret from the keystore invokes a passphrase manager service, which is used to collect the passphrase from some end-user.  This service is implementation specific.  Currently we have implementations for command-line passphrase collection for use in the `hc keyget` command, and also an implementation that routes through the same JSON RPC pathway for other admin commands sent to the conductor for other admin actions.
```
lucksus: We don't have an implementation for passphrase through JSON RPC yet...
Also, this should it should be `keygen` instead of `keyget`, right?
```suggestion
Decrypting a secret from the keystore invokes a passphrase manager service, which is used to collect the passphrase from some end-user.  This service is implementation specific.  Currently we have implementations for command-line passphrase collection for use in the `hc keygen` command, and also a command-line implementation within the conductor.
```

conductor_api/src/conductor/admin.rs
```diff
@@ -381,22 +388,70 @@ impl ConductorAdmin for Conductor {
...
+        if new_config.agents.iter().any(|i| i.id == id) {
             return Err(HolochainError::ErrorGeneric(format!(
                 "Agent with ID '{}' already exists",
-                new_agent.id
+                id
             )));
         }
-        new_config.agents.push(new_agent.clone());
+
+        let (keystore, public_address) = if self.using_dpki() {
```
lucksus: As per https://github.com/holochain/holochain-rust/pull/1199, I think we should omit creating a keystore if `holo_remote_key` is set.

dpki/src/keypair.rs
```diff
@@ -81,6 +85,14 @@ impl KeyPair for SigningKeyPair {
         // Done
         Ok(SigningKeyPair::new(pub_key_b32, priv_sec_buf))
     }
+
+    fn new_from_self(&mut self) -> HcResult<Self> {
+        let mut buf = SecBuf::with_secure(self.private.len());
+        let pub_key = self.public();
+        let lock = self.private().read_lock();
+        buf.write(0, &**lock)?;
```
neonphog: perhaps we should implement clone for secbuf? https://github.com/holochain/holochain-rust/compare/clone-secbuf

conductor_api/src/conductor/admin.rs
```diff
@@ -381,22 +388,73 @@ impl ConductorAdmin for Conductor {
         Ok(())
     }
 
-    fn add_agent(&mut self, new_agent: AgentConfiguration) -> Result<(), HolochainError> {
+    fn add_agent(
+        &mut self,
+        id: String,
+        name: String,
+        holo_remote_key: Option<&str>,
```
maackle:
```suggestion
        holo_remote_key: Option<String>,
```

It's getting converted to a string later anyway

conductor_api/src/conductor/admin.rs
```diff
@@ -381,22 +388,73 @@ impl ConductorAdmin for Conductor {
...
+        if new_config.agents.iter().any(|i| i.id == id) {
             return Err(HolochainError::ErrorGeneric(format!(
                 "Agent with ID '{}' already exists",
-                new_agent.id
+                id
             )));
         }
-        new_config.agents.push(new_agent.clone());
+
+        let (keystore_file, public_address) = if holo_remote_key.is_some() {
```
maackle:
```suggestion
        let (keystore_file, public_address) = if let Some(public_address) = holo_remote_key {
```

conductor_api/src/conductor/admin.rs
```diff
@@ -381,22 +388,73 @@ impl ConductorAdmin for Conductor {
...
+            ("".to_string(), holo_remote_key.unwrap().to_string())
```
maackle:
```suggestion
            ("".to_string(), holo_remote_key)
```

conductor_api/src/dpki_instance.rs
zippy: Done

conductor_api/src/conductor/admin.rs
zippy: Done

dpki/src/keypair.rs
zippy: whoah!  Nice I never saw a PR for that!!  Would much prefer...

zippy: I just made a PR for this.  Somehow I had it in my mind that SecBuf's couldn't be cloned.  Dunno why...

doc/holochain_101/src/dpki.md
```diff
@@ -0,0 +1,54 @@
+# Distributed Public Key Infrastructure (DPKI)
+
+## Context
+
+Like most other distributed systems, Holochain fundamentally relies on public key cryptography. Among other uses, Holochain nodes are identified by their public keys, and thus provenance of messages from nodes can be checked simply by checking a message signature against the node's identifier.  This fundamental use requires nodes to keep the private key secret, lest the node's very agency be compromised.  Keeping such secrets in the digital age is a non-trivial problem.  Additionally, distributed systems create a context of a profusion of nodes with multiple public keys, many of which may wish to be identified as being under the control a of a single actor.  Solutions to this need create yet another layer of keys which sign other keys, and the management of all this can become quite complex.
```
Connoropolous:
```suggestion
Like most other distributed systems, Holochain fundamentally relies on public key cryptography. Among other uses, Holochain nodes are identified by their public keys, and thus provenance of messages from nodes can be checked simply by checking a message signature against the node's identifier.  This fundamental use requires nodes to keep the private key secret, lest the node's very agency be compromised.  Keeping such secrets in the digital age is a non-trivial problem.  Additionally, distributed systems imply an abundance of nodes with public keys, many of which may wish to be identified as being under the control of a single actor.  Solutions to this need create yet another layer of keys which sign other keys, and the management of all this can become quite complex.
```

doc/holochain_101/src/dpki.md
```diff
@@ -0,0 +1,54 @@
...
+
+To deliver on the basics of Distributed Public Key Infrastructure, we need a way to generate keys of various types (revocation, identity, encryption, signing) from seeds, and we need to be able to generate such seeds from primary seeds, so that a human agent can create related "device agents" provably under their control.
+
+After studying a number of uses cases, including initial sign-up, key revocation, etc, the central insight we came to was the need to create a Hierarchical Deterministic Key generation system, based on a Primary Seed, from which additional seeds can be generated which then are in turn used to actually generate many key-pairs.  This allows us, by-convention, to use the first seed generated by the Primary seed as the seed for revocation keys, and subsequent seeds as seeds for keys of separate Holochain devices that can be proven to be under the control of the holder of Primary Seed.
+
+DeepKey
+
+TODO: merge the various docs we developed to explain DeepKey here.
+- https://medium.com/holochain/part-2-holochain-holo-accounts-cryptographic-key-management-and-deepkey-bf32ee91af65
+- https://hackmd.io/UbfvwQdJRKaAHI9Xa7F3VA?view
```
Connoropolous: Are these hackmds editable by anyone who comes upon them?  We should check the permissions, since this will be published

PR #1184, opened 2019-03-30
Making json string to JsonString conversion explicit by removing JsonString From<String>

willemolding:
Re-attempt at #810

Currently using the hdk::call function in zome code requires manually parsing the JsonString since it is impossible to serialize into a ZomeApiResult. This is the first change to JsonString required to make this possible.


willemolding:
Note this will probably be a breaking change for DNA developers if they are using JsonString

conductor_api/wasm-test/src/lib.rs
```diff
@@ -109,7 +109,7 @@ pub extern "C" fn debug_stacked_hello(
     let fish = mem_stack.write_json(TestStruct {
         value: "fish".to_string(),
     });
-    hdk_debug(&mut mem_stack, &JsonString::from("disruptive debug log"));
+    hdk_debug(&mut mem_stack, &JsonString::from(RawString::from(RawString::from("disruptive debug log"))));
```
thedavidmeister: why the double `RawString`?

conductor_api/test-bridge-caller/src/lib.rs
```diff
@@ -12,7 +12,7 @@ use hdk::holochain_core_types::{
 };
 
 fn handle_call_bridge() -> JsonString {
-    hdk::call("test-callee", "greeter", Address::from("token"), "hello", JsonString::from("{}")).unwrap()
+    hdk::call("test-callee", "greeter", Address::from("token"), "hello", JsonString::from_json("{}")).unwrap()
```
thedavidmeister: `JsonString::empty_object()`?

core_types/src/json.rs
```diff
@@ -39,14 +39,16 @@ impl JsonString {
...
+    // Creates a JsonString from stringified json
+    // replaces From<String> for JsonString and requires conversions to be explicit
+    // This is because string types must be handled differently depending on if
+    // they are strinfigied JSON or JSON strings
+    pub fn from_json(s: &str) -> JsonString {
         let cleaned = s
             // remove whitespace from both ends
             .trim()
-            // remove null characters from both ends
+            // remove null characters from both endsi
```
thedavidmeister: typo

core_types/src/json.rs
```diff
@@ -145,38 +147,47 @@ pub trait JsonError {}
...
+    }
+}
+
+impl<E: Into<JsonString>> From<Result<String, E>> for JsonString {
+    fn from(result: Result<String, E>) -> JsonString {
+        result_to_json_string(result.map(|v| RawString::from(v)))
+    }
+}
+
+impl From<Result<String, String>> for JsonString {
```
thedavidmeister: @lucksus lookit ^^ better late than never?

nodejs_waiter/src/waiter.rs
```diff
@@ -417,7 +417,7 @@ mod tests {
     }
 
     fn zf_response(call: ZomeFnCall) -> ExecuteZomeFnResponse {
-        ExecuteZomeFnResponse::new(call, Ok(JsonString::from("")))
+        ExecuteZomeFnResponse::new(call, Ok(JsonString::from_json("{}")))
```
thedavidmeister: `JsonString::empty_object()`

core_types/src/json.rs
lucksus: 👏

core_types/src/json.rs
```diff
@@ -145,38 +147,47 @@ pub trait JsonError {}
...
+    let inner_json: JsonString = match result {
+        Ok(inner) => inner.into(),
+        Err(inner) => inner.into(),
+    };
+    let inner_string = String::from(inner_json);
+    JsonString::from_json(&format!(
+        "{{\"{}\":{}}}",
+        if is_ok { "Ok" } else { "Err" },
+        inner_string
+    ))
```
maackle: This looks correct but doesn't Result serialize to exactly this anyway? `serde_json::to_string(&result)`?

thedavidmeister: i think that's the point right? to extend the types/traits so that they "magically" track default serde behaviour

Issue #980, opened 2019-02-09
Permission Denied when running install script (Windows WSL)


EzraWeller:
Just got this error when running `./scripts/install/auto.sh` in Windows Subsystem for Linux (Windows 10):

```
E: Could not open lock file /var/lib/apt/lists/lock - open (13: Permission denied)
E: Unable to lock directory /var/lib/apt/lists/
E: Could not open lock file /var/lib/dpkg/lock - open (13: Permission denied)
E: Unable to lock the administration directory (/var/lib/dpkg/), are you root?
Error executing command, exiting
E: Could not open lock file /var/lib/apt/lists/lock - open (13: Permission denied)
E: Unable to lock directory /var/lib/apt/lists/
E: Could not open lock file /var/lib/dpkg/lock - open (13: Permission denied)
E: Unable to lock the administration directory (/var/lib/dpkg/), are you root?
```

Worked fine when I ran it with sudo but not sure if that's the intention.

hoijui:
same here (on Debian testing)

The problem are these lines in _scripts/install/linux.sh_:
```bash
# basics
apt-get update
apt-get install -y sudo
sudo apt-get install -y curl git
```
Both `apt-get update` and `apt-get install` need root priviledges. If we are runnig as root, we do not need sudo anymore, and can just run everything without it. if we are not root, we can not install sudo, so these lines make no sense.
Even if they would make sense, it would be very bad if a simple instlal script of a software woudl install `sudo`, which if missing, usually is so because it is deemed a security risc.

hoijui:
proposed fix:
https://github.com/holochain/holochain-rust/pull/1192

lazaridiscom:
@thedavidmeister, closable?

hoijui:
if https://github.com/holochain/holochain-rust/pull/1218 gets merged yes, until then this remains (with slightly different error message).

thedavidmeister:
we don't have any tests around this

last time we changed the sudo setup it caused failed builds in develop and releases, so we should move forward carefully

WSL isn't something we've explicitly tested the install script on at all, or are saying we "officially" support, but if a few minor tweaks covers off what is needed i don't want to block anything :)

#1218 fixing debian doesn't guarantee that it will fix windows WSL iirc, although it's likely to help

hopefully it does fix both! let's not close this off as a dupe until we've merged and can test/confirm the results

hoijui:
what do you think about running TravisCI also with debian and windows?
[...]

PR #1192, opened 2019-03-31
Linux install script: do not install `sudo` [fix] #980

hoijui:
Without `sudo` available, and without being run as root,
we can not install `sudo`,
so previous code was bogus.

Also, a script for installing a software should not
by itsself install `sudo`,
which by some is considered a security risk.

Now it uses `sudo` if available, and otherwise requires to be run by root.


lazaridiscom:
not tested, but code looks very clean.

Confirming the logic.

thedavidmeister:
this breaks all the cold builds on CI - https://circleci.com/gh/holochain/holochain-rust/15741

thedavidmeister:
i think the issue is that circle uses pipefail, so `which sudo > /dev/null` silently kills builds...

thedavidmeister:
```
root@621e9dde120d:~/project# . ./scripts/install/auto.sh 
bash: sudo: command not found
This script requires  or root priviledges
Connection to 34.224.100.93 closed.
```

for some reason this doesn't show up in the GUI but is visible when i ssh in manually

thedavidmeister:
ok i think we're good now, it was two things, fixes pushed to develop for tomorrow's release:

- root user id _is_ equal to `0` so `-eq` not `-ne`
- debian needs an `apt-get update` before attempting to install `curl` and `git`

PR #1197, opened 2019-04-01
Rename silent to nullpass and add quiet mode

maackle:

Using this may look roughly like the following
```
const hc = spawn('hc keygen --quiet')
  console.info("auto-entering passphrase twice...")
  hc.stdin.write(passphrase + '\n')
  hc.stdin.write(passphrase + '\n')
  hc.stdin.end()
  // read answers from hc.stdout
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Adds hdk access to keystore [#1148](https://github.com/holochain/holochain-rust/pull/1148)
 - Adds a `--path` option to `hc keygen` to specify the location of the generated keybundle. [#1194](https://github.com/holochain/holochain-rust/pull/1194)
+- Adds a `--quiet` option to `hc keygen` for machine-readable output, intended for use in scripts. [#1197](https://github.com/holochain/holochain-rust/pull/1197)
 
 ### Changed
 - Performance optimization: don't recalculate DNA hash during handling of every network message but instead cache the DNA hash. [PR#1163](https://github.com/holochain/holochain-rust/pull/1163)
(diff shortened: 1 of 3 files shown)


cli/src/main.rs
```diff
@@ -226,13 +232,17 @@ fn run() -> HolochainResult<()> {
         }
         .map_err(HolochainError::Default)?,
 
-        Cli::KeyGen { silent, path } => {
-            let passphrase = if silent {
+        Cli::KeyGen {
+            path,
+            quiet,
+            nullpass,
+        } => {
```
```suggestion
        } => {
        // a default passphrase is used as the null passphrase, because
        // separately in the Conductor it will attempt to decrypt using the
        // default passphrase first, and then fallback on asking for user input
```

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Adds hdk access to keystore [#1148](https://github.com/holochain/holochain-rust/pull/1148)
 - Adds a `--path` option to `hc keygen` to specify the location of the generated keybundle. [#1194](https://github.com/holochain/holochain-rust/pull/1194)
+- Adds a `--quiet` option to `hc keygen` for machine-readable output, intended for use in scripts. [#1197](https://github.com/holochain/holochain-rust/pull/1197)
```
Connoropolous: This should additionally mention that this PR replaces `--silent` option, and also adds `--nullpass` option

maackle: @Connoropolous disagree, because --silent and now --nullpass are temporary and meant to be undocumented

maackle: --silent is simply renamed to --nullpass. --quiet is the new one.

Connoropolous: Ok. Can do

PR #1202, opened 2019-04-02
Fixing pickledb bugs

AshantiMutinta:


AshantiMutinta:
Dont merge until the profiler is run(can still review) just need to confirm that the space issue has been fixed

nodejs_conductor/native/src/config.rs
```diff
@@ -83,11 +85,13 @@ fn make_config(instance_data: Vec<InstanceData>, logger: LoggerConfiguration) ->
 
         let agent_id = agent_config.id.clone();
         let dna_id = dna_config.id.clone();
+        let temp = tempdir().expect("test was supposed to create temp dir");
```
lucksus: I was looking up [tempdir's docs](https://docs.rs/tempfile/3.0.7/tempfile/fn.tempdir.html) because I was wondering if that is a unique directory since we don't want multiple test instances share the same CAS. It says that the tempdir gets removed when `TempDir` is dropped. This happens a few lines below, right? The conductor will recreate the the path I guess but then nothing will delete it, so it actually is not temporary. This might cause mysterious problems later on.. Not sure if we actually want to merge pickle as default for nodejs conductor into develop.. We might want to if it really is faster than MemoryStorage..

AshantiMutinta: I agree, I think since memory is the fastest and this is for the use in the nodejs we should still go back to memory. I will revert this back asap

AshantiMutinta:
Benchmarks have been run and flame graph is up on realtimeboard

test bench_file_eav_many_to_one   ... bench: 183,641,550 ns/iter (+/- 131,852,367)
test bench_file_eav_one_to_many   ... bench: 201,104,300 ns/iter (+/- 119,277,360)
test bench_memory_eav_many_to_one ... bench:     791,201 ns/iter (+/- 77,859)
test bench_memory_eav_one_to_many ... bench:     827,959 ns/iter (+/- 95,228)
test bench_pickle_eav_many_to_one ... bench:  14,167,780 ns/iter (+/- 14,597,584)
test bench_pickle_eav_one_to_many ... bench:  13,985,950 ns/iter (+/- 10,538,156)

Issue #1191, opened 2019-03-31
New ws rpc "call" method fails if inner params set to null


pythagorean:
Using the new "call" method, if the inner params value is null, the call fails. If empty inner params are explicitly {} it works.

Connoropolous:
that's no fun!

We have a PR (link above this comment) that hasn't yet addressed this, but is starting to integrate it as feedback, up now

Connoropolous:
@willemolding you have something in progress related to this as well now I believe?

dymayday:
Hi! This is fixed in https://github.com/holochain/holochain-rust/pull/1271 :)

PR #1203, opened 2019-04-02
Update mdbook to reflect "call" method changes

maackle:
Partially complete update to mdbook to reflect changes to `call` RPC method. Still needs to be updated in two subsections.

I also propose: 
- we rename "params" to "args" or "arguments", to avoid nested the nested "params". 
- we allow both "params" and "args" in the "call" method handler, deprecating "params".
- then we update the documentation to use "args".

Also, params is a misnomer here anyway. According to [generally](https://en.wikipedia.org/wiki/Parameter_(computer_programming)#Parameters_and_arguments) [accepted](https://www.quora.com/What-is-the-difference-between-parameters-and-arguments-in-Python) [notions](https://stackoverflow.com/a/156787/612758), "parameters" are the placeholders you put in a function definition, and "arguments" are the actual values you pass when invoking a function, and the latter is what we're doing here.

This is what I'm talking about

```javascript
{
    "jsonrpc": "2.0",
    "id": "0",
    "method": "call",
    "params": {  // <--------- from JSON-RPC spec
        "instance_id": "test-instance",
        "zome": "blog",
        "function": "create_blog",
        "params": {  // <---------- proposing we change to "args"
            "blog": {
                "content": "sample content"
            } 
        }
    }
}
```


Connoropolous:
@maackle I also really disliked the `params -> params` 
so I am very in favor of updating to using a key called `args` 👍

zippy:
I'm ok with doing this update, but if we are going to do it, lets convert this PR/branch into not just a documentation change, but do the work here too.  Also we should do it soon.  Someone wanna take this on and add into the tree?

maackle:
I can definitely take it on next week, and maybe this week.

We can bundle up #1191 with this work as well.

maackle:
But, @zippy if by "soon" you mean this week, maybe someone else can take it on

zippy:
Hey @dymayday how about you take these on!?

dymayday:
Sure! I will be more than happy to do it as soon as I'm done with the guidebook, I don't want to mess with my first PR...

PR #1204, opened 2019-04-02
move CoreDevSetup to own file

lazaridiscom:
followup to #1185

Further reduce main README size.

No changes to the text itself.


lazaridiscom:
> This branch is out-of-date with the base branch
> Merge the latest changes from develop into this branch.

Githup process is a bit misleading. No need to update the branch.

thedavidmeister:
@lazaridiscom we have configured the github process so that tests must be passing against any PR with develop merged into it

probably not so important with readme changes, but the general principle makes sense

lazaridiscom:
> even the developer docs that are linked to (the quick start) include steps that are probably not needed for an app dev (e.g. the install script is probably not the best idea if nix-shell can suffice...)

In intentionally change only structure (and not content) at this point ((me currently in the process of identifying the system actors, like e.g. AppDev and CoreDev).

Content needs to be reworked completely (mostly simplification), and this is something a newcomer like me should do (not folks which is familiar with holochain).

The overall time-frame to fix the overall system-documentation (including entry-points like website) is estimated 3 months. Tiny incremental steps, and patience.

zippy:
Hi @lazaridiscom, thanks for you participation.  I do like your simplification too.  If you can resolve the merge conflicts on your branch, then we'll merge!

lazaridiscom:
It's a minimal conflict, will take a look when I'm on my main machine (max 24h)

lazaridiscom:
Git... always fun.

Closing, i'll send a new pr in.

lazaridiscom:
just realized my fault: i merged "master" instead of "develop" locally, then pushed. Ignore this one, use the fresh #1222 instead.

PR #1211, opened 2019-04-03
Optimize WASM calls

lucksus:
# Erasing 81% of run_dna's CPU time

## Context
Sub-items of performance optimization for WASM calls:

Screenshot 2019-04-05 at 20 42 13

## Main changes
In the [old FlameGraph](https://realtimeboard.com/app/board/o9J_kyiXmFs=/?moveToWidget=3074457346553142160) we can see that within `run_dna` most time is spent creating the parsed WASM module from the byte array (`wasmi::Module::from_buffer`). After first trying to hold the whole `wasmi::ModuleInstance` in the nucleus state, for which I also [refactored wasmi in a fork](https://github.com/holochain/wasmi) so that `ModuleInstance` implements `Send`, I realized that it makes much more sense to just hold a parsed `wasmi::Module` inside the DNA, next to the unparsed byte array. We can then still have `run_dna` create a new instance running in a new thread, but we use the already present Module that we find in the DNA. I ended up not using my wasmi fork since the mutex applied there introduced new performance issues for WASM memory access.

Also, with this PR `DnaWasm::code` changes from being a plain `Vec<u8>` to an `Arc<Vec<u8>>` which makes DNA cloning (which happens quite a bit, not yet a major thing but already visible on the FlameGraphs) much cheaper.

Next to those main changes I cleaned up by removing the now unnecessary parameter `wasm` from `run_dna` and the functions that call it. Same with `dna_name` which can be get from the DNA, via the context, where needed.

## Before

Screenshot 2019-04-05 at 20 41 26

See SVG in SoA tree: https://realtimeboard.com/app/board/o9J_kyiXmFs=/?moveToWidget=3074457346553142160
**`Module::from_buffer` makes up 81% for `run_dna`**
## After 

Screenshot 2019-04-05 at 20 41 32

# Outlook

This is a substantial improvement, *[and we can do more](https://realtimeboard.com/app/board/o9J_kyiXmFs=/?moveToWidget=3074457346558621744)*. Instantiating `ModuleInstance` is still more expensive than actually executing the WASM code - at least in these [holochain-basic-chat](https://github.com/holochain/holochain-basic-chat) based profiles. Also, WASM-side initialization like `init_globals` (since it's called with every new ModuleInstance) makes up for a good part of spent CPU time. A promising next step could be re-purposing a fixed pool of ModuleInstances and initializing them from memory snapshots that include the result of `init_globals` etc.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,6 +31,36 @@ jobs:
          echo "$UNPINNED"
          exit 1
         fi
+   # don't allow changelog root items with no PR reference
+   - run:
+      name: no changelogs without PR references
+      command: |
+        # temporary avoid build fails due to greps
+        set +eo pipefail
+        export UNREFERENCED=`nix-shell --run hc-changelog-grep-pr-references`
+        set -eo pipefail
+        if [[ $UNREFERENCED ]]
+        then
+          echo "all root level changelog items need a PR reference:"
+          echo "$UNREFERENCED"
+          exit 1
+        fi
+
+   # flag any references to a nightly version in a readme other than the
+   # ${date} in nix-shell
+   - run:
+      name: no readmes with bad rust nightly versions
+      command: |
+        # temporary avoid build fails due to greps
+        set +eo pipefail
+        export BAD_NIGHTLY=`nix-shell --run hc-readme-grep-nightly`
+        set -eo pipefail
+        if [[ $BAD_NIGHTLY ]]
+        then
+          echo "there is an incorrect nightly version in a readme file:"
+          echo $BAD_NIGHTLY
+          exit 1
+        fi
 
  app-spec-tests:
   docker:
@@ -118,7 +148,7 @@ jobs:
   resource_class: xlarge
   steps:
    - checkout
-   - run: . ./scripts/install/auto.sh
+   - run: ./scripts/install/auto.sh
    - run: make test
 
  # https://github.com/NixOS/nix/issues/1559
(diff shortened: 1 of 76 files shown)


lucksus:
@willemolding, would love to hear how this branch performs in your Holo test setup.

ddd-mtl:
@lucksus I think its a good idea to cache the result of `init_globals` but be careful about some of the globals which are not constants and can change value between dna calls. Like `agent_latest_hash` for example.

thedavidmeister:
yes, i like where this is going and i also have a friendly reminder for future work re: stack based memory not being safe to work with concurrently - we would need to move to `wee_alloc` to support that i believe

lucksus:
Right, thanks @ddd-mtl. So it actually means that we would need to recreate the initialized memory snapshot after any commit to the source chain. Isn't that very related to the "as at", @thedavidmeister, @artbrock?

ddd-mtl:
@lucksus Not "any" commit, only a commit that updates the agentId. Thats what `agent_latest_hash` is tracking. Unless you were referring to a different global?

thedavidmeister:
@lucksus yeha well the summary of that is that the head commit at the start of the function call invalidates the result of the function call if the head has moved by the end of the function call

that is something else that we can avoid thinking about atm with sequential function calls but would need to be considered across concurrent calls

well.... perhaps, we might find benefit in adding "as at" to actions in the action loop... but that's another discussion as we're not doing that yet

PR #1212, opened 2019-04-03
Print DNA hash in hc package output

maackle:


zippy:
This is great because there's an issue that folks need to remember, which is that the hash in the conductor config will eventually get checked against what's actually in source dna file, and what's in the cas at boot-up time.  This helps get it for putting into the config!

PR #1213, opened 2019-04-03
Documentation Changes for pickle db

AshantiMutinta:

Description : Reading this again, I have made some changes to the documentation just so that is more clear

### DIFF
--- a/doc/holochain_101/src/conductor_instances.md
+++ b/doc/holochain_101/src/conductor_instances.md
@@ -22,7 +22,7 @@ A table for configuring the approach to storage of the local source chain and DH
 Select between different storage implementations. There are three so far:
 - `memory`: Persist actions taken in this instance only to memory. Everything will disappear when the Conductor process stops.
 - `file`: Persist actions taken in this instance to the disk of the device the Conductor is running on. If the Conductor process stops and then restarts, the actions taken will resume at the place in the local source chain they last were at.
-- `pickle` : Persists to a fast memory call which is eventually persisted to a file storage every 5 seconds. The actions taken will also resume at the place in the local source chain they were last. It implements a drop trait which allows data to be flushed out whenever an application error occurs.
+- `pickle` : Persists to a fast memory call which is eventually persisted to a file storage every 5 seconds. The actions taken will also resume at the place in the local source chain they were last. If an application error does occur, it will make sure to persist the latest data prior to any shutdown occurring.
 
 #### `StorageConfiguration.path`: `string`
 Path to the folder in which to store the data for this instance.
--- a/doc/holochain_101/src/conductor_instances.md
+++ b/doc/holochain_101/src/conductor_instances.md
@@ -19,10 +19,10 @@ A reference to the given ID of a defined [DNA](./conductor_dnas.md)
 A table for configuring the approach to storage of the local source chain and DHT for this instance
 
 #### `StorageConfiguration.type`: `enum`
-Select between different storage implementations. There are two so far:
+Select between different storage implementations. There are three so far:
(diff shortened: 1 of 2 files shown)


doc/holochain_101/src/conductor_instances.md
```diff
@@ -19,10 +19,10 @@ A reference to the given ID of a defined [DNA](./conductor_dnas.md)
...
 
 #### `StorageConfiguration.type`: `enum`
-Select between different storage implementations. There are two so far:
+Select between different storage implementations. There are three so far:
 - `memory`: Persist actions taken in this instance only to memory. Everything will disappear when the Conductor process stops.
 - `file`: Persist actions taken in this instance to the disk of the device the Conductor is running on. If the Conductor process stops and then restarts, the actions taken will resume at the place in the local source chain they last were at.
-- `pickle` : Persists to a fast rust implementation. This persists every 5 minutes and on a drop trait, actions will continue from where they were last were at if application is every restarted. Takes in path string which is directory where pickledb instance will be persisting to
+- `pickle` : Persists to a fast memory call which is eventually persisted to a file storage every 5 minutes. The actions taken will also resume at the place in the local source chain they were last. It implements a drop trait which allows data to be flushed out whenever an application error occurs.
```
Connoropolous: The last sentence is interesting. What's going on there? Will that actually be problematic?

zippy: From my read of the code it's actually every 5 seconds:

`const PERSISTENCE_INTERVAL: Duration = Duration::from_millis(5000);`

Which I think is better anyways.  It does mean that if someone pulled the plug on the computer at millisecond 4999, then it's possible that any commits in the last 5 seconds would be lost, AND they may have already made it out to the DHT.  So we are going to have to stress test this and see what it means.

AshantiMutinta: Oh wooops that is every 5 seconds.

doc/holochain_101/src/conductor_instances.md
```diff
@@ -19,10 +19,10 @@ A reference to the given ID of a defined [DNA](./conductor_dnas.md)
...
 
 #### `StorageConfiguration.type`: `enum`
-Select between different storage implementations. There are two so far:
+Select between different storage implementations. There are three so far:
 - `memory`: Persist actions taken in this instance only to memory. Everything will disappear when the Conductor process stops.
 - `file`: Persist actions taken in this instance to the disk of the device the Conductor is running on. If the Conductor process stops and then restarts, the actions taken will resume at the place in the local source chain they last were at.
-- `pickle` : Persists to a fast rust implementation. This persists every 5 minutes and on a drop trait, actions will continue from where they were last were at if application is every restarted. Takes in path string which is directory where pickledb instance will be persisting to
+- `pickle` : Persists to a fast memory call which is eventually persisted to a file storage every 5 seconds. The actions taken will also resume at the place in the local source chain they were last. It implements a drop trait which allows data to be flushed out whenever an application error occurs.
```
Connoropolous:
```suggestion
- `pickle` : Persists to a fast memory call which is eventually persisted to a file storage every 5 seconds. The actions taken will also resume at the place in the local source chain they were last. If an application error does occur, it will make sure to persist the latest data prior to any shutdown occuring.
```

doc/holochain_101/src/conductor_instances.md
thedavidmeister: @zippy yes i'm concerned about this too, i wouldn't expect that we'd be optimising for write speeds so much that we'd accept permanent corruption of local state due to power loss on local

that kind of breaks pickledb in my mind...

PR #1214, opened 2019-04-04
Update docs for changes to JsonString

willemolding:
Makes doc changes for #1184 regarding the removal of From<String> for JsonString

### DIFF
--- a/doc/holochain_101/src/json_string.md
+++ b/doc/holochain_101/src/json_string.md
@@ -339,7 +339,7 @@ important edge cases that we need to cover with additional techniques/tooling.
 
 #### String handling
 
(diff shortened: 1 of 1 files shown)


doc/holochain_101/src/json_string.md
```diff
@@ -339,15 +339,17 @@ important edge cases that we need to cover with additional techniques/tooling.
 
 #### String handling
 
-`JsonString::from` assumes any `String` or `&str` passed to it is already a
-serialized JSON value.
+`JsonString::from_json(&str)` the `&str` passed to it is already a
```
thedavidmeister:
```suggestion
`JsonString::from_json(&str)` requires the `&str` passed to it is already a
```

PR #1218, opened 2019-04-04
linux install script: update apt cache as root [fix]

hoijui:
APT cache is stored system globally,
and thus in a place where only root has access to,
so updating it requires root access.


thedavidmeister:
this works on all the cold builds currently, so sudo isn't required there at least (likely because CI uses root)

adding sudo to the `npm install` previously seems to have broken our deployment scripts (still investigating this)

can we demonstrate a failing case showing sudo is needed here?

hoijui:
I also think it is run as root on CI, so `as_root=""` (It would break with `sudo`).
on my local debian, running the script as non root, without this commit, I get:

> …/holochain-rust $ ./scripts/install/auto.sh
Reading package lists... Done
E: Could not open lock file /var/lib/apt/lists/lock - open (13: Permission denied)
E: Unable to lock directory /var/lib/apt/lists/
W: Problem unlinking the file /var/cache/apt/pkgcache.bin - RemoveCaches (13: Permission denied)
W: Problem unlinking the file /var/cache/apt/srcpkgcache.bin - RemoveCaches (13: Permission denied)
[sudo] password for user:

lazaridiscom:
Works on my side when applying patch.

Confirming logic.

thedavidmeister:
ok, we'll let's ship this and see how it impacts `0.0.11` :)

Connoropolous:
@thedavidmeister is this still relevant?

hoijui:
yes, it should still be merged.
as said, wihtout this, install on linux will fail when run by non-root.

Connoropolous:
Thanks @hoijui . 

There is however, a slightly broader framing for the question, which involves a lot of other work that @thedavidmeister has been doing on the installation process, that I am wondering whether it actually does away with the necessity for this whole script.

hoijui:
ok. But even if it will be deleted soon, I can't see how it could make more sense to delete a broken script then to delete a working one. In the worst case, he wil lget a merge conflict when merging in his changes, and then has to decide to delete this fie anyway.

thedavidmeister:
i have merged, if cold builds break i will rollback as we have a release scheduled soon

Issue #1223, opened 2019-04-04
Thrown exceptions in tape hang tests


MightyAlex200:
Any uncaught exceptions in the function passed to `runTape` will cause tests to hang after running all tapes. I think this was supposed to be fixed by #918, but the problem has returned.

Minimum case: (`test/index.js`)
```javascript
const { Config, Scenario } = require("@holochain/holochain-nodejs")
Scenario.setTape(require("tape"))

const dnaPath = "./dist/test.dna.json"
const agentAlice = Config.agent("alice")
const dna = Config.dna(dnaPath)
const instanceAlice = Config.instance(agentAlice, dna)
const scenario = new Scenario([instanceAlice])

scenario.runTape("simplified example of test hanging", async (t, { alice }) => {
  throw new Error();
});
```

I think the problem is that it's doing something like this:
```javascript
try {
    asyncFunctionThatThrows();
} catch (e) {
    console.log(`Caught. ${e}`);
}
```

When it should be doing something like this:
```javascript
try {
    await asyncFunctionThatThrows();
} catch (e) {
    console.log(`Caught. ${e}`);
}
```

PR #1225, opened 2019-04-05
Fix tests hanging when runTape function throws exception

MightyAlex200:

I think this was just a small thing that went unnoticed when changing the function to `async`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


maackle:
@MightyAlex200 thanks for this PR, I didn't notice it until now! I wound up implementing a fix myself (#1232) after seeing your issue #1223, but before seeing this. That fix already got merged, and approaches the problem in a different way. You identified the problem correctly -- my fix just handles the possible rejection further upstream.

Actually, if you feel so inclined, could you take a look at #1232 and see if that solves the problem, and if the changes make sense?

lucksus:
#1232 got merged and develop has conflicting changes that look like an even more sophisticated solution to me. Do we still need this? @MightyAlex200, could you check if the version in develop right now solves all your needs so we can close this?

MightyAlex200:
Yep, it's fixed, exceptions no longer hang tests.

PR #1244, opened 2019-04-11
fix dev pulse sync regex

thedavidmeister:

this is a quick fix to the regex that was scooping up too much of the dev pulse if "Details" appears later


shell.nix
```diff
@@ -498,7 +498,7 @@ All binaries are for 64-bit operating systems.
    # gets a markdown version of pulse
    # greps for everything from summary to details (not including details heading)
    # deletes null characters that throw warnings in bash
-   PULSE_NOTES=$( curl -s https://md.unmediumed.com/${pulse-url} | grep -Pzo "(?s)(###.*Summary.*)(?=###.*Details)" | tr -d '\0' )
+   PULSE_NOTES=$( curl -s https://md.unmediumed.com/${pulse-url} | grep -Pzo "(?s)(###.*Summary.*)(?=###\s+\**Details)" | tr -d '\0' )
```
pdaoust:
```suggestion
   PULSE_NOTES=$( curl -s https://md.unmediumed.com/${pulse-url} | grep -Pzo "(?s)(###\s+\**Summary.*)(?=###\s+\**Details)" | tr -d '\0' )
```

@thedavidmeister three thoughts:

1. it seems good to me to apply the same fix to the 'Summary' header as well. (See suggested change)
2. re: `\**` what is this fixing? Is the 'Details' element sometimes wrapped in bold or italic, _in addition to_ being a header? (I can totally believe it; just making sure I understood that piece of regex.)
3. Might be good to future-proof the `###` against the styling idiosyncrasies of future social media managers by going `#{1,}` instead. May be paranoid, but give the way that people use WYSIWYG editors, I reckon paranoia is entirely justified :D

pdaoust:
Meta-discussion about this regex that's unrelated to the current fix: do we need the activity snapshot in the release notes? I feel like it's not really germane to the release itself, or at least it's not terribly interesting to those who merely want to download a binary, which is the most likely reason to be on the releases page.

shell.nix
pdaoust: oh blast -- I didn't refresh this page and discover it'd already been merged :grimacing: whoops.

thedavidmeister:
@pdaoust 

> it seems good to me to apply the same fix to the 'Summary' header as well. (See suggested change)

i'll pick this up in next release, good spot thanks!

> Is the 'Details' element sometimes wrapped in bold or italic, in addition to being a header? 

yes

> Might be good to future-proof the ### against the styling idiosyncrasies of future social media managers by going #{1,} instead. May be paranoid, but give the way that people use WYSIWYG editors, I reckon paranoia is entirely justified :D

yeah, right now everything relies on medium as we suck everything out of https://unmediumed.com to get markdown - it's all tightly coupled to medium

medium are quite disciplined about their approach to WYSIWYG handling, notably mapping styles <> DOM:

https://medium.engineering/why-contenteditable-is-terrible-122d8a40e480

i'm not tooo worried atm, although i agree with your concerns if we ever open this up more

PR #1245, opened 2019-04-11
vendor openssl where needed (static)

thedavidmeister:
## changes

vendors (static link) openssl on nixos and other linux et al

continues to use `native-tls` logic on windows and mac os x, which uses native infra and not openssl at all

note that this doesn't work with wasm, but also shouldn't need to (there was confusion around this due to the way the AV builds were running)

### changelog


thedavidmeister:
it's working if somewhere in the compilation we see:

```
   Compiling openssl-src v111.2.1+1.1.1b
```

thedavidmeister:
i tested this on nixos and ubuntu (in a docker) locally

i think perl is already on ubuntu/debian

PR #1250, opened 2019-04-11
Links Implement Get Headers

AshantiMutinta:


AshantiMutinta:
Planning to do these in small digestible chunks so working on the crud_status links in parallel to this.

app_spec/test/test.js
```diff
@@ -440,6 +440,29 @@ scenario1.runTape('my_posts_immediate_timeout', async (t, { alice }) => {
   t.equal(JSON.parse(result.Err.Internal).kind, "Timeout")
 })
 
+scenario2.runTape('get_soruces_from_link', async (t, { alice, bob }) => {
```
zippy:
```suggestion
scenario2.runTape('get_sources_from_link', async (t, { alice, bob }) => {
```

app_spec/test/test.js
```diff
@@ -440,6 +440,29 @@ scenario1.runTape('my_posts_immediate_timeout', async (t, { alice }) => {
...
+
+  const bob_posts = alice.call("blog","authored_posts_with_sources",
+  {
+    "agent" : bob_address.Ok
+  });
+  t.equal(bob.agentId,bob_posts.Ok.headers[0].provenances[0][0]);
+  t.equal(alice.agentId,alice_posts.Ok.headers[0].provenances[0][0]);
+
+})
+
```
zippy: can we add a test that confirms that if two people add the same link, we get both sources?

AshantiMutinta: Yes I think that is a good idea. I will do just that

app_spec/test/test.js
```diff
@@ -440,6 +440,59 @@ scenario1.runTape('my_posts_immediate_timeout', async (t, { alice }) => {
...
+);
+  
+  const alice_posts = bob.call("blog","authored_posts_with_sources",
+  {
+    "agent" : alice_address.Ok
+  });
+
+  const bob_posts = bob.call("blog","authored_posts_with_sources",
+  {
+    "agent" : alice_address.Ok
```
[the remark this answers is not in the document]
lucksus: Ah, I see below you added `create_post_with_agent`. Did you mean to call that function above?

app_spec/test/test.js
```diff
@@ -440,6 +440,59 @@ scenario1.runTape('my_posts_immediate_timeout', async (t, { alice }) => {
...
+  });
+
+ console.log("provananced for alice" + JSON.stringify(alice_posts.Ok));
+ console.log("provananced for alice" + JSON.stringify(bob_posts.Ok));
+
+  t.equal(bob.agentId,alice_posts.Ok.headers[0].provenances[0][0]);
+  t.equal(alice.agentId,alice_posts.Ok.headers[1].provenances[0][0]);
+  t.equal(bob.agentId,bob_posts.Ok.headers[0].provenances[0][0]);
+  t.equal(alice.agentId,bob_posts.Ok.headers[1].provenances[0][0]);
```
lucksus: I'm a bit surprised that this test actually passes...

app_spec/test/test.js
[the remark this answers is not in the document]
AshantiMutinta: yeah meant to call that one instead, will make that change thanks!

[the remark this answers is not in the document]
AshantiMutinta: also Im using alice address because they have only been added with alice agent address, so I don't think we need a bob addess

PR #1255, opened 2019-04-12
Makefile cleanup: the current Makefile does not work, and needs to be fixed

pjkundert:
The current Makefile is not useful (doesn't know how to build 'holochain', incorrectly builds 'nodejs_conductor', fails for non-Nix-Shell Native toolchain/library builds, etc.)  However, many grey-beards still look to Makefile for build advice; I feel that it is important to maintain -- at least, as an interface to document how to invoke the alternative build approaches, eg. Nix Shell.

Adds support for both Nix/Native build. Adds support for building the main hc, holochain, WASM and NodeJS conductor targets via Nix Environment, as well as for selecting the Native libsodium library options to use for native builds (defaults to same build approach as used by shell.nix, with documentation to support other platforms, eg. Debian).

The appropiate Nix Shell targets required to build the Holochain Rust components is included in the Makefile for documentation, selection and execution.  This doesn't replace the direct Nix Shell approach; it just makes it accessible and documented for Makefile users.

Likewise, the selection of system libsodium library is foundational to the success of building/running the NodeJS tests. The options for selecting/building that are detailed here, as well -- again, the shell.nix approach is the "default", and alternatives are clearly laid out for platforms where that default approach does not work.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -249,7 +249,7 @@ wasm_build: ensure_wasm_target
 	  hdk-rust/wasm-test \
 	  wasm_utils/wasm-test/integration-test; do \
 	    echo -e "\033[0;93m## Building $${target}... ##\033[0m"; \
-	    ( cd core/src/nucleus/actions/wasm-test && $(CARGO) build --release --target wasm32-unknown-unknown ); \
+	    ( cd $${target} && $(CARGO) build --release --target wasm32-unknown-unknown ); \
 	done
 
 .PHONY: build_holochain libsodium_version
--- a/Makefile
+++ b/Makefile
@@ -1,32 +1,90 @@
 # holochain-rust Makefile
-# currently only supports 'debug' builds
+# - Build Holochain targets, either via Nix Shell (recommended), or using local toolchain/libs
 
+SHELL = /bin/bash
+
+# Native Toolchain {build,install,help}
+# 
+# Using the native toolchain and installed libraries may require system configuration (eg. see
+# RUST_SODIUM..., below)
 .PHONY: all build install help
-
 all: build
 
-build: build_holochain build_cli build_nodejs
+build: build_holochain build_cli build_node_conductor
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -165,12 +224,14 @@ test_cli: build_cli
 	cd cli && RUSTFLAGS="-D warnings" $(CARGO) test
 
 test_app_spec: RUST_VERSION=$(CORE_RUST_VERSION)
-test_app_spec: version_rustup ensure_wasm_target install_cli build_nodejs_conductor
+test_app_spec: version_rustup ensure_wasm_target install_cli build_node_conductor
```
lucksus: I don't think this is right. The native conductor is not used when running app-spec tests, but the nodejs-conductor is. This will change in the future though, but currently nodejs_conductor is a dependency for app-spec, `holochain` executable is not.

Makefile
```diff
@@ -181,32 +242,47 @@ test_c_ci: c_build c_binding_tests ${C_BINDING_TESTS}
...
-	$(CARGO) build --all --exclude hc
+	@$(CARGO) build --all --exclude hc
 
 .PHONY: build_cli
 build_cli: core_toolchain ensure_wasm_target
 	@echo -e "\033[0;93m## Building hc command... ##\033[0m"
-	$(CARGO) build -p hc
+	@$(CARGO) build -p hc --release
 
-.PHONY: build_nodejs
```
lucksus: Not sure why we would want to remove this?

thedavidmeister:
i have provided reviews of these code changes elsewhere that broadly mirror @zippy's review

this PR is blocked until it is split into at least the 3 conceptual components ^^

thedavidmeister:
i'm closing this @zippy @pjkundert 

@pjkundert please read the most recent comments and reopen as smaller/focussed PRs if needed

PR #1259, opened 2019-04-12
Allow specification of bridges in Config.conductor & new call format

maackle:


nodejs_conductor/test/test-config.js
```diff
@@ -66,6 +63,84 @@ test('config construction', t => {
...
+        bridges: [
+            C.bridge('bridge-forward', instance1, instance2),
+            C.bridge('bridge-backward', instance2, instance1),
+        ],
+        debugLog: true,
+    })
+    commonConfigAssertions(t, config, path)
+    t.ok(isFullDebugLogger(config.logger))
+    t.deepEqual(config.bridges, [
+        {handle: 'bridge-forward', caller_id: 'alessia', callee_id: 'bartolini'},
```
juliolrmonteiro: Just to make sure: the `caller_id` and `callee_id` here are instance_ids, not agent_ids, right?

maackle: @juliolrmonteiro yes, this is the same as in the conductor config. It just so happens the instance IDs here are the same as the agent names :)

doc/holochain_101/src/testing_configuration.md
```diff
@@ -112,36 +112,104 @@ console.log(dnaConfig)
...
+
+Takes three arguments: the bridge handle, the caller, and the callee (both instances)
+
+___
+
+**Name** handle
+
+**Type** `string`
+
+**Description** The desired bridge handle, as required by your DNA. See the [bridging section of the docs](./bridging.md) for more detail
```
Connoropolous: Can you state a very simple explanation of what the 'bridge handle' is used for?

doc/holochain_101/src/testing_configuration.md
```diff
@@ -112,36 +112,104 @@ console.log(dnaConfig)
...
+
+___
+
+**Name** callerInstanceConfig
+
+**Type** `object`
+
+**Description** A config object with a `name` property, as produced by `Config.instance`
```
Connoropolous: Can you state a very simple explanation of what the 'caller' is, in a bridging scenario?

doc/holochain_101/src/testing_configuration.md
```diff
@@ -112,36 +112,104 @@ console.log(dnaConfig)
...
+
+___
+
+**Name** calleeInstanceConfig
+
+**Type** `object`
+
+**Description** A config object with a `name` property, as produced by `Config.instance`
```
Connoropolous: Can you state a very simple explanation of what the 'callee' is, in a bridging scenario?

doc/holochain_101/src/testing_configuration.md
```diff
@@ -112,36 +112,104 @@ console.log(dnaConfig)
...
+*/
+```
+
 ## Full Conductor Configuration
 
+### `Config.conductor(conductorOptions)` => `object`
+
 ### `Config.conductor(instancesArray, [conductorOptions])` => `object`
 
+There are two ways to construct a valid Conductor configuration from these `Config` helpers. Using the first way, you put all the config data into a single required object. Using the second "shorthand" style, you specify an array of `Config.instance` data, along with an optional object of extra options. The second way can be more convenient when you are just trying to set up a collection of instances with nothing extra options.
```
Connoropolous:
```suggestion
There are two ways to construct a valid Conductor configuration from these `Config` helpers. Using the first way, you put all the config data into a single required object. Using the second "shorthand" style, you specify an array of `Config.instance` data, along with an optional object of extra options. The second way can be more convenient when you are just trying to set up a collection of instances with no extra options.
```

doc/holochain_101/src/testing_configuration.md
zippy:
```suggestion
**Description** A config object with a `name` property, as produced by `Config.instance`, where the name specifies which instance will be receiving calls over the bridge.
```

doc/holochain_101/src/testing_configuration.md
zippy:
```suggestion
**Description** A config object with a `name` property, as produced by `Config.instance`, where the name specifies which instance will be making calls over the bridge.
```

maackle:
@AshantiMutinta you're right, it probably should be a Result. There are two possible panics here, one for the keybundle that got added recently, and the other one is the `make_dna_config` function which actually currently cannot panic. However since we're hopefully retiring the nodejs conductor soon, maybe we can just leave it as is for now since this code will go away.

maackle:
@Connoropolous I updated the docs using most of @zippy's suggestions

PR #1265, opened 2019-04-14
2019 04 14 binary building script

thedavidmeister:
## PR summary

This brings the binary production component of the openssl work that was done for @samrose back into develop.

- adds `hc-build-release-artifacts` to nix
  - everything is compiled for `${linux-release-target}` aka `x86_64-unknown-linux-gnu`
  - builds node conductor for node 8.x
  - builds cli
  - builds rust conductor
- everything is built into `/dist`
- the tarballs no longer extract to a subdirectory as per https://github.com/holochain/holochain-rust/issues/1275

## testing/benchmarking notes

## followups

- plug all this into the travis file so that it uses these commands to builds linux artifacts
- handle other versions of node for the node conductor
- build scripts for other operating systems

## changelog


shell.nix
```diff
@@ -619,6 +621,37 @@ All binaries are for 64-bit operating systems.
...
+  build-release-artifact = path:
+  ''
+   export artifact_name=`sed "s/unknown/generic/g" <<< "${path}-${core-version}-${linux-release-target}"`
+   echo
+   echo "building $artifact_name..."
+   echo
+
+   CARGO_INCREMENTAL=0 cargo rustc --manifest-path ${path}/Cargo.toml --target ${linux-release-target} --release -- -C lto
+   mkdir -p dist/$artifact_name
+   cp target/${linux-release-target}/release/hc ${path}/LICENSE ${path}/README.md dist/$artifact_name
```
Connoropolous: @samrose experienced an issue, that we traced back to this line, and the fact that the `cp` command hardcodes `hc` into the path, when it also needs to accomodate `holochain` for the conductor binary. @thedavidmeister we are hoping you'll use your nix foo to resolve. 

Nevertheless, we got Sam unblocked by just changing the line by hand, to get the binaries built out

thedavidmeister: yes yes i have a solution i will push it sometime today

shell.nix
```diff
@@ -643,6 +685,9 @@ stdenv.mkDerivation rec {
     # for openssl static installation
     perl
 
+    # for openssl static installation
+    perl
+
```
maackle: duplicate?

PR #1266, opened 2019-04-15
Fix unspecified errors that can occur during entry deletion

Connoropolous:
## testing/benchmarking notes

## followups

## changelog


core/src/nucleus/ribosome/api/remove_entry.rs
```diff
@@ -42,10 +42,14 @@ pub fn invoke_remove_entry(runtime: &mut Runtime, args: &RuntimeArgs) -> ZomeApi
         .block_on(get_entry_result_workflow(&context, &get_args));
 
     if let Err(_err) = maybe_entry_result {
+        // TODO: don't just return "Unspecified"
```
Connoropolous: @AshantiMutinta what exactly can this error be?

core/src/nucleus/ribosome/api/remove_entry.rs
```diff
@@ -42,10 +42,14 @@ pub fn invoke_remove_entry(runtime: &mut Runtime, args: &RuntimeArgs) -> ZomeApi
...
+        // pass a relevant error for the error occurring
         return ribosome_error_code!(Unspecified);
     }
     let entry_result = maybe_entry_result.unwrap();
     if !entry_result.found() {
+        // TODO: don't just return "Unspecified"
+        // pass a relevant error for the error occurring
         return ribosome_error_code!(Unspecified);
```
Connoropolous: I discovered `EntryNotFound` in the `update_entry` API code, so I'm thinking that's relevant to use here too

core/src/nucleus/ribosome/api/remove_entry.rs
AshantiMutinta: Probably an error that tells us something went wrong during this get process. I'm not sure of the list of error types but I will look through and if not, we would have to create a new one for it.

PR #1270, opened 2019-04-15
Fix `call()` function's namespacing

pdaoust:
`s/hdk::call/hdk::api::call/g`

## changelog

### DIFF
--- a/doc/holochain_101/src/bridging.md
+++ b/doc/holochain_101/src/bridging.md
@@ -33,10 +33,10 @@ callee_id = "target-instance"
 handle = "sample-bridge"
 ```
 
-Then on the caller DNA you have to initiate the bridge call using `hdk::call` like this:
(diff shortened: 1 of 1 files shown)


pdaoust:
@lucksus thanks for letting me know. I learned something new today! It does bring up the question, though, of what our standards for documentation should be: should we always be explicit about the longest module path, or settle on dropping `api::` everywhere? @Connoropolous do you have any established style yet?

Connoropolous:
"do you have any established style yet?"

No, I agree it might be better were it consolidated for clarity. Probably without API makes more sense since having 'api' in the thing itself seems kinda redundant.

Connoropolous:
I am going to close for now, but we can keep discussing here, regardless.

pdaoust:
@Connoropolous I'm seeing people dropping the `api::` so let's pave the cowpaths (and save five keystrokes, many times over).

PR #1273, opened 2019-04-16
Capability Request available in global on zome function call

zippy:
## PR summary

Context: This is a pre-work PR getting ready for hdk::grant and hdk:claim.  It achieves:

- one public facing thing, which is a new Zome global  CAPABILITY_REQ which makes the request available during all zome calls.  This is important for doing validation of provenance for a zome call that wants to create a capability grant.
- one code reordering thing, which is moving the CapabilityRequest struct into core_types
- app-spec test of CAPABILITY_REQ

## changelog


core_types/src/dna/capabilities.rs
```diff
@@ -44,3 +25,47 @@ pub enum CapabilityType {
     #[serde(rename = "assigned")]
     Assigned,
 }
+
+//--------------------------------------------------------------------------------------------------
+// CapabilityRequest
+//--------------------------------------------------------------------------------------------------
+
+/// a struct to hold the capability information needed to make any capability request,
+/// namely the provenance of the request (the agent address an signature) and the
```
Connoropolous:
```suggestion
/// namely the provenance of the request (the agent address and signature) and the
```

CHANGELOG.md
```diff
@@ -28,6 +28,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Adds CAPABILITY_REQ global into the zome context. This makes the request available during all zome calls and is important for doing validation of provenance for a zome call that wants to create a capability grant. [#1273](https://github.com/holochain/holochain-rust/pull/1273)
```
Connoropolous: @zippy can you reword this?
This makes no sense to me...  I'll break down what I don't understand specifically if that helps...
What's "global into the zome context"?
"the request" which request?
"a capability grant" What's a capability grant, from the perspective of someone who's never heard of it before?

struktured:
It could considering reference prior work on capabilities- I believe we
mean capability in the sense of this:
https://en.wikipedia.org/wiki/Capability-based_security.

zippy:
@Connoropolous I updated the changelog text and added a link to the architecture ADR to explain capabilities grant rather than trying to do it in the change-log text itself, as that would be too verbose I think.

core/src/context.rs
```diff
@@ -279,33 +283,31 @@ impl Context {
...
+        for grant in grants {
+            let addr = grant.entry_address().to_owned();
+            let entry = get_entry_from_cas(&cas, &addr)?
+                .ok_or::<HolochainError>("Can't get CapTokenGrant entry from CAS".into())?;
+            // if entry is the public grant return it
+            if let Entry::CapTokenGrant(grant) = entry {
+                if grant.cap_type() == CapabilityType::Public
+                    && grant.id() == ReservedCapabilityId::Public.as_str()
+                {
+                    return Ok(addr);
```
AshantiMutinta: Just an added comment but we could probably use a fold() to make it look more idiomatic without the early return but no qualms about the way it is done here

core_types/src/entry/cap_entries.rs
```diff
@@ -21,6 +21,32 @@ pub enum CapabilityType {
...
+    /// used for identifying the default public capability
+    Public,
+}
+
+impl FromStr for ReservedCapabilityId {
+    type Err = &'static str;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "hc_public" => Ok(ReservedCapabilityId::Public),
+            _ => Err("Cannot convert string to ReservedCapabilityId"),
```
AshantiMutinta: I like this

PR #1276, opened 2019-04-16
holochain_net: P2pNetwork now waits for #P2P-READY#

ddd-mtl:
## PR summary
Have n3h users wait for the `#P2P-READY#` message.
Increased timeout because n3h now generates keys at init.

## changelog


struktured:
@ddd-mtl does this resolve https://miro.com/app/board/o9J_kyiXmFs=/?moveToWidget=3074457346565169866 and/or https://miro.com/app/board/o9J_kyiXmFs=/?moveToWidget=3074457346565169871? Do you have a good way to test that it does?

ddd-mtl:
@struktured I think so yes, indirectly.
 `#P2P-READY#` is different from `Protocol::P2pReady` though.
 `#P2P-READY#` means n3h spawned successfully. Which means its okay to send IPC messages to n3h, but that doesn't mean p2p networking is ready.
`Protocol::P2pReady` means p2p networking is ready. Messages can be sent over the network.
Currently n3h does wait for network readiness before sending  `#P2P-READY#`.
I don't know how much it will still be the case in the futur because p2p init does take some time.

To test I think you just need to comment the 1 sec sleep in P2pNetwork and see if two simple apps manage to see and talk to each other.

PR #1283, opened 2019-04-17
make portable - c_char is u8 on arm

neonphog:
## PR summary

This bit breaks when building on an arm processor, apparently c_char is a u8 instead of an i8 on those architectures... so use the c_char type directly.

## testing/benchmarking notes

builds / runs on arm64:

```
$ ./holochain --version
holochain 0.0.11-alpha1
```

```
$ readelf -h ./holochain
ELF Header:
  Class:                             ELF64
  Data:                              2's complement, little endian
  Type:                              DYN (Shared object file)
  Machine:                           AArch64
# ... snip
```

## followups

## changelog


thedavidmeister:
this code is very old

this highlights that we need some arm testing, add that to the ops queue ;)

PR #1285, opened 2019-04-18
HDK grant_capabiliy

zippy:
## PR summary

This PR adds `grant_capability(id, cap_type, assignees, functions)` to the hdk.

## changelog

### DIFF
--- a/app_spec/zomes/blog/code/src/lib.rs
+++ b/app_spec/zomes/blog/code/src/lib.rs
@@ -99,6 +99,13 @@ define_zome! {
             handler: blog::handle_request_post_grant
         }
 
+        get_grants: {
+            inputs: | |,
+            outputs: |result: ZomeApiResult<Vec<Address>>|,
+            handler: blog::handle_get_grants
+
+        }
+
         create_memo: {
             inputs: |content: String|,
             outputs: |result: ZomeApiResult<Address>|,
@@ -210,6 +217,6 @@ define_zome! {
     ]
 
     traits: {
-        hc_public [show_env, check_sum, check_send, get_sources, post_address, create_post, delete_post, delete_entry_post, update_post, posts_by_agent, get_post, my_posts,memo_address,get_memo,my_memos,create_memo,my_posts_as_committed, my_posts_immediate_timeout, recommend_post, my_recommended_posts,get_initial_post,get_history_post,get_post_with_options,get_post_with_options_latest,authored_posts_with_sources,create_post_with_agent,request_post_grant]
+        hc_public [show_env, check_sum, check_send, get_sources, post_address, create_post, delete_post, delete_entry_post, update_post, posts_by_agent, get_post, my_posts,memo_address,get_memo,my_memos,create_memo,my_posts_as_committed, my_posts_immediate_timeout, recommend_post, my_recommended_posts,get_initial_post,get_history_post,get_post_with_options,get_post_with_options_latest,authored_posts_with_sources,create_post_with_agent,request_post_grant,get_grants]
     }
 }
--- a/app_spec/zomes/blog/code/src/lib.rs
+++ b/app_spec/zomes/blog/code/src/lib.rs
@@ -210,6 +210,6 @@ define_zome! {
     ]
 
     traits: {
-        hc_public [show_env, check_sum, check_send, get_sources, post_address, create_post, delete_post, delete_entry_post, update_post, posts_by_agent, get_post, my_posts,memo_address,get_memo,my_memos,create_memo,my_posts_as_committed, my_posts_immediate_timeout, recommend_post, my_recommended_posts,get_initial_post,get_history_post,get_post_with_options,get_post_with_options_latest,authored_posts_with_sources,create_post_with_agent]
+        hc_public [show_env, check_sum, check_send, get_sources, post_address, delete_post, delete_entry_post, update_post, posts_by_agent, get_post, my_posts,memo_address,get_memo,my_memos,create_memo,my_posts_as_committed, my_posts_immediate_timeout, recommend_post, my_recommended_posts,get_initial_post,get_history_post,get_post_with_options,get_post_with_options_latest,authored_posts_with_sources,create_post_with_agent,handle_request_post_grant]
     }
 }
--- a/app_spec/zomes/blog/code/src/lib.rs
+++ b/app_spec/zomes/blog/code/src/lib.rs
@@ -210,6 +210,6 @@ define_zome! {
     ]
 
     traits: {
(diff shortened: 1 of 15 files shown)


doc/holochain_101/src/zome/capabilities.md
```diff
@@ -29,3 +30,29 @@ define_zome! {
 
 }
 ```
+
+### Grant Capabilities
+
+You can use the `grant_capabilities` HDK function to create a custom capability grant.  For example, imaging a blogging use-case where you want to grant friends the ability to call the `create_post` function in a `blog` zome.  Assuming the function `is_my_friend(addr)` correctly examines the provenance in CAPABILITY_REQ global which always holds the capability request of the current zome call, then the following code is an example of how you might call `hdk::grant_capability`:
```
lucksus:
```suggestion
You can use the `grant_capability` HDK function to create a custom capability grant.  For example, imaging a blogging use-case where you want to grant friends the ability to call the `create_post` function in a `blog` zome.  Assuming the function `is_my_friend(addr)` correctly examines the provenance in CAPABILITY_REQ global which always holds the capability request of the current zome call, then the following code is an example of how you might call `hdk::grant_capability`:
```

core/src/nucleus/ribosome/api/capabilities.rs
```diff
@@ -0,0 +1,83 @@
...
+use std::convert::TryFrom;
+use wasmi::{RuntimeArgs, RuntimeValue};
+
+pub fn invoke_grant_capability(runtime: &mut Runtime, args: &RuntimeArgs) -> ZomeApiResult {
+    let context = runtime.context()?;
+    // deserialize args
+    let args_str = runtime.load_json_string_from_args(&args);
+    let args = match GrantCapabilityArgs::try_from(args_str.clone()) {
+        Ok(input) => input,
+        Err(..) => return ribosome_error_code!(ArgumentDeserializationFailed),
```
struktured: Wouldn't hurt to add a log statement here about details of the deserialization error. Eventually I would even change the ribosome errors to be more structured but that's outside the scope of this PR.

app_spec/zomes/blog/code/src/lib.rs
```diff
@@ -204,6 +217,6 @@ define_zome! {
     ]
 
     traits: {
-        hc_public [show_env, check_sum, check_send, get_sources, post_address, create_post, delete_post, delete_entry_post, update_post, posts_by_agent, get_post, my_posts,memo_address,get_memo,my_memos,create_memo,my_posts_as_committed, my_posts_immediate_timeout, recommend_post, my_recommended_posts,get_initial_post,get_history_post,get_post_with_options,get_post_with_options_latest,authored_posts_with_sources,create_post_with_agent]
+        hc_public [show_env, check_sum, check_send, get_sources, post_address, create_post, delete_post, delete_entry_post, update_post, posts_by_agent, get_post, my_posts,memo_address,get_memo,my_memos,create_memo,my_posts_as_committed, my_posts_immediate_timeout, recommend_post, my_recommended_posts,get_initial_post,get_history_post,get_post_with_options,get_post_with_options_latest,authored_posts_with_sources,create_post_with_agent,request_post_grant,get_grants]
     }
```
struktured: Minor, but I think we should standardize whether we comma separate this list or not. Surprised the `hc-fmt` tool doesn't enforce this.

core/src/nucleus/ribosome/api/capabilities.rs
```diff
@@ -0,0 +1,83 @@
...
+    };
+
+    let task_result: Result<Address, HolochainError> =
+        match CapTokenGrant::create(&args.id, args.cap_type, args.assignees, args.functions) {
+            Ok(grant) => context.block_on(commit_entry(
```
struktured: This is another example where I think we could compose with futures instead. You are essentially forced to block here because you must return a zome api result. It is outside the scope of the PR, but we could change all these api functions to return a future zome api result, and universally handle the unraveling of these futures at the last moment when serializing over the network. In this model no explicit blocking is ever really necessary- even during initialization, if this is done right. Instead you just transform an existing future which achieves the same behavior.

core/src/nucleus/ribosome/api/capabilities.rs
```diff
@@ -0,0 +1,83 @@
...
+                None,
+                &context.clone(),
+            )),
+            Err(err) => Err(HolochainError::ErrorGeneric(format!(
+                "Unable to commit capability grant: {}",
+                err
+            ))),
+        };
+
+    runtime.store_result(task_result)
```
struktured: This store result invocation is another cross cutting concern these api functions always seem to exhibit. We would benefit from isolating common patterns such as these and leverage a code reuse feature from rust to DRY this out (maybe a trait implementation per api function, or a macro that all functions get wrapped over).

core/src/nucleus/ribosome/api/capabilities.rs
```diff
@@ -0,0 +1,83 @@
...
+        assert_eq!(
+            call_result,
+            JsonString::from_json(
+                &(String::from(JsonString::from(ZomeApiInternalResult::success(
+                    Address::from("Qma8KWBHZwiXNBJ4PBtT4uDUVgPAyUJASHumThZMTPAAJe")
+                ))) + "\u{0}")
+            ),
+        );
+    }
+}
```
struktured: Maybe I'm missing this in the diffs, but is there a test for the access granted or that access is denied if not granted? Or is this a different part of a different PR?

app_spec/zomes/blog/code/src/lib.rs
zippy: good point.  Just added that to the soa tree.

core/src/nucleus/ribosome/api/capabilities.rs
zippy: Hmm interesting.  Adding an item to tech-debt to talk this through.

core/src/nucleus/ribosome/api/capabilities.rs
zippy: There is not one because in app-spec we can't actually make the signed capability request inside the node conductor.  As soon as we get that I'll add a test there.

core/src/nucleus/ribosome/api/capabilities.rs
zippy: Yep, there's a set of SOA nodes about refactoring the way zome api calls are built...

app_spec/zomes/blog/code/src/lib.rs
zippy: Also, that particular one appears to get missed by cargo fmt (I just did it manually) because its in a macro.

PR #1286, opened 2019-04-18
Make Scenario API DPKI-aware

maackle:
## PR summary

Adds ability to easily set `dpki` property of conductor config via Scenario API

### TODO

- [ ] needs docs

## changelog


maackle:
@zo-el this fell through the cracks, but do you still need this?

zo-el:
Yes, definitely. Not that urgent but needed.

PR #1287, opened 2019-04-18
Clarify the DnaHashMismatch error a little

maackle:
## PR summary

Makes an error read better. The previous wording made it feel like the order of arguments was reversed. This feels like a better former/latter flow.

## changelog


lucksus:
@zippy, I've fixed the tests here. Please see if that implements your change-requests.

conductor_api/src/conductor/admin.rs
```diff
@@ -894,8 +894,8 @@ id = 'new-dna'"#,
                 None
             ),
             Err(HolochainError::DnaHashMismatch(
+                "wrong-address".into(),
                 dna.address(),
```
maackle: @lucksus nooo, this defeats the main purpose of this PR :) The test should be changed, not the ordering.

PR #1293, opened 2019-04-18
holochain_net: Peer disconnexion tests

ddd-mtl:
### `net`
Have parent process look for PID sent by n3h and use it kill child properly on windows.

### `test_bin`
Added `P2Node::track_dna() ` and `P2Node::untrack_dna()` in preperation of multidna tests.
Added three nodes disconnexion test.


test_bin/src/connection_workflows.rs
```diff
@@ -103,3 +105,190 @@ pub(crate) fn two_nodes_disconnect_test(
...
+    alex.stop();
+    billy.stop();
+    camille.stop();
+
+    Ok(())
+}
+
+//// Test disconnects with multiple dna per Node
+//#[cfg_attr(tarpaulin, skip)]
+//pub(crate) fn three_nodes_multi_dna_disconnect_test(
```
neonphog: remove?

net/src/ipc/spawn.rs
```diff
@@ -123,16 +140,30 @@ pub fn ipc_spawn(
...
     }));
 
     Ok(out)
 }
+
+#[cfg(windows)]
+fn kill_child(pid: &str) {
+    let mut child_killer = std::process::Command::new("taskkill");
+    child_killer.args(&["/pid", pid, "/f", "/t"]);
+    let _ = child_killer.spawn();
```
neonphog: I think you might want `.status()` here instead of `.spawn()` since we don't want this to be a long-running thing, and may want log reporting if it fails.

ddd-mtl: not much we can do about it if it fails, but okay

PR #1295, opened 2019-04-18
[WIP] - Use procedural macros to replace entry!

willemolding:
Extending upon #941 this is a proposal to also update the entry! macro which creates a ValidatingEntryType struct which the #[zome] macro can register as an app entry.

The current form of the macro can be very verbose and difficult to read in the case where there is complex validation or many different links e.g.

```rust
entry!(
    name: "public_stream",
    description: "A stream of which anyone can become a member and post",
    sharing: Sharing::Public,
    validation_package: || {
        hdk::ValidationPackageDefinition::Entry
    },
    validation: |_validation_data: hdk::EntryValidationData<Stream>| {
        Ok(())
    },
    links: [
        to!(
            "%agent_id",
            tag: "has_member",
            validation_package: || {
                hdk::ValidationPackageDefinition::Entry
            },
            validation: |_validation_data: hdk::LinkValidationData| {
                Ok(())
            }
        ),
        from!(
            "%agent_id",
            tag: "member_of",
            validation_package: || {
                hdk::ValidationPackageDefinition::Entry
            },
            validation: |_validation_data: hdk::LinkValidationData| {
                Ok(())
            }
        ),
        to!(
            "message",
            tag: "message_in",
            validation_package: || {
                hdk::ValidationPackageDefinition::Entry
            },
            validation: |_validation_data: hdk::LinkValidationData| {
                Ok(())
            }
        )
    ]
)
```


willemolding:
A candidate proposal is to keep the same format but streamline it by removing the confusing closure style defs for validation_package and validation to have something like this:

```rust
/// description can be rustdoc style comments which are picked up by the macro
define_entry! {
   name: "public_stream",
   type: Stream,
   sharing: Sharing::Public,
   validation_package: ValidationPackageDefinition::ChainFull,
   validation_callback: validate_stream,
   links_to: ['%agent_id', "message"],
   links_from: ['%agent_id'],
}

pub fn validate_stream() -> Result<()> {
    Ok(())
}
```

We could then split out the definition of the links specification

```rust
link!{
   from: "%agent_id",
   to: "public_stream",
   sharing: Sharing::Public,
   validation_callback: validate_agent_to_stream_link,
}
```

compile time checks could ensure that every links to/from specified in an entry has a corresponding definition

maackle:
I guess if you had compile-time checks for `link!` definitions, you wouldn't even need `links_from` and `links_to` in the entry definition.

I like this because a link does not really belong to any one entry type, it is its own thing that references two entry types, so it's arbitrary which entry type you associate it with. Let it be completely unaffiliated, and do compile time checks against registered entry type names.

Connoropolous:
@willemolding I am big in favor of this direction... it cleans up some of the outstanding ugly/confusing parts of the Rust HDK 👍

Connoropolous:
I wonder why it wasn't done this way in the first place...
```
validation_package: ValidationPackageDefinition::ChainFull,
validation_callback: validate_stream,
```
oh right because we didn't have proc-macros

Connoropolous:
"I like this because a link does not really belong to any one entry type, it is its own thing that references two entry types, so it's arbitrary which entry type you associate it with. Let it be completely unaffiliated, and do compile time checks against registered entry type names."

So true @maackle this has really bugged me, knowing which side to put the link def on, and I am not even sure how/if Holochain knows which side its defined on, and how to check for it

willemolding:
Proposal is to split out links and entries entirely. The resulting `entry!` macro would therefore look like:
```
/// description can be rustdoc style comments which are picked up by the macro
entry! {
   name: "public_stream",
   type: Stream,
   sharing: Sharing::Public,
   validation_package: ValidationPackageDefinition::ChainFull,
   validation_callback: validate_stream,
}
```

thedavidmeister:
> I like this because a link does not really belong to any one entry type,

:100:

thedavidmeister:
yeah i'm keen to start seeing this in action :)

lucksus:
> So true @maackle this has really bugged me, knowing which side to put the link def on, and I am not even sure how/if Holochain knows which side its defined on, and how to check for it

When validating a link, we first get base and target to get their type and the we search in both entry defs. for a matching link.

The history of links: 
* in proto links have been entry types which meant you had to (manually) define a link entry type and then adding a link was committing that link entry which was cumbersome and confusing
* with the Rust refactor we changed links to be an aspect of the entry that we link from and introduced `hdk::link_entries`
* then we realized that while this allows for a cleaner definition, it makes it impossible to link from something that we don't have an explicit entry definition for (like DNA and agent)
* that's why we added the `linked_from` case so we can say in post entry definition for example that posts can be attached to the agent ID.

I think having links be onotologically scoped within makes a lot of sense and allows for cleaner expressions. I would argue that if you have a case where a link really is not an aspect of one of the entries being linked, it actually should be an *entry type* (as was the only possible way in proto) that defines links to the things it connects - i.e. an explicit edge entry type.

Thoughts?

lucksus:
That aside, yay for putting validation callbacks into functions and getting the syntax cleaner!

thedavidmeister:
@lucksus :man_shrugging: i think that's something real world usage will flush out

thedavidmeister:
@lucksus so, even if we go with the "join table entry" style proposal, does it impact this proposal?

having a clean `entry!` and `link!` macro pair with references to callbacks rather than nested inline macrology seems good right?

embrace the flat

lucksus:
> @lucksus so, even if we go with the "join table entry" style pattern as "best practise", does it impact this proposal?

Not sure I get your question, @thedavidmeister.
The "join table entry" style for links is the one thing I don't like about this proposal. I like all other aspects: plain rust functions with annotation as validation functions, less macro-foo.

I reiterate: I think it makes much more sense to regard links as aspects of the entries that are linked. Even stronger: I think it makes sense to regard link types as an aspect of **one** of the entry types it links. I am looking at it from the lens of classes and their associations. As in:

```ruby
class Agent
  has_many :comments
end
```

But: we don't have to lock us into one specific syntax. As with the other procedural macro HDK additions, we could keep the old style as well and allow testing and playing with several syntaxes before making an exclusive decision.

So, I'm totally fine with this proposal being implemented this way :)

PR #1296, opened 2019-04-18
Fix n3h version check

lucksus:
## PR summary

Although I made the pinned n3h version match the version in the n3h branch I'm working on, I couldn't get the conductor to use my local n3h from $PATH. Found out that the version check fails if the string returned from the spawned command does not match the regex, although in my case the string was exactly the version that was pinned.

This adds a plain `res == &N3H.version` test and refactors that function so the flow easier to understand.

### DIFF
--- a/net/src/ipc/n3h.rs
+++ b/net/src/ipc/n3h.rs
@@ -150,11 +150,10 @@ fn check_n3h_version(path: &std::path::PathBuf) -> NetResult<Vec<String>> {
 }
 
 fn sub_check_n3h_version(path: &std::path::PathBuf, out_args: &[&str]) -> NetResult<bool> {
-    let res = exec_output(path, out_args, ".", false)
-        .map_err(|err| {
-            println!("response: {:?}", err);
-            format_err!("n3h not found in path: {:?}", &path)
-        })?;
+    let res = exec_output(path, out_args, ".", false).map_err(|err| {
+        println!("response: {:?}", err);
+        format_err!("n3h not found in path: {:?}", &path)
+    })?;
 
     // `n3h --version` returns only the version number, for example "v0.0.11-alpha"
     if res == N3H_INFO.version {
--- a/net/src/ipc/n3h.rs
+++ b/net/src/ipc/n3h.rs
@@ -167,14 +167,15 @@ fn sub_check_n3h_version(path: &std::path::PathBuf, out_args: &[&str]) -> NetRes
             version = Some(m[1].to_string());
         }
 
-        if version.is_none() || version.as_ref().unwrap() != &N3H_INFO.version {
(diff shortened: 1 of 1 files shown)


net/src/ipc/n3h.rs
```diff
@@ -150,27 +150,33 @@ fn check_n3h_version(path: &std::path::PathBuf) -> NetResult<Vec<String>> {
...
-        let res = res.unwrap();
+    let res = exec_output(path, out_args, ".", false).map_err(|err| {
+        println!("response: {:?}", err);
+        format_err!("n3h not found in path: {:?}", &path)
+    })?;
+
+    // `n3h --version` returns only the version number, for example "v0.0.11-alpha"
+    if res == N3H_INFO.version {
+        Ok(true)
+    } else {
```
ddd-mtl: You can `return Ok(())` and remove the `else`.

net/src/ipc/n3h.rs
```diff
@@ -150,27 +150,33 @@ fn check_n3h_version(path: &std::path::PathBuf) -> NetResult<Vec<String>> {
...
         let re = regex::Regex::new(r#"(?m)#\s+n3h\s+version:\s+"([^"]+)"\s+#$"#)?;
 
         let mut version = None;
         for m in re.captures_iter(&res) {
             version = Some(m[1].to_string());
         }
 
-        if version.is_none() || version.as_ref().unwrap() != &N3H_INFO.version {
+        if version.is_some() && version.as_ref().unwrap() == &N3H_INFO.version {
+            Ok(true)
```
ddd-mtl: same here

net/src/ipc/n3h.rs
```diff
@@ -150,27 +150,33 @@ fn check_n3h_version(path: &std::path::PathBuf) -> NetResult<Vec<String>> {
 }
 
 fn sub_check_n3h_version(path: &std::path::PathBuf, out_args: &[&str]) -> NetResult<bool> {
-    let res = exec_output(path, out_args, ".", false);
```
ddd-mtl: Might as well return `NetResult<()>`, that boolean is not used for anything.

lucksus:
Yes, @neonphog, I see this:
```
lucksus@terence:~$ n3h --version
v0.0.11-alpha
lucksus@terence:~$ 
```
That is from https://github.com/holochain/n3h/pull/71 in which I had merged develop when I worked on this last Thursday.

zippy:
Is this ready to land @neonphog @lucksus ?

lucksus:
Yes, @neonphog, you're correct. With n3h's master in my n3h branch I get the right version output, so yeah, this is not needed. Will close this PR.

PR #1324, opened 2019-04-24
Allow n3h to shutdown gracefully

neonphog:
## PR summary

Rust std child process does not support a graceful shutdown signal, only a hard [kill](https://doc.rust-lang.org/std/process/struct.Child.html#method.kill). This adds a graceful shutdown request on all non-windows systems so the n3h process can do some cleanup / bookkeeping on shutdown.

## testing/benchmarking notes

## followups

## changelog


neonphog:
@Connoropolous 

> can you explain this one in the Summary section of the PR @neonphog ?

done

neonphog:
@Connoropolous 

> @neonphog per your checkmark re. affecting end users... it seems to me this would have come up as an issue for somebody, and that's why we're fixing it. is that not true?

No, so far not an issue. I'm working on a branch in n3h to output some debugging for the test_bin network tests when n3h is shut down, but it never gets the chance to display it because of the kill signal. This PR doesn't change any current functionality, only allows future functionality.

net/src/ipc/spawn.rs
```diff
@@ -158,6 +161,22 @@ pub fn ipc_spawn(
...
+    false
+}
+
+#[cfg(not(windows))]
+fn term_child(pid: u32) -> bool {
+    unsafe {
+        if libc::kill(pid as i32, libc::SIGTERM) == 0 {
+            libc::waitpid(pid as i32, std::ptr::null_mut(), 0);
+            return true;
+        }
```
struktured: What about when kill returns nonzero? Are we concerned about this cases at all, at least perhaps from a logging perspective?

```
The kill() function shall fail if:

[EINVAL]
The value of the sig argument is an invalid or unsupported signal number.
[EPERM]
The process does not have permission to send the signal to any receiving process.
[ESRCH]
No process or process group can be found corresponding to that specified by pid.
```

There is similar error handling possible with `waitpid`, but again not sure how important it is to check for it here.

neonphog: Yeah, it's not very robust, but if the SIGTERM fails, it should still send the KILL.

Added an issue to implement a more robust solution in the future: https://github.com/holochain/holochain-rust/issues/1357

PR #1325, opened 2019-04-25
hdk::commit_capability_claim

zippy:
## PR summary

- adds `hdk::commit_capability_claim()` for storing a capability grant for use in making calls later
- renames `hdk::grant_capability()` -> `hdk::commit_capability_grant()`

## Testing

- includes an app-spec test that demonstrates how to use capabilities inside the context of send/receive. 

## Followups

- this PR revealed a bug somewhere deep in our runtime manager (or in link_entries) because the call to link_entries when made inside the receive handler of the app_spec test fails with an `{"kind":{"ErrorGeneric":"Error calling receive callback: Fail(\\"WASM invocation failed: Trap: Trap { kind: Host(BadCallError) }\\")"}`  error being returned. 
- this PR revealed the need for passing in the from node address as a parameter to the receive call-back.

## changelog


app_spec/zomes/blog/code/src/blog.rs
```diff
@@ -107,17 +116,24 @@ pub fn handle_post_address(content: String) -> ZomeApiResult<Address> {
...
-    addr == Address::from("HcScjwO9ji9633ZYxa6IYubHJHW6ctfoufv5eq4F7ZOxay8wR76FP4xeG9pY3ui")
+    addr == Address::from(BOB_AGENT_ID)
 }
 
 pub fn handle_request_post_grant() -> ZomeApiResult<Option<Address>> {
-    let addr = CAPABILITY_REQ.provenance.source();
+    // if it worked for the nodejs conductor to make calls with different provenances
+    // we could get the caller we want from the CAPABILITY_REQ global like this:
+    //    let addr = CAPABILITY_REQ.provenance.source();
+    // but it doesn't work yet so for this test we are hard-coding the "friend"" to bob
```
maackle: This is an open conversation to loop back to. Last time we talked about it, it sounded like it was beneficial to test it the way you did here, because that would be more like how a real app would work.

app_spec/zomes/blog/code/src/blog.rs
```diff
@@ -132,6 +148,172 @@ pub fn handle_get_grants() -> ZomeApiResult<Vec<Address>> {
...
+        },
+    };
+    json!({
+        "msg_type": "response",
+        "body": response
+    })
+    .to_string()
+}
+
+fn find_claim(_identifier: &str, _grantor: &Address) -> Result<Address, HolochainError> {
```
maackle: Maybe use these args? Or explain why you're not?

core/src/entry/mod.rs
```diff
@@ -12,6 +12,7 @@ impl CanPublish for EntryType {
         match self {
             EntryType::Dna => return false,
             EntryType::CapTokenGrant => return false,
+            EntryType::CapTokenClaim => return false,
```
struktured: This could become `EntryType::CapTokenGrant | EntryType::CapTokenClaim => return false`. In a future PR we should isolate any use of a wildcard when pattern matching and explode out the enum cases.

core_types/src/entry/cap_entries.rs
```diff
@@ -62,18 +62,26 @@ pub type CapTokenValue = Address;
...
+    }
+    pub fn token(&self) -> CapTokenValue {
+        self.token.clone()
+    }
+    pub fn id(&self) -> String {
+        self.id.clone()
     }
-    pub fn token(self) -> CapTokenValue {
-        self.token
+    pub fn grantor(&self) -> Address {
```
struktured: What happen to CapTokenValue?

app_spec/test/test.js
```diff
@@ -17,6 +17,48 @@ const scenario1 = new Scenario([instanceAlice], { debugLog:true })
 const scenario2 = new Scenario([instanceAlice, instanceBob], { debugLog: true })
 const scenario3 = new Scenario([instanceAlice, instanceBob, instanceCarol], { debugLog: true })
 
+scenario2.runTape('capabilities grant and claim', async (t, { alice, bob }) => {
+    /*
+      This is not a perfect test of requesting a grant because currently there
+      is no way in the test conductor to actually pass in the provenance of the
+      call.  That will be added when we convert the test framework to being built
+      on top of the rust conductor.
```
maackle: Is this still the case given the other comment about directly specifying an alternate provenance?

core_types/src/entry/cap_entries.rs
zippy: good point, I think I should delete it completely because it's allways an address and I don't think the type system can help us in any way around when we are treating an address as a token and when not.

app_spec/test/test.js
zippy: good point deleted

app_spec/zomes/blog/code/src/blog.rs
zippy: yep. updated the comment a bit

core/src/entry/mod.rs
zippy: yep, updated

PR #1326, opened 2019-04-25
Macro hdk code review

willemolding:
## PR summary

This PR is for reviewing the actual implementation of the proc macro HDK. All discussions regarding the HDK itself should be posted to #941 

## testing/benchmarking notes

- app spec tests now also run against the proc macro app spec in CI

## changelog

- [ x] this is not a code change, or doesn't effect anyone outside holochain core development


hdk-proc-macros/src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+
+mod code_generators;
+mod into_zome;
+mod to_tokens;
+mod zome_code_def;
+
+use crate::zome_code_def::ZomeCodeDef;
+
+/**
+ * @brief Defines the #[zome] cacro to be used on a Rust module.
```
lucksus:
```suggestion
 * @brief Defines the #[zome] macro to be used on a Rust module.
```

hdk-proc-macros/src/into_zome.rs
```diff
@@ -0,0 +1,503 @@
...
+                });
+                acc
+            })
+    }
+
+    fn extract_entry_defs(&self) -> Vec<syn::ItemFn> {
+        funcs_iter(self)
+            .filter(is_tagged_with(ENTRY_DEF_ATTRIBUTE))
+            .fold(Vec::new(), |mut acc, mut func| {
+                // drop all attributes on the fn. This may cause problems
```
lucksus:
```suggestion
                // TODO: drop all attributes on the fn. This may cause problems
```

app_spec_proc_macro/zomes/blog/code/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    #[receive]
+    pub fn receive(message: String) {
+        json!({
+            "message": message
+        }).to_string()
+    }
+
+    #[zome_fn("hc_public")]
```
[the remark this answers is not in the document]
willemolding: This was to keep the migration from the old app spec as light as possible. There is another problem that the #[zome_fn] attributes only have meaning when used inside a module that has the #[zome] attribute. This way seemed nicer than putting all the full function bodies in the same module.

app_spec_proc_macro/zomes/blog/code/src/post.rs
```diff
@@ -0,0 +1,214 @@
...
+    validation::{EntryValidationData},
+    dna::entry_types::Sharing
+};
+
+/// We declare the structure of our entry type with this Rust struct.
+/// It will be checked automatically by the macro below, similar
+/// to how this happens with functions parameters and zome_functions!.
+///
+/// So this is our normative schema definition:
+#[derive(Serialize, Deserialize, Debug, DefaultJson,Clone)]
```
Connoropolous:
```suggestion
#[derive(Serialize, Deserialize, Debug, DefaultJson, Clone)]
```

hdk-proc-macros/src/code_generators/zome_setup.rs
```diff
@@ -0,0 +1,23 @@
...
+            .iter()
+            .map(|func| func.ident.clone())
+            .clone();
+
+        quote! {
+            #[no_mangle]
+            #[allow(unused_variables)]
+            pub extern "C" fn zome_setup(zd: &mut hdk::meta::ZomeDefinition) {
+                #(
+                    zd.define(#entry_fn_idents ());
```
Connoropolous: What's happening here where there's a space?
`#entry_fn_idents ()`
does that matter?

hdk-proc-macros/src/into_zome.rs
```diff
@@ -0,0 +1,503 @@
...
+impl IntoZome for syn::ItemMod {
+    fn extract_genesis(&self) -> GenesisCallback {
+        // find all the functions tagged as the genesis callback
+        let geneses: Vec<Box<syn::Block>> = funcs_iter(self)
+            .filter(is_tagged_with(GENESIS_ATTRIBUTE))
+            .fold(Vec::new(), |mut acc, func| {
+                acc.push(func.block);
+                acc
+            });
+        // only a single function can be tagged in a valid some so error if there is more than one
```
Connoropolous:
```suggestion
        // only a single function can be tagged in a valid Zome so error if there is more than one
```

hdk-proc-macros/src/into_zome.rs
```diff
@@ -0,0 +1,503 @@
...
+        // if there is None then use the sensible default of Ok(())
```
Connoropolous: This comment doesn't seem to match up with the logic below?

hdk-proc-macros/src/into_zome.rs
```diff
@@ -0,0 +1,503 @@
...
+                        }
+                    })
+                    .collect()
+            }
+            None => Vec::new(),
+        }
+    }
+
+    fn extract_receive_callback(&self) -> Option<ReceiveCallback> {
+        // find all the functions tagged as the genesis callback
```
Connoropolous:
```suggestion
        // find all the functions tagged as the receive callback
```

hdk-proc-macros/src/into_zome.rs
```diff
@@ -0,0 +1,503 @@
...
+mod tests {
+    use super::*;
+    use syn::parse_quote;
+
+    #[test]
+    fn test_extract_genesis_smoke_test() {
+        let module: syn::ItemMod = parse_quote! {
+            mod zome {
+                #[genesis]
+                fn genisis() {
```
Connoropolous: is this intentionally `genisis` not `genesis`?

hdk-proc-macros/src/into_zome.rs
```diff
@@ -0,0 +1,503 @@
...
+        };
+        let _ = module.extract_zome();
+    }
+
+    #[test]
+    fn test_extract_single_trait() {
+        let module: syn::ItemMod = parse_quote! {
+            mod zome {
+                #[genesis]
+                fn genisis() {
```
Connoropolous: Same thing: is this intentionally genisis not genesis?

hdk-proc-macros/src/into_zome.rs
```diff
@@ -0,0 +1,503 @@
...
+            expected_traits
+        }
+    }
+
+    #[test]
+    fn test_extract_function_params_and_return() {
+        let module: syn::ItemMod = parse_quote! {
+            mod zome {
+                #[genesis]
+                fn genisis() {
```
Connoropolous: same

hdk-proc-macros/src/into_zome.rs
```diff
@@ -0,0 +1,503 @@
...
+            }
+        }
+    }
+
+    #[test]
+    fn test_extract_function_with_generic_return() {
+        let module: syn::ItemMod = parse_quote! {
+            mod zome {
+                #[genesis]
+                fn genisis() {
```
Connoropolous: same

hdk-proc-macros/src/into_zome.rs
```diff
@@ -0,0 +1,503 @@
...
+            }
+        }
+    }
+
+    #[test]
+    fn test_single_entry() {
+        let module: syn::ItemMod = parse_quote! {
+            mod zome {
+                #[genesis]
+                fn genisis() {
```
Connoropolous: same

maackle:
Is the plan to replace `app_spec` with `app_spec_proc_macro` once this lands? Seems problematic to have mostly duplicated test code that can get out of sync.

Connoropolous:
@maackle 

hm... if the two really do match, maybe we could have one set of js tests and test both against that...
we were thinking of spending a reasonably long period of time just refining the proc macro hdk before announcing it, fully documenting it, etc, so there will be a period where they both stay in the repo

hdk-proc-macros/src/code_generators/zome_setup.rs
willemolding: Strangely if you expand the macro using `cargo expand` this is how it expands function calls be default. I think I did this the same way so the output is consistent.

hdk-proc-macros/src/into_zome.rs
willemolding: haha no. It actually does't matter what it is called but I better fix it hey

hdk-proc-macros/src/into_zome.rs
willemolding: > @maackle
> 
> hm... if the two really do match, maybe we could have one set of js tests and test both against that...
> we were thinking of spending a reasonably long period of time just refining the proc macro hdk before announcing it, fully documenting it, etc, so there will be a period where they both stay in the repo

That is how it is set up now! The tests are copied from `app_spec` to `app_spec_proc_macro` when the tests run because of the limitation of `hc test` that they must be in the same dir.

willemolding:
> @maackle
> 
> hm... if the two really do match, maybe we could have one set of js tests and test both against that...
> we were thinking of spending a reasonably long period of time just refining the proc macro hdk before announcing it, fully documenting it, etc, so there will be a period where they both stay in the repo

That is how it is set up now! The tests are copied from `app_spec` to `app_spec_proc_macro` when the tests run because of the limitation of `hc test` that they must be in the same dir.

maackle:
> The tests are copied from app_spec to app_spec_proc_macro when the tests run because of the limitation of hc test that they must be in the same dir.

If they're just copied over, was it intended to add the copied stuff to version control? I didn't look into it too deeply yet, I just see all these new files in the diff.

If it's just a limitation of `hc test`, you could also `node test/index.js`.

willemolding:
> > The tests are copied from app_spec to app_spec_proc_macro when the tests run because of the limitation of hc test that they must be in the same dir.
> 
> If they're just copied over, was it intended to add the copied stuff to version control? I didn't look into it too deeply yet, I just see all these new files in the diff.
> 
> If it's just a limitation of `hc test`, you could also `node test/index.js`.

Oh you are totally right. Thanks!

hdk-proc-macros/src/to_tokens.rs
```diff
@@ -0,0 +1,141 @@
...
+        let receive_callback = self.receive_callback();
+        let panic_handler = panic_handler();
+
+        tokens.extend(quote! {
+
+            #(#extra)*
+
+            #(#entry_def_fns )*
+
+            #genesis
```
zippy: small thing: when this gets merged it will need to be updated to `init` and I'd suggest we make it be `init_callback` to match `receive_callback.`

maackle:
Are we still keeping the old macro syntax around for a while? If not, why not just clobber the old app_spec with the new one?

zippy:
I think we should totally keeping the old one around for a bit.  This kind of change needs a deprecation phase, and it seems easy to do.  However we might say that we do future app_spec additions only to the new app_spec...

PR #1333, opened 2019-04-26
Add agent with auto keygen

willemolding:
## PR summary

Allows a "passphrase" field to be passed to RPC calls to admin/agent/add. This prevents the conductor from requesting a passphrase from the command line. It is also totally insecure since the passphrase is sent in plain text. This is for the purpose of creating test agents via the scenario-test.js

Also updates the rustdoc for admin/agent/add which was previously not in sync with the code.

## followups

This should enable the scenario-test.js create test agents in the rust conductor. It is slow to generate keypairs though taking ~13 seconds per agent. This is not ideal overhead for every test run so it might need to be changed.

## changelog

- [x ] this is not a code change, or doesn't effect anyone outside holochain core development

### DIFF
--- a/conductor/example-config/empty-container.toml
+++ b/conductor/example-config/empty-container.toml
@@ -1,8 +1,4 @@
-[[agents]]
-id = "test agent 1"
-name = "Holo Tester 1"
-public_address = "HoloTester1-----------------------------------------------------------------------AAACZp4xHB"
-keystore_file = "holo_tester.key"
+agents = []
 
 dnas = []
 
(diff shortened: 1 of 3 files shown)


willemolding:
@lucksus based on our brief chat this morning I think we need a new RPC function for adding agents specifically for running tests. These agents will use the test_keystore_loader like in the current test setup so we don't need to overhead of creating a valid keystore file (which can take ~15 second per agent).

I suggest we add a new namespace `test` so we can have calls like `test/agent/add` and in the future these could be limited to only special test interfaces.

willemolding:
This has been replaced by #1359

PR #1334, opened 2019-04-26
[WIP] - Rust conductor testing

willemolding:
WIP for starting to develop a testing harness that makes calls and waits using RPC and signals


lucksus:
:clap:

scenario-tests.js/index.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  /**
+   * Calls the conductor RPC functions to initialize it according to the config
+   */
+  async initialize () {
+    // setup the DNAs
+    // dna = {id, path}
+    const call = this.call
+    await this.config.dnas.forEach(async dna => {
```
Connoropolous: @willemolding if we have all the info at time of initialization, why not initialize based on a config file instead? how were you thinking the conductor would get started?

scenario-tests.js/index.js
```diff
@@ -0,0 +1,178 @@
...
+    return this.agentIds[instanceId]
+  }
+
+  dna_address (instanceId) {
+    return this.dnaAddresses[instanceId]
+  }
+
+  register_callback (callback) {
+    throw new Error('Not Implemented')
+  }
```
maackle: The `register_callback` was necessary for neon and will probably go away

maackle:
Not sure how desirable it is to replicate the existing API 1:1, as that was something I put together without any input or guidance from anyone else. We have the chance to improve it and learn from mistakes. Also some of the implementation is hairy here, coming from multiple attempts to improve dev experience around exception handling.

In particular, the call / callSync API may be solid but the Config API could be revisited.

Some things are going to be different in pure JS land. Since we're not limited to the Neon runtime, the Waiter can be rewritten and vastly simplified, and that will change some of the functions here, like `register_callback`.

It's nice to have this to start with, but at the same time I don't want to assume that we'll just pick up where Neon left off. That's how we got ourselves stuck with a nodejs conductor in the first place, we picked up the tool we had available and just kept hacking it into shape.

PR #1335, opened 2019-04-26
Conductor config sanitizing

dymayday:
## PR summary
This PR proposes to add sanity check to the conductor to check the configuration passed by the user.

- Check duplicated storage instances
- Check DNA hash from the conductor config file and the one loaded by the conductor

image

## testing/benchmarking notes

## changelog


conductor_api/src/conductor/base.rs
```diff
@@ -607,6 +608,34 @@ impl Conductor {
...
+                            .to_owned();
+                    // dbg!(dna_conf_hash);
+                    let dna_hash_computed = &dna.address();
+                    // dbg!(loaded_dna_hash);
+                    Conductor::check_dna_consistency(
+                        &dna_hash_from_conductor_config,
+                        &dna_hash_computed,
+                    )
+                    // .map_err(|_| {
+                    .unwrap_or({
```
willemolding: Looks like the error message from the `check_dna_consistency` function is being thrown away. Do I have that right?

conductor_api/src/config.rs
```diff
@@ -350,6 +353,38 @@ impl Configuration {
...
+            })
+            .collect();
+
+        // Here we don't use the already implemented 'detect_dupes' function because we don't need
+        // to keep track of all the duplicated values of storage instances. But instead we use the
+        // return value of 'HashSet.insert()' conbined with the short-circuiting propiety of
+        // 'iter().all()' so we don't iterate on all the possible value once we found a duplicated
+        // storage entry.
+        let mut path_set: HashSet<&str> = HashSet::new();
+        let has_uniq_values = storage_paths.iter().all(|&x| path_set.insert(x));
```
willemolding: I'm going to remember this one :)

conductor_api/src/conductor/base.rs
dymayday: That was just a WIP, the actual implementation was still undefined at that time ^^'

conductor_api/src/config.rs
dymayday: ;)

dymayday:
@lucksus 
Do we want to return an error here when there is a discrepancy between DNA hashes or just a warning?

dymayday:
Investigating why this test `holochain::tests::can_call_commit` is failing... O.o

conductor_api/src/conductor/base.rs
```diff
@@ -1230,6 +1277,187 @@ pub mod tests {
...
+        //                      loaded instance of the Conductor. Please check the logs for additional information")),
+        //     "Conductor failed to boot from config"
+        // );
+
+        let a = HashString::from("QmYRM4rh8zmSLaxyShYtv9PBDdQkXuyPieJTZ1e5GZqeeh");
+        let b = HashString::from("QmZAQkpkXhfRcSgBJX4NYyqWCyMnkvuF7X2RkPgqihGMrR");
+
+        assert_eq!(
+            Conductor::check_dna_consistency(&a, &b),
+            Err(HolochainInstanceError::InternalFailure(
```
dymayday: What do we prefer here?
- `HolochainInstanceError::InternalFailure`
or
- `HolochainError::DnaHashMismatch`

dymayday:
> Investigating why this test `holochain::tests::can_call_commit` is failing locally... O.o

I flushed and everything works fine now.

conductor_api/src/conductor/base.rs
```diff
@@ -78,6 +78,7 @@ pub fn mount_conductor_from_config(config: Configuration) {
 /// In order to not bind this code to the assumption that there is a filesystem
 /// and also enable easier testing, a DnaLoader ()which is a closure that returns a
 /// Dna object for a given path string) has to be injected on creation.
+//#[derive(Debug)]
```
lucksus: Remove this line?

dymayday:
> This all looks good, but it does not solve all three smalls in the screenshot: the check between DNA file and conductor config is missing.

Oh, I thought it was the same as checking the hash of a loaded DNA vs Conductor config. I will look into how to perform a test with a DNA file. Thanks.

dymayday:
arf, **cargo test** was passing, but **hc-rust-test** doesn't. Looking into it.

dymayday:
I founded why locally the tests where passing for a period of time.

This is because hashes for missing DNA files are computed based on time drifting algorithm, and so the tests were passing fine for a short period of time when I was updating the broken test's DNA hashes.

The workaround is to provide empty DNA hashes for missing test dna files in the tests so we can skip them during the check for DNA consistency.

CHANGELOG-UNRELEASED.md
```diff
@@ -5,12 +5,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
 - Error log output added for errors occurring during `hdk::call`, including bridge call errors [#1448](https://github.com/holochain/holochain-rust/pull/1448).
+- Discrepancy between DNA hashes are now checked and reported to the user through logs [#1335](https://github.com/holochain/holochain-rust/pull/1335).
+- Providing empty hashes for DNA in the Conductor config make the consistency check skip it.
 
 ### Changed
-- Added a Vagrant file to support nix-shell compatible VMs on windows etc. [#1433](https://github.com/holochain/holochain-rust/pull/1433)
 
-### Changed
+- Added a Vagrant file to support nix-shell compatible VMs on windows etc. [#1433](https://github.com/holochain/holochain-rust/pull/1433)
```
lucksus: I don't think this line should be added here - looks like a broken merge.. ?

conductor_api/src/conductor/base.rs
```diff
@@ -626,6 +626,59 @@ impl Conductor {
...
+                    let dna_hash_from_conductor_config =
+                        HashString::from(dna_config.hash.clone()
+                                         // .unwrap_or("Missing DNA hash from Conductor config.".to_string()))
+                                         .unwrap_or("".to_string()))
+                                         .to_owned();
+                    let dna_hash_computed = &dna.address();
+
+                    // Let's warn the user that we encounter an empty DNA hash in the Conductor
+                    // config that is mmeant to be skipped by the consistency checking mechanisim
+                    context.log("warn/Conductor: Empty DNA hash encountered in the Conductor configuration.");
```
lucksus: This warning seems to get logged unconditionally, no?

conductor_api/src/conductor/base.rs
```diff
@@ -626,6 +626,59 @@ impl Conductor {
...
+
+                    match Conductor::load_dna(&dna_file) {
```
lucksus: `Conductor::load_dna` is the default `DnaLoader` that gets set if there is no other `DnaLoader` provided with the `Conductor` constructor. But the actual `DnaLoader` to use is stored in `Conductor::dna_loader`. We should definitely use the latter to not break the case where we do not use the default (file) loader - such as tests and the upcoming WASM conductor where we don't have a filesystem.

I wonder how this can work. I would expect tests to currently not really tests the code below.

This line should be:
```suggestion
                    match Arc::get_mut(&mut self.dna_loader).unwrap()(&dna_file) {
```

conductor_api/src/conductor/base.rs
```diff
@@ -626,6 +626,59 @@ impl Conductor {
...
+                            let dna_hash_computed_from_file = HashString::from(dna.address());
+                            Conductor::check_dna_consistency_from_all_sources(
+                                &context,
+                                &dna_hash_from_conductor_config,
+                                &dna_hash_computed,
+                                &dna_hash_computed_from_file, &dna_file)?;
+                        },
+                        Err(_) => {
+                            // But if it doesn't exist, like in every unit test, we warn the user
+                            // and do some partial check
```
lucksus: No, unit tests provide a test-`DnaLoader` to emulate a virtual filesystem - see comment above.

conductor_api/src/conductor/base.rs
```diff
@@ -673,6 +726,89 @@ impl Conductor {
...
+
+                return Err(e);
+            }
+        }
+        Ok(())
+    }
+
+    /// This is where we check for DNA's hashes consistency.
+    /// Only a simple equality check between DNA hashes is currently performed.
+    /// N.B.: Empty hashes are skipped for debugging purposes.
```
lucksus: Wouldn't this result in a config that does not specify the DNA hash to effectively turn of this check? If so I don't think we should do that.

conductor_api/src/config.rs
```diff
@@ -351,6 +354,38 @@ impl Configuration {
...
+
+        if !has_uniq_values {
+            Err(String::from(
+                "Forbidden duplicated file storage value encountered.\
+                 Please fix this in the 'TOML Conductor configuration file'.",
```
lucksus: The second instance seems inappropriate as error message since the config does not necessarily come from a file. See comments above.

conductor_api/src/config.rs
```diff
@@ -351,6 +354,38 @@ impl Configuration {
 
         self
     }
+
+    /// This function checks if there is duplicated file storage from the instances section of a provided
+    /// TOML configuration file. For efficiency purposes, we short-circuit on the first encounter of a
+    /// duplicated values.
+    fn check_instances_storage(&self) -> Result<(), String> {
```
lucksus: This function seems misplaced in `ConductorAdmin`. I think it should be in `Config` and called in `Config::consistency_check`.

conductor_api/src/conductor/base.rs
dymayday: Oups, my bad! Good catch!

conductor_api/src/conductor/base.rs
dymayday: That's exactly the purpose of it, to let the tests pass on DNA hashes, that I suspect to be time based, when the DNA file is missing, precisely during tests. This is only a workaround I found to address all those broken tests, there might be a better one of course.

zippy: I understood that what this fix should do is only fail if not only ignore the consistency check when there is no specified hash in the config but also only if there is no DNA file.  I.e. it's consistent for there to be no hash specified in the case that there is actually no DNA file, and that case is only the case int testing when we are looking at other aspects of config files.  But I'm not sure the code actually does that.

conductor_api/src/config.rs
dymayday: I'm not sure about what you mean.
This is already in the `impl Configuration` block. Do you mean as a free function?
Or in a different directory?

CHANGELOG-UNRELEASED.md
dymayday: Thx, definitely a broken merge.

conductor_api/src/conductor/base.rs
dymayday: Thank you for your precious input on the matter, I made the change accordingly :+1:

conductor_api/src/config.rs
dymayday: You're probably right, thanks :+1:

conductor_api/src/conductor/base.rs
```diff
@@ -629,6 +629,64 @@ impl Conductor {
...
+                        Ok(dna) => {
+                            let dna_hash_computed_from_file = HashString::from(dna.address());
+                            Conductor::check_dna_consistency_from_all_sources(
+                                &context,
+                                &dna_hash_from_conductor_config,
+                                &dna_hash_computed,
+                                &dna_hash_computed_from_file, &dna_file)?;
+                        },
+                        Err(_) => {
+                            // But if it doesn't exist, like in every unit test, we warn the user
```
lucksus: Is this still the case?

dymayday: No it's not, so I rewrote the log message. Thanks!

conductor_api/src/conductor/base.rs
```diff
@@ -629,6 +629,62 @@ impl Conductor {
...
+                // the loaded dna.
+                {
+                    let dna_hash_from_conductor_config = HashString::from({
+                        match dna_config.hash {
+                            Some(v) => v,
+                            None => {
+                                // Let's warn the user that we encountered an empty DNA hash in the Conductor
+                                // config that is meant to be skipped by the consistency checking mechanisim
+                                context.log("warn/Conductor: Invalid DNA hash encountered in the Conductor configuration.");
+                                String::from("")
```
AshantiMutinta: Could we use let dna_hash_from_conductor_config = HashString::from({dna_config.hash.unwrap_or_default())}).to_owned() ? Could be a one liner

zippy: Wouldn't that preclude issuing the logging warning?

AshantiMutinta: I guess there is a context.log warn there.  dna_config.hash.unwrap_or_else(|_| context.log;String::default()); ? what do you think?

maackle: I have a question here, how is a empty string causing the integrity check to be skipped? I don't see empty string being treated as a special case. And if it were, I would recommend keeping it as an `Option`

dymayday: There is no skipping anymore, we found a better solution, using the appropriate DNA loader :)

Thanks @AshantiMutinta for your suggestion, I took care of it :+1:

dymayday:
Indeed @maackle you're right, there is no test on the default hash when it's missing, great catch Sir! :)
I added one [here](https://github.com/holochain/holochain-rust/pull/1335/files#diff-fd0e08594879c416d6982565b28d590aR1534).

It's already using the `HolochainError::DnaHashMismatch` [here](https://github.com/holochain/holochain-rust/pull/1335/files#diff-fd0e08594879c416d6982565b28d590aR883)
Also, @lucksus explicitly asked me to keep it in `onfig`, please see https://github.com/holochain/holochain-rust/pull/1335#discussion_r290254414

maackle:
@dymayday your test does test that the empty string does not match a real DNA, but what I meant was: is it intended that if the user does not provide the optional hash in the DNA config, that the integrity check should always fail? Currently it looks like that's what will happen, since if the user provides None as the hash, then it will result in an empty string, and as your test shows, that will cause a `HolochainError`. Which implies that the hash should not be optional, OR, that the check should be ignored if it is None. That is the kind of test I would like to see -- that the check is ignored if hash is `None`.

Also, when I talked about unifying the two integrity checks -- the line of code I pointed to is a place where there is already an integrity check happening. We shouldn't do the same check twice in two places.

dymayday:
> Is it intended that if the user does not provide the optional hash in the DNA config, that the integrity check should always fail?
Yes it's the intended behavior. I started implementing a simple warning but after discussion we core, we preferred to throw an error instead.

> Also, when I talked about unifying the two integrity checks -- the line of code I pointed to is a place where there is already an integrity check happening. We shouldn't do the same check twice in two places.
Yes I totally agree! I don't like duplicated code as well. And I think this should be part of another PR in order to fix the issue.

What do you May think @maackle? May I finally merge this one?

maackle:
> Yes it's the intended behavior

@dymayday are you sure? I think you are talking about the case where the user specifies the *wrong* hash for a DNA. To be clear, what I'm talking about is if the user passes *no* expected hash at all, i.e. `None` instead of `Some(hash)`. If so, I believe *the integrity check will always fail*. This is because the DNA config, which is an Option, winds up being converted to an empty string if it is `None`, via `unwrap_or_default`. I think this is a bug, which will cause most existing conductor configs to fail.

In summary:

I would expect that when the user specifies no DNA hash, that the integrity check would be skipped.
However, what I see in this code is that when the user specifies no DNA hash, the integrity check will run, and *fail every time*.

Expected behavior: no integrity check
Actual behavior: always failing integrity check

Do you see what I mean? Am I misreading the code? Either way, I still think there should be a test for the case when the user does not specify any hash. There is no test for the `None` case, only the `Some` case.

lucksus:
@maackle, I too think these changes provide us with what we need, and what aligns with what I regard as intended behavior.

> I would expect that when the user specifies no DNA hash, that the integrity check would be skipped.
However, what I see in this code is that when the user specifies no DNA hash, the integrity check will run, and fail every time.

I don't think we should add this special case for absent DNA hashes. I guess what is confusing right now is the `Option` and that we should make the DNA hash non-optional in the struct, since this is the effect of these changes.

I believe that by forcing users to set explicit hashes in the conductor config, we actually help them to not run into problems that are hard to debug. The main reason is this: the DNA file gets read only once when we initialize a new instance. It is then copied into the source chain and retrieved from there. If the user then changes the DNA file this won't affect the DNA copy in the source chain and the conductor will run the old version without any complaint. We had several instances of this within our team which wasted many person-hours. Explicitly setting the DNA hash means we can notify the user of a situation that is very likely not what they intend.

maackle:
@lucksus I assumed that since the `hash` param is using `Option`, it was intended to be optional, in which case there is a bug. But, you're saying it's intended to not be optional, so my bug report no longer applies, but I still think we should make changes, i.e. change `hash` from `Option<HashString>` to `HashString`. If there's no reason why a user should omit the hash, why let them?

If we leave it as `Option`, then if a user omits the hash, they'll get a totally unhelpful error message every time, saying something like `"Qmblahblahblah" is not equal to ""` and they won't know where that's coming from. Even if they read the code it would be hard to find. Much better to say "`hash` is required", no?

lucksus:
Yes, fully agree, @maackle.

dymayday:
Thank you both for your help and precious comments! I gonna make the changes you pointed out :+1:

nmrshll:
With this new change (`hash` no more optional, and checked by the `holochain` command), what workflow do you suggest when developing with the `holochain` conductor ?

It seems that we're going to have to change the hash in the config file for every change to the zome code, is there a way around that ? 

(of course it can be automated, but that's a workaround that doesn't feel right)

PR #1337, opened 2019-04-26
Get links crud

AshantiMutinta:
## PR summary
-This PR allows for obtaining crud information with the get_links. The crud information is implied from the link type and we can query this information based on LinkStatusRequestKind
- a tuple of a link and crud_status was chosen for being able to reuse the crud_status as well as saving the space by also not serializing the name of the fields.

## testing/benchmarking notes
Testing of this was implemented by modification and addition of app_spec and united tests.
## followups
-One of the tests in the intergration suite uses a string which makes it hard to debug, we should switch to a struct.
-We should offset the querying of the link to the node end rather than the client. This would mean modifying some of the structs on the network side

- [ x] this is a code change that effects some consumer (e.g. zome developers) of holochain core so there is an 'Unreleased' CHANGELOG item, with the format `- summary of change [PR#1234](https://github.com/holochain/holochain-rust/pull/1234)`


core/src/workflows/get_link_result.rs
```diff
@@ -33,42 +26,85 @@ pub async fn get_link_result_workflow<'a>(
...
+                        address: link.clone(),
+                        headers,
+                        crud_status: link_entry.entry_with_meta.crud_status,
+                        crud_link: link_entry.entry_with_meta.maybe_link_update_delete,
+                    })
+                })
+            })
+            .unwrap_or(None)
+            .unwrap_or(Err(HolochainError::ErrorGeneric(
+                "Could not crud information for link".to_string(),
```
lucksus: This sentence a verb ;)

core/src/workflows/get_link_result.rs
```diff
@@ -33,42 +26,85 @@ pub async fn get_link_result_workflow<'a>(
...
+                .collect(),
         ))
     } else {
         Err(HolochainError::ErrorGeneric(
             "Could not get links".to_string(),
         ))
     }
 }
+
+pub fn get_latest_entry(
```
lucksus: What exactly is the purpose of this function and why can't we reuse `get_entry_result_workflow`: https://github.com/holochain/holochain-rust/blob/cbd91ef03bf0d06dcc42bf8d0e794ebc2f8ec723/core/src/workflows/get_entry_result.rs#L55 ?

core/src/workflows/get_link_result.rs
```diff
@@ -33,42 +26,85 @@ pub async fn get_link_result_workflow<'a>(
...
+                    };
+                    Ok(LinksResult {
+                        address: link.clone(),
+                        headers,
+                        crud_status: link_entry.entry_with_meta.crud_status,
+                        crud_link: link_entry.entry_with_meta.maybe_link_update_delete,
+                    })
+                })
+            })
+            .unwrap_or(None)
```
lucksus: I don't think the above is logically correct. `get_links` above in line 19 returns the target addresses of all matching links. This means that `get_latest_entry` in line 29 gets the target of the link, which in turn means that the `headers` that are **returned are the headers of the link target** and not the headers of the link (i.e. the LinkAdd entries). Or am I confused?

AshantiMutinta:
I implemented this because we do  it have a status request for entries on the getlinks. I forgot to mention it but linkstatus request seems to he on the link level as a whole and entry seems to be referring to the link targets. In general it is just a recursive and non mutable way of doing whatever get_entry_result does and would prefer if we used this one instead but we would need some refactoring

core/src/workflows/get_link_result.rs
AshantiMutinta: I implemented this because we do it have a status request for entries on the getlinks. I forgot to mention it but linkstatus request seems to he on the link level as a whole and entry seems to be referring to the link targets. In general it is just a recursive and non mutable way of doing whatever get_entry_result does and would prefer if we used this one instead but we would need some refactoring

core/src/workflows/get_link_result.rs
AshantiMutinta: You are absolutely right, GetLinks is returning link targets because it returns this list of link targets this includes CrudStatus and CrudLink. It is essentially doing the same thing as it was doing in the last pull request which gets headers(as per app_spec results). I think there was some miscommunication in which headers we were talking about(I think I even asked this in heatbeat)

-It brings about questions. When we mention crud_stauts of link what are we talking about? target or linkAdd itself.
-CrudLink? Target or itself?
I ask these because for a system type such as linkadd these will mostly never change so there is a question of why should they be returned in the first place. I think there might have been a breakdown of requirements

AshantiMutinta:
Failure after inclusion of clustered tests, need to fix that

core/src/action.rs
```diff
@@ -150,7 +150,7 @@ pub enum Action {
     /// Last string is the stringified process unique id of this `hdk::get_links` call.
     GetLinks(GetLinksKey),
     GetLinksTimeout(GetLinksKey),
-    RespondGetLinks((FetchMetaData, Vec<Address>)),
+    RespondGetLinks((FetchMetaData, Vec<(Address, CrudStatus)>)),
```
struktured: @AshantiMutinta Is a tuple future proof enough? Do you imagine other stuff getting associated with address here?

core/src/network/reducers/respond_get_links.rs
```diff
@@ -3,15 +3,15 @@ use crate::{
...
+use holochain_core_types::{cas::content::Address, crud_status::CrudStatus, error::HolochainError};
 use holochain_net::connection::json_protocol::{FetchMetaData, FetchMetaResultData, JsonProtocol};
 
 /// Send back to network a HandleFetchMetaResult, no matter what.
 /// Will return an empty content field if it actually doesn't have the data.
 fn reduce_respond_get_links_inner(
     network_state: &mut NetworkState,
     get_dht_meta_data: &FetchMetaData,
-    links: &Vec<Address>,
+    links: &Vec<(Address, CrudStatus)>,
```
struktured: Similar to other comment, is a tuple good enough here for immediate future?

core/src/network/mod.rs
```diff
@@ -298,8 +298,10 @@ pub mod tests {
         assert_eq!(links.len(), 2, "links = {:?}", links);
         // can be in any order
         assert!(
-            (links[0] == entry_addresses[1] || links[0] == entry_addresses[2])
-                && (links[1] == entry_addresses[1] || links[1] == entry_addresses[2])
+            (links[0] == (entry_addresses[1].clone(), CrudStatus::Live)
+                || links[0] == (entry_addresses[2].clone(), CrudStatus::Live))
+                && (links[1] == (entry_addresses[1].clone(), CrudStatus::Live)
+                    || links[1] == (entry_addresses[2].clone(), CrudStatus::Live))
```
struktured: This test is getting heavy. If it fails it would be probably be a pain to debug. I think it's outside the scope of your PR to fix it however.

AshantiMutinta: Yes it would be nice to break out in follow up pr

core/src/action.rs
AshantiMutinta: I think for now it works because serialization will save space because no field name values. When it grows, should definitely use a struct

AshantiMutinta: I think for now it works because serialization will save space because no field name values. When it grows, should definitely use a struct

PR #1339, opened 2019-04-26
WIP: circle ci scaffold test

Connoropolous:
## testing/benchmarking notes

## followups

## changelog


Connoropolous:
@thedavidmeister any advice on how to solve the `hc command not found` issue?

thedavidmeister:
@Connoropolous yeah, `~/.cargo/bin` is not _necessarily_ on the user's path (in this case it is not)

this could be fixed by vendoring bins into the repo, or otherwise changing the install scripts

i think the hack/workaround here is to re-export the path to include the built binaries, i'll put a fix up

thedavidmeister:
@Connoropolous ok i got your greens

saw this:

```
With Rust already installed switch to it by running the following commands:
$ rustup toolchain install nightly-2019-01-24
$ rustup default nightly-2019-01-24
```

that's gonna need to stop recommending `rustup` in a followup (e.g. for the reason you discovered)

thedavidmeister:
@Connoropolous for reference - https://github.com/NixOS/nix/issues/971

but we should be using the official holochain docker that has nix user issues resolved :thinking:

thedavidmeister:
imma poke around in ssh

thedavidmeister:
@Connoropolous kk i got it running green

i also added a mac option too but it's too slow to run on every PR, if you're into it we can shift it to develop only

there is this output also, but it doesn't fail builds

```
Failed to load instance alice from storage: ErrorGeneric("State could not be loaded due to NoneError")
```

Connoropolous:
Good stuff @thedavidmeister !

There's one more thing before this PR is ready to merge though actually, we need to use `npm link` to connect the built version of the holochain-nodejs to the initialized app

Connoropolous:
Oh, and @thedavidmeister this is intentional... you're not the only one to mistake this for an "error?"

> Failed to load instance alice from storage: ErrorGeneric("State could not be loaded due to NoneError")

I tried many times to improve this messaging, but it still looks like some kind of error. It just means its initializing a chain for the first time, not rebooting it

Connoropolous:
I don't know exactly how to implement this, but here's the criteria:

`npm link` needs to be run within `nodejs_conductor` folder

`npm link @holochain/holochain-nodejs` needs to be run within the `test` folder of an `hc init`ed app

thedavidmeister:
@Connoropolous right, so it must be trying to do something resulty with the init state, which is wrong probably but nothing to do with this

i'll look at the link

Connoropolous:
@maackle this PR would be a good place for us to update the scaffolding code to Diorama based, since this would also introduce ongoing CI testing for it

Connoropolous:
and @thedavidmeister we can also be coming back to this, and working with it as part of the moving of nodejs_conductor to a different repo

Connoropolous:
@thedavidmeister I am blocked on knowing how I would update this... my blocker is:

I don't know to edit holonix to add the command that we'd added here, and do some kind of local development on it?

PR #1340, opened 2019-04-26
Adds missing breaking change in CHANGELOG

maackle:
## PR summary

I ran into this problem with a UI that broke when moving to 0.0.12-alpha1, this should be mentioned int he changelog

## changelog

- [X] this is a change to the changelog :)


Connoropolous:
Arg. Sorry @maackle we've had some messups with the changelog because Git merges have been messing with it, when we merge develop branch back into WIP branches... 
@thedavidmeister has a solution in the works. Shouldn't happen again that anything gets missed in the CHANGELOG after that

PR #1358, opened 2019-04-29
holochain_net: multidna test framework refactor

ddd-mtl:
## PR summary
Renamed P2pNode to TestNode. Gave it its own folder and broke things down in different files.
(Note: might use the P2pNode name for a future trait).
Have TestNode handle multiple DNAs:
 - Holds a HashMap of DnaStores
 - Has a 'current_dna' to keep things as simple as before, since 90% of test are mono-dna.

## changelog


test_bin/src/basic_workflows.rs
```diff
@@ -386,7 +389,13 @@ pub fn dht_test(alex: &mut P2pNode, billy: &mut P2pNode, can_connect: bool) -> N
 }
 
 /// Sending a Message before doing a 'TrackDna' should fail
-pub fn no_setup_test(alex: &mut P2pNode, billy: &mut P2pNode, _connect: bool) -> NetResult<()> {
+pub fn no_setup_test(alex: &mut TestNode, billy: &mut TestNode, _connect: bool) -> NetResult<()> {
+    alex.track_dna(&DNA_ADDRESS_A, true)
+        .expect("Failed sending TrackDna message on alex");
+    alex.untrack_current_dna()
+        .expect("Failed sending UntrackDna message on alex");
+    alex.set_current_dna(&DNA_ADDRESS_A);
```
neonphog: are these backwards? should we set the current first, before trying to untrack the current?

neonphog: mmm, I'm guessing from below that this is a hack to make sure the current isn't `None`. Might deserve a comment here.

ddd-mtl: `untrack_current_dna()` will also 'unset' the `current_dna` so we need to re-set it, otherwise an assert will pop when calling `send_message()`.
I concede this edge case test is a bit messed up.

ddd-mtl:
@neonphog Yes, no multidna tests for now. This PR was already quite big, so tests will be in follow-up PRs.

PR #1359, opened 2019-04-30
adds new rpc function test/agent/add

willemolding:
## PR summary

Adds a new RPC method to the admin interface `test/agent/add`.

This is used for adding a test agent that uses the test_keystore much like the agents in the node testing interface. This is much much faster than attempting to add agents with real keystores. This is a better solution that #1333.

## changelog

- [ x] this is a code change that effects some consumer (e.g. zome developers) of holochain core so there is an 'Unreleased' CHANGELOG item, with the format `- summary of change [PR#1234](https://github.com/holochain/holochain-rust/pull/1234)`


conductor_api/src/interface.rs
```diff
@@ -473,14 +473,14 @@ impl ConductorApiBuilder {
...
                 let properties = params_map.get("properties");
-                conductor_call!(|c| c.install_dna_from_file(
+                let dna_hash = conductor_call!(|c| c.install_dna_from_file(
                     PathBuf::from(path),
                     id.to_string(),
                     copy,
                     expected_hash,
                     properties
                 ))?;
-                Ok(json!({"success": true}))
```
maackle: Envoy is actually checking for `success` to see if the request happened successfully. Maybe it's enough to check for `Ok`, i.e. not `Err`. Maybe leave `success` in until we standardize this? Thoughts from others?

willemolding: Yeah lets leave it in why not

conductor_api/src/conductor/test_admin.rs
```diff
@@ -0,0 +1,57 @@
...
+            id: id.clone(),
+            name: name.clone(),
+            public_address: public_address.clone(),
+            keystore_file: name.clone(),
+            holo_remote_key: None,
+        };
+        new_config.agents.push(new_agent);
+        new_config.check_consistency()?;
+        self.config = new_config;
+        // self.save_config()?; we don't actually want to save it for tests
```
Connoropolous: we could just pull this line out?

zippy: Well it might be good to keep in because it reminds us that we Don't actually want to save the config, unlike in other code chunks..

PR #1363, opened 2019-04-30
EAV Tombstone for links

AshantiMutinta:
## PR summary
-[x] Added tombstone for removal of links
-[x] LinkData contains Top Chain Header and AgentID
-[x] Corrected Get Headers

## testing/benchmarking notes
modification of unit tests and app_spec tests. Regression tests have been reflected for the different modes of the UTC Dispatcher

## followups

## changelog


core/src/nucleus/ribosome/api/remove_link.rs
```diff
@@ -33,8 +38,51 @@ pub fn invoke_remove_link(runtime: &mut Runtime, args: &RuntimeArgs) -> ZomeApiR
...
+                        .map(|entry| match entry {
+                            Entry::LinkAdd(link_data) => link_data.link().target() == link.target(),
+                            _ => false,
+                        })
+                        .unwrap_or(false),
+                    _ => false,
+                })
+                .unwrap_or(false)
+        })
+        .collect::<Vec<_>>();
```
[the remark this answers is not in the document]
lucksus: That change happened in this PR by @willemolding: https://github.com/holochain/holochain-rust/pull/1412/files

core/src/workflows/get_link_result.rs
```diff
@@ -56,36 +46,79 @@ pub async fn get_link_result_workflow<'a>(
...
             link_results.into_iter().map(|s| s.unwrap()).collect(),
         ))
+    } else {
+        Err(HolochainError::ErrorGeneric(
+            "Could not get links".to_string(),
+        ))
+    }
+}
+
+async fn get_link_caches<'a>(
```
lucksus: I was confused by this name and trying to understand in what way it caches anything. I do understand this function to be necessary exactly because of the performance problem I mentioned above: `get_links` returns addresses of `Entry::LinkAdd` entries, so this function gets those link entries. Should we rename it to something like:
```suggestion
async fn get_link_add_entries<'a>(
```
?

core_types/src/cas/storage.rs
```diff
@@ -628,6 +638,71 @@ impl EavTestSuite {
...
+                .fetch_eavi(&EaviQuery::new(
+                    Some(one.address()).into(),
+                    None.into(),
+                    Some(two.address()).into(),
+                    IndexFilter::LatestByAttribute,
+                    Some(expected_last_attribute.into())
+                ))
+                .unwrap()
+        );
+    }
```
lucksus: I don't understand this test - that is, I don't understand how this tests a tombstone set implementation.

You are only adding links with `Attribute::LinkTag`. I don't see `Attribute::RemoveLink` used at all. Isn't that what should add a link to the tombstone set?

Can you maybe add some comments to help understand what's going on?

AshantiMutinta: I will change this but the RemoveLink aspect of this algorithm is irrelevant because it is able to fold on the desired Attribute + Content. There is a can_remove_links that illustrates this but will change this to RemoveLink.

The fold() function in line 55 is essentially the heart of how the tombstone works.

comments are good and will add them!

core/src/dht/dht_reducers.rs
```diff
@@ -82,9 +84,14 @@ pub(crate) fn reduce_add_link(
...
     action_wrapper: &ActionWrapper,
 ) -> Option<DhtStore> {
-    let link = unwrap_to!(action_wrapper.action() => Action::AddLink);
+    let (link, entry) = unwrap_to!(action_wrapper.action() => Action::AddLink);
     let mut new_store = (*old_store).clone();
-    let res = reduce_add_remove_link_inner(&mut new_store, link, LinkModification::Add);
+    let res = reduce_add_link_inner(
+        &mut new_store,
+        link,
+        &entry.address(),
```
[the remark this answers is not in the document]
AshantiMutinta: The reason for this is being able to differentiate between different types of links. That is links with different authors and different timestamps.  IF we only add the address of the link we would get duplicates that tombstone algorithm wont allow another add

AshantiMutinta:
Changed Get_Links it is now piggy backing off get_entry_result and the optimizations that the workflow has.

app_spec/test/regressions.js
```diff
@@ -118,11 +118,11 @@ scenario.runTape('create & publish post -> recommend to other agent', async (t,
...
   })
   console.log("linked: ", linked)
-  t.equal(linked.Ok, "QmUFkn3kTXuFmBvynEMQ4ox3WyvkkG4GtfvPkNqb5b7Ge7")
+  t.ok(linked.Ok);
 
   const recommendedPosts = tash.call('blog', 'my_recommended_posts', {})
   console.log("recommendedPosts", recommendedPosts)
   console.log('agent addresses: ', alice.agentId, tash.agentId)
-
+  t.equal(recommendedPosts.Ok.links[0].address,"QmaERphTvXExCBcVdCRzGBE2d1dnRNWDjthwfyBuq9Mf6a");
```
maackle: Can we keep these hashes out of the test? Aren't these going to change every time we update app spec? Is there a less brittle way we can do the same check?

conductor_api/src/config.rs
```diff
@@ -382,6 +385,24 @@ pub struct DnaConfiguration {
     pub hash: Option<String>,
 }
 
+#[derive(Deserialize, Serialize, Clone, Debug, PartialEq)]
+pub enum ISOActive {
+    On,
+    Off,
+}
```
maackle: Hmm why not just a boolean for this? Do you anticipate other variants?

core/src/action.rs
```diff
@@ -107,13 +107,13 @@ pub enum Action {
 
     /// Adds a link to the local DHT shard's meta/EAV storage
     /// Does not validate, assumes link is valid.
-    AddLink(Link),
+    AddLink(LinkData),
```
maackle: I have actually been curious about our use of `LinkData` for add/remove actions. `LinkData` is just a `Link` plus a `LinkActionKind` which can be either `ADD` or `REMOVE`, isn't that redundant in the context of an action that's meant for adding links?

core_types/src/eav/query.rs
```diff
@@ -20,6 +21,7 @@ impl<'a> Default for EaviQuery<'a> {
             Default::default(),
             Default::default(),
             IndexFilter::LatestByAttribute,
+            None,
```
maackle: I was hoping EaviQuery would stay general, and focused on only E, A, V, and I. If we start tacking on extra special-case filters, I'm afraid it will just grow and grow. Is there any reason why the tombstone filter can't be implemented as part of a simple AttributeFilter, i.e. use the `attribute` param?

AshantiMutinta: This was a space decision. Because RemoveLink has to add a RemoveLink for every other link that has this same LinkData, it will mean a tombstone will have to be set for all the link data as well. This allows a runtime tombstone and saves a lot more space, I think we might need to develop a small DSL for the EAVI imo

core/src/action.rs
AshantiMutinta: LinkData contains AgentID and ISOTime right about now.

conductor_api/src/config.rs
AshantiMutinta: Anticipating a Fixed Time variant which the can specify which UTC to use

app_spec/test/regressions.js
AshantiMutinta: hmmm maybe find a way to generate them and then check against them?

core_types/src/cas/storage.rs
lucksus: Ok, let me check if I get this right:
1. You have two entries (addressable contents used as E and V ;))
2. You create several links (EAVIs) between them but with differing link type and link tag.
3. For one of those type/tag combinations you also create a RemoveLink EAVI and another new LinkAdd EAVI.
4. Then you query the EAVI storage for anything between those two entries that has the RemoveLink attribute and you assert that you get back exactly that one EAVI item.

Is that what this code is doing? I still don't understand how that tests tombstone semantic..

app_spec/test/regressions.js
maackle: Is this really a useful test though? Would we really expect a linked address to be different from what we would expect? Wouldn't that mean holochain is broken? Does that check belong in an end-to-end test?

core/src/action.rs
maackle: Hmm not sure if we're talking about the same `LinkData`. I'm referring to this, which is just a `Link` with an enum attached. https://github.com/holochain/holochain-rust/blob/develop/core_types/src/link/link_data.rs#L13

app_spec/test/regressions.js
AshantiMutinta: This was an existing test just with a changed hash, we had a existing conversation with nico and I am probably going to remove some of the hashes because with the chain header it is not going to be deterministic anymore. 

Pros of links imo is the determinism because of the data where are putting in, we know what we are getting out, but if we already have coverage at the lower levels we can just check the number of links it gets back.

I don't really know why the regression.js is for to be honest because same setup + looks just like what we are doing on the app_spec side. Would be better if regression.js had the rust conductor

core/src/action.rs
AshantiMutinta: meant in this pull request ;) The LinkData will be used against the Entry::LinkAdd for generating a hash :)

You are right though, previous uses deem in redundant

app_spec/test/test.js
```diff
@@ -344,26 +344,64 @@ scenario('my_memos_are_private', async (s, t, { alice, bob }) => {
 
 scenario('delete_post', async (s, t, { alice, bob }) => {
 
-  //create post
-  const alice_create_post_result = await alice.callSync("blog", "create_post",
-    { "content": "Posty", "in_reply_to": "" }
+  //creates a simple link with alice as author with chain header
```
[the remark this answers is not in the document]
AshantiMutinta: supposed to be 'author with new chain header' will change that

app_spec/zomes/simple/code/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+}
+
+fn simple_entry(content: String) -> Entry {
+    Entry::App("simple".into(), Simple::new(content).into())
+}
+
+
+pub fn handle_create_my_link(base: Address,content : String) -> ZomeApiResult<()> {
+    let address = hdk::commit_entry(&simple_entry(content))?;
```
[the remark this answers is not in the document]
AshantiMutinta: We need to commit an address before we can link it to the link_entry, I wouldn't want to seperate the two because it would cause a lot of noise for the app_spec part since we care mostly about the links rather than the entry

app_spec/test/test.js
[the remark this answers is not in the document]
AshantiMutinta: can_remove_links has a test that does the tombstone set, maybe just needs a few comments but it is situated on the EAVFilter

core_types/src/eav/query.rs
```diff
@@ -96,8 +151,12 @@ impl<'a> EaviQuery<'a> {
     pub fn index(&self) -> &IndexFilter {
         &self.index
     }
+    pub fn tombstone(&self) -> &Option<AttributeFilter<'a>> {
+        &self.tombstone
+    }
 }
 
+/// Represents a fitler type which takes in a function to match on
```
maackle:
```suggestion
/// Represents a filter type which takes in a function to match on
```

core/src/action.rs
maackle: I still am not sure we're talking about the same thing. I am just wondering why we use `LinkData` instead of just `Link`. But my question isn't really necessary to answer for this PR, so I'll ask it elsewhere.

maackle: OK I just got it. We are talking about the same thing. I didn't see that you added two fields to `LinkData`. But I am specifically wondering if we need `action_kind`. And that is a question for another day.

core/src/dht/dht_reducers.rs
```diff
@@ -230,57 +259,168 @@ pub mod tests {
...
         );
+
+        //add new link with same chain header
+        let action_link_add = ActionWrapper::new(Action::AddLink(link_data));
+        let new_dht_store = reduce(store.dht(), &action_link_add);
+
+        //fetch from dht after link with same chain header is added
+        let storage = new_dht_store.meta_storage();
+        let fetched = storage.read().unwrap().fetch_eavi(&EaviQuery::new(
+            Some(entry.address()).into(),
```
maackle: Please DRY up this big `EaviQuery` that gets constructed into a function, it gets used in several places and I'm afraid the test will go out of date if we change the implementation in `dht_store::get_links()`. OR, could you just use `get_links` in the test instead of constructing the `EaviQuery` manually here?

core/src/dht/dht_reducers.rs
```diff
@@ -230,57 +259,168 @@ pub mod tests {
...
+            test_chain_header(),
+            test_agent_id_with_name("new_agent"),
+        );
+        let entry_link_add = Entry::LinkAdd(link_data.clone());
+        let action_link_add = ActionWrapper::new(Action::AddLink(link_data));
+        let _new_dht_store = reduce(store.dht(), &action_link_add);
+
+        //after new link has been added return from fetch and make sure tombstone and new link is added
+        let storage = new_dht_store.meta_storage();
+        let fetched = storage.read().unwrap().fetch_eavi(&EaviQuery::new(
```
maackle: another use of a big manually constructed EaviQuery

core/src/dht/dht_store.rs
```diff
@@ -56,32 +56,38 @@ impl DhtStore {
...
     ) -> Result<BTreeSet<EntityAttributeValueIndex>, HolochainError> {
         // interpret link tags and types as regex
-        let link_type = Regex::new(&link_type)
+        let link_type_regex = Regex::new(&link_type)
             .map_err(|_| HolochainError::from("Invalid regex passed for type"))?;
-        let tag =
+        let tag_regex =
             Regex::new(&tag).map_err(|_| HolochainError::from("Invalid regex passed for tag"))?;
 
         let filtered = self.meta_storage.read()?.fetch_eavi(&EaviQuery::new(
```
maackle: Big EaviQuery that looks similar to those constructed manually in tests

PR #1369, opened 2019-05-01
bump n3h version to v0.0.12-alpha

neonphog:
@Connoropolous @ddd-mtl 

## testing/benchmarking notes

## followups

## changelog


CHANGELOG-UNRELEASED.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Changes `LinkAdd` and `RemoveEntry` so that they return a hash instead of a null [#1343](https://github.com/holochain/holochain-rust/pull/1343)
+- Updated linked [n3h](https://github.com/holochain/n3h) version to v0.0.12-alpha
```
Connoropolous: Here we should expand this with minor details re. 
- persistent transport Ids
- MDNS

lucksus: +1

neonphog:
@Connoropolous @lucksus - I don't know if I missed an onboarding announcement regarding the changelog or if there wasn't such a thing, but I don't really understand the process.

Moreover, I don't think trying to cram all the n3h changes into the minor holochain-rust PR that just updates the pinning is appropriate. In reality, this should probably be an automated CI thing that is triggered everytime an n3h release is published.

As a totally separate discussion, perhaps a changelog should be tracked in the n3h repo. That would make more sense to me than doing this here.

Connoropolous:
@neonphog an n3h changelog would be fantastic...
then we could link to it from here when we do our version bumps or something, but either way, consumers of `holochain` certainly need to know what's changing in the networking stack, as that's the area users are struggling the most at this point

Connoropolous:
@neonphog the PR template is supposed to explain the changelog process... what doesn't make sense about it?

PR #1378, opened 2019-05-02
Trace signals conditionally exposed

maackle:
## PR summary

* Rename `Signal::Internal` to `Signal::Trace`
* Add `expose_trace_signals` bool to `Configuration`, if false then Trace signals will never be emitted over any interface
* Set `expose_trace_signals` to true for nodejs conductor config

**The first commit is noisy, it's just the sweeping rename to `Trace`. Look at the second commit for the actual filtering logic**

## changelog


zippy:
Hey @maackle what's the scoop with this PR, should we be moving it forward?

maackle:
I redid this one as #1428 since my first attempt at find-and-replace was overzealous.

PR #1384, opened 2019-05-07
Make keystore_file optional, for test agents

maackle:
## PR summary

Add-on to #1359, which attempts to load a bogus keystore_file on instance creation. This makes `keystore_file` optional, with `None` representing a test agent using the `test_keystore`.

## changelog


willemolding:
Hmm yeah I starting doing it this way then figured out I could fix it by just registering the test keystore properly in #1385. I don't know which way is better though..!

maackle:
#1385 is the way

PR #1389, opened 2019-05-09
Refactor dht reducers

willemolding:
## PR summary

Dries up the DHT reducer functions be refactoring them to have `inner_reducer` functions which perform the required mutations on the state object. Some of these are reused in multiple reducers which is nice. inner_reducers are also moved to their own module `dht_inner_reducers.rs`.

Other general tidying of the `dht_reducers.rs` file as well including removing all of the unwrap calls.

## testing/benchmarking notes

Since this was a refactor no additional tests were added

## changelog


core/src/dht/dht_inner_reducers.rs
```diff
@@ -0,0 +1,135 @@
...
+            store.meta_storage().write()?.add_eavi(&eav)?;
+            Ok(link.base().clone())
+        }
+        false => Err(HolochainError::ErrorGeneric(String::from(
+            "Base for link not found",
+        ))),
+    }
+}
+
+pub(crate) fn reduce_update_entry_inner(
```
willemolding: I am a little bit unsure about this one. I am worried that it is possible to get into an invalid state if the crud-status EAV entry is added but the link fails. Is this correct and similar to how it worked before?

core/src/dht/dht_inner_reducers.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    (*meta_storage.write()?).add_eavi(&new_status_eav)?;
+
+    // Update crud-link
+    let crud_link_eav = create_crud_link_eav(latest_deleted_address, deletion_address)
+        .map_err(|_| HolochainError::ErrorGeneric(String::from("Could not create eav")))?;
+    (*meta_storage.write()?).add_eavi(&crud_link_eav)?;
+
+    Ok(latest_deleted_address.clone())
+}
```
AshantiMutinta: Awesome, I think because it is now an inner function we could utilize the '?' and get rid of some to those early 'return error' blocks.

core/src/dht/dht_inner_reducers.rs
```diff
@@ -0,0 +1,135 @@
...
+    }
+}
+
+pub(crate) fn reduce_add_remove_link_inner(
+    store: &mut DhtStore,
+    link: &Link,
+    link_modification: LinkModification,
+) -> HcResult<Address> {
+    match (*store.content_storage().read()?).contains(link.base())? {
+        true => {
```
AshantiMutinta: I think it might be clearer to use an if / else statement here? What do you think about that suggestion?

willemolding: Great suggestion. I made the change and it is more readable now

core/src/dht/dht_inner_reducers.rs
willemolding: I was able to get rid of some the early returns using `.map_err` but I left the ones where they return early based on a boolean condition (e.g. 
```
if entry.entry_type().to_owned().is_sys() {
        return Err(HolochainError::ErrorGeneric(
            "trying to remove a system entry type".into()
        ));
 }
``` 

I couldn't figure out a more readable way to write this. Any ideas?

AshantiMutinta: imo I think this is fine to leave in! It was nice to clean out some of those other returns tho

This looks great

core/src/dht/dht_reducers.rs
```diff
@@ -53,279 +50,90 @@ fn resolve_reducer(action_wrapper: &ActionWrapper) -> Option<DhtReducer> {
...
-        Action::Commit((entry, _, _)) => reduce_store_entry_common(context, old_store, &entry),
-        Action::Hold(EntryWithHeader { entry, header }) => {
-            reduce_store_entry_common(context.clone(), old_store, &entry).and_then(|state| {
-                state.add_header_for_entry(&entry, &header).ok()?;
-                Some(state)
-            })
+    let (entry, _, _) = unwrap_to!(action_wrapper.action() => Action::Commit);
+    let mut new_store = (*old_store).clone();
+    match reduce_store_entry_inner(&mut new_store, entry) {
+        Ok(()) => Some(new_store),
```
lucksus: Hm, I don't think it makes sense to add an entry to the DHT store without its header. Either we don't store entries here on commit or we store them the same way as we would when holding. So I would suggest we stick to reducing `Action::Commit` with `reduce_hold_entry`.

What was the reasoning behind separating the handling of commit actions like this?

willemolding: This was already the case I just refactored it to make it more explicit. While there was a single reducer for both `Action::Commit` and `Action::Hold` the reducer matched on the action type and only added the header in the hold case. [See here](https://github.com/holochain/holochain-rust/blob/c065bbdbc4af29ac7f837efb1531a6011b60f86d/core/src/dht/dht_reducers.rs#L61).

If we do want to change how it works it might be worth making a separate PR to discuss it

lucksus: Oh, wow. Ok...

PR #1398, opened 2019-05-12
WIP: Link type tagging

jdeepee:
## PR summary

This pull requests attempts to implement type tagging on links for the purpose of validation on tag based indexing.

Currently I am building a social network (@juntofoundation) on Holochain. One of the things we wish to be able to support in our application is "advanced querying". The ability to query an entry based on more than one parameter - for example being to filter all entries by a given timestamp and category. My current research leads me to believe there is two ways to do this on holochain (please correct me if I am wrong). 

1/ Get links on each query item and then compare results to find matches in the link results. 
2/ Saving all possible permutations of queries as links from query anchors to entry(s) we wish to be able to query for. 

The first implementation of querying would work fine when you only have a few links on each query parameter you care for, but in a public production setting where you may wish to be able to query against a large number of entries this could mean filtering and matching over millions of items. The second implementation introduces added link complexity due to saving all permutation of query values in the tag(s), but the query you wish to make could happen instantly - no sorting would need to happen.

Currently writing validation rules for the kind of behaviour outlined in implementation 2 is not possible. There is no way to associate validation to the given links as the tag is mostly unknown. If we could specify a type for links this would allow validation to run on groups of links rather than validation running on one specific link that has a matching tag. 

## Current status

I have added an example in app_spec detailing when this ability may be utilised. I have also modified all function/data types I could find which interact with link commit's and validation. However upon running my test file in app_spec: `query.js` I receive the following error when `hdk::link_entries` is called: 
```
Zome function failure: Argument deserialization failed, when calling: CallbackFnCall { id: ProcessUniqueId { prefix: 12, offset: 0 }, zome_name: "blog", fn_name: "__hdk_get_validation_package_for_link", parameters: JsonString("{\"entry_type\":\"post\",\"tag\":\"authored_posts\",\"direction\":\"From\"}") }
```

My understanding of how all parts of `holochain-rust` function is limited so help implementing this feature as well as an audit of the work/proposal as it is would be wonderful.

## changelog

### DIFF
--- a/app_spec/test/query.js
+++ b/app_spec/test/query.js
@@ -1,0 +1,36 @@
+const path = require('path')
+const { Config, Conductor, Scenario } = require('../../nodejs_conductor')
+Scenario.setTape(require('tape'))
+
+const dnaPath = path.join(__dirname, "../dist/app_spec.dna.json")
+const dna = Config.dna(dnaPath, 'app-spec')
+const agentAlice = Config.agent("alice")
+
+const instanceAlice = Config.instance(agentAlice, dna)
+
+const scenario = new Scenario([instanceAlice], { debugLog: true })
+
+scenario.runTape('query posts based using time index', async (t, {alice}) => {
+    let date = new Date(); //Get current timestamp - post should be linked to the same timestamps
+    date.toISOString();
+    let year = date.getFullYear().toString();
+    let month = date.getMonth().toString();
+    let day = date.getDate().toString();
+    let hour = date.getHours().toString();
+    
+    //create post which will create index links on following timestamp anchors: year, month, day, hour
+    const create1 = await alice.callSync("blog","create_post", {content: 'hi', in_reply_to: null})
+    t.ok(create1)
+
+    //Get the hash of the month timestamp we want to start our query from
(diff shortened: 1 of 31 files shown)


willemolding:
To make sure we are on the same page here are the current tasks around links in our SoA tree

link-tags-soa-tree

So what is currently known as **tag** becomes **type**. You must define all the link types at compile time (much like entry_types) along with their unique validation callback functions. This is no different to how it currently works so only some renaming is required.

The new thing called **tag** will be a string that can be added to a link at creation time. This will be available in the validation callback so you can ensure they follow certain rules (e.g. < 100 chars) but they do not need to be predefined.

HDK calls will allow you to retrieve links either by type or tag (or both e.g. type-with-tag). The filtering operation will be performed by the node you are retrieving the links from.

This also enables the possibility of custom tag filtering callbacks encoded into the DNA. For example if you tag each link with a timestamp you could write a filter which filters within a daterange. This would be part of the DNA so every agent has a copy and can execute these on request for a given base. This last one isn't currently on the task tree but is certainly something that will be possible if there is demand for it.

PR #1400, opened 2019-05-13
Rename link tag to type

willemolding:
## PR summary

Renames the `tag` aspect of links to `link_type` in all parts of the code including the HDK and docs.

This is to enable the next iteration which is re-implementing the `tag` as a true tag (e.g. something you can define dynamically at run time) like it was in Holochain proto.

This will be a minor breaking change for zome developers so it may be worth waiting until the entire new set of link features are added before including this in a release.

## changelog

### DIFF
--- a/core_types/src/eav/eavi.rs
+++ b/core_types/src/eav/eavi.rs
@@ -72,8 +72,8 @@ impl fmt::Display for Attribute {
             Attribute::EntryHeader => write!(f, "entry-header"),
             Attribute::Link => write!(f, "link"),
             Attribute::LinkRemove => write!(f, "link_remove"),
-            Attribute::LinkTag(tag) => write!(f, "link__{}", tag),
-            Attribute::RemovedLink(tag) => write!(f, "removed_link__{}", tag),
+            Attribute::LinkTag(link_type) => write!(f, "link__{}", link_type),
+            Attribute::RemovedLink(link_type) => write!(f, "removed_link__{}", link_type),
             Attribute::PendingEntry => write!(f, "pending-entry"),
         }
     }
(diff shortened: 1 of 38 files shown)


hdk-rust/src/api.rs
```diff
@@ -1107,10 +1107,10 @@ pub fn remove_entry(address: &Address) -> ZomeApiResult<Address> {
     Dispatch::RemoveEntry.with_input(address.to_owned())
 }
 
-/// Consumes three values; the address of an entry get get links from (the base), the tag of the links
+/// Consumes three values; the address of an entry get get links from (the base), the type of the links
 /// to be retrieved, and an options struct for selecting what meta data and crud status links to retrieve.
-/// Note: the tag is intended to describe the relationship between the `base` and other entries you wish to lookup.
-/// This function returns a list of addresses of other entries which matched as being linked by the given `tag`.
+/// Note: the type is intended to describe the relationship between the `base` and other entries you wish to lookup.
+/// This function returns a list of addresses of other entries which matched as being linked by the given `type`.
```
maackle:
```suggestion
/// This function returns a list of addresses of other entries which matched as being linked by the given `link_type`.
```

hdk-rust/src/api.rs
```diff
@@ -803,19 +803,19 @@ pub fn get_entry_result(
...
-        tag: tag.into(),
+        link_type: link_type.into(),
     })
 }
 
 /// Commits a LinkRemove entry to your local source chain that marks a link as 'deleted' by setting
 /// its status metadata to `Deleted` which gets published to the DHT.
 /// Consumes three values, two of which are the addresses of entries, and one of which is a string that removes a
-/// relationship between them, called a `tag`. Later, lists of entries.
+/// relationship between them, called its `type`. Later, lists of entries.
```
maackle:
```suggestion
/// relationship between them, called its `link_type`. Later, lists of entries.
```

core_types/src/eav/eavi.rs
```diff
@@ -72,8 +72,8 @@ impl fmt::Display for Attribute {
             Attribute::EntryHeader => write!(f, "entry-header"),
             Attribute::Link => write!(f, "link"),
             Attribute::LinkRemove => write!(f, "link_remove"),
-            Attribute::LinkTag(tag) => write!(f, "link__{}", tag),
-            Attribute::RemovedLink(tag) => write!(f, "removed_link__{}", tag),
+            Attribute::LinkTag(link_type) => write!(f, "link__{}", link_type),
```
maackle: Should it be `LinkType` now? Or is this what actually enables real tags?

willemolding: Yes I think this should definitely be Attribute::Link but only if it is the case that it isn't used already

willemolding:
In order to prevent this branch being added to a release before the rest of links is complete I am closing this PR and continuing in #1402

PR #1402, opened 2019-05-14
Add dynamic link tagging

willemolding:
## PR summary

Depends on #1412

This PR adds tags to links with the following properties:
- [x] All links must have a tag
- [x] The tag of a link will be retrieved when hdk::get_links is called
- [x] It is possible to only get links that exactly match a given tag
- [x] It is possible to retrieve all links irrespective of the tag (and it will return all the tags)

Works great except introduced some flaky tests which need sorting out before merging. It is ready for review though


core/src/action.rs
```diff
@@ -237,8 +237,11 @@ pub struct GetLinksKey {
     /// The address of the Link base
     pub base_address: Address,
 
+    /// The link type
+    pub link_type: String,
+
     /// The link tag
-    pub tag: String,
+    pub tag: String, // this makes it tricky to retrieve all links
```
willemolding: I believe this is causing it to timeout while waiting for a get_links response to return when the returning tag is different (e.g. like when the wildcard tag is used)

core/src/dht/dht_store.rs
```diff
@@ -59,22 +59,31 @@ impl DhtStore {
...
     ) -> Result<BTreeSet<EntityAttributeValueIndex>, HolochainError> {
         let filtered = self.meta_storage.read()?.fetch_eavi(&EaviQuery::new(
             Some(address).into(),
-            EavFilter::multiple(vec![
-                Attribute::LinkTag(tag.clone()),
-                Attribute::RemovedLink(tag),
-            ]),
+            EavFilter::predicate(move |attr| {
+                match attr {
+                    Attribute::LinkTag(query_link_type, query_tag) | Attribute::RemovedLink(query_link_type, query_tag) => {
```
willemolding: Highlight: This is where filtering is done based on if a tag is provided (so match both tag and type) or only a type is provided (return all tags of that type)

core/src/workflows/get_link_result.rs
```diff
@@ -49,6 +50,7 @@ pub async fn get_link_result_workflow<'a>(
                                         Ok(LinksResult {
                                             address: link.clone(),
                                             headers,
+                                            tag: "?".into(),
```
willemolding: Need to figure out how to get access to the tag here

core/src/nucleus/ribosome/api/get_links.rs
```diff
@@ -138,4 +170,49 @@ pub mod tests {
         );
     }
 
+    #[test]
+    fn test_with_same_target_and_tag_dedup() {
```
willemolding: These two tests are currently failing but illustrate the way it should work

core/src/network/actions/get_links.rs
```diff
@@ -17,11 +17,13 @@ use std::{pin::Pin, sync::Arc, thread};
 pub async fn get_links(
     context: Arc<Context>,
     address: Address,
-    tag: String,
+    link_type: String,
```
lucksus: Hm, I guess we also want this to be an Option
```suggestion
    link_type: Option<String>,
```
so that we could requests all links with the same tag but different types. That would match the `get_links` semantic we had in proto, no?
@zippy

willemolding: So you think we should make `link_type` optional as well? I guess this is possible but not sure if it makes sense...

I am also not sure if we should be forcing every link to be tagged. This is how it works currently but I suppose we could let people pass a `None` to `link_entries` and have the HDK add a sensible default behind the scenes. There are a few usability things to sort out.

lucksus: I guess it makes sense to have no tag, i.e. `None`.

It does make sense to have the type optional in the `get_links` signature so that we can requests all links with the same tag but from different types.

Imagine a use-case where agents can add all different kinds of comments on entries, each being a link with a tag `comment`. But the reason why that comment is a valid statement could differ which would best be modeled by having multiple link types with different validation rules. But for rendering all those comments a UI might not care about why something is a valid comment, not caring for the type but instead requesting any link with the tag `comment` independent of the type.

willemolding:
Added the ability to retrieve based on type, tag, type+tag or everything.

Also added a test for getting the exact match tag (then there are other links to filter out) in the app spec and rust tests

CHANGELOG-UNRELEASED.md
```diff
@@ -6,12 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Option to show NPM output when pulling deps during `hc test` [PR#1401](https://github.com/holochain/holochain-rust/pull/1401)
-- Adds scaffolding/skeleton for a future WASM conductor [#894](https://github.com/holochain/holochain-rust/pull/894)
```
lucksus: These lines must stay in! Changelogs from other PRs that got merged before!

CHANGELOG-UNRELEASED.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Renames link tags to link_type. Adds new link tag which can be any string. This is available in validation of links and links can be retrieved based on their tag+type, just tag, just type or retrieve all.  [#1402](https://github.com/holochain/holochain-rust/pull/1402).
```
zippy: I think this change also needs a "breaking change" note re `hdk::link_entries` and `hdk::get_links`  needing extra params no?

core_types/src/link/mod.rs
```diff
@@ -62,17 +69,22 @@ pub mod tests {
...
+        link::{Link, LinkActionKind, LinkTag, LinkType},
     };
 
-    pub fn example_link_tag() -> LinkTag {
-        LinkTag::from("foo-tag")
+    pub fn example_link_type() -> LinkType {
+        LinkType::from("foo-link-type")
+    }
+
+    pub fn example_link_tag() -> LinkType {
```
zippy: Isn't the return type here supposed to be `LinkTag`?

core/src/network/state.rs
```diff
@@ -52,7 +52,7 @@ pub struct NetworkState {
     pub get_entry_with_meta_results: HashMap<GetEntryKey, GetEntryWithMetaResult>,
 
     /// Here we store the results of GET links processes.
-    /// The key of this map is the base address and the tag name for which the links
+    /// The key of this map is the base address and the link_type name for which the links
```
zippy: isn't this actually the link_type/link_tag pair?  Should this comment be expanded?

doc/holochain_101/src/first_steps.md
```diff
@@ -205,7 +205,7 @@ fn handle_create_list(list: List) -> ZomeApiResult<Address> {
 
 The `hdk::commit_entry` function is how a zome can interact with holochain core to add entries to the DHT or local chain. This will trigger the validation function for the entry and if successful will store the entry and return its hash/address.
 
-The `add_item` function requires the use of holochain links to associate two entries. In holochain-proto this required the use of a commit with a special Links entry but it can now be done using the HDK function `link_entries(address1, address2, tag)`. The add item handler accepts a `ListItem` and an address of a list, commits the `ListItem`, then links it to the list address:
+The `add_item` function requires the use of holochain links to associate two entries. In holochain-proto this required the use of a commit with a special Links entry but it can now be done using the HDK function `link_entries(address1, address2, link_type)`. The add item handler accepts a `ListItem` and an address of a list, commits the `ListItem`, then links it to the list address:
```
zippy: Missing the `link_tag` parameter in `link_entries`?  Also I think it would be good to increase the documentation here to explain the difference between type and tag.  I.e. that instead of creating a new entry type (which we needed for figuring out which validation to run) that's now encoded in the `link_type` string passed directly to the `link_entries` funciton.

doc/holochain_101/src/first_steps.md
```diff
@@ -223,7 +223,7 @@ fn handle_add_item(list_item: ListItem, list_addr: HashString) -> ZomeApiResult<
 
 At the moment there is no validation done on the link entries. This will be added soon with an additional validation callback.
 
-Finally, `get_list` requires us to use the HDK function `get_links(base_address, tag)`. As you may have guessed, this will return the addresses of all the entries that are linked to the `base_address` with a given tag. As this only returns the addresses, we must then map over each of then and load the required entry.
+Finally, `get_list` requires us to use the HDK function `get_links(base_address, link_type)`. As you may have guessed, this will return the addresses of all the entries that are linked to the `base_address` with a given link_type. As this only returns the addresses, we must then map over each of then and load the required entry.
```
zippy: same here, missing the `link_tag` param?

doc/holochain_101/src/zome/api_functions.md
```diff
@@ -147,7 +147,7 @@ Caller can request additional metadata on the entry such as type or sources
 
 Canonical name: `get_links`
 
-Consumes two values, the first of which is the address of an entry, base, and the second of which is a string, tag, used to describe the relationship between the base and other entries you wish to lookup. Returns a list of addresses of other entries which matched as being linked by the given tag. Links are created in the first place using the Zome API function [link_entries](#link-entries). Once you have the addresses, there is a good likelihood that you will wish to call [get_entry](#get-entry) for each of them.
+Consumes two values, the first of which is the address of an entry, base, and the second of which is a string, link_type, used to describe the relationship between the base and other entries you wish to lookup. Returns a list of addresses of other entries which matched as being linked by the given link type. Links are created in the first place using the Zome API function [link_entries](#link-entries). Once you have the addresses, there is a good likelihood that you will wish to call [get_entry](#get-entry) for each of them.
```
zippy: needs updating, consumes 3 values now.

doc/holochain_101/src/zome/api_functions.md
```diff
@@ -159,7 +159,7 @@ Consumes two values, the first of which is the address of an entry, base, and th
 
 Canonical name: `link_entries`
 
-Consumes three values, two of which are the addresses of entries, and one of which is a string that defines a relationship between them, called a `tag`. Later, lists of entries can be looked up by using `get_links`. Entries can only be looked up in the direction from the `base`, which is the first argument, to the `target`, which is the second. This function returns a hash for the LinkAdd entry on completion.
+Consumes three values, two of which are the addresses of entries, and one of which is a string that defines a relationship between them, called a `link_type`. Later, lists of entries can be looked up by using `get_links`. Entries can only be looked up in the direction from the `base`, which is the first argument, to the `target`, which is the second. This function returns a hash for the LinkAdd entry on completion.
```
zippy: needs updating, consumes 4 values now.

hdk-rust/src/api/get_links.rs
```diff
@@ -26,28 +28,34 @@ use holochain_wasm_utils::api_serialization::{
...
+pub fn get_links(
+    base: &Address,
+    link_type: Option<String>,
+    tag: Option<String>,
+) -> ZomeApiResult<GetLinksResult> {
+    get_links_with_options(base, link_type, tag, GetLinksOptions::default())
 }
 
-/// Retrieves data about entries linked to a base address with a given tag. This is the most general version of the various get_links
+/// Retrieves data about entries linked to a base address with a given type. This is the most general version of the various get_links
```
zippy:
```suggestion
/// Retrieves data about entries linked to a base address with a given type and tag. This is the most general version of the various get_links
```

hdk-rust/src/api/link_entries.rs
```diff
@@ -3,10 +3,18 @@ use error::ZomeApiError;
 use holochain_core_types::cas::content::Address;
 use holochain_wasm_utils::api_serialization::link_entries::LinkEntriesArgs;
 
-/// Adds a named, directed link between two entries on the DHT.
-/// Consumes three values, two of which are the addresses of entries, and one of which is a string that defines a
-/// relationship between them, called a `tag`. Later, lists of entries can be looked up by using [get_links](fn.get_links.html). Entries
+/// Adds a named, tagged, directed link between two entries on the DHT.
```
zippy: This is a great description, we also need to get this version into the mdbook part of the docs, not just here...

CHANGELOG-UNRELEASED.md
willemolding: I'll get it right one day :man_facepalming:

PR #1403, opened 2019-05-15
add conductor booting to circle

thedavidmeister:
## PR summary

hardcodes a few conductors into circle so that we can point the app spec tests at them

refactoring the app spec tests is a followup that @maackle will tackle later today

this is a very short-term solution that will be replaced with the work @maackle and @Connoropolous are doing around the orchestrator

this is just enough to unblock popping the nodejs conductor out of the main repo

## testing/benchmarking notes

## followups

## changelog


app_spec/conductor/bobo.toml
```diff
@@ -0,0 +1,21 @@
...
+[[interfaces]]
+admin = true
+id = "websocket interface"
+instances = []
+[interfaces.driver]
+type = "websocket"
+port = 8080
+
+[logger]
+type = "debug"
```
maackle: these configs don't have any networking set up, are they going to gossip to one another?

lucksus: I guess not. Without network section the conductor will default to the in-memory mock network.

lucksus:
Hm, so now a month later without merging this I wonder, do we need this at all? Both Diorama and Cluster Test are spawning conductors from JS..

PR #1404, opened 2019-05-15
WIP: app_spec link indexing showcase & changes to hdk crate to support tag option

jdeepee:
## PR summary

Implemented a showcase in app_spec which shows how links can be used for indexing of entries when link types are supported (and dynamic tag creation). Showcase uses topics which can be specified when making a post - topics are created from topic data received in `create_post` function - then link index's are made between topics and post entry. Posts can be retrieved by topics at: `get_posts_by_topics`. Accepts a vector of topics and will return any entries that are present in all topics in vector. 

Also made small changes to HDK crate so that functions associated to getting/linking of entries can accept a tag w/ type: `Option`. If option is `None` we use the following string for link data: `"*"` otherwise string inside `Some()` is used. 

Currently something funny is happening with `link_entries` & `get_links` functions. They don't seem to fully execute and don't return an error or success - just `undefined`. 

## changelog

### DIFF
--- a/app_spec/zomes/blog/code/src/blog.rs
+++ b/app_spec/zomes/blog/code/src/blog.rs
@@ -22,7 +22,9 @@ use hdk::{
 
 use memo::Memo;
 use post::Post;
-use topic::Topic;
+use topic::{
+    Topic, QueryType,
+};
 use std::{
     collections::BTreeMap,
     convert::{TryFrom, TryInto},
@@ -324,8 +326,6 @@ pub fn handle_memo_address(content: String) -> ZomeApiResult<Address> {
 pub fn create_topics(topics: &Vec<String>) -> ZomeApiResult<Vec<Address>> {
     let mut topic_addresses = vec![];
     for topic in topics{
-        hdk::debug("creating topic")?;
-        hdk::debug(topic.clone())?;
         let entry = Entry::App("topic".into(), Topic{topic: topic.to_string()}.into());
         let topic_entry_address = hdk::entry_address(&entry)?;
         match hdk::get_entry(&topic_entry_address)? {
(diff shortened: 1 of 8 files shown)


jdeepee:
Ive just added a query_type parameter to get_posts by ID. This allows the function to either get entries where the entry is present in all topics specified or to get posts from each topic specified no matter if that post is also in other specified topics.

Not sure if this is necessary in app_spec at all - but I had the code already written so thought I would include it.

app_spec/zomes/blog/code/src/blog.rs
```diff
@@ -319,20 +323,83 @@ pub fn handle_memo_address(content: String) -> ZomeApiResult<Address> {
...
+    Ok("Link indexs created".to_string())
+}
+
+pub fn handle_create_post(content: String, in_reply_to: Option<Address>, topics: Vec<String>) -> ZomeApiResult<Address> {
     let address = hdk::commit_entry(&post_entry(content))?;
+    create_topics(&topics)?;
+    create_topic_indexs(&address, topics)?;
 
-    hdk::link_entries(&AGENT_ADDRESS, &address, "authored_posts")?;
+    hdk::link_entries(&AGENT_ADDRESS, &address, "authored_posts", None)?;
```
willemolding: I think I was wrong with what I mentioned before. The plan is for every link to have a tag even if it is an empty string. Therefore the last parameter to `hdk::link_entries` cannot be none. 

What I mean to say was that the last parameter for `hdk::get_links` will be optional where the `None` type means to retrieve all links regardless of the tag. 

Sorry about that

app_spec/zomes/blog/code/src/blog.rs
```diff
@@ -385,11 +452,11 @@ pub fn handle_delete_post(content: String) -> ZomeApiResult<Address> {
 }
 
 pub fn handle_posts_by_agent(agent: Address) -> ZomeApiResult<GetLinksResult> {
-    hdk::get_links(&agent, "authored_posts")
+    hdk::get_links(&agent, "authored_posts", None)
```
willemolding: Looks good

hdk-rust/src/api.rs
```diff
@@ -804,12 +805,16 @@ pub fn link_entries<S: Into<String>>(
     base: &Address,
     target: &Address,
     link_type: S,
+    tag: Option<S>,
```
willemolding:
```suggestion
    tag: S,
```

hdk-rust/src/api.rs
```diff
@@ -1128,26 +1133,34 @@ pub fn remove_entry(address: &Address) -> ZomeApiResult<Address> {
...
     base: &Address,
     link_type: S,
+    tag: Option<S>,
     options: GetLinksOptions,
 ) -> ZomeApiResult<GetLinksResult> {
     Dispatch::GetLinks.with_input(GetLinksArgs {
         entry_address: base.clone(),
         link_type: link_type.into(),
-        tag: Some("".into()), // TODO: Expose actual parameter
+        tag: match tag {
```
willemolding: `tag.map(|t| t.into())` might work as well

willemolding:
Other than the few comments it looks great. We should keep this PR open for now but I will merge this branch into #1402 when it is ready

app_spec/zomes/blog/code/src/blog.rs
jdeepee: What about if we still allow passing an option to link_entries but then map None to an empty string. That way tag will be an option the same as it is when using get_links. I think using either option or just a string for both get_links and link_entries functions might make the most sense - but maybe I am missing something that this approach would cause in the deeper implementation @ core.

PR #1412, opened 2019-05-17
Storing address of LinkEntry (rather than target) in EAV

willemolding:
## PR summary

This is something I think we need to implementing link tags and also link CRUD correctly. Before going further I want to get some input as to if this is indeed the correct approach and any potential gochas we might encounter. 

~~This implementation appears to work and tests pass. I had to add some extra logic (copy pasted for the moment) in the reduce_add_link to store the link entry in the DHT. I am a bit confused as to why this was needed as I thought the LinkEntry was stored already?~~ Didn't actually need that.

Would appreciate any other feedback or comments and then I will make a proper PR


willemolding:
It turned out this could be done with very few changes. Most of this PR is a refactor of some of the `get_links` code.

PR #1414, opened 2019-05-20
Sunset node conductor for app spec tests

### LINKED ISSUE #1419: Adding bridges through admin interface is impossible
While we can add instances and bridges through admin interface methods, the integrity checks and order of operations for instance instantiation makes it impossible in practice to actually get things hooked up properly.

* As soon as `admin/instance/add` is called, the instance gets instantiated. This process also checks for bridges, and adds the bridge handles as instance identifiers.
    * If `admin/bridge/add` is called after `admin/instance/add`, the bridge handle is not registered as an instance identifier, and bridge calls fail
    * If bridges are added before the instances, not only does the integrity check fail, but the bridge dependency resolver doesn't know what to do since it can't find the instances the bridge is referring to, as soon as the first instance is added

There are several ways we could fix this, and I'm not sure which is best.

- we could have an admin function that sets up multiple instances and bridges atomically, to avoid invalid intermediate states
- we could rework instance creation, so the instance is not fully instantiated until it `start`s
- we could disable integrity checks when using admin interface methods
- and many more...

maackle:
## PR summary

Uses [hc-playbook](https://github.com/holochain/hc-playbook) for scenario tests, removing the dependency on `nodejs_conductor`. Playbook, in turn, uses [scenario-waiter](https://github.com/holochain/scenario-waiter) to handle eventual-consistency checks.

TODOs:

- [x] Re-enable bridging tests (fixes #1419)
- [x] Implement nodejs Waiter
- [ ] find a solution to #1446, which will cause app spec flakiness
- [ ] Update cli `hc test` (if needed)
- [ ] ~Scrub remaining nodejs_conductor references~ (we don't actually want to remove nodejs_conductor from repo yet, that will be a followup)

## followups

* Since we're eventually pulling nodejs_conductor out of this repo, we should transplant it to a new home, a repo of its own.
* Maybe change some of the nix scripts, now that there is only one conductor, can we just have e.g. `hc-conductor-install`?

## changelog


.circleci/config.yml
```diff
@@ -75,16 +75,12 @@ jobs:
 
    - run:
        name: install node conductor
-       command: nix-shell --run hc-conductor-node-install
+       command: nix-shell --run hc-conductor-rust-install
```
Connoropolous: instead of replacing this, but still calling it 'install node conductor' I think it'd be better to just remove this command, and point whatever was calling this command to the existing job

Connoropolous: actually I see that you replaced it already in `holonix/app-spec/src/test.nix` so just remove this command instead

thedavidmeister:
@maackle i'm going to try and split this into some smaller chunks as per our discussion and https://hackmd.io/F-f5uSuGRtydh9pvmSYIxw

maackle:
Broke this up into smaller pieces, the largest of which is https://github.com/holochain/holochain-rust/pull/1475

PR #1422, opened 2019-05-22
DONOTMERGE - Init has required functionality

willemolding:
## PR summary

Depends on #1417 

Moves the calling of the zome init callbacks from the `initialize` action to its own `call_init` action which is called in the application `initialize` workflow. This allows it to now be called after the network is initialized so it is able to commit entries, make hdk:send calls and bridge calls.

Also adds some tests to the wasm-test and app-spec illustrating calling of these hdk functions in init

## changelog


core/src/workflows/application.rs
```diff
@@ -17,15 +18,16 @@ pub async fn initialize(
...
             err
         ));
-        let dna = dna.ok_or(HolochainError::DnaMissing)?;
         context.log("dna/initialize: Initializing new chain from given DNA...");
-        await!(initialize_chain(dna, &instance_context))?;
+        await!(initialize_chain(dna.clone(), &instance_context))?;
     }
-    await!(initialize_network::initialize_network(&instance_context))?;
+    await!(initialize_network(&instance_context))?;
+    await!(call_init(dna, &instance_context))?;
```
willemolding: Highlight: The `call_init` now happens after the `initialize_network`

core/src/nucleus/actions/call_init.rs
```diff
@@ -0,0 +1,36 @@
...
+            CallbackResult::Fail(error_string) => Some((zome_name.to_owned(), error_string)),
+            _ => None,
+        })
+        .collect();
+
+    if errors.is_empty() {
+        Ok(())
+    } else {
+        Err(HolochainError::ErrorGeneric(format!(
+            "At least one zome init returned error: {:?}",
```
willemolding: Highlight: It will now tell you which zomes init functions are producing the error

PR #1425, opened 2019-05-22
add uuid option to admin/dna/install_from_file

maackle:
## PR summary

TODO:

- [x] write a test

## changelog


zippy:
I'm not sure I understand the motivation for this PR.  How does the UUID "overrride" what's in the DNA, does it re-write the DNA?  That would change the hash.  Is that the point?

maackle:
@zippy yes, it changes the hash, which is the main motivation. It's similar to setting a property on the DNA through the admin interface, which we already have an admin interface method for. It's needed for [Diorama](https://github.com/holochain/diorama) (the new scenario test harness) to ensure that each test runs in an independent network with a different DNA hash. I didn't want to alter `properties` because the DNA actually cares about its properties, so a test harness shouldn't be injecting randomness into a semantically relevant chunk of data. UUID seems like the vehicle for making uniquely hashed copies of the same DNA.

Connoropolous:
@maackle I think you included this work in your other PR? We can close?

lucksus:
Why WIP? This looks good, can we merge if tests pass?

maackle:
I know I neglected this branch but my intention was to review this separately and merge it in.

PR #1429, opened 2019-05-24
link_meta shouldn't use println

Connoropolous:
## PR summary
Found this by disabling logging, but I still saw this being output, so I looked it up.

## testing/benchmarking notes

## followups

## changelog


lucksus:
In https://github.com/holochain/holochain-rust/pull/1399 I removed `context` from all reducers which made our current logging solution unavailable in reducers. I had removed noisy debug/trace logs int that PR first. @struktured requested to at least log somewhere that we lost those logs. I had then argued that we most likely don't need those debug-logs and that we could easily re-add them but then I finally re-added them already in that PR to get it merged.

I still think we can just loose this println-log.
And, new logging that would work in reducers without context is underway (coming form @dymayday)

Connoropolous:
Ok @lucksus I liked your suggestion to just remove it ;) that's what I've done. You approve?

PR #1431, opened 2019-05-25
Consistency signals

maackle:
## PR summary

Adds a new type of signal, `Signal::Consistency`, which is used to enable consistency checks on scenario tests

This signal contains a subset of the data present in `Action`. Each Action that gets reduced may produce 0 or 1 Consistency signals which get emitted on the usual signal channels (broadcasters over admin interfaces, and test signal channels, if applicable). A Consistency signal contains a `ConsistencyEvent` which corresponds to the Action which produced it, and a list of future "pending" `ConsistencyEvents` to wait for.

A Waiter implementation then listens for each consistency signal, and if the signal contains pending events, it puts them in watch-list. As new signals come in that match, they are removed from the watchlist, and whenever the watchlist is empty, consistency has been achieved. The `ConsistencyEvent`s are compared directly in their opaque serialized form, requiring no special Holochain-specific logic in the Waiter implementation itself.

See https://github.com/holochain/scenario-waiter for the initial Waiter implementation (used in https://github.com/holochain/hc-playbook)

**Note to reader**: unfortunately we have a lot of picky tests around config, so the one little config required many lines of change to the tests, making this PR artificially long. Most of the interesting stuff is in two files, especially [consistency.rs](https://github.com/holochain/holochain-rust/pull/1431/files#diff-1b0244be5624b8a8ef74bbf81e8cc775R1), so just look for it.

## followups

In a conversation with @lucksus, we determined that this signal can likely be replaced with a new approach, once we have network signals that alert a node once another node has held an entry it has published. For now though, this is the only way to do consistency checks.

## changelog


core/src/instance.rs
```diff
@@ -161,18 +162,19 @@ impl Instance {
         rx_action: Receiver<ActionWrapper>,
         rx_observer: Receiver<Observer>,
     ) {
-        let sync_self = self.clone();
+        let mut sync_self = self.clone();
```
maackle: Necessary because of the stateful nature of the `ConsistencyModel` (it needs to keep track of Commits so that it can emit that data when the entry is later Published)

core/src/consistency.rs
```diff
@@ -0,0 +1,148 @@
...
+                    Entry::LinkRemove(link_data) => Some(RemoveLink(link_data.clone().link)),
+                    // Question: Why does Entry::LinkAdd take LinkData instead of Link?
+                    _ => None,
+                });
+                let mut pending = vec![hold];
+                meta.map(|m| pending.push(m));
+                let signal =
+                    ConsistencySignal::new_pending(Publish(address.clone()), Validators, pending);
+                self.commit_cache.insert(address, signal);
+                None
```
lucksus: Why is this signal not returned? Does the JS waiter not need to see commits so it would start waiting for the according pending signal? Maybe I'm misunderstanding something here.

core/src/consistency.rs
```diff
@@ -0,0 +1,148 @@
...
+                let hold = Hold(address.clone());
+                let meta = crud_link.clone().and_then(|crud| match entry {
+                    Entry::App(_, _) => Some(UpdateEntry(crud, address.clone())),
+                    Entry::Deletion(_) => Some(RemoveEntry(crud, address.clone())),
+                    Entry::LinkAdd(link_data) => Some(AddLink(link_data.clone().link)),
+                    Entry::LinkRemove(link_data) => Some(RemoveLink(link_data.clone().link)),
+                    // Question: Why does Entry::LinkAdd take LinkData instead of Link?
+                    _ => None,
+                });
+                let mut pending = vec![hold];
```
lucksus: This code does not account for private entries that don't imply a hold.

core/src/consistency.rs
```diff
@@ -0,0 +1,148 @@
...
+                meta.map(|m| pending.push(m));
+                let signal =
+                    ConsistencySignal::new_pending(Publish(address.clone()), Validators, pending);
+                self.commit_cache.insert(address, signal);
+                None
+            }
+            Action::Publish(address) => self.commit_cache.remove(address).or_else(|| {
```
lucksus: Ok, I don't think I understand how this is meant to work. Above we construct a pending list including metas, but remove the whole `ConsistencySignal` including that list of meta-pendings when we see the according `Publish` ?

maackle: @lucksus I should definitely add a comment. My assumption for dealing with private entries is thus: The signal that specifies that nodes should await a `Hold` is only fired on `Publish`, rather than on `Commit` as in nodejs_conductor. The complexity here is that the `Publish` action only contains an address, while most of the important data comes with the `Commit` action. 

So, when the consistency model sees a Commit, it constructs the `ConsistencySignal` that should be sent by a `Publish`, and stores it in this hashmap. The `remove` actually returns the value removed, it could really be called `fetch_and_remove`.

It occurs to me that private entries will clutter up the commit_cache since they'll never get published. So maybe I should check if they are publishable before even putting them in the cache.

Does that all sound right?

lucksus: Ah, that clarifies it! Sorry, got a bit confused. Yeah, private entries will fill the hash map, but I see how this works now.

maackle: @lucksus just added a commit which checks for can_publish before caching signals, and makes this more readable.

maackle:
@Connoropolous I think this is only integration-testable. The types say it all. The actual functionality is serialized data that gets spewed out through interfaces, and it's out of core's control what the consumer does with it. The actual test will be the new app spec tests.

maackle:
# Don't merge

The last commit is causing panics, because `can_publish` can panic, and in this case it is. I'm going to ask about it at the sync, and if we can't find a good solution, I will roll back that change. Unfortunately that means we'll be accumulating unpublishable commit data in this in-memory cache.

maackle:
Added a workaround for the panic which only partially accomplishes what was intended (to avoid caching things unnecessarily). Maybe this can be cleaned up in the context refactor pt 2.

PR #1441, opened 2019-05-27
cluster integration test

Connoropolous:
## PR summary

This PR introduces the ability to fully test the system, from hc-web-client through to app-spec through to n3h (and eventually lib3h).

It introduces a new holonix command `hc-app-spec-cluster-test`. This command does the following:
- compiles the new `hc_cluster_test` typescript subfolder/project into regular JS
- builds the app_spec DNA
- Runs the app_spec DNA inside as many different Conductor instances as its told to
- Checks (via a signal subscription) that it detects as many occurrences of the Hold action as it expects to, across Conductors.

`hc-app-spec-cluster-test` can be used as a local development tool for `n3h` development as well.

The following command in `/holonix/app-spec-cluster/src/test.nix` can be tweaked to use different parameters: 
`( EMULATION_HOLOCHAIN_BIN_PATH=./.cargo/bin/holochain node ./app_spec/cluster_test/index.js 2 false)`

The first argument of number of nodes can be changed:
`( EMULATION_HOLOCHAIN_BIN_PATH=./.cargo/bin/holochain node ./app_spec/cluster_test/index.js 4 )` instead of the default of 2
A second argument of whether to show debugging logs can be added:
`( EMULATION_HOLOCHAIN_BIN_PATH=./.cargo/bin/holochain node ./app_spec/cluster_test/index.js 4 true)`

## testing/benchmarking notes

Look here for an example of logs with debugging turned on, and with 4 nodes:
https://gist.github.com/Connoropolous/55b87b2fc4eb84c3ba805eafea7e7cb5

## changelog


app_spec/cluster_test/index.js
```diff
@@ -0,0 +1,69 @@
...
+  let enteringShutdown = false
+
+  // wait a maximum of 5 seconds for
+  // all the expected signals to arrive
+  // otherwise consider it a timed out failure
+  setTimeout(async () => {
+    if (!enteringShutdown) {
+      enteringShutdown = true
+      console.log('after 5 seconds, all nodes should be holding all entries and all links')
+      console.log(`There are only ${countHolding} after 5 seconds.`)
```
maackle:
```suggestion
      console.log(`There are ${countHolding} after 5 seconds.`)
```

There might be no holds barred, right?

Connoropolous: The idea of this is that if all HOLDs that were expected were found, it wouldn't be exiting with status 1

hc_cluster_test/conductor_handle.ts
```diff
@@ -0,0 +1,93 @@
+import { dnaInterfaceId } from './config'
+import { connect } from '@holochain/hc-web-client'
+import * as kill from 'tree-kill'
+
+type ChildProcess = any
+type WebsocketConnection = any
+
+type Call = (...segments: Array<string>) => (params: any) => Promise<any>
```
Connoropolous: @willemolding I just borrowed the type defs for these functions from hc-web-client, but is there some way that I could import or reference them instead?

app_spec/cluster_test/index.js
maackle: Ah I missed the setTimeout

hc_cluster_test/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "hc_cluster_test",
+  "version": "0.0.1",
+  "description": "",
+  "main": "conductor_cluster.js",
+  "author": "",
+  "license": "ISC",
```
Connoropolous: @zippy what license should this be?

PR #1450, opened 2019-05-29
Fix BadCallError

lucksus:
## PR summary

This fixes two bugs that made it impossible to call `hdk::call` or use globals like `hdk::AGENT_ADDRESS` in any non-zome-function, i.e. callbacks like receive, genesis or validation functions.

## Context
The implementations of HDK API functions like `invoke_call` or `init_globals` can't have direct access to the instance or its state/context. Instead, a `Runtime` object is passed in by WASMi. In our case, that contains a `CallData` enum which has specializations for three different cases:
* zome function call
* callback call
* direct call
The last is for cases where WASM gets executed outside the context of an instance, like in `hc --package` when the DNA JSON is created from config info in WASM.

The other two distinguish zome function calls from callbacks.

`hdk::call` and `init_globals` both assumed to always find the zome function call specialization and explicitly failed with `BadCallError` otherwise.

This was unnecessary and gets removed in this PR.

## changelog


core/src/nucleus/ribosome/api/init_globals.rs
```diff
@@ -25,7 +27,16 @@ pub fn invoke_init_globals(runtime: &mut Runtime, _args: &RuntimeArgs) -> ZomeAp
...
-        cap_request: runtime.zome_call_data()?.call.cap.clone(),
+        cap_request: runtime
+            .zome_call_data()
+            .map(|zome_call_data| zome_call_data.call.cap.clone())
+            .unwrap_or_else(|_| {
+                CapabilityRequest::new(
+                    Address::from("none"),
+                    Address::from("none"),
+                    Signature::fake(),
+                )
```
lucksus: @zippy: I'm just creating a bogus cap_request here. Should that field be an `Option<CapabilityRequest>` so we can set it to `None` in case we are in a callback...

zippy: The context here is that the the app should be able to find the provenance of who made a given call, and that's inside the capabilityrequest of the call.  So yes, I think we should use None to indicate the context that the call came from self as a callback.  I guess we could have some enum to make this more explicit.  But we definitely shouldn't have a bogus CapabilityRequest.

PR #1458, opened 2019-05-31
(tiny) Fix bug in get_links_and_load

willemolding:
## PR summary

Calling get_links_and_load when one of the linked entries had been deleted caused it to error. This has been fixed and an app spec test added to show it working after entry deletion.

Probably a rewrite of the get_links* variants is needed soon but this is a bugfix for the mean time


willemolding:
Oh I forgot we have to add proc_macro tests now as well :P

PR #1461, opened 2019-05-31
Bridge config/runtime checks

lucksus:
## PR summary

[*from changelog*] **BREAKING:** Conductor configuration checks for bridges added #1461. Conductor will bail with an error message if the configuration of bridges between instances does not match the bridge requirements defined in the caller instance's DNA (required bridge missing, DNA hash mismatch, trait mismatch) or if a bridge with the handle specified in the config can not be found in the caller's DNA.

## Detail
So far, bridges were working but the configuration was a bit raw without any sanity checks. 
This PR fixes that by introducing:

 - [x] Check that any required bridges as defined in the caller's DNA are configured and the callee is running, before starting the caller instance
 - [x] Check that the callee's DNA hash is as expected for `BridgeReference::Address` bridges when running `Config::check_consistency` (needs to pass the `dna_loader` as parameter)
 - [x] Check that the callee implements the required zome functions for `BridgeReference::Trait` bridges

## Bridge configuration in DNA

In order to make it possible to configure a bridge, any of the caller's DNA's zomes now needs to define a bridge with handle (=name) that matches the one that is provided in the conductor config and which is also the name by which the caller references the callee.

In order to make the app-spec tests pass again, I had to add this JSON file to the blog zome directory, which adds a `bridges` section to the zome DNA:
```json
{
  "bridges": [
    {
      "presence": "optional",
      "handle": "test-bridge",
      "reference": {
        "traits": {}
      }
    }
  ]
}
```

Without it the bridge test would fail because the conductor config mentions a bridge with handle `test-bridge` which fails if it can't find a definition of a `test-bridge` it in the caller's DNA.

This definition makes the presence of the bridge optional so that all other tests that don't setup a bridge in the config still work.

The reference goes by trait with an empty trait list which results in no further checks against the callee's DNA. This makes sense for not breaking tests but in a real-world application we would either want to bind statically to specific DNA hash, or at least require the callee to implement a certain trait that we would specify through function signatures like so:

```json
"presence": "optional",
"handle": "Vault",
"reference": {
  "traits": {
    "persona_management": {
      "functions": [
        {
          "name": "get_persona",
          "inputs": [{"name": "domain", "type": "string"}],
          "outputs": [{"name": "persona", "type": "json"}]
        }
      ]
    }
  }
}
```

This would make the conductor check that any callee connected to this DNA via bridge implements a trait `persona_management` that has at least one function with the signature: `get_persona(domain : string) -> persona : JSON` and reject the conductor config if it doesn't.

See [test code in dna/mod.rs](https://github.com/holochain/holochain-rust/blob/1f746880eab10fd0d382fbabbdc80eb801c9e75d/core_types/src/dna/mod.rs#L831) for more examples.

## Follow-up

Currently bridges need to be configured by manually adding a definition to the JSON code. Next: add syntactic sugar to our HDK macros.

## changelog


conductor_api/src/conductor/base.rs
```diff
@@ -320,14 +322,63 @@ impl Conductor {
...
-            .map(|(id, hc)| {
-                notify(format!("Starting instance \"{}\"...", id));
-                hc.write().unwrap().start()
-            })
+        self.config
+            .instances
+            .iter()
+            .map(|instance_config| instance_config.id.clone())
+            .collect::<Vec<String>>()
+            .iter()
```
AshantiMutinta: hmmm do we have to collect and call another iter()? Can it continue from the previous iter?

conductor_api/src/conductor/base.rs
```diff
@@ -1419,26 +1487,39 @@ pub mod tests {
...
-            .functions
-            .push("hello".into());
+        let fn_declaration = bridge_call_fn_declaration();
+
+        {
+            let zome = dna.zomes.get_mut("greeter").unwrap();
+            zome.fn_declarations.push(fn_declaration.clone());
+            zome.traits
+                .get_mut("hc_public")
+                .unwrap()
```
AshantiMutinta: The idea I guess is these unwraps are unrecoverable? Small change but would prefer expects since they give a little bit more context

conductor_api/src/config.rs
```diff
@@ -147,13 +153,35 @@ impl Configuration {
...
+            dna_config.is_some().ok_or_else(|| {
                 format!(
                     "DNA configuration \"{}\" not found, mentioned in instance \"{}\"",
                     instance.dna, instance.id
                 )
             })?;
+            let dna_config = dna_config.unwrap();
+            let dna =
+                Arc::get_mut(&mut dna_loader).unwrap()(&PathBuf::from(dna_config.file.clone()))
+                    .map_err(|_| format!("Could not load DNA file \"{}\"", dna_config.file))?;
```
AshantiMutinta: get_mut will return a panic if they are Arc references to the same value. Is this what we want or can we switch to a mutex?

conductor_api/src/conductor/base.rs
zippy: yah, but remember this is in a test, so if it fails something really weird is going on.

AshantiMutinta: Oh yeah it is, Dismiss this comment @lucksus

conductor_api/src/config.rs
```diff
@@ -222,6 +235,129 @@ impl Configuration {
...
+            format!(
+                "DNA configuration \"{}\" not found, mentioned in instance \"{}\"",
+                caller_config.dna, caller_config.id
+            )
+        })?;
+
+        let caller_dna_file = caller_dna_config.file.clone();
+        let caller_dna =
+            Arc::get_mut(&mut dna_loader).unwrap()(&PathBuf::from(caller_dna_file.clone()))
+                .map_err(|_| format!("Could not load DNA file \"{}\"", caller_dna_file))?;
```
zippy: Can we have the actual error message here?  That's often helpful, i.e. perms, file not found, etc:

```suggestion
                .map_err(|err| format!("Could not load DNA file \"{}\"; error was: {}", caller_dna_file, err))?;
```

conductor_api/src/config.rs
```diff
@@ -222,6 +235,129 @@ impl Configuration {
...
+            format!(
+                "DNA configuration \"{}\" not found, mentioned in instance \"{}\"",
+                callee_config.dna, callee_config.id
+            )
+        })?;
+
+        let callee_dna_file = callee_dna_config.file.clone();
+        let callee_dna =
+            Arc::get_mut(&mut dna_loader).unwrap()(&PathBuf::from(callee_dna_file.clone()))
+                .map_err(|_| format!("Could not load DNA file \"{}\"", callee_dna_file))?;
```
zippy: ditto

```suggestion
                .map_err(|err| format!("Could not load DNA file \"{}\"; error was: {}", callee_dna_file, err))?;
```

conductor_api/src/conductor/base.rs
lucksus: Yes, we actually have to because we are borrowing `self` in the first line and then again in the second `map`. The `collect()` and `iter()` ends the first borrow and enables the second one.

conductor_api/src/config.rs
lucksus: So far we are fine since the dna_loader is only used during boot-up of the conductor, which happens in one thread.

But you might right that we should switch anyway. And I would not do that within the scope of this PR :)

PR #1464, opened 2019-06-03
Adds conversions from json string to generic Results

willemolding:
## PR summary

Adds some of the missing conversion implementations for JsonString. It is now possible to `try_into` a result type directly from a JsonString.

This was NOT possible before meaning bridge calls were really difficult to get right without resorting to serde_json::Value.

It needs extra implementations for all variations of result containing a `String` since this no longer implements `Into<JsonString>`

## changelog


core_types/src/json.rs
```diff
@@ -449,4 +501,24 @@ pub mod tests {
             String::from(JsonString::from(RawString::from(1))),
         );
     }
+
+    #[test]
+    fn result_from_json_string() {
```
willemolding: @philipbeadle this should be very similar to how you want to use it in P&P

pjkundert:
Great work!  This solves exactly the problem I ran into with both HoloFuel and the App Spec tests, when trying to serialize a Result<...> response for the hdk::send/receive round-trip.  Looking forward to seeing this merged!

Issue #1446, opened 2019-05-28
Interfaces sometimes self-collide, restarting before fully stopping when adding/remove instances


maackle:
Infrequently, when adding/removing an instance from a WebSocket interface, the interface will trip over itself when restarting, apparently not completely shutting down before restarting.

Sample output (with some extra [C] markings from a different context that I'm running the conductor in)
```
[C] Stopping instance "alice-11"...
Removed instance "alice-11".

[C] Stopping interface test-interface-40481

[C] Starting interface 'test-interface-40481'.

[C] Stopping instance "bob-11"...
Removed instance "bob-11".

[C] Stopping interface test-interface-40481

[C] Starting interface 'test-interface-40481'.

!C! thread 'jsonrpc-eventloop-0' panicked at 'called `Result::unwrap()` on an `Err` value: "Address already in use (os error 98)"', src/libcore/result.rs:997:5
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

[C] 2019-05-28 15:11:37 ThreadId(1):conductor: err/conductor: Error running interface 'test-interface-40481': Address already in use (os error 98)
```

PR #1475, opened 2019-06-03
Diorama-based app spec tests

maackle:
## PR summary

Based on #1425 and https://github.com/holochain/holochain-rust/tree/reset-conductor-api-after-adding-bridge (PR forthcoming)

Uses [diorama](https://github.com/holochain/diorama) instead of nodejs_conductor to run app-spec tests. This is in preparation for removing nodejs_conductor from this repo and ceasing to use it for app-spec tests.

## followups

- [ ] find a solution to #1446, which will cause app spec flakiness
- [ ] move nodejs_conductor to its own repo [&raquo; hackmd](https://hackmd.io/F-f5uSuGRtydh9pvmSYIxw)

## changelog


CHANGELOG-UNRELEASED.md
```diff
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Error log output added for errors occurring during `hdk::call`, including bridge call errors [#1448](https://github.com/holochain/holochain-rust/pull/1448).
+- New `uuid` parameter for `admin/dna/install_from_file`, to set the UUID of the installed DNA, changing its hash
```
Connoropolous: add in PR ref

conductor_api/src/conductor/admin.rs
```diff
@@ -16,8 +16,13 @@ use std::{
     fs::{self, create_dir_all},
     path::PathBuf,
     sync::{Arc, RwLock},
+    thread::sleep,
+    time::Duration,
 };
 
+/// how many milliseconds sleep all bugs under rugs
```
Connoropolous: lol, but also. 

Can I get an update/explainer? What's happening that we need to do this?

core_types/build.rs
```diff
@@ -1,5 +1,8 @@
-use std::process::Command;
+use std::{path::Path, process::Command};
 fn main() {
+    if Path::new("../.git/HEAD").exists() {
```
Connoropolous: what's this?

Connoropolous:
@maackle I notice the attempt to do `which holochain` and `which hc`

They produce: 
```
./build_and_test.sh: line 6: which: command not found
```
in the CI. I'm assuming they would do the same locally. 

I need to double check to verify which `holochain` is on the PATH

core_types/build.rs
AshantiMutinta: Can we use pathbuf instead? Just handles the seperators better

maackle:
@Connoropolous 

> the attempt to do `which holochain`...

Yeah, @thedavidmeister you think we can add `which` to nix-shell, or should I take this out? It's a harmless warning but no use keeping it in if it works for nobody (works for me on my host machine though of course)

conductor_api/src/conductor/admin.rs
maackle: This is a workaround for https://github.com/holochain/holochain-rust/issues/1446

Connoropolous: Ok cool, well maybe we could put that into the code itself so that there's a good record

app_spec/build_and_test.sh
```diff
@@ -1,8 +1,11 @@
-#! /bin/bash
+#!/usr/bin/env bash
 mkdir -p dist
 echo "===================================================================================="
 echo "RUNNING cargo test for zomes"
+echo "Using conductor binary: `which holochain`"
+echo "Using cli binary:       `which hc`"
```
maackle: One of the sources of whichiness @thedavidmeister

core_types/build.rs
maackle: Isn't Path just a slice of a PathBuf?

AshantiMutinta: Pathbuf is to string as Path is to &str , so what we would want is to PathBuf::().join() so that we dont have to handle the seperator logic for different Oses

thedavidmeister: @maackle @AshantiMutinta yes and this is my code and i hate it whoops

PR #1476, opened 2019-06-03
Reset caller's conductor api after adding bridge dynamically

### LINKED ISSUE #1419: Adding bridges through admin interface is impossible

lucksus:
## PR summary

These changes make bridging work in the case of a purely dynamically created config, as needed for the Rust-conductor based testing framework.

Before it was impossible to have bridges working if they were not configured in the config file during start-up. The reason was that bridges are called through a conductor API that is exposed to the caller instance. These conductor APIs were only created on boot-up.

This PR breaks that code out into a separate function that also gets called when a bridge is added through an admin interface after the conductor finished booting.

closes #1419 

## changelog

### DIFF
--- a/conductor_api/src/conductor/base.rs
+++ b/conductor_api/src/conductor/base.rs
@@ -37,6 +37,7 @@ use std::{
 };
 
 use conductor::passphrase_manager::{PassphraseManager, PassphraseServiceCmd};
+use config::AgentConfiguration;
 use holochain_net::{
     ipc::spawn::{ipc_spawn, SpawnResult},
     p2p_config::P2pConfig,
@@ -44,7 +45,6 @@ use holochain_net::{
 use interface::{ConductorApiBuilder, InstanceMap, Interface};
 use signal_wrapper::SignalWrapper;
 use static_file_server::StaticServer;
-use config::AgentConfiguration;
 
 lazy_static! {
     /// This is a global and mutable Conductor singleton.
@@ -605,7 +605,11 @@ impl Conductor {
             })
     }
 
-    pub fn build_conductor_api(&mut self, instance_id: String, config: &Configuration) -> Result<IoHandler, HolochainError> {
+    pub fn build_conductor_api(
+        &mut self,
+        instance_id: String,
+        config: &Configuration,
+    ) -> Result<IoHandler, HolochainError> {
         let instance_config = config.instance_by_id(&instance_id)?;
         let agent_id = instance_config.agent.clone();
         let agent_config = config.agent_by_id(&agent_id)?;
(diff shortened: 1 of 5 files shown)


conductor_api/src/conductor/base.rs
```diff
@@ -504,8 +505,10 @@ impl Conductor {
     pub fn instantiate_from_config(
         &mut self,
         id: &String,
-        config: &Configuration,
+        maybe_config: Option<&Configuration>,
     ) -> Result<Holochain, String> {
+        let self_config = self.config.clone();
+        let config = maybe_config.unwrap_or(&self_config);
```
maackle: I wonder if this is possible:

```suggestion
        let config = maybe_config.unwrap_or_else(|| &self.config.clone());
```

lucksus: Guess it would be possible, but why clone when we want a reference anyway?

conductor_api/src/conductor/base.rs
```diff
@@ -653,6 +605,77 @@ impl Conductor {
...
+        if let Some(true) = agent_config.holo_remote_key {
+            // !!!!!!!!!!!!!!!!!!!!!!!
+            // Holo closed-alpha hack:
+            // !!!!!!!!!!!!!!!!!!!!!!!
+            api_builder = api_builder.with_outsource_signing_callback(
+                self.agent_config_to_id(&agent_config)?,
+                self.config
+                    .signing_service_uri
+                    .clone()
+                    .expect("holo_remote_key needs signing_service_uri set"),
```
AshantiMutinta: If we are returning a result, should we not handle this error better?

conductor_api/src/conductor/base.rs
```diff
@@ -653,6 +605,77 @@ impl Conductor {
...
+                r#"
+                    We have to create instances ordered by bridge dependencies such that we
+                    can expect the callee to be present here because we need it to create
+                    the bridge API"#,
+            );
+
+            api_builder =
+                api_builder.with_named_instance(bridge.handle.clone(), callee_instance.clone());
+            api_builder =
+                api_builder.with_named_instance_config(bridge.handle.clone(), callee_config);
```
AshantiMutinta: Not to be applied but you could also probably make use of the question mark operator here by folding

config.bridge_dependencies(id.clone()).iter().fold(OK(api_builder),|bridge,current_result|{
current_result?;
config.instance_by_id(&bridge.callee_id)?;
self.instances.get(&bridge.callee_id)?;
api_builder=api_builder.with_named_instance(bridge.handle.clone(), callee_instance.clone());
api_builder =api_builder.with_named_instance_config(bridge.handle.clone(), callee_config);
OK(api_builder)
})?;

AshantiMutinta: Just to reduce the number of expects here

lucksus: The expect above is one of those few examples were I really think an expect is a more appropriate choice than returning an error. The reason is that it really is an invariant of the code that the callee instance should be live already. Instances are created ordered by their bridge dependencies. If this invariant should be violated through some future changes, I want an `expect` to notify the developer. It is not a runtime error that could be recovered from by a developer mistake. That is exactly the semantic attached to explicit panics. Maybe an `assert!` would make this even clearer?

conductor_api/src/conductor/base.rs
lucksus: This code was not changed but just moved. I chose an expect here because it is in the scope of the Holo closed-alpha hack and should go away soon. But maybe you're right and we should change it to return an error. I will go ahead and merge this anyway since this discussion is not really in scope for this PR.

conductor_api/src/conductor/base.rs
AshantiMutinta: Assert would be clearer I think because returning a result but doing expect makes it look like we are going around the type system.

I guess it is also a question of "what do we do with the result of this whole process?" cause if we expect it then we could also return a result and then expect it from the outside? I think if it is developer notification then  yes maybe an assert would be better

PR #1477, opened 2019-06-04
Unwrap Party

AshantiMutinta:
## PR summary

Just removing some unwraps 

## changelog

- [ x] this is not a code change, or doesn't effect anyone outside holochain core development


AshantiMutinta:
> This is fine by me, but I'm just worried that this cleanup work will get skipped by @struktured moving of this code into lib3h. Maybe good to do it there?

This is a good point, I will move this the other side

PR #1482, opened 2019-06-05
updated old version numbers that got missed by release process

zippy:
## PR summary

It appears that sometimes old version numbers didn't get updated by the release process, probably if stuff was stuck in branches that were worked on over a long time and then merged poorly.

For this PR, I did a manual search for all version numbers from "0.0.7-alpha" to "0.0.17-alpha" and fixed them to be 0.0.18-alpha1 which is what's current at the time of this branch.  Thus this should get merged before next release so the scripts don't miss these updates again!

## testing/benchmarking notes

## followups

## changelog

### DIFF
--- a/doc/holochain_101/src/first_steps.md
+++ b/doc/holochain_101/src/first_steps.md
@@ -517,4 +517,4 @@ Pro tip: [Pipe the output to tap-spec](https://github.com/scottcorgan/tap-spec) 
 
 And there we have it! A simple Zome created with Holochain using the Rust HDK.
 
-The [complete working version of this project is available on github](https://github.com/willemolding/holochain-rust-todo). This builds under the 0.0.18-alpha1 release but as the API and HDK are changing it will likely fail under newer releases.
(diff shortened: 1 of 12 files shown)


doc/holochain_101/src/first_steps.md
```diff
@@ -1,7 +1,7 @@
 # First steps writing Holochain hApps with Rust
 
 ___
-This tutorial builds for the 0.0.9-alpha release but as the API and HDK are changing it will likely fail under newer releases.
+This tutorial builds for the 0.0.18-alpha1 release but as the API and HDK are changing it will likely fail under newer releases.
```
Connoropolous: @zippy this one can't be updated unless the tutorial is actually updated too, which I think someone is doing that open source in another PR

Connoropolous:
@zippy yah we would prbly also want to make the relevant changes to holonix release scripts to make sure it happens ongoing

zippy:
I think holonix looks for the "previous" value, but they got skipped some times.  I did add a note about this in the release process.

doc/holochain_101/src/first_steps.md
jamesray1: Agreed. What I'll do is I'll update to the latest release and check that the tutorial works fine.

PR #1487, opened 2019-06-06
Make bridge and zome calls consistent

willemolding:
## PR summary

This addresses an issue we have been having with bridge calls in developing Personas/Profiles.

Prior to this if calling a function with a signature `fn_name(pamams) -> ZomeApiResult<SomeType>` using `hdk::call` from another zome in the same DNA the result would be a JsonString-ified `ZomeApiResult<SomeType>`. If calling the same function over a bridge with `hdk::call` the result would be a JsonString-ified `SomeType`.

I understand that in the past it was necessary to do this unwrapping in code as it was difficult to do in the zome but the conversion is now trivial with #1464 .

In the current implementation not only is it inconsistent but it does not account for the case where a zome call may not return a ZomeApiResult (such as functions which cannot fail)

This PR removes the unwrapping in the core and updates the app spec to reflect the change. The updated app spec now shows consistency between zome and bridge calls.

@lucksus in particular I would like to discuss this with you

## changelog

### DIFF
--- a/core/src/nucleus/ribosome/api/call.rs
+++ b/core/src/nucleus/ribosome/api/call.rs
@@ -6,10 +6,7 @@ use crate::{
         ZomeFnCall,
     },
 };
-use holochain_core_types::{
-    error::HolochainError,
-    json::{JsonString},
-};
+use holochain_core_types::{error::HolochainError, json::JsonString};
 use holochain_wasm_utils::api_serialization::{ZomeFnCallArgs, THIS_INSTANCE};
 use jsonrpc_lite::JsonRpc;
 use snowflake::ProcessUniqueId;
--- a/core/src/nucleus/ribosome/api/call.rs
+++ b/core/src/nucleus/ribosome/api/call.rs
@@ -8,11 +8,10 @@ use crate::{
 };
 use holochain_core_types::{
     error::HolochainError,
-    json::{JsonString, RawString},
+    json::{JsonString},
 };
 use holochain_wasm_utils::api_serialization::{ZomeFnCallArgs, THIS_INSTANCE};
 use jsonrpc_lite::JsonRpc;
-use serde_json::Value;
 use snowflake::ProcessUniqueId;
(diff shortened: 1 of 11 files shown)


core/src/nucleus/ribosome/api/call.rs
```diff
@@ -139,32 +135,8 @@ fn bridge_call(runtime: &mut Runtime, input: ZomeFnCallArgs) -> Result<JsonStrin
...
-                    // one last time:
-                    let maybe_parsed_value: Result<Value, serde_json::error::Error> =
-                        serde_json::from_str(&sanitized_response);
-                    match maybe_parsed_value {
-                        Ok(parsed_value) => Ok(JsonString::from(parsed_value)),
-                        Err(_) => Ok(JsonString::from(RawString::from(sanitized_response))),
-                    }
-                })
+            // This should be returned as a JsonString for handling in the zome code.
+            Ok(JsonString::from_json(&sanitized_response))
```
willemolding: This is the main change to core. The unwrapping code is dropped and the JsonString result returned directly as for a zome call

philipbeadle:
How does this affect the Holo apps that bridge? I agree we should be consistent across the calls.

philipbeadle:
This works so well!! The code in basic-chat to call P&P across the bridge and across zomes is super simple and readable now.

lucksus:
I'm very glad if we can do this unwrapping/deserializing in WASM now - and even that easy!
I had added the unwrapping code by responding to a pull from Holo and Julio in particular. With that I was initially making bridge and zome calls consistent, but, yes, only for the Results known by that unwrapping code.. ;)

I would like to ping people in Holo who are using bridges right now if they agree this is better. They will have to make the same changes using `.try_into()` as you did here in app-spec.

But I'm all for this change!

@peeech, @pjkundert, @artbrock (who else?)

zippy:
whoa, something broke in the last last merge develop

PR #1494, opened 2019-06-07
WIP: Fix Memory Leaks

lucksus:
## PR summary

Based on #1475.

Instances and all that belongs to them (mainly the state) are not dropped through `admin/instance/remove` because the `InstanceMap` in the conductor holds `Arc<RwLock<Holochain>>` references. And `Instance` holds `Arc<RwLock<State>>` references. As long as there is an `Arc` having a reference, the `RwLock` and it's content is not dropped. The redux action loop lives in its own thread and so far was never cleaned up.

This PR adds `Instance::stop_action_loop()` that kills the redux thread, and `Holochain::kill()` that drops its `Instance` and `Context` reference.

#1495 got merged into this already.

## changelog


conductor_api/src/interface.rs
```diff
@@ -163,7 +163,8 @@ impl ConductorApiBuilder {
             let func_name = Self::get_as_string("function", &params_map)?;
 
             let cap_request = {
-                let context = hc.context();
+                let context = hc.context()
+                    .expect("Reference to dropped instance in interface handler. This should not happen since interfaces should be rebuild when an instance gets removed...");
```
zippy:
```suggestion
                    .expect("Reference to dropped instance in interface handler. This should not happen since interfaces should be rebuilt when an instance gets removed...");
```

conductor_api/src/holochain.rs
```diff
@@ -168,26 +168,51 @@ impl Holochain {
...
+        }
+    }
+
+    fn check_active(&self) -> Result<(), HolochainInstanceError> {
+        if !self.active {
+            Err(HolochainInstanceError::InstanceNotActiveYet)
+        } else {
+            Ok(())
+        }
+    }
```
maackle: Now we have two ways to return `InstanceNotActiveYet`, one if the `active` bool is false, another if the instance or context is not initialized yet. They mean different things, because an instance can be initialized and yet not active.

However, I don't think we actually use the notion of "active" for anything meaningful yet. What would it mean for an instance to be initialized and not active? Currently the only thing that changes is this boolean. It feels like a placeholder for something more meaningful. If we are going to add a more meaningful notion of active/inactive, maybe we should do it now, or soon as a followup. I imagine deactivating an instance also has to do with stopping its action loop, without totally tearing it down, which is related to this work.

maackle:
Let's look at #1495 and maybe merge that into here first

conductor_api/src/holochain.rs
lucksus: Yes. „active“ should reflect if the action loop is running. I was reusing InstanceNotActiveYet but intended to add a new error.

core_types/src/ugly.rs
```diff
@@ -0,0 +1,23 @@
+use std::fmt::Debug;
```
zippy: I'm not very happy with `ugly` in core_types.  This seems to mean that we are explicitly taking on some tech-debt, and if so I'd like to see that justified and documented in the PR under followups, and even better in the tech-debt section of the tree.

lucksus:
I've remove the `println!`s and added `HolochainInstanceError::InstanceNotInitialized` as requested, but put this PR back to WIP as it actually does not fix the memory leak (yet).

core_types/src/ugly.rs
maackle: Actually I think now that these printlns are removed, we can do away with `lax_send`. This was part of the path of discovery. I think in the current approach, it's actually acceptable to send on a channel and ignore whether it succeeded or not, because in each of these cases, that's now handled by the corresponding Future. I'll move these functions elsewhere.

I still like the idea of an `ugly` module, because if we're going to have tech-debt, I want it to be highly visible and slightly uncomfortable. But after this change, it would be empty.

maackle: @zippy where would be a good place to move `lax_send_sync`? I want to keep the function but move it out of `ugly`

PR #1501, opened 2019-06-10
reference holonix

thedavidmeister:
## testing/benchmarking notes

## followups

## changelog


default.nix
```diff
@@ -1,18 +1,19 @@
 let
+ holonix-release-tag = "expecting-bad-ozone";
+ holonix-release-sha256 = "0x6bma8p1kmlzsr2dd5zjyvw809hjmycdk8n8bknigx7vd88v52m";
 
-  pkgs = import ./holonix/nixpkgs/nixpkgs.nix;
-  darwin = import ./holonix/darwin/config.nix;
-  openssl = import ./holonix/openssl/config.nix;
-  rust = import ./holonix/rust/config.nix;
-
+ holonix = import (fetchTarball {
```
thedavidmeister: this is where the magic happens

default.nix
```diff
@@ -1,18 +1,19 @@
 let
+ holonix-release-tag = "expecting-bad-ozone";
```
thedavidmeister: not using semver

Connoropolous:
@thedavidmeister can we get a little explainer of what the outcomes of this PR are?

thedavidmeister:
@Connoropolous 

- moved the holonix directory to the holonix repo
- reworked the holonix output slightly to be a set rather than a derivation
  - means that downstream consumers can provide a context relevant derivation and use holonix as a variable
  - means that downstream consumers can reference the holonix repo by URL as a tarball
  - core is the first downstream so `foo` becomes `holonix.foo`
  - node js will be the second downstream so we can pull it out
- plan for future is that holonix repo will be refactored to be only the "commons" and downstreams will implement their own workflows e.g. `hc-*`

thedavidmeister:
@Connoropolous for example... the conflicts in dist right now should sit in core!

Connoropolous:
thanks @thedavidmeister

PR #1505, opened 2019-06-10
Use lib3h persistence crates

struktured:
## PR summary

This PR incorporates the new persistence crates migrated from `holochain-core` and generalized into the `lib3h` repo. 

  - the `cas` and `eav` apis are stripped out from `holochain_core_types` into `holochain_persistence_api` within holochain-persistence repository
  - `core_types_derive` is moved into holochain-serialization as `holochain_json_derive`
  - `cas_implementations` is removed and factored into `holochain_persistence_[mem|file|pickle]` in the holochain-persistence repository.
  - attribute validation is done by wrapping calls to `EntityAttributeValueIndex::new` with a factory function in `eavi.rs` (previously was done internally).
## testing/benchmarking notes

- existing cas / eav tests still exist

## followups
 - make separate repos for both persistence and json crates.
 
## changelog


lucksus:
Also, @AshantiMutinta's changes in https://github.com/holochain/holochain-rust/pull/1363 would get lost by this.

struktured:
> Also, @AshantiMutinta's changes in #1363 would get lost by this.

I've merged in @AshantiMutinta's changes in https://github.com/holochain/holochain-rust/pull/1505/commits/7b703d2ad183681e1a816a58342365c08f766091 and https://github.com/holochain/lib3h/pull/109

struktured:
> Woah! I‘m quite surprised by this PR.
> 
> * Why renaming all of this to lib3h_persistence? I thought lib3h would use the same CAS/EAV from core.

The names are now:
 - `holochain_json_derive` (previously `core_types_derive`)
 - `holochain_json_api` (extracted from `core_types`)
 - `holochain_persistence_api` (extracted from `core_types`)
 - `holochain_persistence_[file|mem|pickle]` (previously `cas_implementations`)

> * Why ripping out all the implementations? I thought lib3h would get a trait object in the same way core just gets a trait object from the conductor. That way we keep core clean from file system access code and alike that won’t compile to WASM and confine future compile-time switches to only conductor level.

You're right that in essence the trait is the only thing that is needed to be shared and perhaps that would have been a simpler approach to start with. However from a separation of concerns perspective it's beneficial to factorize out the shared implementations and since the work is done we might as well take advantage of it.

As for what repo the code should exist, the ideal is to make separate repositories for persistence _and_ json, but at the moment they live in `lib3h` for simplicity. If this is an immediate pain point for some, we could easily start migrating the code to the appropriate repos after (or even before) the PR is merged (since everything uses crates.io anyhow).

> I would like to talk about the applied strategy in this PR. I‘m pretty sure we can do this (providing storage to lib3h) with much less changes and much less consequences.

@lucksus We had a few overflow discussions to reach agreement- are there any other concerns left on your end?

struktured:
> i'm still pretty confused about the idea of "lib3h persistence crates"

Renamed to `holochain_persistence_[api|pickle|mem|file]`

PR #1506, opened 2019-06-10
Create per-agent n3h persistence directory for `hc run`; fixes #1338

### LINKED ISSUE #1338: `hc run ...` doesn't allocate a temporary work directory; all agents share one
Starting multiple agents via `hc run ...` causes communications failure between the agents, because some communications details are persisted to the directory, and both agents end up using the same persistence directory and using the same communications details.

When running an agent via `hc run ...`, there are 2 major use cases:

1. Starting the agent for testing, with intent to discard the resultant source chain and other data
2. Persisting the source chain and other data between `hc runs ...` invocations

The default, without specifying options to the contrary, should be to create a temporary directory that contains the source chain and data, which is discarded at the end of execution.

A `--persist=<directory>` option should be added, to allow the explicit provision of a directory to persist source chain and other necessary data into.

pjkundert:
## PR summary

When you start multiple instances of `hc run`, every instance of n3h spawned uses the same persistence directory.   This is almost certainly not what you expect, and will cause n3h to fail with strange error messages.

This pull is almost certainly not the correct way to fix this: n3h should persist per-agent data in files or directories specific to the Agent it is representing.  But, this gives us something to begin working with -- at least, by default, if you start `HC_AGENT=Agent-1 hc run` and `IPC=<...> P2P=<...> HC_AGENT=Agent-2 hc run` (referencing the ipc: and p2p: data output by the first instance), they will successfully start and communicate with eachother.

Presently, you need to add `HC_N3H_WORK_DIR=$$( mktemp -d ) ... hc run` to each per-agent instance, in order for multi-agent tests to work.

## testing/benchmarking notes

## followups

## changelog


pjkundert:
Probably out of date by now; if nobody else is running into this, it mustn't be a problem...

PR #1510, opened 2019-06-12
remove nodejs conductor

thedavidmeister:
## PR summary

node conductor moving back to https://github.com/holochain/holochain-nodejs/pull/17

## testing/benchmarking notes

## followups

reinstate over at other repo

## changelog


maackle:
@Connoropolous I guess as long as the npm package is published, the scaffold will also be fine, because the scaffold is just importing from npm anyway. When we change the scaffold to use diorama, that will mark the official switch/deprecation, but people are still free to use whichever test runner they want.

Issue #1499, opened 2019-06-10
Update all HDK methods to use generics to allow more ergonomic, idiomatic DNA code


pospi:
I've been experiencing some friction when attempting to add additional robustness to DNA code, as [outlined in this SO question](https://stackoverflow.com/questions/56471878/how-do-i-coerce-types-of-iterator-elements-to-from-a-newtype).

It would appear that the HDK has been coded against Rust best practises and that [this is what](https://stackoverflow.com/a/40006220/141881) makes implementing such things cumbersome or inefficient. For example, this handling in `GetLinksResult`:

```rust
pub fn addresses(&self) -> Vec<Address> {
    self.links.iter().map(|s| s.address.clone()).collect()
}
```

should be changed to something like:

```rust
pub fn addresses<'a>(&self) -> &[Address] {
    self.links
}
```

which would mean I can simply do:

```rust
let links: Vec<AgentAddress> = links_result.addresses()
```

instead of:

```rust
let links_raw = links_result.addresses();
let links: Vec<AgentAddress> = links_raw.iter().map(|s| s.clone().into()).collect();
```

Seems like an easy efficiency win to take a pass over the codebase and make these changes. Perhaps better done sooner rather than later in case any of it breaks calling code? I wouldn't think  it would, but... Rust n00b :P

pdaoust:
> It would appear that the HDK has been coded against Rust best practises

'against' as in 'holding itself to', or as in 'in contravention of'? ;)

I've seen a recent trend in the HDK towards accepting string params as generic `Into<String>()` params as well, and then the HDK itself can do all the `into()`s rather than the DNA developer. This feels like a win for dev ergonomics too, and something that would be great to standardise across the entire API.

pospi:
Been discussing this a bit with @willemolding. Interestingly, it seems that [iterators are more general than slices](https://stackoverflow.com/a/33189335/141881), so for certain operations that might be a better approach. I realise my example above is not a great one, because `self.links` is not a vector of `Address`, but a vector of `GetLinksResult`. So there is no contiguous memory to pass through efficiently. Perhaps the most efficient pattern for cases like these (where a different memory structure is returned) is to return an iterator rather than a slice... but then most calling code is going to need to call `.collect()` itself, which seems silly.

It's entirely possible we can't cover all bases with singular function types, and might want to have multiple methods- in this case `addresses()` & `addresses_iter()`.

pospi:
[...]

PR #1515, opened 2019-06-14
WIP - HDK standardization

willemolding:
## PR summary

Response to #1499 

The HDK functions were somewhat inconsistent (e.g. functions take ownership or take reference pretty much randomly). This is the start of some proposed changes to the HDK functions. Ideally we should adhere to [this checklist](https://rust-lang-nursery.github.io/api-guidelines/checklist.html)

Only changes to the HDK itself are included for the moment. Once we have fully agreed on the API then I will make the changes to the app spec etc.

TODOs:

- [ ] functions are consistent and always accept references
- [ ] functions don't accept String or Vec but rather use slices (&[<type>]) (or Into<String>) to allow flexibility of calling code
- [ ] Iterators are returned rather than Vecs
- [ ] `get_links_load` variations return the address as well as the entry

### DIFF
--- a/hdk-rust/src/api/get_entry.rs
+++ b/hdk-rust/src/api/get_entry.rs
@@ -32,7 +32,7 @@ use holochain_wasm_utils::api_serialization::get_entry::{
 /// # }
 /// ```
 pub fn get_entry(address: &Address) -> ZomeApiResult<Option<Entry>> {
-    let entry_result = get_entry_result(address, GetEntryOptions::default())?;
+    let entry_result = get_entry_result(address, &GetEntryOptions::default())?;
 
     let entry = if !entry_result.found() {
         None
@@ -48,7 +48,7 @@ pub fn get_entry(address: &Address) -> ZomeApiResult<Option<Entry>> {
 pub fn get_entry_initial(address: &Address) -> ZomeApiResult<Option<Entry>> {
     let entry_result = get_entry_result(
         address,
-        GetEntryOptions::new(StatusRequestKind::Initial, true, false, Default::default()),
+        &GetEntryOptions::new(StatusRequestKind::Initial, true, false, Default::default()),
     )?;
     Ok(entry_result.latest())
 }
@@ -59,7 +59,7 @@ pub fn get_entry_initial(address: &Address) -> ZomeApiResult<Option<Entry>> {
 pub fn get_entry_history(address: &Address) -> ZomeApiResult<Option<EntryHistory>> {
     let entry_result = get_entry_result(
         address,
-        GetEntryOptions::new(StatusRequestKind::All, true, false, Default::default()),
+        &GetEntryOptions::new(StatusRequestKind::All, true, false, Default::default()),
     )?;
     if !entry_result.found() {
         return Ok(None);
@@ -75,10 +75,10 @@ pub fn get_entry_history(address: &Address) -> ZomeApiResult<Option<EntryHistory
 /// The data returned is configurable with the GetEntryOptions argument.
 pub fn get_entry_result(
     address: &Address,
-    options: GetEntryOptions,
+    options: &GetEntryOptions,
 ) -> ZomeApiResult<GetEntryResult> {
     Dispatch::GetEntry.with_input(GetEntryArgs {
         address: address.clone(),
-        options,
+        options: options.to_owned(),
     })
 }
--- a/hdk-rust/src/api/get_links.rs
+++ b/hdk-rust/src/api/get_links.rs
@@ -35,14 +35,14 @@ use holochain_wasm_utils::api_serialization::{
 /// }
 /// # }
 /// ```
-pub fn get_links_with_options<T1: Into<LinkMatch<String>>, T2: Into<LinkMatch<String>>>(
+pub fn get_links_with_options(
     base: &Address,
-    link_type: T1,
-    tag: T2,
+    link_type: LinkMatch<&str>,
+    tag: LinkMatch<&str>,
     options: &GetLinksOptions,
 ) -> ZomeApiResult<GetLinksResult> {
-    let type_re = link_type.into().to_regex_string()?;
-    let tag_re = tag.into().to_regex_string()?;
+    let type_re = link_type.to_regex_string()?;
+    let tag_re = tag.to_regex_string()?;
 
     Dispatch::GetLinks.with_input(GetLinksArgs {
         entry_address: base.clone(),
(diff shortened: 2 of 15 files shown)


hdk-rust/src/api/get_links.rs
```diff
@@ -39,7 +39,7 @@ pub fn get_links_with_options(
     base: &Address,
     link_type: LinkMatch<&str>,
     tag: LinkMatch<&str>,
```
pdaoust: @willemolding I seem to recall you saying Rust's type inference foo isn't strong enough to make these two `LinkMatch<&S>` where `S: Into<String>`, right?

pdaoust: (oh, and you said the most likely case was that people would be using it with literals anyway, right?)

hdk-rust/src/api/keystore.rs
```diff
@@ -24,13 +24,13 @@ pub fn keystore_derive_seed<S: Into<String>>(
     src_id: S,
     dst_id: S,
     context: S,
```
pdaoust: @willemolding I notice that for `call` and `link_entries` you give the three stringish params their own generic type (I presume so that they can each be different implementations of `Into<String>`). Would this be a useful approach for the keystore functions too?

pospi:
I switched my HDK to this branch to see what would happen with our app (which isn't a super detailed example as it's really just got some basic entry & link stuff going on in there atm, but nevertheless...)

So far, this doesn't actually break all that much. Everywhere where I was passing a `String` or `Address` I get an error, but those are very easy to fix because the error message is sane ("consider borrowing here"), and the fix is just adding an `&` in front. The places where it's slightly more verbose are where `JsonString`s are passed in for serialisation, for example what used to be:

```rust
let result: JsonString = call(target_dna_id, zome_name, cap_token, fn_name, RemoteEntryLinkRequest {
    base_entry: base_entry.clone().into(),
    target_entries: target_entries.clone().into(),
}.into()).unwrap();
```

now has to be:

```rust
let result: JsonString = call(target_dna_id, zome_name, &cap_token, fn_name, &(RemoteEntryLinkRequest {
    base_entry: base_entry.clone().into(),
    target_entries: target_entries.clone().into(),
}).into()).unwrap();
```

Which is *slightly* weird syntax, and I've seen others having problems with this and needing to introduce a local variable to get around it because they're not aware of the "anonymous variable reference" shorthand. There may also be another trait that can be implemented for auto-referencing the struct to remove this requirement, ie `&hdk::holochain_core_types::json::JsonString: std::convert::From<RemoteEntryLinkRequest>`. For me it's certainly worth the reduced overhead- IMO we want to reduce memory usage as much as possible, because the way hApps are designed to run the effect of every micro-optimisation is exponentially compounded by running multiple apps.

hdk-rust/src/api/get_links.rs
willemolding: Yeah I spent ages trying to figure this out. You can do it using `LinkMatch<S>` where `S: Into<String>` and it works great unless you pass `LinkMatch::Any`. Then it can't infer `S` even though it doesn't need to and you have to write `LinkMatch<String>::Any` or something like that every time you want to use the `Any` variant.

I decided to go with this way where you only need to do a type conversion when calling it with a `String` which I figure is the least common case.

hdk-rust/src/api/keystore.rs
willemolding: yes! Looks like I missed this one

willemolding:
> I switched my HDK to this branch to see what would happen with our app (which isn't a super detailed example as it's really just got some basic entry & link stuff going on in there atm, but nevertheless...)
> 
> So far, this doesn't actually break all that much. Everywhere where I was passing a `String` or `Address` I get an error, but those are very easy to fix because the error message is sane ("consider borrowing here"), and the fix is just adding an `&` in front. The places where it's slightly more verbose are where `JsonString`s are passed in for serialisation, for example what used to be:
> 
> ```rust
> let result: JsonString = call(target_dna_id, zome_name, cap_token, fn_name, RemoteEntryLinkRequest {
>     base_entry: base_entry.clone().into(),
>     target_entries: target_entries.clone().into(),
> }.into()).unwrap();
> ```
> 
> now has to be:
> 
> ```rust
> let result: JsonString = call(target_dna_id, zome_name, &cap_token, fn_name, &(RemoteEntryLinkRequest {
>     base_entry: base_entry.clone().into(),
>     target_entries: target_entries.clone().into(),
> }).into()).unwrap();
> ```
> 
> Which is _slightly_ weird syntax, and I've seen others having problems with this and needing to introduce a local variable to get around it because they're not aware of the "anonymous variable reference" shorthand. There may also be another trait that can be implemented for auto-referencing the struct to remove this requirement, ie `&hdk::holochain_core_types::json::JsonString: std::convert::From<RemoteEntryLinkRequest>`. For me it's certainly worth the reduced overhead- IMO we want to reduce memory usage as much as possible, because the way hApps are designed to run the effect of every micro-optimisation is exponentially compounded by running multiple apps.

First up I think `call` should accept something more like an `Into<JsonString>` so you can pass structs that implement `DefaultJson` straight in

hdk-rust/src/api/get_entry.rs
```diff
@@ -75,10 +75,10 @@ pub fn get_entry_history(address: &Address) -> ZomeApiResult<Option<EntryHistory
 /// The data returned is configurable with the GetEntryOptions argument.
 pub fn get_entry_result(
     address: &Address,
-    options: GetEntryOptions,
+    options: &GetEntryOptions,
 ) -> ZomeApiResult<GetEntryResult> {
     Dispatch::GetEntry.with_input(GetEntryArgs {
         address: address.clone(),
```
zippy: So I think I'm confused about this PR.  In the [checklist CALLER-CONTROL](https://rust-lang-nursery.github.io/api-guidelines/flexibility.html#c-caller-control) seems to indicate that we shouldn't be doing exactly this.  Am I mis-understanding?

AshantiMutinta:
I like this, interesting but I also think it opens the doors to a lot of smalls.

-Namely, if we are going to be using iterators, I think we could go as far as the persistence which means CAS and EAV should implement those.
-Also I think if we are using references we shouldn't copy them in the method body but use them all the way into the the core code

pdaoust:
@AshantiMutinta 

> Namely, if we are going to be using iterators, I think we could go as far as the persistence which means CAS and EAV should implement those.

That's been on my mind too. Opens up lots of opportunities for persistence implementations to make wise decisions about how to do things in a performant way. And eventually devs will be clamoring for filtered queries and performance improvements thereupon. But I'm ignorant about the WASM boundary and whether it scuttles all performance optimisations.

PR #1516, opened 2019-06-15
Signals MVP

lucksus:
## PR summary

This adds an HDK function:
```
hdk::emit_signal("my signal", args.into());
```
which will send `Signal::UserSignal`s over the signal channel and websocket interface to the UI where it can be listened to like so:

```javascript
const { onSignal } = await this.webClientConnect({url})

onSignal((msg) => {
  console.log(msg.signal) // -> { signal_type: 'User', name: 'test-signal', arguments: '{"message":"test message"}' }
})
```

So it is a bare-bones signal implementation without all the checks and DNA defintions that we were hinting towards in the [Signals ADR](https://github.com/holochain/holochain-rust/blob/develop/doc/architecture/decisions/0013-signals-listeners-model-and-api.md).

But it should be enough to get Basic-Chat rolling with a UI that refreshes automatically and without polling :) (pinging @philipbeadle and @willemolding)

Is using [some changes to Diorama](https://github.com/holochain/diorama/pull/10) to actually test if signals arrive at the interface client.

## followups

* Use this in Basic-Chat or Hylo :)
* Implement the rest of Signals as described in the ADR which will make it possible to introspect about signals and set them up as bridges etc.

## changelog


app_spec/test/test.js
```diff
@@ -847,4 +847,13 @@ scenario('request grant', async (s, t, { alice, bob }) => {
     t.equal(result.Ok, grants.Ok[0])
 })
 
+  scenario('emit signal', async (s, t, { alice }) => {
+
+    const result = await alice.callSync("simple", "test_emit_signal", {message: "test message"})
+    // TODO: Here we need some way to read all the received UserSignals...
```
lucksus: @maackle, I've already spotted where Diorama processes signals. Can you point me into a direction where and how to add signals to the object that is `alice` here? I guess that is a `DnaInstance`? How can I access all those from https://github.com/holochain/diorama/blob/65dd181cd72d15813d08272631426be5acf37921/src/diorama.ts#L61.

app_spec/zomes/simple/code/src/lib.rs
```diff
@@ -63,6 +63,10 @@ pub fn handle_get_my_links(base: Address) -> ZomeApiResult<GetLinksResult> {
     hdk::get_links(&base, LinkMatch::Exactly("authored_posts"), LinkMatch::Any)
 }
 
+pub fn handle_test_emit_signal(message: String) -> ZomeApiResult<()> {
+    hdk::emit_signal("test-signal", JsonString::from_json(&format!("{{\"message\": {}}}", message)))
```
willemolding: I think a good improvement would be to use a struct that implements DefaultJson rather than this string formatting. 

e.g. 
```rust
#[derive(..., DefaultJson)]
struct SignalPayload {
   message: String
}
...
let payload = SignalPayload{message}
hdk::emit_signal("test-signal", payload.into())
```

Only because people are going to copy what is in the app spec and I don't think this is the best practice.

willemolding:
This is amazing. So excited to play around with it.

app_spec/zomes/simple/code/src/lib.rs
lucksus: Yes, good point, will add that!

PR #1528, opened 2019-06-18
Add stress test with 10 agents

maackle:
## PR summary

Adds simple Diorama scenario where 10 nodes are created, one is singled out as an anchor, and every node places a link on that anchor. Then the anchor attempts to retrieve all links and verify that there are 10 of them.

This test is flaky. Usually one of the zome function calls simply hangs up and never returns, though I have also seen a case where the zome calls do return, but the hachiko waiter never receives all the signals it needs to proceed.

## changelog

### DIFF
--- a/app_spec/test/many-agent-tests.js
+++ b/app_spec/test/many-agent-tests.js
@@ -1,0 +1,28 @@
+module.exports = scenario => {
+
+
+scenario('can run a test with many agents', async (s, t, instanceMap) => {
+  const nodes = Object.values(instanceMap)
+  const anchor = nodes[0]
+
+  t.ok(anchor.agentAddress)
+
+  for (const [name, node] of Object.entries(nodes)) {
+    await node.call('simple', 'create_link', {
+      base: anchor.agentAddress,
+      target: node.agentAddress,
+    })
+  }
+  await s.consistent()
+  const links = await anchor.call('simple', 'get_my_links', {
+    base: anchor.agentAddress,
+  })
+
+  console.log('linx', links.Ok.links)
+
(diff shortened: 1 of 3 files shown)


app_spec/test/many-agent-tests.js
```diff
@@ -0,0 +1,28 @@
...
+
+
+scenario('can run a test with many agents', async (s, t, instanceMap) => {
+  const nodes = Object.values(instanceMap)
+  const anchor = nodes[0]
+
+  t.ok(anchor.agentAddress)
+
+  for (const [name, node] of Object.entries(nodes)) {
+    await node.call('simple', 'create_link', {
```
lucksus: Should this be
```suggestion
    await node.callSync('simple', 'create_link', {
```
or is the new Diorama call function sync already and callSync just an alias?

maackle: @lucksus the waiting happens now through `s.consistent()`. The `callSync` in the current app spec test is actually syntactic sugar added by the `backwardsCompabilityMiddleware`, used to minimize the changes needed. The intention of this test is to have each create_link happen as quickly as possible, though, without waiting in between.

jdeepee:
Has there been any more progress made with the cause of this issue? Im looking to run multi agent tests of my application as well as stress tests involving a large number of agents. Would be great to get this resolved. I even sometimes have issues calling tests with two agents and have to re-run the tests many times before they all the zome functions get called.

lucksus:
Closing in favour of: https://github.com/holochain/holochain-rust/pull/1755

PR #1533, opened 2019-06-19
Updating the bootstrap process for the conductor while using dpki

zo-el:
## PR summary
This is just a suggestion pr. Initially created for me to move forward in testing the deepkey app.
Problems faced
### 1. Issue during the boot process, 
  - The `dpki_bootstrap` fn was called which needed the deepkey happ to be running so that it could make a zome call 
  - So I initially start the deepkey happ and then shut it down after the bootstrap process is done

### 2. Issue in the `dpki_instance`
  - Here while calling the `is_initialized` zome fn, the return was expecting a bool after running `result.try_into()?`
  - But it was returning a struct in the form 
  ```
  let result: Result<bool, HolochainError> = result.try_into()?;
  ```
## testing/benchmarking notes

## followups

## changelog

- [ ] this is a code change that affects some consumer (e.g. zome developers) of holochain core so it is added to the CHANGELOG-UNRELEASED.md (linked above), with the format `- summary of change [PR#1234](https://github.com/holochain/holochain-rust/pull/1234)`
- [ ] this is not a code change or doesn't affect anyone outside holochain core development


conductor_api/src/dpki_instance.rs
```diff
@@ -74,7 +80,8 @@ impl DpkiInstance for Holochain {
             DPKI_TRAIT_FN_IS_INITIALIZED,
             params,
         )?;
-        let result: bool = result.try_into()?;
-        Ok(result)
+        let result: IsInitializedResult = result.try_into()?;
```
zippy: I don't thing you need to create a special Struct, you can just `try_into` a result:
```suggestion
        let result: Result<bool, HolochainError> = result.try_into()?;
```
This may cause confusions with different levels of errors because the issue is that the try_into itself may produce an err (hence the ?), and the `call` returns a json-encoded result as well, which may be an err.

conductor_api/src/dpki_instance.rs
```diff
@@ -40,6 +40,12 @@ fn dpki_cap_request(
     ))
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, DefaultJson)]
```
zippy: I think you don't need this anymore right?

conductor/src/main.rs
```diff
@@ -54,6 +54,27 @@ fn main() {
...
+                    println!("Starting dpki instances...");
+                    conductor
+                        .start_dpki_happ_instance()
+                        .expect("Could not start dpki happ instance!");
+
+                    conductor
+                        .dpki_bootstrap()
+                        .expect("Could not bootstrap dpki!");
+
+                    // Need to stop the dpki instance for now.
```
zippy: Instead of stopping the DPKI instance, how about passing in a boolean "dpki_instance_already_started" that skips starting it?  Would that work?

conductor_api/src/conductor/base.rs
```diff
@@ -1003,8 +1022,11 @@ impl Conductor {
 
     /// bootstraps the dpki app if configured
     pub fn dpki_bootstrap(&mut self) -> Result<(), HolochainError> {
+        println!("Starting DPKI Bootstrap ...");
```
lucksus:
```suggestion
        notify("Starting DPKI Bootstrap ...");
```

conductor_api/src/conductor/base.rs
```diff
@@ -1003,8 +1022,11 @@ impl Conductor {
...
+        println!("Cheking if conductor is using DPKI : {}", self.using_dpki());
```
lucksus:
```suggestion
        notify("Checking if conductor is using DPKI : {}", self.using_dpki());
```

conductor/src/main.rs
lucksus: I would suggest we change `Conductor::start_all_instances()` to ignore *already-running* errors and then leave the DPKI instance running here...

conductor_api/src/dpki_instance.rs
zo-el: :+1:

zo-el:
@zippy @lucksus, Could you take a quick look.
I have made all the changes suggested

conductor_api/src/holochain.rs
```diff
@@ -182,9 +182,6 @@ impl Holochain {
 
     /// activate the Holochain instance
     pub fn start(&mut self) -> Result<(), HolochainInstanceError> {
-        if self.active {
-            return Err(HolochainInstanceError::InstanceAlreadyActive);
-        }
```
lucksus: I don't think we should remove this here. Rather handle the error where this is called within `Conductor`...

conductor/src/main.rs
```diff
@@ -54,6 +54,7 @@ fn main() {
                     "Successfully loaded {} instance configurations",
                     conductor.instances().len()
                 );
+
```
lucksus:
```suggestion
```

zo-el:
Hi @lucksus, any Idea why the app-spec-tests fail?

zo-el:
Hi @zippy @lucksus 
Can we get this branch merged into develop?

PR #1534, opened 2019-06-19
Hdk encrypt/decrypt functions

AshantiMutinta:
## PR summary

In this Pr we implement and refactor crypto functions in the hdk

## testing/benchmarking notes

## followups

## changelog


AshantiMutinta:
More refactors need to be done to DRY up these functions

conductor_api/src/holo_signing_service.rs
```diff
@@ -2,31 +2,34 @@ use boolinator::Boolinator;
...
     payload: &String,
-    signing_service_uri: &String,
+    service_uri: &String,
 ) -> Result<String, HolochainError> {
     let body_json = json!({"agent_id": agent_id.address(), "payload": payload});
     let client = reqwest::Client::new();
-    let url = reqwest::Url::parse(signing_service_uri).map_err(|_| {
+    let url = reqwest::Url::parse(service_uri).map_err(|_| {
         HolochainError::ConfigError(format!(
             "Can't parse signing service URI: '{}'",
```
zippy:
```suggestion
            "Can't parse service URI: '{}'",
```

core/src/conductor_api.rs
```diff
@@ -7,34 +7,54 @@ use std::{
...
+    let response = JsonRpc::parse(&response)?;
+
+    match response {
+        JsonRpc::Success(_) => Ok(String::from(
+            response.get_result()?[request_reponse.1].as_str()?,
+        )),
+        JsonRpc::Error(_) => Err(HolochainError::ErrorGeneric(serde_json::to_string(
+            &response.get_error()?,
+        )?)),
+        _ => Err(HolochainError::ErrorGeneric("Signing failed".to_string())),
```
zippy: error message should include the request response so that it's more generic

core/src/conductor_api.rs
```diff
@@ -7,34 +7,54 @@ use std::{
...
+    let request = format!(
+        r#"{{"jsonrpc": "2.0", "method": "agent/{}", "params": {{"payload": "{}"}}, "id": "{}"}}"#,
+        request_reponse.0,
+        payload,
+        ProcessUniqueId::new(),
+    );
+
+    let response = handler
+        .handle_request_sync(&request)
+        .ok_or("Conductor sign call failed".to_string())?;
```
zippy: error message should include the request response so that it's more generic

core/src/nucleus/ribosome/api/crypto.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    // deserialize args
+    let args_str = runtime.load_json_string_from_args(&args);
+
+    let crypto_args = match CryptoArgs::try_from(args_str.clone()) {
+        Ok(entry_input) => entry_input,
+        // Exit on error
+        Err(_) => {
+            context.log(format!(
+                "err/zome: invoke_sign failed to deserialize SignArgs: {:?}",
```
zippy:
```suggestion
                "err/zome: invoke_crypto failed to deserialize SignArgs: {:?}",
```

dpki/src/keypair.rs
```diff
@@ -174,6 +174,33 @@ impl EncryptingKeyPair {
     }
 
     // TODO: Encrypt and decrypt functions
```
zippy: remove outdated comment?

test_utils/src/mock_signing.rs
```diff
@@ -72,12 +72,119 @@ pub fn mock_signer(payload: String, agent_id: &AgentId) -> String { TEST_AGENT_K
         .unwrap()
 }
 
+/// This is a local mock for the `agent/sign` conductor API function.
```
zippy:
```suggestion
/// This is a local mock for the `agent/encrypt` conductor API function.
```

test_utils/src/mock_signing.rs
```diff
@@ -72,12 +72,119 @@ pub fn mock_signer(payload: String, agent_id: &AgentId) -> String { TEST_AGENT_K
...
+            let mut message_signed = keybundle.encrypt(&mut message).expect("Mock signing failed.");
+            let message_signed = message_signed.read_lock();
+
+            // Return as base64 encoded string
+            base64::encode(&**message_signed)
+        })
+        .unwrap()
+}
+
+/// This is a local mock for the `agent/sign` conductor API function.
```
zippy:
```suggestion
/// This is a local mock for the `agent/decrypt` conductor API function.
```

dpki/src/keypair.rs
```diff
@@ -174,6 +174,33 @@ impl EncryptingKeyPair {
...
+        let mut nonce = SecBuf::with_insecure(lib3h_sodium::aead::NONCEBYTES);
+        let mut cipher = SecBuf::with_insecure(data.len() + lib3h_sodium::aead::ABYTES);
+        lib3h_sodium::aead::enc(data, &mut self.private, None, &mut nonce, &mut cipher)?;
+        Ok(cipher.clone())
+    }
+
+    /// encrypt some arbitrary data with the signing private key
+    /// @param {SecBuf} data - the data to sign
+    /// @return {SecBuf} signature - Empty SecBuf to be filled with the signature
+    pub fn decrypt(&mut self, cipher: &mut SecBuf) -> HcResult<SecBuf> {
```
neonphog: I think we need the output to be a parameter, so that the implementer can decide if the data need to be protected or not.

something like:

```rust
pub fn decrypt(&mut self, cipher: &mut SecBuf, result: &mut SecBuf) -> HcResult<()> {
```

``` rust
let res = SecBuf::with_secure(cipher.len() + ABYTES);
store.decrypt(&mut cipher, &mut res)?;
```

For symmetry, we could also make the output of encrypt be a param, but that's not strictly needed, since the encrypted data shouldn't need to be protected.

dpki/src/keypair.rs
```diff
@@ -174,6 +174,33 @@ impl EncryptingKeyPair {
     }
 
     // TODO: Encrypt and decrypt functions
+
+    /// encrypt some arbitrary data with the signing private key
+    /// @param {SecBuf} data - the data to sign
+    /// @return {SecBuf} signature - Empty SecBuf to be filled with the signature
+    pub fn encrypt(&mut self, data: &mut SecBuf) -> HcResult<SecBuf> {
+        let mut nonce = SecBuf::with_insecure(lib3h_sodium::aead::NONCEBYTES);
```
neonphog: It is not secure to encrypt two different messages with the same nonce.

We've got two options.

1 - If we want to keep the simple API, we could randomize the nonce here, and increase the cipher length by `NONCEBYTES`, appending it to the output, so it can be read / used by the decrypt function

or

2 - We could promote nonce to a parameter in both encrypt and decrypt so that the implementor can be put in charge of doing the randomizing / tracking.

(or we could provide both apis so people can choose which one to use?)

dpki/src/keypair.rs
AshantiMutinta: The reason I implemented it this way was because we have a few unsafe spots deep in the encryption code and thought it would be better to reduce the area of mutability by just returning the new encrypted data. That way that way they have a new cleaner block of data that minimizes the side effects

dpki/src/keypair.rs
AshantiMutinta: I like the idea of coming back and implementing with nonce because I think eric and I talked about this and decided it wasn't really needed at this point. But definitely need nonce to be more secure

zippy: Yep, lets just add in the randomization for now, and then open up the API later.

doc/holochain_101/src/zome/api_functions.md
```diff
@@ -82,6 +82,22 @@ Not yet available.
 
 Enables the signing of some piece of data, with the private keys associated with the acting agent.
 
+### Encrypt
+
+Canonical name: `encrypt`
+
+Not yet available.
```
lucksus: Hm? Isn't this PR implementing it??

doc/holochain_101/src/zome/api_functions.md
```diff
@@ -82,6 +82,22 @@ Not yet available.
...
+
+Enables the encryption of some piece of data, with the private keys associated with the acting agent.
+
+### Decrypt
+
+Canonical name: `decrypt`
+
+Not yet available.
```
lucksus: Dito.

dpki/src/keypair.rs
lucksus: +1 for random nonce now and extra parameter later.

dpki/src/keypair.rs
neonphog: I don't understand this reasoning. The unsafe code is executed in both cases, either by us or by the user of this api. We need to make sure that is solid through whatever testing is needed, but is also not related to my comment above.

The current code does not allow the user to memory protect the results of the decryption. If they are decrypting sensitive data from somewhere, they may need to use `with_secure` instead of `with_insecure` which the current api does not allow.

AshantiMutinta: I guess the comment was related to the output type itself, I would still like to output a new piece of data just to be a bit closer to determinism but i guess it is okay since rust compiler will always take care of shared mutability. 

How do you feel about an enum just to make it more rust style?

decrypt(&mut self, cipher,CryptoOptions::Secure) -> HcResult<String>

and crypto options could be Secure and InSecure?

neonphog: I swear I'm not trying to be argumentative, lol : ) But, I don't see how the determinism is different in either case, and I don't like the dissonance of implementing the HDK api in a different manner than the underlying crypto api.

I feel like I'm missing some context for why this is a problem. Is this specifically because of the unsafe copy in the write function: `std::ptr::copy(data.as_ptr(), (**b).as_mut_ptr().add(offset), data.len());`?

I misspoke earlier when I said that would happen in either case. You are correct, it doesn't get called when creating a new SecBuf. That code was originally a for-loop copy of the bytes, but that can't take advantage of memory block size optimizations in the processor, the ptr copy is far more efficient, and there is an explicit bonds check above it, so there shouldn't be any real unsafety (though if you think of additional unit tests needed, please feel free to add them : )

Ultimately, though, even rust std code uses mem copy techniques at its lowest levels, if there's an efficiency reason to do it, I don't feel like we should fear diving in.

I'm just making things up at this point... please let me know if I'm totally off the mark. Thanks!

AshantiMutinta: I have made the changes though and even though I have my opinions on rust and unsafety, you have worked on this side of the code more and I trust your expertise!

Issue #43, opened 2018-06-23
address redux state model bloat


zippy:
In the current implementation of #71 each function call results in two additions to the state via the actions:     ExecuteZomeFunction(FunctionCall),
ReturnZomeFunctionResult(FunctionResult),

this means the state object will grow in size for each function call, which means that eventually it has to get cleaned up or garbage collected.

thedavidmeister:
@zippy do you know off the top of your head if this still happening? i remember someone commenting this week that our memory usage seems to grow over time

PR #1540, opened 2019-06-21
lib3h config option

zippy:
## PR summary

This PR implements the use of Lib3h configuration in core, and stubs out the ability to choose either lib3h or n3h as the p2p networking engine.  See: [lib3h:#43](https://github.com/holochain/lib3h/issues/43)

*Breaking Change*: You must add `type="n3h"` for current config files to work.  eg:
e.g. current n3h based configurations need to look something like like:
```
...

    [network]
    type="n3h"
    bootstrap_nodes = ["wss://192.168.0.11:64519/?a=hkYW7TrZUS1hy-i374iRu5VbZP1sSw2mLxP4TSe_YI1H2BJM3v_LgAQnpmWA_iR1W5k-8_UoA1BNjzBSUTVNDSIcz9UG0uaM"]
    n3h_persistence_path = "/Users/cnorris/.holochain/n3h_persistence"
    networking_config_file = "/Users/cnorris/.holochain/network_config.json"
    n3h_log_level = "d"
```

## changelog


net/src/p2p_network.rs
```diff
@@ -90,8 +97,8 @@ impl P2pNetwork {
         let connection =
             NetConnectionThread::new(wrapped_handler, worker_factory, None).map_err(|e| {
                 format_err!(
-                    "Failed to obtain a connection to a p2p network module w/ config: {}: {}",
-                    p2p_config.as_str(),
+                    "Failed to obtain a connection to a p2p network module w/ config: {}: ",
+                    //   p2p_config.as_str(),
```
ddd-mtl: dead code

zippy: fixed

willemolding:
The tests that are failing are these ones
```
can_get_all_links_of_any_tag_or_type
get_links_with_non_existent_tag_returns_nothing
get_links_with_non_existent_type_returns_nothing
```
Interestingly these are tests I have had issues with before when I didn't give the test instance a unique network name
```rust
test_instance(dna, netname)
                              ^^^^^^^^ - This was not unique
```
In that case they failed with the exact same timeout errors.

I am wondering if the network name is no longer being handled correctly resulting in conflicting in-memory networks.

zippy:
That's exactly what I thought, and went and traced through the netname stuff but couldn't find things out of place.  An now it looks like things pass.  I can't see how your commit could have fixed it.

willemolding:
I'm pretty sure it was using the `json!` macro incorrectly previously. To my knowledge you can't pass a formatted JSON string into it you have to use their macro/json style. I wonder if this meant that the netname was  'null' in every case

PR #1556, opened 2019-06-27
Get validation package defs for links only from the link type

lucksus:
...instead of fetching base and target and infer it from their types

## PR summary

In #1514 I've simplified the validation process for entry types that don't need the author's source chain for validation (`ValidationPackageDefinition::Entry`) which makes it possible to gossip these entries without caching validation packages.

That helped for custom entry types but not for links since links' entry types used to be inferred through their base and target types. But since we added explicit link types recently, we don't need to depend on base and target (which had to be retrieved to tell if a link can be validated locally).

These changes simplify this by just searching for a link definition that matches the type string.

**Beware:** this introduces the assumption that link entry type names are unique across the whole DNA - which is a reasonable assumption but one we did not buy in with the old code that inferred the type through base, target and tag.

## Clean-up

While working on this PR and making tests pass again, I found that we were not using the pending validation loop for anything but newly created entries and links - not for entry updates, removals nor link removals. I've added those in this PR and harmonized the signatures of all workflow functions.

## changelog

### DIFF
--- a/core/src/nucleus/ribosome/callback/links_utils.rs
+++ b/core/src/nucleus/ribosome/callback/links_utils.rs
@@ -60,72 +60,11 @@ pub struct LinkDefinitionPath {
     pub link_type: String,
 }
 
-/// This function tries to find the link definition for a link given by base type,
-/// link type and target type.
-///
-/// It first looks at all "links_to" definitions in the base entry type and checks
-/// for matching link type and target type.
-///
-/// If nothing could be found there it iterates over all "linked_form" definitions in
-/// the target entry type.
+/// This function tries to find the link definition for a link given by link type.
+/// It assumes that link type names are unique and thus just iterates through
+/// zomes, entry types and their links and returns the first match.
 ///
 /// Returns a LinkDefinitionPath to uniquely reference the link definition in the DNA.
-pub fn find_link_definition_in_dna(
-    base_type: &EntryType,
-    link_type: &String,
-    target_type: &EntryType,
-    context: &Arc<Context>,
-) -> Result<LinkDefinitionPath, HolochainError> {
-    let dna = context.get_dna().expect("No DNA found?!");
-    match base_type {
-        EntryType::App(app_entry_type) => dna
-            .get_entry_type_def(&app_entry_type.to_string())
-            .ok_or(HolochainError::ErrorGeneric(String::from(
-                "Unknown entry type",
-            )))?
-            .links_to
-            .iter()
-            .find(|&link_def| {
-                link_def.target_type == String::from(target_type.clone())
-                    && &link_def.link_type == link_type
-            })
-            .and_then(|link_def| {
-                Some(LinkDefinitionPath {
-                    zome_name: dna.get_zome_name_for_app_entry_type(app_entry_type)?,
-                    entry_type_name: app_entry_type.to_string(),
-                    direction: LinkDirection::To,
-                    link_type: link_def.link_type.clone(),
-                })
-            }),
-        _ => None,
-    }
-    .or(match target_type {
-        EntryType::App(app_entry_type) => dna
-            .get_entry_type_def(&app_entry_type.to_string())
-            .ok_or(HolochainError::ErrorGeneric(String::from(
-                "Unknown entry type",
-            )))?
-            .linked_from
-            .iter()
-            .find(|&link_def| {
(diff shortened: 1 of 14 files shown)


core/src/nucleus/validation/link_entry.rs
```diff
@@ -33,19 +33,14 @@ pub async fn validate_link_entry(
         }
     };
     let link = link.link().clone();
-    let (base, target) = links_utils::get_link_entries(&link, context).map_err(|_| {
+    let (_base, _target) = links_utils::get_link_entries(&link, context).map_err(|_| {
```
lucksus: So, we don't need base and target anymore to find out what type of link this is, since we have explicit link types now.

We also don't put base and target into the `ValidationData` struct when validating links, so the validation code would (currently) need to explicitly get it anyway if the link validation is dependent of something in the base.

Question: should we either:
* add base and target to `LinkValidationData`
* or remove it completely and also don't treat it as a dependency, which means we could write DNAs that validate a link without having the base
?
@zippy, @artbrock, @maackle

lucksus: I would suggest we keep the hardcoded dependency (which will sub-consciously trigger the pending validation logic) to base and target but add both entries to `LinkValidationData` so that they are available in the validation code, which means DNA devs don't have to `hdk::get_entry` in there to get the linked entries.

lucksus: If we don't want to do that, we could also make the pending validation logic more accessible to DNA devs and have them explicitly say that they depend on base/target, by adding some hdk sugar that returns `ValidationError::UnresolvedDependencies` from the validation callback. (i.e. implementing what we've discussed months ago, @maackle, @artbrock).

Still, I would say we keep hadcoded base/target dependency for links but add them to the validation data now, and maybe take the next step later.

lucksus: Ok, nevermind. We do need base and target for the sub-conscious check of the link valiation anyway. Added that now. I would still suggest to add base and target to `LinkValidationData` - but in another PR.

core/src/nucleus/ribosome/callback/links_utils.rs
```diff
@@ -60,68 +60,45 @@ pub struct LinkDefinitionPath {
...
-            .and_then(|link_def| {
-                Some(LinkDefinitionPath {
-                    zome_name: dna.get_zome_name_for_app_entry_type(app_entry_type)?,
-                    entry_type_name: app_entry_type.to_string(),
-                    direction: LinkDirection::To,
-                    link_type: link_def.link_type.clone(),
-                })
-            }),
-        _ => None,
+    for (zome_name, zome) in dna.zomes.iter() {
```
willemolding: Not so big on the nested for loops. I made a different implementation using iterators #1557 which is probably more idiomatic but not sure which one you think is more readable

AshantiMutinta: I agree about the nested loop here, I think we should incorperate refactors. We could use iter().skip_while() for some of these operations as well

core/src/workflows/hold_entry_update.rs
```diff
@@ -39,12 +57,34 @@ pub async fn hold_update_workflow<'a>(
...
         validation_data,
         &context
-    ))?;
+    ))
+    .map_err(|err| {
+        if let ValidationError::UnresolvedDependencies(dependencies) = &err {
+            context.log(format!("debug/workflow/hold_update: Entry update could not be validated due to unresolved dependencies and will be tried later. List of missing dependencies: {:?}", dependencies));
+            add_pending_validation(
+                entry_with_header.to_owned(),
+                dependencies.clone(),
```
AshantiMutinta: I think we could also DRY out this pending validation operation

lucksus: What hinders just putting it in a function is that the `ValidationWorkflow` specilization as well as the log messages depend on the workflow this is in and are different for each...

lucksus:
@willemolding, @AshantiMutinta, I've commented on Willem's PR that I find the nested for-loops (in this case) much better readable than the alternative. Also because they reflect the the structure of the DNA that is parsed over - much better than the functional call chains. But I would have just merged it if it would pass tests. Just reran them to see if it just was our flaky app-spec tests. If that branch does not pass CI, are you ok with me merging this? We could still merge the refactoring once it passes..

PR #1557, opened 2019-06-27
Simple link validation package rev1

willemolding:
Rewrites some of the nested for loops using iterators. I think this probably more Rusty but totally up to you which one you think is more readable


lucksus:
Hm, interesting. In this case, I actually find the nested for-loops much better readable/comprehensible than those filter_map/chain calls... Tests were failing, just reran to see if it was a remainder of the recent flaky-ness.

PR #1558, opened 2019-06-28
Entry def properties

willemolding:
## PR summary

In order for zomes to become more self describing a properties field has been added to the entry definitions. This replaces the description field and rather than just being a string it is intended to hold a JsonString so that entry definitions can have complex properties.

An example of this might be to include indexing hints, UI rendering hints, example entries as well as a description field.

The properties of an entry definition can be retrieved using the new  `entry_type_properties` HDK function. This could allow bridging zomes to inspect entries to know how to handle them.

In order for this not to be a breaking change I have field name in the `entry!` macro as `description:`. Since we are planning to re-implement entry definitions in the macro HDK soon I figure it makes sense to not make it a breaking change until then,

## changelog


hdk-rust/src/api/entry_type_properties.rs
```diff
@@ -0,0 +1,27 @@
...
+}
+
+// Returns the properties defined with an entry type
+// It is encouraged to using JSON to encode structured properties
+// with an entry
+pub fn entry_type_properties(name: &EntryType) -> ZomeApiResult<JsonString> {
+    let mut zd = ZomeDefinition::new();
+    unsafe { zome_setup(&mut zd) };
+
+    let entry_def = zd.entry_types.iter().find(|elem| {
```
willemolding: Is iterating through the zome def the best way to find the given entry? Not sure about this one

PR #1559, opened 2019-07-01
bugfix: CLI handling of .json files in the project root

willemolding:
## PR summary

Previously the CLI would use the first file it found in the root directory as the template for the .dna.json. Other .json files found would be converted to base64 and included in the .dna.json.

This lead to difficult to find/debug issues if for some reason you added another .json file to the project root and it became random as to which one would be chosen as the template.

This PR adds a check, if no .json files are in the root it starts packaging from an empty object. If it finds one it uses that as the template object and if it finds more than one it notifies the user of the ambiguity.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -23,6 +23,9 @@ extern crate serde_json;
 extern crate ignore;
 extern crate rpassword;
 
+// #[cfg(test)]
+// extern crate assert_cmd;
+
 mod cli;
 mod config_files;
 mod error;
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -25,7 +25,6 @@ extern crate rpassword;
 
 // #[cfg(test)]
 // extern crate assert_cmd;
-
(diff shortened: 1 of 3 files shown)


cli/src/main.rs
```diff
@@ -23,6 +23,9 @@ extern crate serde_json;
 extern crate ignore;
 extern crate rpassword;
 
+// #[cfg(test)]
```
lucksus:
```suggestion
```

cli/src/main.rs
```diff
@@ -23,6 +23,9 @@ extern crate serde_json;
...
+// extern crate assert_cmd;
```
lucksus:
```suggestion
```

cli/src/main.rs
```diff
@@ -23,6 +23,9 @@ extern crate serde_json;
...
+
```
lucksus:
```suggestion
```

PR #1563, opened 2019-07-01
Make calls to validation callback consistent

willemolding:
## PR summary

In the current implementation when validating a new entry the local commit function is called twice. Once as a result of the `author_entry` workflow and once as part of the `hold_entry` workflow. 

This in itself is not the problem. What is problematic is that the first time the validation function is called the new entry is NOT included in the local chain passed in the validation_data and the second time it is. Furthermore when the other agents who must hold this entry run the validation callback I believe the entry IS included in the local chain then.

My issue with this is that it is possible to write a validation callback where a public entry can be added to the local chain but will not pass validation to be held by you or any other agent. This results in essentially an invalid state.

Possible solutions:
- 1. Optimistically add the new entry to the local chain data passed to the validation callback when validating if the entry should be added to the local chain. This makes the data passed to the function consistent in both cases.

- 2. Ensure that the new entry is NOT included in the piece of local chain you send to agents when they request a validation package. I guess this falls into the realm of As-At

- 3. Leave it as it and document the fact that you should not write validation functions that might pass/fail based on additional things being included in the local chain (this came up in the dev camp because I was using the most recent entry in the local chain to validate which agents turn it is to make a move. If was fairly easy to workaround by making an explicit check if the top entry is the current entry being validated)


thedavidmeister:
@willemolding my vote is for "as at" because my personal experience is that including time/order in data is pretty much the only way to make things reliable in these types of situations

if multiple function invocations are supposed to be pure and reliable then we better feed them the same data across invocations :sweat_smile:

core/src/workflows/author_entry.rs
```diff
@@ -34,6 +34,7 @@ pub async fn author_entry<'a>(
     ));
 
     // 1. Build the context needed for validation of the entry
+    // Should we add the current entry to the chain in the validation package?
```
willemolding: This would be option i.

core/src/nucleus/actions/build_validation_package.rs
```diff
@@ -125,6 +125,7 @@ pub async fn build_validation_package<'a>(
                         Entry => ValidationPackage::only_header(entry_header),
                         ChainEntries => {
                             let mut package = ValidationPackage::only_header(entry_header);
+                            // What if we make another function all_public_chain_entries_as_of(entry) which returns the local chain up to and not including the given entry?
```
willemolding: This would be option ii.

lucksus:
Hm, have to think about this more, but I don't like altering the chain (as seen in the validation call) for either case, so I would prefer neither doing i) nor ii).

I guess we will need as-at sooner rather than later anyway (since there could be other entries in the source that got created after this one, but new DHT nodes are validating an old entry). But, until we have that, I would go with option iii). Also note that we do pass the information about the entry lifecylce in which this validation runs into the callback: https://github.com/holochain/holochain-rust/blob/b70d98307926138a30246b73e0c49ec497b03c1b/core_types/src/validation.rs#L129. So if an app needs to distinguish authoring (which happens before the entry is added to the chain) from DHT nodes holding it, it can already.

pjkundert:
I ran into the difference between the lifecycle `Dht` and `Chain` when implementing the validators for HoloFuel.  In order to compute a trial ledger balance, if lifecycle is `Chain` (ie. we're validating the Entry *before* it is committed to the local source-chain), I had to manually add the candidate Entry to the full source-chain being validated.

I was wondering, at the time, whether it makes sense to simply always add the candidate Entry being  validated, onto the end of the full source chain.  After all, that's the source-chain "state" being validated -- its either valid or it isn't.  If it is not validated, the provided candidate source-chain will not be committed to storage.

As for option iii), I'm pretty sure that a DHT Validator should only be passed a source-chain up to and including the Entry that it is validating.  If every validator has to be resilient to arbitrarily long chains that contain the candidate Entry somewhere within it (or maybe at the end in `Dht` lifecycle, or maybe not there at all, if the  lifecycle is `Chain`), it'll add significant complexity to the validation logic -- pretty much where we should be minimizing complexity...

willemolding:
This interesting I didn't know about the lifecyle enum being passed to the validation callback. I will use that to make my game code more explicit.

That said I agree with Perry that the validation callbacks should ideally receive the exact same chain data in every case (local, remote, remote at a later time). The local chain either up to OR up to and including the entry being validated. It seems a case of making impossible states impossible (entries that are valid sometimes and not other times) rather than forcing the developer to consciously code to avoid it. I can't think of a case where this isn't what you would want.

lucksus:
Hm, I think you might be right that this (getting the exact same chain in every validation call) is what always makes sense.

I would translate that to: yes, we need the *as-at* logic (getting a source chain at a specific hash).
I would suggest then that we always provide the source chain *without* the new entry, just as it were before that entry was authored (as-at the top chain hash that the zome call saw when the new entry was committed).

The reason that the entry is included in the DHT validation is that currently we don't have any of this. DHT validation callbacks that happen in 1 month after the entry was authored would potentially see lots of new entries after the one being validated.

lucksus:
Hm, thinking more about this: the building of validation packages is a rather specific and confined use-case (and code). It would be easy to spearhead any as-at stuff with this by always creating the source chain *as at <previous entry header>*.

I guess I have change my vote to: let's do ii)
:)

@zippy, what say you?

willemolding:
> Hm, thinking more about this: the building of validation packages is a rather specific and confined use-case (and code). It would be easy to spearhead any as-at stuff with this by always creating the source chain _as at previous-entry-header_.
> 
> I guess I have change my vote to: let's do ii)
> :)
> 
> @zippy, what say you?

Yeah I think a naive implementation of this would be very trivial. I'll commit something for review

willemolding:
Ok that seems to do what we want. I propose we also remove the lifecycle field from ValidationData so that the validation functions take the exact same data in both cases and cannot possibly fail one and not the other

core/src/nucleus/actions/build_validation_package.rs
```diff
@@ -332,13 +355,82 @@ mod tests {
...
+
+    #[test]
+    fn test_all_chain_headers_before_header_entry_dht_validation() {
+        let (_instance, context) = instance(None);
+        // entry is added to the local chain
+        let chain_header = commit(test_entry_package_chain_full(), &context);
+        let headers = all_chain_headers_before_header(&context, &chain_header);
+        // entry should not appear in the validating chain
+        assert_eq!(headers.contains(&chain_header), false);
+        assert_eq!(headers.len(), 2) // includes the DNA and agent entries
```
lucksus: Do we actually include the DNA in validation packages? But we are at least filtering out the WASM, no? Valiation packages would be way to big otherwise...

zippy: I think we do with the data redacted.

core/src/workflows/author_entry.rs
```diff
@@ -43,6 +43,7 @@ pub async fn author_entry<'a>(
     }
 
     // 1. Build the context needed for validation of the entry
+    // Should we add the current entry to the chain in the validation package?
```
zippy: I think we can remove this comment now that we've settled on option ii.

PR #1565, opened 2019-07-02
Move zome templates out of core repo and support third party templates

willemolding:
## PR summary

Maintaining the template zomes produced by `hc generate` in the core repo was awkward. This PR:

- Moves the templates to their [own repos](https://github.com/holochain/rust-proc-zome-template)
- Uses a proper templating crate (Tera)
- Allows users to specify the URLs to their own custom template zome repos by running `hc generate path/to/zome https://github.com/username/template-repo`. As long as they follow our templating convention ({{ version }}, {{ author }} and {{ name }} for the moment) it will fill these fields.
- Updates the .hcbuild. Previously this was populated using cargo which used absolute paths but this caused issues when the .hcbuild files are commited/shared. It now uses relative paths.

## followups

If there is pull for it we could also add support for user defined template fields which are passed to `hc generate` as an extra param.

## changelog


willemolding:
Not sure about the best way to go about testing. I don't want to slow down the current CI cycle but it would be nice to have the peace of mind that using `hc init` and `hc generate` produces something that actually builds and passes dna tests.

cli/src/config_files/build.rs
```diff
@@ -24,13 +24,13 @@ impl Build {
         Ok(build)
     }
 
-    pub fn save_as<T: AsRef<Path>>(&self, path: T) -> DefaultResult<()> {
-        let file = File::create(path)?;
+    // pub fn save_as<T: AsRef<Path>>(&self, path: T) -> DefaultResult<()> {
```
Connoropolous: Why doing just commented out code?

cli/src/main.rs
```diff
@@ -71,16 +72,19 @@ enum Cli {
...
     Generate {
         #[structopt(
             help = "The path to the zome that should be generated (usually in ./zomes/)",
             parse(from_os_str)
         )]
         zome: PathBuf,
-        #[structopt(help = "The language of the generated zome", default_value = "rust")]
-        language: String,
+        #[structopt(
+            help = "Either the name of a build in template (rust, rust-proc) or the url to a git repo containing a zome template.",
```
```suggestion
            help = "Either the name of a built-in template (rust, rust-proc) or the url to a git repo containing a Zome template.",
```

thedavidmeister:
@willemolding oh i think this needs the merge conflicts solved, the diff looks pretty messy atm?

PR #1568, opened 2019-07-02
Get links Count

AshantiMutinta:
## PR summary
Allow the zome developer to be able to get the link count 
## testing/benchmarking notes

## followups

## changelog


AshantiMutinta:
I propose a follow up story in which GetLinks uses the NetworkQuery for the querying of the GetLinks

AshantiMutinta:
Still missing get_by_tag functions but that should be coming soon

AshantiMutinta:
Feature complete : Now in process of refactor

core/src/network/reducers/get_links_count.rs
```diff
@@ -0,0 +1,130 @@
...
+) -> Result<(), HolochainError> {
+    network_state.initialized()?;
+    let query_json: JsonString = NetworkQuery::GetLinksByTag(key.tag.clone(), crud_status).into();
+    send(
+        network_state,
+        JsonProtocol::QueryEntry(QueryEntryData {
+            requester_agent_id: network_state.agent_id.clone().unwrap().into(),
+            request_id: key.id.clone(),
+            dna_address: network_state.dna_address.clone().unwrap(),
+            entry_address: HashString::from(String::default()),
```
AshantiMutinta: Not sure about the ramifications of returning an empty hash @lucksus

AshantiMutinta:
Have problems running app_spec on my local with new merge.

app_spec/test/test.js
```diff
@@ -108,7 +108,7 @@ scenario('show_env', async (s, t, { alice }) => {
   t.equal(result.Ok.agent_id, '{"nick":"alice::app","pub_sign_key":"' + alice.app.agentId + '"}')
   t.equal(result.Ok.properties, '{"test_property":"test-property-value"}')
 
-  // don't compare the public token because it changes every time we change the dna.
+  // don't compare the public token because it changes every time we change the dna
   t.deepEqual(result.Ok.cap_request.provenance, [ alice.app.agentId, 'HFQkrDmnSOcmGQnYNtaYZWj89rlIQVFg0PpEoeFyx/Qw6Oizy5PI+tcsO8wYrllkuVPPzF5P3pvbCctKkfyGBg==' ]
```
AshantiMutinta: I changed this just to trigger the CI again

core/src/action.rs
```diff
@@ -157,8 +157,15 @@ pub enum Action {
     /// Last string is the stringified process unique id of this `hdk::get_links` call.
     GetLinks(GetLinksKey),
     GetLinksTimeout(GetLinksKey),
+    GetLinksTimeoutByTag(GetLinksKeyByTag),
+    GetLinksCount((GetLinksKey, Option<CrudStatus>)),
+    GetLinksCountByTag((GetLinksKeyByTag, Option<CrudStatus>)),
     RespondGetLinks((QueryEntryData, Vec<(Address, CrudStatus)>, String, String)),
+    RespondGetLinksCount((QueryEntryData, usize, String, String)),
+    RespondGetLinksCountByTag((QueryEntryData, usize, String)),
```
AshantiMutinta: This could have to be more comonized but having explicit enums like this tells you exactly what it is doing.

core/src/dht/dht_store.rs
```diff
@@ -71,6 +71,31 @@ pub fn create_get_links_eavi_query<'a>(
...
+        IndexFilter::LatestByAttribute,
+        Some(EavFilter::predicate(move |attr: Attribute| {
+            match attr.clone() {
+                Attribute::RemovedLink(_, query_tag) => tombstone_tag_regex.is_match(&query_tag),
+                _ => false,
+            }
+        })),
+    ))
+}
+
```
AshantiMutinta: This is a regex for link tags only

core/src/network/reducers/get_links_count.rs
lucksus: It won't work in non-full-sync scenario.
I didn't know where to put it, so I'll start here: **you are missing the link base everywhere.** That is what has to go in here!

AshantiMutinta:
> field that has to be provided: lib3h will use that to find the right node to send this query to. Currently n3h gossips all data with everybody and sends the query just to self, which is why this code passes your tests.
> 
> Regarding the design, I would prefer not adding separate actions but rather subsume all those under `Query` actions (`SendQuery`, `HandleQuery`).
> 
> But I also don't think we need three different HDK functions. `hdk::get_links_count` shou

No problem, will just revert the changes from the tag part

AshantiMutinta:
GetLinksByTag has been removed as it won't work in situations where base won't be available

core/src/action.rs
```diff
@@ -160,8 +160,11 @@ pub enum Action {
     /// Last string is the stringified process unique id of this `hdk::get_links` call.
     GetLinks(GetLinksKey),
     GetLinksTimeout(GetLinksKey),
+    GetLinksCount((GetLinksKey, Option<CrudStatus>)),
```
lucksus: We discussed merging Get* / RespondGet* to Query / RespondQuery. That would make this PR already smaller as we would need all that cruft including a new future etc.

app_spec/test/test.js
```diff
@@ -851,6 +851,137 @@ scenario('get_links_crud', async (s, t, { alice, bob }) => {
   t.equal("deleted",bob_posts_all.Ok.links[1].status);
 
 
+})
+
+scenario('get_links_crud_count', async (s, t, { alice, bob }) => {
```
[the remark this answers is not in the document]
AshantiMutinta: I thought app_spec_proc copies over this from test.js?

core/src/action.rs
AshantiMutinta: I thought about this and just used GetLinks with an enum because one thing I was worried about was the disconnect between what happens on the from the request point of view and the sever point of view. The reason why GetLinksCount was implemented explicitly because following it down the workflow is easy in terms of "new developers" and someone who isn't just new to the system. But I just used an enum as a parameter instead because doesn't over complicate what is happening there.

RespondGet merges and RespondQuery I think is something that should come into play as a later PR imo because some of this stuff can really propagate through a lot of changes making the scope much bigger as it goes.

core/src/network/query.rs
```diff
@@ -2,14 +2,26 @@ use holochain_core_types::{crud_status::CrudStatus, entry::EntryWithMetaAndHeade
...
+pub enum GetLinksNetworkResult {
+    Count(usize),
+    Links(Vec<(Address, CrudStatus)>),
+}
+
 #[derive(Debug, Serialize, Deserialize, PartialEq, DefaultJson, Clone)]
 pub enum NetworkQuery {
     GetEntry,
-    GetLinks(String, String),
+    GetLinks(String, String, Option<CrudStatus>, GetLinksNetworkQuery),
```
lucksus: I regard link counts as a different query, instead of a parameter to `GetLinks`. I think it would be much cleaner to represent it here in that way too: instead of introducing those enums above, why don't we have a `GetLinksCount(String, String)` specialization here?

app_spec/test/test.js
```diff
@@ -853,6 +853,73 @@ scenario('get_links_crud', async (s, t, { alice, bob }) => {
...
+  // get all posts with a deleted status from alice
+  const alice_posts_deleted = await alice.app.call("simple","get_my_links_count",
+  {
+    "base" : alice.app.agentId,
+    "status_request" : "Deleted"
+  });
+
+  //make sure  it is equal to 1
+  t.equal(1,alice_posts_deleted.Ok.count);
+  t.equal(1,bob_posts_deleted.Ok.count);
```
lucksus: These assertions actually fail: https://circleci.com/gh/holochain/holochain-rust/31280?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

core/src/action.rs
```diff
@@ -158,10 +161,10 @@ pub enum Action {
 
     /// get links from entry address and link_type name
     /// Last string is the stringified process unique id of this `hdk::get_links` call.
-    GetLinks(GetLinksKey),
+    GetLinks((GetLinksKey, Option<CrudStatus>, GetLinksNetworkQuery)),
     GetLinksTimeout(GetLinksKey),
-    RespondGetLinks((QueryEntryData, Vec<(Address, CrudStatus)>, String, String)),
-    HandleGetLinksResult((Vec<(Address, CrudStatus)>, GetLinksKey)),
+    RespondGetLinks((QueryEntryData, GetLinksNetworkResult, String, String)),
+    HandleGetLinksResult((GetLinksNetworkResult, GetLinksKey)),
```
lucksus: I would love to see these actions and the corresponding action creators and reducers in the network module to not know about links at all and just operate on the generic level of queries.

PR #1569, opened 2019-07-03
Full-integration app-spec tests with multiple conductors and n3h

lucksus:
## PR summary

This combines the [Try-o-rama PR](https://github.com/holochain/holochain-rust/pull/1554) and [my earlier attempt to fix the memory leak](https://github.com/holochain/holochain-rust/pull/1494).

The high-level intention of theses changes is simple: **have app-spec exercise our production networking implementation** instead of the in-memory mock.

There was quite some work needed to get there:

### Try-o-rama
Diorama was built to remote control one single conductor in the same way we were using one nodejs conductor for our app-spec tests before.

In order to tests networking, we have to spawn at least two conductors as separate processes and have them connect via n3h. But we need to have handles for both conductor processes (and the HC core instances they run) available in our testing JS context.

This switch from being a remote control for one conductor to orchestrating many conductors (potentially even on other devices) is what made @maackle change the name from Diorama to Triorama, which works even better as Try-o-rama since it's a testing framework ;)

In Try-o-rama, we always speak about whole conductor configs and are able to configure several different conductor setups:
```js
const commonConductorConfig = {
  instances: {
    app: dna,
  },
}

const orchestratorSimple = new Orchestrator({
  conductors: {
    alice: commonConductorConfig,
    bob: commonConductorConfig,
    carol: commonConductorConfig,
  },
  debugLog: false,
  executor: tapeExecutor(require('tape')),
  middleware: backwardCompatibilityMiddleware,
})
```
The above creates a Try-o-rama orchestrator that works with three conductors (=nodes, =agents) - here `alice`, `bob` and `carol`.

When running this orchestrator instance, it will wait for three conductors to connect. This could either happen on different machines through a HTTP callback, or, as in app-spec in this PR, by calling `orchestratorSimple.registerConductor(..)`. This is what you will find in app_spec/test/index.js:
```js
const run = async () => {
  const alice = await spawnConductor('alice', 3000)
  await orchestratorSimple.registerConductor({name: 'alice', url: 'http://0.0.0.0:3000'})
  const bob = await spawnConductor('bob', 4000)
  await orchestratorSimple.registerConductor({name: 'bob', url: 'http://0.0.0.0:4000'})
  const carol = await spawnConductor('carol', 5000)
  await orchestratorSimple.registerConductor({name: 'carol', url: 'http://0.0.0.0:5000'})

  const delay = ms => new Promise(resolve => setTimeout(resolve, ms))
  console.log("Waiting for conductors to settle...")
  await delay(5000)
  console.log("Ok, starting tests!")

  await orchestratorSimple.run()
  alice.kill()
  bob.kill()
  carol.kill()
  process.kill()
}
```

Talking about conductors instead of instances has another consequence to how we write scenarios: nodes/agents (like `alice`) can have several instances. So instead of
```js
const result = alice.call("blog", "get_grants", {})
```
we now call zome functions by accessing a conductors instance and then `call`:
```js
const result = alice.app.call("blog", "get_grants", {})
```
(where `app` is the instance name as defined in the Orchestrator's constructor)

### Memory-leak

With Try-o-rama we are reusing conductor and n3h processes and just tear-down and re-initialize instances between tests. This made us aware of a memory leak that would let the memory consumption of one conductor grow to about 8GB before completing the app-spec test suite.

I had started work on that 3 weeks ago but didn't complete it. This PR contains a symptomatic fix that decreases the impact of the leak - but I was not able to find the root cause yet.

*"Memory leak in Rust?!"*
Yes. We use `Arc`s and `Arc<RwLock<..>>`s a lot. Those owned contents only get dropped if there is no reference left.. Although not confirmed, I'm pretty certain that we keep references to `Context` objects after instances got stoped and removed. 

What I could confirm is that the `State` of an instance did not get dropped after the instance was dropped. Since that makes for most of the leaked memory, I've made it so that we can explicitly drop the state when the instance gets dropped.

I've wrapped `State` in
```rust
pub struct StateWrapper {
    state: Option<State>,
}
```
and put that in the `Arc<RwLock<>>`, so that we can set the option to `None`, which is what `fn StateWrapper::drop_inner_state()` does. With that we can do:
```rust
impl Drop for Instance {
    fn drop(&mut self) {
        self.stop_action_loop();
        self.state.write().unwrap().drop_inner_state();
    }
}
```

`StateWrapper` exposes the same functions `State` has, but causes a panic if the inner state is `None`:
```rust
    pub fn nucleus(&self) -> Arc<NucleusState> {
        Arc::clone(
            &self
                .state
                .as_ref()
                .expect("Tried to use dropped state")
                .nucleus,
        )
    }
```
My intention with this was to find zombie-threads that are still working with the context and state after the instance got dropped - but that did not happen. I could run app-spec without seeing this panic once (- it did make a Rust test fail where the instance gets dropped too early).

I then decided to leave this in since any rogue thread that was spawned in the context of an instance and that has outlived its instance *actually should panic*.

With this, the conductor process stays below 1GB of memory even after all the >200 assertions of our app-spec suite - which means we still have a leak as several 100MBs is still way too much, but this should be good enough to wrap up the network-integration-app-spec story...

## followups
Fix the memory leak for good.

## changelog


app_spec/test/spawn_conductors.js
```diff
@@ -0,0 +1,81 @@
+const child_process = require('child_process')
```
Connoropolous: This does feel a bit awkward, now that we're at this point, of duplicating most of this code... roughly in three places now `hc_cluster_test`, here, and in `diorama`

I wonder if we can do some easy deduping

app_spec/test/index.js
```diff
@@ -1,51 +1,85 @@
...
+  const alice = await spawnConductor('alice', 3000)
+  await orchestratorSimple.registerConductor({name: 'alice', url: 'http://0.0.0.0:3000'})
+  const bob = await spawnConductor('bob', 4000)
+  await orchestratorSimple.registerConductor({name: 'bob', url: 'http://0.0.0.0:4000'})
+  const carol = await spawnConductor('carol', 5000)
+  await orchestratorSimple.registerConductor({name: 'carol', url: 'http://0.0.0.0:5000'})
+
+  const delay = ms => new Promise(resolve => setTimeout(resolve, ms))
+  console.log("Waiting for conductors to settle...")
+  await delay(5000)
```
Connoropolous: could this come back to bite us? What is this waiting, more specifically?

conductor_api/src/holochain.rs
```diff
@@ -108,14 +107,15 @@ use holochain_core_types::{
...
 use jsonrpc_core::IoHandler;
 use std::sync::Arc;
 
 /// contains a Holochain application instance
 pub struct Holochain {
-    instance: Instance,
+    instance: Option<Instance>,
     #[allow(dead_code)]
-    context: Arc<Context>,
+    context: Option<Arc<Context>>,
```
maackle: I would like to reinstate the `ugly::Initable` type for these, and other places where a `None` corresponds to an invalid state rather than a truly optional value that was omitted. Then we can be reminded of that fact, and hopefully eventually refactor this so that we can't represent these invalid states anymore, and can avoid them altogether. In the meantime, I figure we will still run into panics every now and then, it's very hard to catch all the edge cases.

app_spec/test/spawn_conductors.js
lucksus: Diorama: yes, but we are not using Diorama here anymore. Try-o-rama does not spawn conductors anymore. (Though this might change with the next step).

Concerning cluster-test, let's talk about that. Would cluster-test at some piont be completely merged with app-spec? If not, yes we should at least merge the spawn_conductor code. Copied it here because I was to lazy to setup TypeScript for app-spec..

app_spec/test/index.js
lucksus: I believe this is mainly about having n3h fully booted and nodes connected. First tests were failing sometimes without this.

maackle: I wonder if this is still needed. At one point hachiko was not hooked up properly and we were diagnosing the problem. I can't currently run these tests locally so can't confirm. Maybe we try removing it and running this in CI?

I get this error running locally: `error: Test conductor aborted: {"code":-32603,"message":"No space left on device (os error 28)"}`

core/src/nucleus/actions/call_zome_function.rs
```diff
@@ -15,6 +15,7 @@ use holochain_core_types::{
     },
     error::HolochainError,
     signature::{Provenance, Signature},
+    ugly::lax_send_sync,
```
maackle: I think we decided that is no longer "ugly", i.e. it's actually a pretty reasonable thing to do in our case. I could move this to a less derogatory module.

core/src/state.rs
```diff
@@ -215,6 +215,142 @@ impl State {
     }
 }
 
+//impl Drop for InnerState {
+//    fn drop(&mut self) {
+//        println!("==================STATE DROPPED!=======================");
+//    }
+//}
```
maackle: remove?

conductor_api/src/holochain.rs
lucksus: Well, these options don't represent invalid states. They enable an explicit drop of the content:
```rust
    pub fn kill(&mut self) {
        let _ = self.stop();
        self.instance = None;
        self.context = None;
    }
```

PR #1572, opened 2019-07-04
Network usage improvements

lucksus:
## PR summary

There a few smaller deltas between how core uses network and how it should be used:
* [x] Don't ask self for validation packages of our entries 
* [x] Untrack on shutdown
* [x] Authoring/Gossip lists are implemented
* [x] Network Keep-alive (from #1582 )

## changelog


core/src/network/handler/lists.rs
```diff
@@ -0,0 +1,89 @@
...
+            "handle_authoring_list/{}",
+            ProcessUniqueId::new().to_string()
+        ))
+        .spawn(move || {
+            let mut address_map = HashMap::new();
+            for entry in get_all_public_chain_entries(context.clone()) {
+                address_map.insert(
+                    entry.clone(),
+                    get_all_aspect_addresses(&entry, context.clone())
+                        .expect("Error getting entry aspects of authoring list"),
```
zippy: Can we really have an expect here?  It seems possible that this could fail.

lucksus: Yep, and then the thread that we spawned 6 lines above should stop. This expect won't make the main thread panic.

PR #1590, opened 2019-07-12
Hachiko lax mode

maackle:
## PR summary

Uses a minor change to hachiko, where when hachiko times out, instead of inducing a test failure (by raising an exeption), it just gives up waiting and lets the rest of the test run to completion.

If we see that tests fail after a timeout, then we can at least get some more insight into what causes the timeout. If the tests pass, then either hachiko was waiting for no good reason, or the test was not testing the right failure mode.

## testing/benchmarking notes

Run CI a bunch, this may help diagnose the CI errors

## changelog


maackle:
I tried to make a change to hachiko too quickly, and made it worse. Don't use this PR, or hachiko v0.3.1, I'll fix it next week.

maackle:
This seems mergeable now, with try-o-rama@0.0.2, hachiko@0.3.2

PR #1593, opened 2019-07-13
Update interface difference

pegaltier:
http vs websocket

## PR summary

just additional doc information

### DIFF
--- a/doc/holochain_101/src/development_conductor.md
+++ b/doc/holochain_101/src/development_conductor.md
@@ -39,7 +39,7 @@ Select a particular JSON-RPC interface to serve your DNA instance over.
 
 The JSON-RPC interface will expose, via a port on your device, a WebSocket or an HTTP server. It can be used to make function calls to the Zomes of a DNA instance. These are covered in depth in the [JSON-RPC interfaces article](./json_rpc_interfaces.md).
 
-The default interface is `websocket`.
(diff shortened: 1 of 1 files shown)


pegaltier:
I will try to fix it soon, thank for all the information!

pegaltier:
@willemolding it should be better now, thanks!

doc/holochain_101/src/development_conductor.md
```diff
@@ -39,7 +39,7 @@ Select a particular JSON-RPC interface to serve your DNA instance over.
 
 The JSON-RPC interface will expose, via a port on your device, a WebSocket or an HTTP server. It can be used to make function calls to the Zomes of a DNA instance. These are covered in depth in the [JSON-RPC interfaces article](./json_rpc_interfaces.md).
 
-The default interface is `websocket`.
+The default interface is `websocket`. `http` can be simpler and faster for prototyping but `websocket` is recommended and for sure more adapted to the production usage thanks to his bi-directional socket connection to the conductor. In fact with http it's only possible to pull data from the conductor. With websocket both are possible, pull data from the conductor and conductor can push data to the user. This will be very useful in many cases, for instance to receive a confirmation when a certain data has been validated or even to debug with the `admin = true` flag.
```
lucksus:
```suggestion
The default interface is `websocket`. `http` can be simpler and faster for prototyping but `websocket` is recommended and for sure more adapted to the production usage thanks to its bi-directional socket connection to the conductor. In fact with http it's only possible to pull data from the conductor. With websockets both directions, push and pull, are possible. Push is used by Holochain with `signals` which can be emitted explicitly from the DNA code, and which also are send via admin interfaces implicitly for debugging purposes representing each core action and thus state mutation.  
```

PR #1595, opened 2019-07-14
Replace ui web server

willemolding:
## PR summary

This addresses issues with the conductor UI webserver not correctly setting content-type headers among other things. This is actually a full replacement of the static file server with a more stable and mature crate.

PR contains the following:
- Refactor static server so we can add other implementations if we need to again
- Implements static server using the Nickel crate
- Adds extra configuration to the conductor config so that it can be bound to different addresses and can set the default rerouting behavior (e.g. to support SPAs with virtual routing)

## testing/benchmarking notes

It would be great to get this tested out with a few different static sites and in particular with the Cypress testing framework @philipbeadle 

## changelog


thedavidmeister:
@willemolding hey would this be better handled upstream in holonix with apache or nginx?

philipbeadle:
Ill try it out :)

willemolding:
@thedavidmeister @lucksus I definitely agree we want to remove this from the conductor when we can. I think having a dev server as part of holonix and something else for end users running Holochain is the way. As far as I can see this is just extra features/code to maintain/support for no good reason really. 

I propose we merge this and then see where everyone else sits on this issue.

We also need to get in contact with Holo team as I think they were planning to use this for the admin UI on the holoports but it should not be a big deal for them to switch to a websever in the holoportOS environment.

conductor_api/Cargo.toml
```diff
@@ -35,13 +35,12 @@ maplit = "=1.0.1"
 lazy_static = "=1.2.0"
 json-patch = "=0.2.2"
 hyper = "=0.12.25"
```
maackle: Do we need hyper at all now?
```suggestion
```

conductor_api/src/conductor/ui_admin.rs
```diff
@@ -349,6 +352,8 @@ id = 'test-bundle-id'"#,
                 port: 4000,
                 bundle: "test-bundle-id".into(),
                 dna_interface: None,
+                reroute_to_root: true,
+                bind_address: "127.0.0.1".to_string()
```
AshantiMutinta: I see the local host literal come up a few times. Is it worth it putting it in a static variable?

conductor_api/src/config.rs
```diff
@@ -693,6 +693,26 @@ pub struct UiInterfaceConfiguration {
...
+    /// Default = "127.0.0.1"
+    pub bind_address: String,
+}
+
+fn default_reroute() -> bool {
+    true
+}
+
+fn default_address() -> String {
+    "127.0.0.1".into()
```
AshantiMutinta: might be worth it to implement a default trait as well

thedavidmeister:
ok i've cross posted this to https://github.com/holochain/holonix/issues/28

happy to merge a stopgap solution here :)

removing `hyper` is great as it (does? did?) contains a nasty `build.rs` file that forces recompilation of all its dependencies which breaks incremental compilation

zippy:
> We also need to get in contact with Holo team as I think they were planning to use this for the admin UI on the holoports but it should not be a big deal for them to switch to a websever in the holoportOS environment.

I heard this morning that the holoports are going to have nginx bundled with them.  But not sure for serving exactly which parts

conductor_api/Cargo.toml
zippy: It looks like it's still being used for http::uri and for Request

willemolding:
Yeah I think we still need to keep Hyper. It is used by Nickel internally anyway so we don't save anything by dropping the explicit dep.

PR #1597, opened 2019-07-15
Clippy Party

lucksus:
## PR summary

**"submitting to the wrath of clippy"**

1.  [x] Fix all Clippy warnings
2.  [x] Add Clippy to CI steps as required

## Join the party
Thanks to the effort of @thedavidmeister, we can run Clippy from holonix with a simple command:
```
hn-rust-clippy
```

Let's submit our code base to the wrath of clippy!

## changelog


core/src/action.rs
```diff
@@ -92,6 +92,7 @@ impl Hash for ActionWrapper {
 /// All Actions for the Holochain Instance Store, according to Redux pattern.
 #[derive(Clone, PartialEq, Debug, Serialize)]
 #[serde(tag = "action_type", content = "data")]
+#[allow(clippy::large_enum_variant)]
```
AshantiMutinta: There might be some truth here about the size of our enums and I think we could make them a bit smaller? Some of them already look like they are organized by sections :)

thedavidmeister:
i added the clippy command to circle builds as a step under fmt

conductor/src/main.rs
```diff
@@ -91,25 +91,28 @@ fn main() {
...
-                            conductor
-                                .expect("No conductor running")
-                                .shutdown()
-                                .expect("Error shutting down conductor");
-                        }
-                        break;
-                        // NB: conductor is dropped here and should shut down itself
+
+                    // Wait forever until we get one of the signals defined above
+                    termination_signals.forever();
```
maackle: I think this is an iterator, so you at least need to call `next()` on it.

AshantiMutinta:
This looks good but they are a lot of places where we are allowing large enum variants. I think while we are here we might as well break them up into smaller enums and use Box to reference. 

The issue is that an enum allocates memory according to the largest variant so smaller variants are given too much space.

lucksus:
@AshantiMutinta, I surely wouldn't mind having the enums be broken up, but I also would like to get this in soon so that any new changes are clippy-checked by CI automatically. There were lots of merge conflicts this morning since develop went ahead while this was stale. 

Do you want to go ahead and break up the enums?
Great if you do. And I would say we should merge this within the next 24h even without that.

PR #1598, opened 2019-07-16
Persisted test agents

### LINKED ISSUE #1596: Test agents get saved to config, but conductor doesn't know how to load them from config
When calling the `test/agent/add` admin API method, the agent is not saved to config. However, when calling another admin method, the previously created test agent is still part of the `Configuration` object and gets dumped at this point. However when the conductor is reloaded, it attempts to recreate this as a real agent, which fails because the test agent doesn't have a real file-backed keystore.

Probably we should filter out test agents whenever saving config. Or give them a proper representation in the config.

willemolding:
## PR summary

Closes #1596 

Test agents can now be stored and serialized in the conductor-config. This does not break existing config files.

## changelog


conductor_api/src/conductor/base.rs
```diff
@@ -1603,6 +1612,37 @@ pub mod tests {
         )
     }
 
+    #[test]
```
willemolding: @maackle Can you confirm this test covers the required use case?

conductor_api/src/config.rs
```diff
@@ -559,6 +559,8 @@ pub struct AgentConfiguration {
     /// If set to true conductor will ignore keystore_file and instead use the remote signer
     /// accessible through signing_service_uri to request signatures.
     pub holo_remote_key: Option<bool>,
+    /// If true this agent will use dummy keys rather than a keystore file
+    pub test_agent: Option<bool>,
```
maackle:
```suggestion
    #[serde(default)]
    pub test_agent: bool,
```

conductor_api/src/conductor/base.rs
maackle: nice test! :+1:

conductor_api/src/config.rs
willemolding: I'm kinda a fan of the `Option<bool>`. It has the added benefit of not serializing it if it is None so all the existing tests that test against hard coded TOML still work.

maackle: I feel like that's a reason to make those hard coded TOML tests more lax! It's a huge PITA to add anything to the config.

struktured: @willemolding  @maackle Instead of this boolean field, why can't we just have a test agent load an embedded key store file? This worked fine with lib3h when testing tls encryption, for instance.

lucksus: I would prefer @maackle's version without the `Option` here.
@struktured, I don't understand your suggstion. With a boolean field, how would out know what is a test agent?

struktured: @lucksus I am suggesting we make (static?) agent configuration instances that point to embedded key stores.  Although maybe I missing something- is there a place that requires absolutely knowing you are running a test agent vs a normal agent?

willemolding:
Re-posting here from the Mattermost chat:

> @struktured - My intuition says you shouldn't need a field called "test_agent" but instead have configurations that are strictly for testing

This seems pretty logical to me. Can anything think of a case where you would want to mix test agents and regular agents? I can't so maybe we should introduce the idea of test configs instead.

willemolding:
Merging as is to unblock testing team but should revisit at some state to simplify the config

PR #1599, opened 2019-07-16
Signal shutdown

maackle:
## PR summary

Allows conductor to shut down gracefully upon SIGINT (Ctrl+C) or SIGKILL, which also allows n3h to shut down gracefully

Necessary for try-o-rama to kill and respawn a conductor during a test

## testing/benchmarking notes

There is no test written for this. It will have to be tested manually, and later, implicitly through try-o-rama tests

## changelog


conductor_api/src/conductor/base.rs
```diff
@@ -434,15 +435,14 @@ impl Conductor {
 
     /// Stop and clear all instances
     /// @QUESTION: why don't we care about errors on shutdown?
```
maackle: ANSWER: we do!

conductor/src/main.rs
```diff
@@ -66,9 +70,24 @@ fn main() {
...
-            // TODO wait for a SIGKILL or SIGINT instead here.
-            loop {
-                sleep(Duration::from_secs(1))
+            for _sig in termination_signals.forever() {
+                let mut conductor_guard = CONDUCTOR.lock().unwrap();
+                let conductor = std::mem::replace(&mut *conductor_guard, None);
+                let refs = Arc::strong_count(&CONDUCTOR);
+                if refs == 1 {
+                    println!("Gracefully shutting down conductor...");
+                } else {
```
Connoropolous: what's the difference between 'graceful' shutdown and 'explicit' shutdown?

(context, I've lost the ability to fully track/understand some of the more sophisticated Rust code that's being written by y'all now)

Connoropolous:
@maackle could you toss a changelog item on here?

conductor/Cargo.toml
```diff
@@ -8,6 +8,7 @@ holochain_core_types = { path = "../core_types" }
 holochain_conductor_api = { path = "../conductor_api" }
 lib3h_sodium = "=0.0.7"
 holochain_common = { path = "../common" }
+signal-hook = "0.1.9"
```
AshantiMutinta: Change this to "=0.1.9"?

conductor/src/main.rs
```diff
@@ -66,9 +70,24 @@ fn main() {
                     .expect("Could not start UI servers!");
             }
 
-            // TODO wait for a SIGKILL or SIGINT instead here.
-            loop {
-                sleep(Duration::from_secs(1))
+            for _sig in termination_signals.forever() {
+                let mut conductor_guard = CONDUCTOR.lock().unwrap();
+                let conductor = std::mem::replace(&mut *conductor_guard, None);
+                let refs = Arc::strong_count(&CONDUCTOR);
```
AshantiMutinta: do we care about dropping the value in the conductor_guard?

conductor_api/src/conductor/base.rs
```diff
@@ -446,16 +449,14 @@ impl Conductor {
     }
 
     /// Stop and clear all instances
-    /// @QUESTION: why don't we care about errors on shutdown?
-    pub fn shutdown(&mut self) {
-        let _ = self
-            .stop_all_instances()
-            .map_err(|error| notify(format!("Error during shutdown: {}", error)));
+    pub fn shutdown(&mut self) -> Result<(), HolochainInstanceError> {
+        let _ = self.stop_all_instances()?;
```
AshantiMutinta: can this just be self.stop_all_instances()?

conductor/Cargo.toml
willemolding: That is probs why the fmt is failing
```suggestion
signal-hook = "=0.1.9"
```

lucksus:
```suggestion
signal-hook = "=0.1.9"
```

conductor/src/main.rs
lucksus: I guess the idea is that if the reference count for the `CONDUCTOR` Arc is 1 there is nobody else (no other thread) alive and running for this conductor and we don't need to explicitly call `conductor.shutdown()` as that is called in the `Drop` implementation.

But I don't completely get the difference either since we take ownership of the conductor in line 75 with `mem::replace`, right? So that object will be dropped at the end of the for-body's scope either way.. Isn't that right, @maackle?

maackle: Yes, this bit is a little confusing, but I figured it's better to have something than nothing. I was thinking that other threads would still have access to a conductor in the explicit shutdown case, but @lucksus is right that the conductor gets dropped either way. However, if there are other threads referencing CONDUCTOR at this point, they would wind up with a `None` when they expected `Some(Conductor)`, so this warning at least explains why. I could modify the warning message a little, but I think it still mostly makes sense like this.

conductor/src/main.rs
maackle: @AshantiMutinta it will get dropped at the end of the block. What's your concern?

PR #1601, opened 2019-07-17
State dump debugging

lucksus:
## PR summary

Adds a state dump for debugging that if activated gets printed as log every 10 seconds and is human readable. It looks like this:
```
=============STATE DUMP===============

Nucleus:
========
Running zome calls: []
-------------------
Pending validations: []
--------------------

Network:
--------
Running GET ENTRY flows: []
------------------------
Running GET LINKS flows: []
------------------------
Running VALIDATION PACKAGE requests: []
------------------------------------
Running DIRECT MESSAGES: []

Dht:
====
Holding:
* HcSCJCqoIY3uwiw34acyvNmJMyzkk4y9groHdYKBekqp7y48mvwfVTQQkzcjnfz: [%agent_id] bob::app
* HcSCIv3cPT5kegjoqgXM7nVU8rFbd9pyg5oOYUz9PSryp5mb7DKhCsXCS768pua: [%agent_id] alice::app
* HcScIjHzecIuensj8zcPFAdrY6if7ydywAYgpreQVz469qjdV75s99uCOWStxgr: [%agent_id] carol::app
* QmWY98zemN4Mi9A9SgmC8vQufaQUf5a13Xe81PfWDvvnn9: [post] {"content":"hi","date_created":"now"}
* QmWPiVMcq1EfpgErRdS4AgB297hahbVqNgdLAtq8Uyc6wP: [%link_add] (authored_posts#)
	HcSCIv3cPT5kegjoqgXM7nVU8rFbd9pyg5oOYUz9PSryp5mb7DKhCsXCS768pua => QmWY98zemN4Mi9A9SgmC8vQufaQUf5a13Xe81PfWDvvnn9
--------
```

It gets activated through a boolean config flag in the conductor config under `logger`:
```
[logger]
state_dump = true
```

## app-spec

This branch also activate this for app-spec test so we can do forensics for CI runs that fail non-deterministically.

## changelog


core/src/nucleus/reducers/return_zome_function_result.rs
```diff
@@ -18,6 +18,20 @@ pub fn reduce_return_zome_function_result(
...
+/// a zome call is running
+pub fn reduce_signal_zome_function(
+    state: &mut NucleusState,
+    _root_state: &State,
+    action_wrapper: &ActionWrapper,
+) {
+    let action = action_wrapper.action();
+    let call = unwrap_to!(action => Action::SignalZomeFunctionCall);
+    state.zome_calls.insert(call.clone(), None);
+    println!("debug/reduce_signal_zome_function: zome call started: {:?}", call);
```
maackle: is this println necessary since it should be kinda captured in the state dump?

lucksus: Not anymore. I used it for debugging. Will remove.

lucksus:
I intended to make the holding list less opaque by actually printing the entry contents but I wrapped this up and un-WIPed this PR so it could be merged if needed to help diagnosis app-spec flakyness already. Will add detailed holding list next week either in this PR or a new one.

maackle:
Looks like a bunch of tests are failing because of the addition to the config. Are these tests maybe a little too specific? I have felt the pain of adding a simple item to the config before too, I don't know how much value we're getting in asserting that the config is generated exactly a certain way, with all the items in a particular order...

Overall I am OK with this PR but I am wary about a proliferation of new config items if we need to add even more diagnostic logging. I wonder if this could be incorporated into the logging framework with new LogRules. As long as we keep that in mind for future additions I'm Ok with this one once tests pass.

core/src/scheduled_jobs/state_dump.rs
```diff
@@ -0,0 +1,137 @@
...
+        .map(|(key, _)| key)
+        .cloned()
+        .collect();
+
+    let validation_package_flows: Vec<Address> = state_lock.network().get_validation_package_results
+        .iter()
+        .filter(|(_, result)| result.is_none())
+        .map(|(address, _)| address)
+        .cloned()
+        .collect();
```
AshantiMutinta: I think we could eliminate a lot of code if we DRY this out. Maybe pass the iterator in a function?

Also I think we could use filter_map because it packs all we are doing into one thing

core/src/scheduled_jobs/state_dump.rs
```diff
@@ -0,0 +1,137 @@
...
+    let pending_validation_strings = state_lock
+        .nucleus()
+        .pending_validations
+        .keys()
+        .map(|pending_validation_key| {
+            let maybe_content = address_to_content_and_type(&pending_validation_key.address, context.clone());
+            if let Ok((content_type, content)) = maybe_content {
+                format!("<{}> [{}] {}: {}", pending_validation_key.workflow.to_string(), content_type, pending_validation_key.address.to_string(), content)
+            } else {
+                format!("<{}> [UNKNOWN] {}: Error trying to get type/content: {}", pending_validation_key.workflow.to_string(), pending_validation_key.address.to_string(), maybe_content.err().unwrap())
```
AshantiMutinta: I hope I am not being too  pendantic but we could also eliminate some lines of code in this if we did something like this

maybe_content
.map(|(content_type,content)| format!())
.unwrap_or_else(|_|format!())

core/src/scheduled_jobs/state_dump.rs
```diff
@@ -0,0 +1,137 @@
...
+    } else {
+        Ok((String::from("UNKNOWN"), raw_content.to_string()))
+    }
+}
+
+pub fn state_dump(context: Arc<Context>) {
+    let state_lock = context.state().expect("No state?!");
+
+    let running_calls: Vec<ZomeFnCall> = state_lock.nucleus().zome_calls
+        .iter()
```
willemolding: Could use `into_iter()` and drop the `cloned()` call (I'm pretty sure).

core/src/scheduled_jobs/state_dump.rs
willemolding: +1 for `filter_map`!

I don't think its an issue that this isn't totally DRY. It is very readable.

lucksus: Hm, I tried `filter_map` but IMHO it makes it worse because I then have to have an complex `if` inside that closure. I prefer leaving it the way it is.

core/src/scheduled_jobs/state_dump.rs
lucksus: Doesn't work without having to implement some `Copy` traits..

PR #1497, opened 2019-06-10
Add validation for agent entry

willemolding:
## PR summary

To enable membranes there must be a way for agents to validate who is allowed to join. This PR adds some new functionality to enable this by making the AgentId (second entry on the local chain) a validate-able entry with a user defined callback in the zome. 

This is a breaking change to existing zomes. Zome developers must add a `validate_agent` callback either using the old HDK or the proc-macro HDK.

## changelog

Please check one of the following, relating to the [CHANGELOG-UNRELEASED.md](https://github.com/holochain/holochain-rust/blob/develop/CHANGELOG-UNRELEASED.md)

- [x] this is a code change that effects some consumer (e.g. zome developers) of holochain core so it is added to the CHANGELOG-UNRELEASED.md (linked above), with the format `- summary of change [PR#1234](https://github.com/holochain/holochain-rust/pull/1234)`
- [ ] this is not a code change, or doesn't effect anyone outside holochain core development

PR #1604, opened 2019-07-18
Re-enable accidentally commented app spec tests

maackle:
## PR summary

https://github.com/holochain/holochain-rust/pull/1497 had a few extra commits that were merged into develop after all reviewers approved, one of which accidentally disabled almost all app spec tests. This PR restores the tests, and adds a simple meta-test to see if we have about as many tests as we expect to have.

## followups

We might want to discuss our code review policy when a PR undergoes more changes after passing CR. In the most strict case, we should redo CR for any new commits, but maybe we can have something more lax?

## changelog

### DIFF
--- a/app_spec/test/index.js
+++ b/app_spec/test/index.js
@@ -48,39 +48,39 @@ const orchestratorMultiDna = new Orchestrator({
   callbacksPort: 8888,
 })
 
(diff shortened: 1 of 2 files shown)


maackle:
app spec tests are back to failing in the usual way, calling it good and merging.

PR #1607, opened 2019-07-22
Link optimization

AshantiMutinta:
## PR summary
Leveraging some of the link processing to the DHT node
## testing/benchmarking notes

## followups

## changelog

- [ x] this is a code change that effects some consumer (e.g. zome developers) of holochain core so it is added to the CHANGELOG-UNRELEASED.md (linked above), with the format `- summary of change [PR#1234](https://github.com/holochain/holochain-rust/pull/1234)`


core/src/network/handler/query.rs
```diff
@@ -8,28 +8,49 @@ use crate::{
...
+                targets.iter().last().map(|last_target|{
+                   Ok((eav_crud.0,eav_crud.1,last_target.value())) 
+                }).unwrap_or(Err(HolochainError::ErrorGeneric("Could not find cached target".to_string())))
+            }).unwrap_or(Err(HolochainError::ErrorGeneric("Could not find cached target".to_string())))
+        }).partition(Result::is_ok);
+
+        //if can't find target don't return
+        if error.is_empty()
+        {
+            Err(HolochainError::ErrorGeneric("Could not find targets in local dht".to_string()))
```
willemolding: It might be nice to include the actual errors that were thrown in the output rather here

core/src/network/handler/query.rs
```diff
@@ -8,28 +8,49 @@ use crate::{
...
+            .get_link_targets(eav_crud.0.clone())
+            .map(|targets|{
+                targets.iter().last().map(|last_target|{
+                   Ok((eav_crud.0,eav_crud.1,last_target.value())) 
+                }).unwrap_or(Err(HolochainError::ErrorGeneric("Could not find cached target".to_string())))
+            }).unwrap_or(Err(HolochainError::ErrorGeneric("Could not find cached target".to_string())))
+        }).partition(Result::is_ok);
+
+        //if can't find target don't return
+        if error.is_empty()
```
willemolding: Return an error if error is empty? Is this an error or am I not understanding it right?

core/src/network/handler/query.rs
```diff
@@ -5,31 +5,85 @@ use crate::{
...
+        .partition(Result::is_ok);
+
+        //if can't find target throw error
+        if error.is_empty()
+        {
+            Ok(get_link.iter().map(|s|s.clone().unwrap()).collect::<Vec<_>>())
+        }
+        else
+        {
+            Err(HolochainError::ErrorGeneric("Could not find targets in local dht".to_string()))
```
willemolding: Might be nice to include the thrown errors in the output

core/src/network/handler/query.rs
```diff
@@ -5,31 +5,85 @@ use crate::{
...
+                Attribute::LinkTag(_,tag) => Ok(tag),
+                Attribute::RemovedLink(_,tag) => Ok(tag),
+                _ =>  Err(HolochainError::ErrorGeneric("Could not get tag".to_string()))
+            }.expect("INVALID ATTRIBUTE ON EAV GET, SOMETHING VERY WRONG IN EAV QUERY");
+            (eav_crud.0,eav_crud.1,tag)
+        })
+        //get targets from dht
+        .map(|eav_crud|{
+            dht_store
+            .get_link_targets(eav_crud.0.clone())
```
willemolding: I'm not sure if it is safe to be getting the targets from the store like this. With full sync there is a good chance that the target are held by the same node as the base but this might not be the case with sharding right? Maybe there is something I am not understanding

AshantiMutinta: I think with this situation because of how every node that stores the base also stores the meta and target(check dht_reduers_inner line 59), this should be fine. If no base is found in the local dht it won't even reach this far

core/src/network/handler/query.rs
AshantiMutinta: Good catch

core/src/network/handler/query.rs
AshantiMutinta: @willemolding You were right with this on the aspect that getting the targets like this would work because of the link_caching but getting the entries would be the problem. Just cause one has the link doesn't mean they have the linkadd entry(is a part I overlooked) switched this around so take a look when you can!

core/src/workflows/hold_link.rs
```diff
@@ -99,6 +99,12 @@ pub async fn hold_link_workflow(
     // 3. If valid store the entry in the local DHT shard
     await!(add_link(&link_add, &context))?;
     context.log(format!("debug/workflow/hold_link: added! {:?}", link));
+
+    //4. store link_add Entry
+    await!(hold_entry_workflow(&entry_with_header, context.clone()))?;
+    context.log(format!("debug/workflow/hold_entry: added! {:?}", entry_with_header));
+
+    //4. EAV add target to address where we only have one look up
```
willemolding: This comment seems to be misplaced now and there are two step number 4

AshantiMutinta:
Fixing app_spec after merge

core/src/workflows/hold_link.rs
```diff
@@ -99,6 +99,12 @@ pub async fn hold_link_workflow(
     // 3. If valid store the entry in the local DHT shard
     await!(add_link(&link_add, &context))?;
     context.log(format!("debug/workflow/hold_link: added! {:?}", link));
+
+    //4. store link_add Entry
+    await!(hold_entry_workflow(&entry_with_header, context.clone()))?;
```
willemolding: Just to clarify does this now mean that the nodes that hold the base also hold the link meta on the base and the link_add entry?

AshantiMutinta: Yes.

For the link meta => so that we can get the headers
Link Add Entry => So that we can get the target

core/src/network/query.rs
```diff
@@ -1,17 +1,47 @@
...
 pub enum GetLinksNetworkQuery {
     Count,
-    Links,
+    Links(GetLinksQueryConfiguration)
 }
 
+#[derive(Debug, Serialize, Deserialize, PartialEq, DefaultJson, Clone)]
+pub struct GetLinkData
+{
+    pub address : Address,
```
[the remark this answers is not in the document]
AshantiMutinta: address of the link_add :) will be used to filter for the remove_link

core/src/workflows/hold_link.rs
willemolding: I think this seems ok. It does mean that LinkAdd entries will be twice as dense in the DHT space when there is sharding so I guess this is a storage vs. retrieval time tradeoff.

PR #1613, opened 2019-07-25
WIP - Add header publishing

willemolding:
## PR summary

Published headers as entries to the DHT when publishing an app entry. This is pre-work to allow for constructing an agents local chain when they are offline for validation or data recovery purposes.

**WIP** Currently only public entries headers are published. This is incorrect and will prevent proper validation. Also cannot seem to load headers to the HDK although tests show they are successfully being published.

### DIFF
--- a/core/src/network/reducers/publish.rs
+++ b/core/src/network/reducers/publish.rs
@@ -13,6 +13,7 @@ use holochain_core_types::{
     crud_status::CrudStatus,
     entry::{entry_type::EntryType, Entry},
     error::HolochainError,
+    chain_header::ChainHeader,
 };
 use lib3h_protocol::{
     data_types::{EntryData, ProvidedEntryData},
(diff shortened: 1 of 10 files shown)


core/src/network/reducers/publish.rs
```diff
@@ -129,7 +152,11 @@ fn reduce_publish_inner(
     let entry_with_header = fetch_entry_with_header(&address, root_state)?;
     match entry_with_header.entry.entry_type() {
         EntryType::AgentId => publish_entry(network_state, &entry_with_header),
-        EntryType::App(_) => publish_entry(network_state, &entry_with_header).and_then(|_| {
+        EntryType::App(_) => publish_entry(network_state, &entry_with_header)
+        .and_then(|_|{
+            publish_header(network_state, &entry_with_header.header)
```
willemolding: The trouble with doing this here is this publish call is only made for public entries. Need to move it elsewhere

PR #1615, opened 2019-07-26
Feature add header publishing

willemolding:
## PR summary

Adds automatic publishing of all headers as special header entries in the DHT. This is to allow reconstruction of any agents local chain for validation and data recovery purposes.

Particularly this is pre-work for offline source agent validation.

- Changes `author_entry` workflow to call publish for all entries (even private)
- Changes `reduce_publish_inner` to publish headers always and only the entry itself in particular cases
- Rewrites `get_entry_from_cas` to fail if deserialization fails. Previously this returned None in this case making it difficult to track down bugs.
- Adds publishing of headers for Dna and AgentId entries in the initialization workflow
- Rewrites `create_new_chain_header` to take a &State rather than StateWrapper 

- Adds tests that headers are correctly published and communicated over the test network (`test_commit_with_dht_publish_header_is_published`)
- Adds app spec test to ensure that headers can be retrieved by another agent
- Adds test that an agents local chain can be fully recovered from the DHT entries

## changelog


core/src/entry/mod.rs
```diff
@@ -40,6 +41,40 @@ impl CanPublish for EntryType {
         }
         true
     }
+
+    fn can_publish_from_state(&self, state: &State) -> bool {
```
willemolding: This is a bit of a hack for now. @lucksus any idea what the best way to call `can_publish` from inside `reduce_publish` where the context is not available?

lucksus: This seems perfectly fine to me. The `context` is mainly a means to make the state of an instance available outside of the reducers...

core/src/network/handler/store.rs
```diff
@@ -40,7 +41,23 @@ pub fn handle_store(dht_data: StoreEntryAspectData, context: Arc<Context>) {
...
+                    ))
+                    .spawn(move || {
+                        match context
+                            .block_on(hold_header_workflow(&header, context.clone()))
+                        {
+                            Err(error) => context.log(format!("err/net/dht: {}", error)),
+                            _ => (),
+                        }
+                    })
+                    .expect("Could not spawn thread for storing EntryAspect::Header");
```
lucksus: This does not seem right to me. Publishing headers so you can get them by their address means they are entries themselves, not an aspect of the entry the header is about, which is what this `EntryAspect::Header` means..

core/src/network/reducers/publish.rs
```diff
@@ -43,6 +45,29 @@ fn publish_entry(
...
+) -> Result<(), HolochainError> {
+    let header_entry = Entry::ChainHeader(chain_header.clone());
+    send(
+        network_state,
+        Lib3hClientProtocol::PublishEntry(ProvidedEntryData {
+            space_address: network_state.dna_address.clone().unwrap().into(),
+            provider_agent_id: network_state.agent_id.clone().unwrap().into(),
+            entry: EntryData {
+                entry_address: header_entry.address().clone(),
+                aspect_list: vec![EntryAspect::Header(
```
lucksus: Hah, interesting. So the tests do pass because you've used the header's address as the `entry_address` here, but then it is a bit weird to use the `EntryAspect::Header`, since this data is rather the content of the entry that is the header.

core/src/workflows/hold_header.rs
```diff
@@ -0,0 +1,34 @@
...
+    let entry_with_header = EntryWithHeader::new(chain_header_entry.clone(), chain_header.clone());
+    await!(hold_entry(&entry_with_header, context.clone()))?;
+
+    context.log(format!(
+        "debug/workflow/hold_header: HOLDING: {}",
+        chain_header_entry.address()
+    ));
+
+    Ok(())
+}
```
lucksus: Related to my comments above, I don't think we need a `hold_header` workflow for header publishing - instead headers should be first class entries that get stored through the regular `hold_entry` workflow.

There might a different future use-case for `hold_header` that works with the `EntryAspect::Header` that is actually storing *only* the header of an entry. This makes sense should we choose/need to implement something like *"sacriledging"* entries, i.e. making it a crime to speak about an entry, because the author has requested to have it deleted really because of privacy issues.. in that case we would still need to store the header and potentially other `EntryAspects` of that entry, to know that there was an entry with that address, but just not the content.

willemolding:
@lucksus Take a look at the latest changes. I switched it over to treating headers more like entries and using the `hold_entry` workflow for them as well. This feel much nicer and allows us to add validation to the header entries as well which will be important.

core/src/network/reducers/publish.rs
```diff
@@ -43,6 +46,38 @@ fn publish_entry(
...
+fn publish_header(
+    network_state: &mut NetworkState,
+    root_state: &State,
+    chain_header: &ChainHeader,
+) -> Result<(), HolochainError> {
+    let header_entry = Entry::ChainHeader(chain_header.clone());
+    let header_entry_header = create_new_chain_header(
+        &header_entry,
+        &root_state.agent(),
+        &StateWrapper::from(root_state.clone()),
```
willemolding: @lucksus This is one thing I am not sure about. Can I create a `&StateWrapper` from a `&State` without cloning it? Also I am not sure if it is ok to clone the state like this...

core/src/workflows/author_entry.rs
```diff
@@ -129,31 +126,71 @@ pub mod tests {
...
+        let header_entry = Entry::ChainHeader(header);
 
-        let x: String = json.unwrap().to_string();
+        // try and load it by its address as Jack. This means it has been communicated over the mock network
+        let mut entry: Option<Entry> = None;
+        let mut tries = 0;
+        while entry.is_none() && tries < 10 {
+            tries = tries + 1;
+            {
+                entry = get_entry_from_dht(&context2, &header_entry.address()).expect("Could not retrieve entry from DHT");
```
willemolding: This isn't quite right. We need to be able to retrieve ChainHeader entries by the hash of the `ChainHeader` not by the hash of the `Entry::ChainHeader`. Without this we can't follow the link to the previous header in the chain which is the whole reason for doing this in the first place.

willemolding: Fixed!

core/src/network/reducers/publish.rs
willemolding: Great idea. This is done now

core/src/workflows/application.rs
```diff
@@ -26,7 +35,18 @@ pub async fn initialize(
         context.log("dna/initialize: Initializing new chain from given DNA...");
         await!(initialize_chain(dna.clone(), &instance_context))?;
     }
+
+    // 3. Initialize the network
     await!(initialize_network(&instance_context))?;
+
+    // 4. Call publish on the agent and DNA entries. 
+    // This is to trigger the publishing of their headers not the entries themselves
+    let dna_entry = Entry::Dna(Box::new(dna.clone()));
```
zippy: Isn't this going to get memory expensive because it's a full clone of the code?  Any way to stub that because the DNA entry is not actually used?

core/src/workflows/author_entry.rs
```diff
@@ -129,31 +127,150 @@ pub mod tests {
...
+            .address();
+        thread::sleep(time::Duration::from_millis(500));
+
+        // collect Jills local chain
+        let state = &context1.state().unwrap();
+        let jill_headers: Vec<ChainHeader> = state
+            .agent()
+            .iter_chain()
+            .collect();
+        let header = jill_headers.first().expect("Must be at least one header in chain");
```
lucksus: I’m not sure about ‘first()’ here. Wouldn’t that five the top chain header as the chain iterator walks back the chain starting with the last one? That would mean that this test is not showing that you can get the whole chain, but just the first one..

core/src/workflows/application.rs
lucksus: It's not too bad as the WASM binary is hold in an `Arc` inside `DnaWasm` here: https://github.com/holochain/holochain-rust/blob/38837a4472387e2f14a312984ba5fa64e039b513/core_types/src/dna/wasm.rs#L96

core/src/workflows/author_entry.rs
lucksus: Only think keeping me from approving this PR, @willemolding.

willemolding: I thought that as well and originally had `.last()` but actually that is incorrect. For some reason they are the other way around to what you would expect.

As it is it definitely retrieves the whole chain (from the top two App entry headers plus the AgentId and DNA entries. For clarity I will add an assertion that the retrieved chain has length=5.

core/src/workflows/application.rs
```diff
@@ -26,7 +35,18 @@ pub async fn initialize(
         context.log("dna/initialize: Initializing new chain from given DNA...");
         await!(initialize_chain(dna.clone(), &instance_context))?;
     }
+
+    // 3. Initialize the network
     await!(initialize_network(&instance_context))?;
+
+    // 4. Call publish on the agent and DNA entries. 
+    // This is to trigger the publishing of their headers not the entries themselves
```
lucksus: Ah!
I guess this is what breaks hachiko's assumptions and makes it wait for the DNA entry to be held by everybody since it sees the publish action but then no subsequent hold!

PR #1620, opened 2019-07-29
Fallback build validation packages using published entries in the DHT

willemolding:
## PR summary

Depends on #1640 

This should finally allow for author agents to drop on and off the network and have data gossiped and persisted in the case where the validation package is `ChainHeaders` or `ChainFull`.

- Adds validation packages constructed from published DHT entries
- Adds tests that these packages are constructed as expected


app_spec/test/files/offline-validation.js
```diff
@@ -0,0 +1,103 @@
...
+      //       t.ok(header_bob.Ok)
+
+      //       t.deepEqual(header_alice.Ok, header_bob.Ok)
+
+      //       chain_headers.push(header_bob.Ok)
+      //   }
+      //   t.comment(JSON.stringify(chain_headers))
+      // })
+
+      scenario('Can perform validation of an entry while the author is offline', async (s, t) => {
```
maackle: could just use `scenario.only` to make this the only test, without commenting a bunch of stuff out

willemolding:
So it looks like everything *should theoretically work(TM)* but the app spec test is letting it down. 

I believe there are two main issues:

- The `await s.consistency()` doesn't work when agents are coming online and offline dynamically. This causes the test to end prematurely (before the validation is complete)

- Even when adding an `await delay(..)` it takes an unreasonable (possibly infinite) time for the validations to resolve. This is because an entry that requires a full chain package to validate is essentially dependent on all of the header entries being validated and held. This is similar to a LinkAdd entry being dependent on the base and target but actually much worse because there could be a large number of header entries required. These all need to be validated first but the current implementation just picks things up randomly. 

The solution to the second point is to add a validation dependency graph and topologically sorting to validate leaves first. This PR will be parked until that is complete.

PR #1624, opened 2019-07-31
Action merge

AshantiMutinta:
## PR summary
This PR refactors and merges actions using enums

## followups

## changelog


AshantiMutinta:
Looks like there is an issue with my cargo fmt

core/src/action.rs
```diff
@@ -87,6 +87,27 @@ impl Hash for ActionWrapper {
     }
 }
 
+///This describes a key for the actions
+#[derive(Clone, PartialEq, Debug, Serialize, Eq, Hash)]
+pub enum Key {
```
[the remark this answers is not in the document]
AshantiMutinta: Yes! I think that is a very good suggestion

core/src/action.rs
```diff
@@ -130,36 +151,26 @@ pub enum Action {
     /// (only publish for AppEntryType, publish and publish_meta for links etc)
     Publish(Address),
 
-    /// Get an Entry on the network by address
-    GetEntry(GetEntryKey),
+    ///Performs a Network Get Action based on the key and payload, used for links and Entries
+    Get((Key, GetPayload)),
```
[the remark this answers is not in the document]
AshantiMutinta: Absolutely, I think that works out much better! Thank you!

core/src/network/actions/get.rs
```diff
@@ -0,0 +1,128 @@
+use crate::{
```
lucksus: This file should be `query.rs` then also.

core/src/network/actions/get.rs
```diff
@@ -0,0 +1,128 @@
...
+use holochain_wasm_utils::api_serialization::get_links::{GetLinksArgs, LinksStatusRequestKind};
+
+/// FetchEntry Action Creator
+/// This is the network version of get_entry that makes the network module start
+/// a look-up process.
+///
+/// Returns a future that resolves to an ActionResponse.]
+
+#[derive(Clone, PartialEq, Debug, Serialize)]
+pub enum GetMethod {
```
lucksus:
```suggestion
pub enum QueryMethod {
```

core/src/network/actions/get.rs
```diff
@@ -0,0 +1,128 @@
...
+    Entry(Address),
+    Link(GetLinksArgs, GetLinksNetworkQuery),
+}
+
+pub async fn get(
```
lucksus:
```suggestion
pub async fn quert(
```

core/src/network/actions/get.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    await!(GetFuture {
+        context: context.clone(),
+        key: key.clone(),
+    })
+}
+
+/// GetEntryFuture resolves to a HcResult<Entry>.
+/// Tracks the state of the network module
+pub struct GetFuture {
```
lucksus:
```suggestion
pub struct QueryFuture {
```
etc...

core/src/network/reducers/get.rs
```diff
@@ -0,0 +1,436 @@
+use crate::{
```
lucksus: This file and all the functions etc. in here should also be `query` instead of `get`

core/src/network/reducers/get.rs
```diff
@@ -0,0 +1,436 @@
...
+}
+
+pub fn reduce_get(
+    network_state: &mut NetworkState,
+    _root_state: &State,
+    action_wrapper: &ActionWrapper,
+) {
+    let action = action_wrapper.action();
+    let (key_type, payload) = unwrap_to!(action => crate::action::Action::Query);
+    let result = match key_type {
```
lucksus: Why do we need to match on the key_type here? Can't we just pass the query we've got to over to the network as it is?

core/src/network/reducers/get.rs
```diff
@@ -0,0 +1,436 @@
...
+}
+
+fn reduce_get_links_inner(
+    network_state: &mut NetworkState,
+    key: &GetLinksKey,
+    get_links_query: &GetLinksNetworkQuery,
+    crud_status: &Option<CrudStatus>,
+) -> Result<(), HolochainError> {
+    network_state.initialized()?;
+    let query_json: JsonString = NetworkQuery::GetLinks(
```
lucksus: Ah, this is why we still have those separate inner reducers. The `NetworkQuery` gets created here. I've overlooked this above in the action creator, but I thought we could move that over there. So create this `NetworkQuery` there and pass it over to here as the parameter of the `Query` action.

core/src/scheduled_jobs/state_dump.rs
```diff
@@ -38,35 +41,60 @@ fn address_to_content_and_type(address: &Address, context: Arc<Context>) -> Resu
...
         .filter(|(_, result)| result.is_none())
-        .map(|(key, _)| key)
+        .filter(|(key, _)| match key {
+            QueryKey::Links(_) => true,
+            _ => false,
+        })
+        .map(|(key, _)| match key {
+            QueryKey::Links(key) => key.clone(),
+            _ => panic!("Wrong Enum Variant should never be reached"),
+        })
```
lucksus: Ah, it actually doesn't really make sense to keep these artificially separated here just to keep the state dump output the same. I think we should rather change the dump to not show separate get_entry/get_link_flows, but instead just running queries...

core/src/action.rs
```diff
@@ -87,6 +87,27 @@ impl Hash for ActionWrapper {
...
+    Entry,
+    Links((Option<CrudStatus>, GetLinksNetworkQuery)),
+}
+
+///This is a payload from a response
+#[derive(Clone, PartialEq, Debug, Serialize)]
+pub enum RespondQueryPayload {
+    Entry(Option<EntryWithMetaAndHeader>),
+    Links((GetLinksNetworkResult, String, String)),
+}
```
lucksus: I don't think we need these two `Payload*` enum anymore, or do we? If we would use `NetworkQuery` and `NetworkQueryResult` (in `core/src/network/query.rs`) we also wouldn't need to translate these types over into `NetworkQuery*` in the according reducer below.

PR #1644, opened 2019-08-08
Make test registerer not hide `.only`

maackle:
## PR summary

The "registerer" introduced in #1604 wound up eclipsing the `registerScenario.only` function, this makes it visible again.

## changelog

### DIFF
--- a/app_spec/test/regressions.js
+++ b/app_spec/test/regressions.js
@@ -1,6 +1,6 @@
 module.exports = scenario => {
 
-scenario.only('calling get_links before link_entries makes no difference', async (s, t, {alice}) => {
+scenario('calling get_links before link_entries makes no difference', async (s, t, {alice}) => {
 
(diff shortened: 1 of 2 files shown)


app_spec/test/regressions.js
```diff
@@ -1,6 +1,6 @@
 module.exports = scenario => {
 
-scenario('calling get_links before link_entries makes no difference', async (s, t, {alice}) => {
+scenario.only('calling get_links before link_entries makes no difference', async (s, t, {alice}) => {
```
zippy: Looks like you committed your test of this by accident?

```suggestion
scenario('calling get_links before link_entries makes no difference', async (s, t, {alice}) => {
```

maackle: Let's see if the registerer catches that in CI

maackle: Well, unfortunately it didn't...that registerer sure has room for improvement

PR #1645, opened 2019-08-08
(do not merge) - Sunset hdk docs and update to proc-macro

willemolding:
## PR summary

This PR updates the mdbook and other docs to use the proc-macro HDK in preparation for it becoming the officially supported method for authoring zomes.

This should not be merged until we actually decide to go ahead with the deprecation of the old HDK

## changelog


doc/holochain_101/src/bridging.md
```diff
@@ -56,28 +56,23 @@ pub fn handle_sample_function(some_param: String) -> ZomeApiResult<Address> {
...
 }
 
-define_zome! {
-    entries: []
-
-    init: || { Ok(()) }
-
-    validate_agent: |validation_data : EntryValidationData::<AgentId>| {
+#[zome]
+mod zome {
```
jmday: suggest `mod my_zome {` or something that illustrates the ability to name the zome whatever ya want.

doc/holochain_101/src/zome/define_zome.md
```diff
@@ -28,33 +28,33 @@ What are Rust macros? Generally speaking, they are code that will actually gener
...
-define_zome! {
-    entries: []
-
-    init: || {
+#[zome]
+mod my_zome {
+    
+    #[init]
+    fn init() -> ZomeApiResult<()> {
         Ok(())
```
jmday: Important but not urgent:
Showing the simplest useful example of how `init()` would be used (e.g. some dis-allowed states?) would be better than just returning `Ok(())` for all calls in an introductory example like this.

doc/holochain_101/src/zome/define_zome.md
```diff
@@ -28,33 +28,33 @@ What are Rust macros? Generally speaking, they are code that will actually gener
...
     }
 
-    validate_agent: |validation_data : EntryValidationData::<AgentId>| {
+    #[validate_agent]
+    fn validate_agent(validation_data : EntryValidationData::<AgentId>) -> ZomeApiResult {
         Ok(())
```
jmday: Important but not urgent:
Showing the simplest useful example of how `validate_agent()` would be used (e.g. blacklist of dis-allowed agents or something?) would be better than just returning `Ok(())` for all calls in an introductory example like this.

doc/holochain_101/src/zome/define_zome.md
```diff
@@ -28,33 +28,33 @@ What are Rust macros? Generally speaking, they are code that will actually gener
...
 ```
 
-`entries` represents the validating entry type definitions. Note that it is an array, because there can be many. What validating entry types are will be [explained next](./entry_type_definitions.md).
+`init` tags the previously mentioned `init` callback that Holochain expects from every Zome. [Skip here for details.](./init.md)'
 
-`init` represents the previously mentioned `init` callback that Holochain expects from every Zome. [Skip here for details.](./init.md)
+`validate_agent` tags the callback that Holochain also expects from every Zome and is used to control who can join.
 
-`functions` is where the functions are defined. [Skip here for details.](./zome_functions.md)
+`functions` must be defined inside the `my_zome` module. [Skip here for details.](./zome_functions.md)
```
jmday: This statement could be interpreted as the name `my_zome` is required for the module. Is there another way to state this that provides the audience with a clear understanding that zome modules are allowed to have any name?

doc/holochain_101/src/zome/define_zome.md
```diff
@@ -28,33 +28,33 @@ What are Rust macros? Generally speaking, they are code that will actually gener
...
 
-These are the three *required* properties of `define_zome!`.
+`entries` must also be defined within the `my_zome` module. Entries are defined by annotating a function that returns a `ValidatingEntryType` with `#[entry_def]`. What validating entry types are will be [explained next](./entry_type_definitions.md).
```
jmday: This statement could be interpreted as the name my_zome is required for the module. Is there another way to state this that provides the audience with a clear understanding that zome modules are allowed to have any name?

doc/holochain_101/src/zome/entry_type_definitions.md
```diff
@@ -208,97 +152,27 @@ Further reading can be found [here](./entry_validation.md).
...
         description: "A blog post entry which has an author",
         sharing: Sharing::Public,
-
         validation_package: || {
-            hdk::ValidationPackageDefinition::Entry
+            ValidationPackageDefinition::Entry
         },
-
-        validation: |_validation_data: Validation<Post>| {
+        validation: |validation_data: EntryValidationData<Post>| {
```
jmday: double-checking: is rust okay with not using the param validation_data? Would this benefit from a preceeding underscore (i.e. `_validation_data`)?

doc/holochain_101/src/bridging.md
```diff
@@ -56,28 +56,23 @@ pub fn handle_sample_function(some_param: String) -> ZomeApiResult<Address> {
...
+    #[init]
```
jmday: Is there a default `#[init]` implementation so that it's not required to define this to just get this simple `Ok(())` result?

doc/holochain_101/src/bridging.md
```diff
@@ -56,28 +56,23 @@ pub fn handle_sample_function(some_param: String) -> ZomeApiResult<Address> {
...
     }
 
-    functions: [
-        sample_function: {
-            inputs: |some_param: String|,
-            outputs: |result: ZomeApiResult<Address>|,
-            handler: handle_sample_function
-        }
-    ]
+    #[validate_agent]
```
jmday: Is there a default `#[validate_agent]` implementation so that it's not required to define this to just get this simple `Ok(())` result?

doc/holochain_101/src/first_steps.md
```diff
@@ -143,13 +135,16 @@ use hdk::{
 
 The `Serialize` and `Deserialize` derived traits allow the structs to be converted to and from JSON, which is how entries are managed internally in Holochain. The DefaultJson derived trait comes from the holochain HDK itself and allows for seamless converting between data stored in the DHT and rust structs.
 
-These structs on their own are not yet valid Holochain entries. To create these we must include them in the `define_zome!` macro by using the `entry!` macro:
+These structs on their own are not yet valid Holochain entries. To create an entry we must write a functio n that return a `ValidatingEntryType` and tag it using the `#[entry_def]` attribute. The `entry!` macro makes it easy to define a `ValidatingEntryType`.
```
zippy:
```suggestion
The List and ListItem structs on their own are not yet valid Holochain entries. To create an entry we must write a function that returns a `ValidatingEntryType` and tag it using the `#[entry_def]` attribute. The `entry!` macro makes it easy to define a `ValidatingEntryType`.
```

doc/holochain_101/src/first_steps.md
```diff
@@ -178,23 +177,25 @@ define_zome! {
...
 ```
 
-Take note of the `native_type` field of the macro which gives which Rust struct represents the entry type. The `validation_package` field is a function that defines what data should be passed to the validation function through the `ctx` argument. In this case we use a predefined function to only include the entry itself, but it is also possible to pass chain headers, chain entries or the full local chain. The validation field is a function that performs custom validation for the entry. In both our cases we are just returning `Ok(())`.
+Take note of the use of the struct types (`List` and `ListItem`) in the parameters to the validation function. This will automatically add validation to ensure these entries always match the structures. The `validation_package` field is a function that defines what data should be passed to the validation function through the `validation_data` argument. In this case we use a predefined function to only include the entry itself, but it is also possible to pass chain headers, chain entries or the full local chain. The validation field is a function that performs custom validation for the entry. In both our cases we are just returning `Ok(())`.
```
zippy:
```suggestion
Take note of the use of the struct types (`List` and `ListItem`) in the parameters to the validation function. This will automatically add validation to ensure these entries always match the structures, i.e. if a rouge node attempts to publish a List or ListItem that doesn't contain the data in the Struct definition, this will get caught by system level validation automatically.

The `validation_package` field is a function that defines what data should be passed to the validation function through the `validation_data` argument. In this case we use a predefined function to only include the entry itself, but it is also possible to pass chain headers, chain entries or the full local chain. The validation field is a function that performs custom validation for the entry. In both our cases we are just returning `Ok(())`.
```

doc/holochain_101/src/first_steps.md
```diff
@@ -178,23 +177,25 @@ define_zome! {
...
 
-For each of these functions we must define a handler, which is a Rust function that will be executed when the conductor calls the function. (For more on conductors, read Nico's recent post.) It is best practice for functions to always return a `ZomeApiResult<T>`, where `T` is the type the function should return if it runs without error. This is an extension of the Rust Result type and allows zome functions to abort early on errors using the `?` operator. At the moment the handler function names cannot be the same as the function itself so we will prefix them with `handle_`. This will be fixed in an upcoming release. The handler for `create_list` could be written as:
+Rust functions inside the zome module can be annotated with `#[zome_fn("hc_public")]` which will expose them to be callable from the conductor. The `"hc_public"` defines the capability required to call this function. This guide will not go in depth on capabilities but just know that hc_public means these functions are callable externally with no added security.
```
zippy:
```suggestion
Rust functions inside the zome module can be annotated with `#[zome_fn("hc_public")]` which will expose them to be callable from the conductor. The `"hc_public"` defines the capability required to call this function. This guide will not go in depth on capabilities but just know that hc_public means these functions are callable externally with no added security.  You can read more about capabilities [here](./zome/capabilities.md).
```

willemolding:
@jmday I'm a bit apprehensive about providing a default implementations of `validate_agent` that is fully permissive. Making a typo in one of the annotations would then result in the default being provided and the zome being made fully open.

It feels to me like requiring devs to be explicit when access to the zome should be open to everyone makes more sense. 

I guess we could do a default implementation for `init` since this is just a callback but something feels better to me about leaving it in. Anyone else have strong feelings @zippy @lucksus ?

doc/holochain_101/src/zome/define_zome.md
willemolding: I have a pretty good example of that in [init.md](https://github.com/holochain/holochain-rust/blob/06bbe308143325282583dbdeb2ad148208853201/doc/holochain_101/src/zome/init.md). Do you think this is ok or should I move it to here?

jmday: I'd recommend either
- showing simple example code that helps the reader understand how to use `#[validate_agent]` effectively
- ensure `#[validate_agent]` is optional with a reasonable default and discuss its use as a more advanced topic (the example in `init.md` is great)

I'd recommend the same for `#[init]`.

jmday:
> @jmday I'm a bit apprehensive about providing a default implementations of `validate_agent` that is fully permissive. Making a typo in one of the annotations would then result in the default being provided and the zome being made fully open.
> 
> It feels to me like requiring devs to be explicit when access to the zome should be open to everyone makes more sense.

I agree that code typos that lead to unfortunate defaults is an important consideration. I sense minimizing the amount of work/thought/understanding required to create a zome would help us drive dev adoption, though. And I envision devs testing that any custom agent validation they create works (I would, and I stink!) and the default of fully open seems right.

willemolding:
> > @jmday I'm a bit apprehensive about providing a default implementations of `validate_agent` that is fully permissive. Making a typo in one of the annotations would then result in the default being provided and the zome being made fully open.
> > It feels to me like requiring devs to be explicit when access to the zome should be open to everyone makes more sense.
> 
> I agree that code typos that lead to unfortunate defaults is an important consideration. I sense minimizing the amount of work/thought/understanding required to create a zome would help us drive dev adoption, though. And I envision devs testing that any custom agent validation they create works (I would, and I stink!) and the default of fully open seems right.

I'll admit it would be super cool to have the simplest possible zome be
```rust
#[zome]
mod zome { }
```

If we were to take this further we should probably also provide defaults for entry validation which could be nice. I'm still not sure if it feels right though. @zippy any thoughts?

jmday:
> I'll admit it would be super cool to have the simplest possible zome be
```
#[zome]
mod zome { }
```

That would be super-super-cool, indeed. And we're so close!

PR #1646, opened 2019-08-08
Passphrase via IPC

lucksus:
## PR summary

This adds a new config item to the conductor config: `passphrase_service`.
It is optional and the default setting is the same behaviour as before: passphrases are prompted via command line (i.e. stdout/stdin).

With this setting a new passphrase service `UnixSocket` can be selected like this:
```rust
[passphrase_service]
type = "unixsocket"
path = "/home/lucksus/.config/Holoscape/conductor-passphrase.sock"
```

That is exactly how [Holoscape](https://github.com/holochain/holoscape/) initializes the conductor config so that it can implement a desktop/windows passphrase prompt.

## testing/benchmarking notes

## followups

## changelog

### DIFF
--- a/app_spec/test/index.js
+++ b/app_spec/test/index.js
@@ -68,16 +68,9 @@ const registerAllScenarios = () => {
   // test cases if there is any Promise awaiting in between test declarations.
   let numRegistered = 0
 
-  const registerer = orchestrator => {
-    const f = (...info) => {
-      numRegistered += 1
-      return orchestrator.registerScenario(...info)
-    }
-    f.only = (...info) => {
-      numRegistered += 1
-      return orchestrator.registerScenario.only(...info)
-    }
-    return f
+  const registerer = orchestrator => (...info) => {
+    numRegistered += 1
+    return orchestrator.registerScenario(...info)
   }
 
   require('./regressions')(registerer(orchestratorSimple))
--- a/app_spec/test/index.js
+++ b/app_spec/test/index.js
@@ -68,9 +68,16 @@ const registerAllScenarios = () => {
   // test cases if there is any Promise awaiting in between test declarations.
   let numRegistered = 0
 
-  const registerer = orchestrator => (...info) => {
-    numRegistered += 1
-    return orchestrator.registerScenario(...info)
+  const registerer = orchestrator => {
+    const f = (...info) => {
+      numRegistered += 1
+      return orchestrator.registerScenario(...info)
+    }
+    f.only = (...info) => {
+      numRegistered += 1
+      return orchestrator.registerScenario.only(...info)
+    }
+    return f
   }
 
   require('./regressions')(registerer(orchestratorSimple))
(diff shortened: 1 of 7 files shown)


conductor_api/src/conductor/passphrase_manager.rs
```diff
@@ -125,3 +127,80 @@ impl PassphraseService for PassphraseServiceMock {
...
+        while self.stream.lock().unwrap().is_none() {
+            log_debug!("No one connected via socket yet. Waiting...");
+            thread::sleep(Duration::from_millis(500));
+        }
+
+        log_debug!("We have an open connection to a passphrase provider.");
+
+        // Request and read passphrase from socket
+        let mut passphrase_string = {
+            let mut stream_option = self.stream.lock().unwrap();
```
willemolding: A lot of unwraps and expects in this block. Since the containing function returns a Result it might be better to early return the errors

lucksus: I've replaced all `unwrap`s with `expect`s which should give enough context if triggered. In this case I regard expects to be the right choice since the conductor can't start and do its job if there is no way to get a passphrase from the user, and this way we have the full stack trace when that happens, which we would not if we return an error that then lets the process panic somewhere else...

app_spec/test/index.js
```diff
@@ -68,16 +68,9 @@ const registerAllScenarios = () => {
...
-    const f = (...info) => {
-      numRegistered += 1
-      return orchestrator.registerScenario(...info)
-    }
-    f.only = (...info) => {
-      numRegistered += 1
-      return orchestrator.registerScenario.only(...info)
-    }
-    return f
+  const registerer = orchestrator => (...info) => {
```
zippy: It looks like a merge mistake here removing the .only

conductor_api/src/conductor/passphrase_manager.rs
```diff
@@ -125,3 +129,86 @@ impl PassphraseService for PassphraseServiceMock {
...
+    fn drop(&mut self) {
+        std::fs::remove_file(self.path.clone()).unwrap();
+    }
+}
+
+#[cfg(unix)]
+impl PassphraseService for PassphraseServiceUnixSocket {
+    fn request_passphrase(&self) -> Result<SecBuf, HolochainError> {
+        log_debug!("Passphrase needed. Using unix socket passphrase service...");
+        while self.stream.lock().unwrap().is_none() {
```
zippy: I think we probably should timeout here at some point no?

conductor_api/src/config.rs
```diff
@@ -107,6 +107,34 @@ pub struct Configuration {
     /// Which signals to emit
     #[serde(default)]
     pub signals: SignalConfig,
+
+    /// How should the conductor prompt the user for the passphrase to lock/unlock keystores?
```
zippy:
```suggestion
    /// Configure how the conductor should prompt the user for the passphrase to lock/unlock keystores.
```

zippy:
Actually @lucksus this needs to be fixed for windows (see the failures in appveyor) with non-used includes.

conductor_api/src/conductor/passphrase_manager.rs
lucksus: The only reasonable thing to do then would be to panic.. why not keep waiting?

app_spec/test/index.js
lucksus: yeah, fixed.

PR #1647, opened 2019-08-08
Log buffer flush & perf improvements

dymayday:
## PR summary

Add the possibility to flush the buffer writer from the logging thread and bring back the use of [`BufWriter`](https://doc.rust-lang.org/std/io/struct.BufWriter.html) for performance reasons.

## testing/benchmarking notes

## followups

## changelog

### DIFF
--- a/logging/src/lib.rs
+++ b/logging/src/lib.rs
@@ -474,7 +474,6 @@ impl<'a> FastLoggerBuilder {
 
     /// Registers a [FastLogger] as the comsumer of [log] facade so it becomes static and any further
     /// mutation are discarded.
-    // #[allow(clippy::let_and_return)]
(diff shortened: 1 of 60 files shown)


logging/src/lib.rs
```diff
@@ -420,41 +506,66 @@ impl<'a> FastLoggerBuilder {
                             .append(true)
                             .open(&file_path)
                             .unwrap_or_else(|_| panic!("Fail to log to {:?}.", &file_path));
-                        // We have some strange behavior with BufWriter and file writing: The file
-                        // ends up empty most of the time during testing, so we don't use a buffer
-                        // at all like : `io::BufWriter::new(file_stream)`
-                        file_stream
+                        // Here we use `writeln!` instead of println! in order to avoid
```
zippy: I think this is an extra copy of the writln comment

PR #1652, opened 2019-08-12
Fix log filtering

dymayday:
## PR summary

This is related to [this issue](https://github.com/holochain/lib3h/issues/244) and propose a fix by disabling the filtering on the `args` log record to only apply it to the `target` log record. 
Which I think is still OK and less error prone / confusing from the user perspective.

## testing/benchmarking notes

## followups

## changelog


dymayday:
PR from wrong branch

PR #1658, opened 2019-08-13
Hdk warning version mismatch

AshantiMutinta:
## testing/benchmarking notes

## followups

## changelog


AshantiMutinta:
Found something, not ready for review yet

AshantiMutinta:
Rewrote warning message. Ready for review now

willemolding:
One high level problem that might come up is assumpton that all zome code was compiled by the Rust HDK

If I understand correctly `hc package` would expect to be able to call `meta` on the zomes and they would return the HDK version they were compiles against. Rather than assuming the Rust HDK could we instead have a zome report that it *was* compiled with the Rust HDK and a particular version. Then later we can add support for other languages checking their HDK versions.

cli/src/cli/package.rs
```diff
@@ -220,6 +241,36 @@ impl Packager {
...
+                            .map(|mut read_file|{
+                                let mut contents = String::new();
+                                read_file
+                                .read_to_string(&mut contents)
+                                .map(|_|{
+                                    hdk_version_compare(&HDK_VERSION,&*contents)
+                                    .map(|hdk_match|{
+                                        if let false = hdk_match
+                                        {
+                                            eprintln!("There has been a version mismatch in the current zome. Current HDK Version is {}",HDK_VERSION.to_string())
```
zippy: Here's a wording suggestion for the error, and also we should include the file name.  If possible, can we get the version found out of the contents to print that too?  i.e. hdk_version_compare could return the value found so we can print it in the warning?  Something like:
```suggestion
                                            eprintln!("WARNING: The HDK version found in {:?} does not match the current version.\n If you are seeing compilation problems, update the version in your Cargo.toml files to the current version: {}", read_path, HDK_VERSION.to_string())
```

cli/src/cli/package.rs
```diff
@@ -332,14 +383,11 @@ fn unpack_recurse(mut obj: Object, to: &PathBuf) -> DefaultResult<()> {
...
 // too slow!
-#[cfg(feature = "broken-tests")]
 mod tests {
     use super::*;
-    use crate::cli::init::tests::gen_dir;
-    use assert_cmd::prelude::*;
-    use std::{path::PathBuf, process::Command};
 
     #[test]
+    #[cfg(feature = "broken-tests")]
```
zippy: Why is this test broken now?

cli/src/cli/scaffold/rust.rs
```diff
@@ -46,11 +46,8 @@ fn generate_cargo_toml(name: &str, contents: &str, template: &str) -> DefaultRes
     let edition_default = Value::from("\"TODO\"");
 
     let maybe_version = EnvVar::ScaffoldVersion.value().ok();
-    let version_default = if maybe_version.is_some() {
-        maybe_version.unwrap()
-    } else {
-        String::from("tag = \"0.0.25-alpha1\"")
-    };
+    let version_default =
```
zippy: This is great because it means we can stop trying to remember to update it...

core/src/nucleus/ribosome/api/meta.rs
```diff
@@ -0,0 +1,68 @@
...
+pub fn invoke_meta(runtime: &mut Runtime, args: &RuntimeArgs) -> ZomeApiResult {
+    let context = runtime.context()?;
+
+    let args_str = runtime.load_json_string_from_args(&args);
+    let meta_args = match MetaArgs::try_from(args_str.clone()) {
+        Ok(args) => args,
+        // Exit on error
+        Err(error) => {
+            log_error!(context,
+                "zome: invoke_emit_signal failed to \
```
zippy: This error text looks wrong.. 
```suggestion
                "zome: invoke_meta failed to \
```

cli/src/cli/package.rs
AshantiMutinta: Absolutely, will implement that right now

cli/src/cli/package.rs
AshantiMutinta: The whole package test suite has been broken for a while. In order to add another test I had to remove the root "broken tests" feature, add a new test but then add this feature to every other test

AshantiMutinta:
> One high level problem that might come up is assumption that all zome code was compiled by the Rust HDK
> 
> If I understand correctly `hc package` would expect to be able to call `meta` on the zomes and they would return the HDK version they were compiled against. Rather than assuming the Rust HDK could we instead have a zome report that it _was_ compiled with the Rust HDK and a particular version. Then later we can add support for other languages checking their HDK versions.

I think this is more of a holochain core version rather than an HDK one, it could be a problem high level definitely as to where HDK versions are behind but the core version is in front. I think this works for now but something that we could address once we start exploring HDKs in other languages

pdaoust:
hey, naïve question... I know that some libs' version numbers don't sync with Holochain proper (for instance, `holochain_json_derive`. I'm wondering:

* Does it matter re: HDK and Holochain versions if someone's targeting HDK 0.0.27 but `holochain_json_derive` 0.0.2 instad of 0.0.14?
* If so, should this PR also check versions of dependent libs?
* Regardless, there may be a parallel issue of dependency management, and I'd argue that we don't want to put this burden on app devs. Inquiry: how do we manage this? I don't want to clutter up this PR, so I'm going to create a separate issue.

cli/src/cli/package.rs
zippy: ok, got it, that makes sense.

zippy:
> One high level problem that might come up is assumption that all zome code was compiled by the Rust HDK
> 
> If I understand correctly `hc package` would expect to be able to call `meta` on the zomes and they would return the HDK version they were compiled against. Rather than assuming the Rust HDK could we instead have a zome report that it _was_ compiled with the Rust HDK and a particular version. Then later we can add support for other languages checking their HDK versions.

The version being reported here isn't the version that the zome was compiled under, it's the version that the conductor was compiled under.

cli/src/cli/package.rs
```diff
@@ -35,6 +39,23 @@ pub const META_CONFIG_SECTION_NAME: &str = "config_file";
 
 pub type Object = Map<String, Value>;
 
+fn hdk_version_compare(hdk_version: &HDKVersion, cargo_toml: &str) -> DefaultResult<bool> {
+    let toml: Value = toml::from_str(cargo_toml)?;
+    let dependancies = toml
+        .get("dependencies")
+        .ok_or(format_err!("Could not get dependancies"))?;
```
lucksus:
```suggestion
        .ok_or(format_err!("Could not get dependencies"))?;
```

PR #1659, opened 2019-08-13
Update ConsistencyModel to be aware of InitializeChain

maackle:
## PR summary

Updates the ConsistencyModel with the knowledge that upon `InitializeNetwork`, every other Validator of the AgentId should eventually `Hold` that AgentId.

TODO:

- [ ] find out if and where AgentId is published (just examine all actions reduced in a scenario test with println)

## changelog


zippy:
reverted to println for the short term.

Could we add a new set of tests like this one:

```

scenario('agent_id_gets_gossiped_on_startup', async (s, t, { alice, bob }) => {
  //confirm that bob can see alices' agent id
  const alice_agent_id = await bob.app.callSync("simple", "get_entry", { "address":alice.app.agentId })
  console.log("ALICE AGENT ID:"+alice.app.agentId)
  console.log(JSON.stringify(alice_agent_id.Ok))
  t.ok(alice_agent_id.Ok)
})
```
that confirm this model?

maackle:
I can add such a test in @zippy but the larger problem is it's failing because of logger initialization

maackle:
I think this is actually wrong. When I run this locally, this line panics every time: https://github.com/holochain/holochain-rust/pull/1659/files#diff-1b0244be5624b8a8ef74bbf81e8cc775R184

core/src/consistency.rs
```diff
@@ -166,6 +179,16 @@ impl ConsistencyModel {
             Action::ReturnZomeFunctionResult(result) => Some(ConsistencySignal::new_terminal(
                 ReturnZomeFunctionResult(result.call().id()),
             )),
+            Action::InitializeChain(_) => {
+                let agent_id = self.agent_id.as_ref().expect(
+                    "ConsistencyEvent::InitializeChain encountered before AgentId committed",
+                );
```
maackle: `InitializeChain` is what commits the `AgentId` so this will never work!

maackle:
This PR is probably unnecessary if AgentId is published, because the consistency model already handles the causality between Publish and Hold.

Here is where the AgentId is committed: https://github.com/holochain/holochain-rust/blob/quadrama/core/src/nucleus/actions/initialize.rs#L94
Where is it published? Is it?

PR #1673, opened 2019-08-19
Spike - Add `call_and_measure` to conductor for Holo service logger

willemolding:
## PR summary

This is a spike of a possible implementation for resource measurement for calls made to the conductor. 

It adds a new RPC method `call_and_measure` to conductor admin interfaces. This is identical to the regular `call` except it spawns a monitor thread which polls the resource usage every 100ms. When the call is completed it generates a report. The measuring and reporting is done via the [self-meter](https://docs.rs/self-meter/0.6.0/self_meter/) crate. The [schema of a report](https://docs.rs/self-meter/0.6.0/self_meter/struct.Report.html) is used directly from that crate for now and contains many fields some of which may be useful.

The response from `call_and_measure` contains the regular call result and the resource report. An example response looks as follows:
```json
{
	"call_response": {...},
	"resource_usage": {
           "timestamp": 0
	    "duration": 0
	    "start_time": 0
	    "system_uptime": 0
	    "global_cpu_usage": 0
	    "process_cpu_usage": 0
	    "gross_cpu_usage": 0
	    "memory_rss": 0
	    "memory_virtual": 0
	    "memory_swap": 0
	    "memory_rss_peak": 0
	    "memory_virtual_peak": 0
	    "memory_swap_peak": 0
	    "disk_read": 0
	    "disk_write": 0
	    "disk_cancelled": 0
	    "io_read": 0
	    "io_write": 0
	    "io_read_ops": 0
	    "io_write_ops": 0
	} 
}
```

Since this method for getting usage measurements is only supported on Linux this RPC function will only be available on that platform. This should be ok for Holo since it only needs to work on HoloportOS.

Currently this spike is up for feedback from relevant parties as to whether this provides an acceptable MVP for use by the service logger.

### DIFF
--- a/conductor_api/src/interface.rs
+++ b/conductor_api/src/interface.rs
@@ -203,61 +203,70 @@ impl ConductorApiBuilder {
         });
     }
 
-    pub fn with_call_with_metrics(mut self) -> Self {
+    /// Adds a call and measure function to the conductor. This is currently only available on linux
+    #[cfg(target_os = "linux")]
+    pub fn with_call_and_measure(mut self) -> Self {
         let instances = self.instances.clone();
         let instance_ids_map = self.instance_ids_map.clone();
 
-        self.io.add_method("call", move |params| {
-            let params_map = Self::unwrap_params_map(params)?;
-            let public_id_str = Self::get_as_string("instance_id", &params_map)?;
+        self.io.add_method("call_and_measure", move |params| {
+            if cfg!(target_os = "linux") {
+                let params_map = Self::unwrap_params_map(params)?;
+                let public_id_str = Self::get_as_string("instance_id", &params_map)?;
 
-            let id = instance_ids_map
-                .get(&PublicInstanceIdentifier::from(public_id_str))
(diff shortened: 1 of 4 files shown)


conductor_api/src/interface.rs
```diff
@@ -138,74 +187,90 @@ impl ConductorApiBuilder {
...
+
+                let hc_lock = instance.clone();
+                let hc_lock_inner = hc_lock.clone();
+                let mut hc = hc_lock_inner.write().unwrap();
+
+                let (zome_name, func_name, args_string) = Self::get_call_args(&params_map)?;
+                let cap_request = Self::get_cap_request(hc.context().unwrap(), &func_name, &params_map, &args_string)?;
+
+                // setup the self-meter in another thread to monitor this one
+                let mut meter = self_meter::Meter::new(Duration::from_millis(100)).unwrap();
```
willemolding: This is the core of it here. A monitor thread is started which scans every 100ms and computes a report when the call has finished which is averaged from the scans.

PR #1675, opened 2019-08-20
WIP: DPKI root- and device-seed generation through hc

lucksus:
## PR summary
Currently: 
- Adds parameters  `--root_seed` and `--device_derivation_index`to `hc keygen` that will make it derive a device seed from that root seed and then derive the agent key from that device seed.

- If only the device derivation index is provided it will prompt the user to input their root seed securely through stdin.

- Also adds new command `hc dpki-init` which will generate a new random root seed and output the mnemonic to stdout

- Adds new command `hc sign` which will request a root seed and return the signature of a given message with the revocation key. This is a feature for the upcoming demo only and will be rejigged in the future

Still needed:
 - [ ] Encrypting generated root seed with passphrase
 - [ ] Decrypting root seed with passphrase

## testing/benchmarking notes

## followups

## changelog


zo-el:
@lucksus @willemolding I was wondering why we call the cmd to create a root seed as `hc dpki-init`? is it doing anything more than that?

Because I think that would create confusion because people could expect it to return values that are needed to be passed as params in the `[dpki]` conductor settings. Also, the dpki trait (zome function) is now called `init_dpki`

zo-el:
Also, Could we update the `hc keygen` to default generate a `device_seed` instead of  `root_seed`.
Because this cmd would be used to generate a quick `device_seed` to spin up a quick conductor.

What do you guys think?

cli/src/cli/sign.rs
```diff
@@ -0,0 +1,34 @@
+use crate::cli::keygen::get_root_seed;
+use holochain_core_types::error::HcResult;
+use lib3h_sodium::secbuf::SecBuf;
+
+
+pub fn sign(root_seed_string: Option<String>, message: String) -> HcResult<String> {
+    let mut root_seed = get_root_seed(root_seed_string, &String::from(""), false)?;
+    let mut revocation_keypair = root_seed.generate_revocation_key()?;
```
zo-el: @willemolding, We may need to be able to pass in the revocation key. Because the revocation key can be changed, so the revocation key won't have a fixed derivation path.

Also, I don't think it's wise to always have to use your master seed for revocation.

willemolding:
Ok it sounds like we need to take a step back and actually figure out the requirements. All these things are just re-jigs of what is there so it should be a really quick thing once we have spec-ed.

@zo-el would you mind writing something up real quick with the requirements for hc for the demo with the inputs and outputs (and formats).

PR #1688, opened 2019-08-27
Feature/dpki cli commands

willemolding:
## PR summary

Adds some new commands to the CLI tool for managing DPKI keys.

- `hc dpki genroot` creates a new random root seed
- `hc dpki keygen` is similar to `hc keygen` but will use a dpki root seed mnemonic so that keys can be recovered
- `hc dpki genrevoke/genauth` produce revocation/auth seeds from which keys can be derived. 
- `hc dpki sign` use a revocation/auth seed to produce a key then sign a message (which can be an agent key). This is the process by which keys can be authorized and revoked

Each time a mnemonic is produced it is optionally encrypted with a passphrase

TODO: DRY up the auth/revoke stuff

## testing/benchmarking notes

## followups

## changelog


cli/src/cli/keygen.rs
```diff
@@ -1,46 +1,105 @@
...
+        (Some(s), false) => s,
+        (Some(_), true) => panic!(
+            "Invalid combination of args. Cannot pass --nullpass and also provide a passphrase"
+        ),
+        (None, false) => {
+            // prompt for the passphrase
+            user_prompt(
+                "Please enter a secret passphrase below. You will have to enter it again
+when unlocking the keybundle to use within a Holochain conductor.\n",
+                quiet,
```
AshantiMutinta: What do you think of utilizing combinators for this one and just utilizing the null_pass boolean?

keystore_passphrase
.map(|s|{
    if nullpass = true{panic!("do this)}else{s}
})
.unwrap_or_else(||{
     //userprompt
	 if nullpass = true
     {
	     get_secure_string_double_check("keystore Passphrase", quiet)
	     .expect("Could not retrieve passphrase")
     }
     else
     {
         String::from(holochain_common::DEFAULT_PASSPHRASE)
	 }	 
	 
})

cli/src/util.rs
```diff
@@ -1,13 +1,80 @@
...
 use std::{
     fs,
-    io::ErrorKind,
+    io::{self, ErrorKind, Write},
     path::PathBuf,
     process::{Command, Stdio},
 };
 
+pub fn get_secure_string_double_check(name: &str, quiet: bool) -> HcResult<String> {
+    if !quiet {
```
AshantiMutinta: can we use the user_prompt here?

dpki/src/seed.rs
```diff
@@ -227,6 +245,18 @@ impl DeviceSeed {
...
+    /// Generate Auth Seed
+    /// @param {number} index - the index number in this seed group, must not be zero
+    pub fn generate_auth_seed(
+        &mut self,
+        index: u64,
+    ) -> HcResult<AuthSeed> {
+        let seed_context = SeedContext::new(AUTH_CTX);
+        let seed_buf =
+            generate_derived_seed_buf(&mut self.inner.buf, &seed_context, index, SEED_SIZE)?;
+        Ok(AuthSeed::new(seed_buf))
```
AshantiMutinta: It is overkill but we could introduce an enum for the request and result in the end we have

fn generate_seed(&mut self, index:i64, SeedType) -> SeedResult; 

where SeedType contains Auth, Revocation variants and the SeedResult contains the same, however not necessary if they are only two imo but could be a possibility if they are more

dpki/src/seed.rs
```diff
@@ -271,6 +301,78 @@ impl DevicePinSeed {
...
+    /// Generate a revocation key
+    pub fn generate_auth_key(&mut self, derivation_index: u64) -> HcResult<KeyBundle> {
+        let mut ref_seed_buf = SecBuf::with_secure(SEED_SIZE);
+        let context = SeedContext::new(DEVICE_CTX);
+        let mut context = context.to_sec_buf();
+        kdf::derive(&mut ref_seed_buf, derivation_index, &mut context, &mut self.inner.buf)?;
+        Ok(KeyBundle::new_from_seed_buf(&mut ref_seed_buf)?)
+    }
+}
+
```
AshantiMutinta: A few parallels between the code I think we could actually introduce a single Seed Enum which has Auth and Revocation variants so that for any difference we need we would just match on

pjkundert:
When holochain-core has upgrade to lib3h version 0.0.11, we can add the `feature/dpki-cli-commands-get-mnemonic` branch's changes to `dpki/src/seed.rs` here.  This allows the MnemonicableSeed's get_mnemonic method to operate on immutable Seed's SecBuf, allowing us to write Debug, Display and De/Serialize Trait methods (which require immutable references).

pjkundert:
It seems that the keys produced by `hc keygen` are not compatible with `holochain`; after generating a key using `hc keygen`, starting a `holochain` conductor using a config file that references that key produces:
```
Error while trying to boot from config: ConfigError("Could not decrypt \'primary_keybundle:sign_key\': ErrorGeneric(\"Invalid SigningKey Blob Format: v0 != v1\")")
```

We'll have to ensure that this pull corrects that.

PR #1691, opened 2019-08-31
Serialize `ConsistencySignal` before transmitting across signal channel

maackle:
## PR summary

Hachiko expects observable "events" as strings which match up exactly, coming from different sources at different times. Currently the serialization is happening on the JS side via `JSON.stringify`, but this is not great since `JSON.stringify` is not deterministic. This PR uses serde to do the serialization, under the assumption that serde_json is deterministic.

## testing/benchmarking notes

This will probably break current app spec tests since try-o-rama currently does the serialization itself. I need to update the npm package for try-o-rama to not do this for this PR.

## changelog

### DIFF
--- a/core/src/consistency.rs
+++ b/core/src/consistency.rs
@@ -2,44 +2,26 @@ use crate::{
     action::Action, context::Context, entry::CanPublish,
     network::entry_with_header::EntryWithHeader,
 };
-use holochain_core_types::{entry::Entry, link::link_data::LinkData};
+use holochain_core_types::{agent::AgentId, entry::Entry, link::link_data::LinkData};
 use holochain_persistence_api::cas::content::{Address, AddressableContent};
+use serde::Serialize;
 use std::{collections::HashMap, sync::Arc};
 
-#[derive(Clone)]
-pub struct ConsistencyModel {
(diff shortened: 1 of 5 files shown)


core/src/signal.rs
```diff
@@ -10,7 +10,7 @@ use std::thread;
 #[serde(tag = "signal_type")]
 pub enum Signal {
     Trace(ActionWrapper),
-    Consistency(ConsistencySignal),
+    Consistency(ConsistencySignal<String>),
```
maackle: This, and the new From impl which goes from enum to String, is the main crux of the PR.

core/src/consistency.rs
```diff
@@ -51,6 +33,26 @@ impl ConsistencySignal {
...
+                .into_iter()
+                .map(|p| PendingConsistency {
+                    event: serde_json::to_string(&p.event)
+                        .expect("ConsistencySignal serialization cannot fail"),
+                    group: p.group,
+                })
+                .collect(),
+        }
+    }
+}
```
maackle: This is the main new part, most of the rest of the PR is just moving stuff around.

maackle:
NB: I had to rerun tests 3 times to get this to pass

PR #1692, opened 2019-09-03
Update test agent config with correct public_address

maackle:
## PR summary

Modify the config so that the public_address is correct. When loading from config, the public_address is simply ignored for test_agents, as it is generated from the agent's name instead of read from a physical keyfile. However, `admin/agent/list` uses the value from the config, which is, in general, wrong. So, this makes the value in the config right so that the admin method can report the right value.

## changelog


CHANGELOG-UNRELEASED.md
```diff
@@ -7,7 +7,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 * Adds EncryptedSeed and seed.encrypt() allow for easy passphrase encrypting/decrypting of any of the existing seed types. Adds the MnemonicableSeed trait allows seeds to be converted to/from BIP39 mnemonics. [#1687](https://github.com/holochain/holochain-rust/pull/1687) 
-- added nix for `hc-conductor-install` and `hc-conductor-uninstall` based on `cargo`
+* added nix for `hc-conductor-install` and `hc-conductor-uninstall` based on `cargo`
```
zippy: looks like this got added back in incorrectly.

maackle: I just changed the `-` to `*` for consistency

PR #1695, opened 2019-09-04
Attempting to turn back some tests on

AshantiMutinta:
## testing/benchmarking notes

## followups

## changelog


core/src/workflows/hold_link.rs
```diff
@@ -127,37 +122,32 @@ pub mod tests {
...
-        // Address of the original DNA
-        let dna_address = dna.address();
 
-        let (_, context1) =
-            test_instance_with_spoofed_dna(hacked_dna, dna_address, "alice").unwrap();
-        let netname = Some("test_reject_invalid_link_on_remove_workflow");
+        let mut dna = test_dna();
+        dna.uuid = "test_reject_invalid_link_on_hold_workflow".to_string();
+        let netname = Some("test_reject_invalid_link_on_hold_workflow, the network");
+        let (_instance1, context1) = instance_by_name("jill", dna.clone(), netname);
```
maackle: Hmm is this really the same test? In the new version you're cloning the same DNA twice, but in the old it seems to use a hacked version that always validates, which is missing in the new version.

AshantiMutinta: This test doesn't run, I was trying to fix it but after further looking at it, it was not worth it atm

core/src/workflows/hold_link.rs
```diff
@@ -128,7 +123,7 @@ pub mod tests {
     /// hold_link_workflow is then expected to fail in its validation step
     fn test_reject_invalid_link_on_hold_workflow() {
         // Hacked DNA that regards everything as valid
-        let hacked_dna = create_test_dna_with_wat("test_zome", Some(&test_wat_always_valid()));
+         let hacked_dna = create_test_dna_with_wat("test_zome", Some(&test_wat_always_valid()));
```
maackle: whitespace

core/src/workflows/hold_link.rs
```diff
@@ -169,12 +165,13 @@ pub mod tests {
             entry: link_entry,
             header,
         };
-
+         println!("before hold link");
```
maackle: remove printlns?

AshantiMutinta: From the broken test still but will remove it

PR #1697, opened 2019-09-05
AppSpec Migration : Misc Phase

AshantiMutinta:
## PR summary
-Tests migration of misc tests, tests in app_spec down from 49 to 45
## testing/benchmarking notes

## followups

## changelog


maackle:
I still see `(_, _,_)`, maybe fmt doesn't hit the tests? It's OK though we can get it in another PR.

AshantiMutinta:
> I still see `(_, _,_)`, maybe fmt doesn't hit the tests? It's OK though we can get it in another PR.

I think it does but I remember app-spec was broken for this distribution. almost like it doesnt go the whole way but you usually get this :

thread 'main' panicked at 'bad span: `.`: `)`', src/tools/rustfmt/src/source_map.rs:52:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

question of whether it goes the full way because of that panic

PR #1702, opened 2019-09-10
Intergration waiter

AshantiMutinta:
## PR summary
Lightweight Waiter For the Intergration Test
## testing/benchmarking notes

## followups

## changelog


hdk-rust/tests/integration_test.rs
```diff
@@ -302,10 +303,67 @@ fn example_valid_entry_address() -> Address {
...
+                    let content = cons.clone().event();
+                    let event: ConsistencyEvent =
+                        serde_json::from_str(&content).expect("Could not convert serde");
+                    match event {
+                        ConsistencyEvent::Hold(_) => Some(event),
+                        ConsistencyEvent::AddLink(_) => Some(event),
+                        ConsistencyEvent::UpdateEntry(_, _) => Some(event),
+                        ConsistencyEvent::RemoveEntry(_, _) => Some(event),
+                        ConsistencyEvent::RemoveLink(_) => Some(event),
+                        _ => None,
```
maackle: Why are you filtering out certain signals? Wouldn't it be OK for the user to specify any signal to wait for?

hdk-rust/tests/integration_test.rs
```diff
@@ -302,10 +303,67 @@ fn example_valid_entry_address() -> Address {
...
+                    }
+                }
+                _ => None,
+            })
+            //check if signal matches expected signal
+            .find(|recv| {
+                recv == &expected_signal
```
maackle: Since this is not using the ConsistencySignal `pending` field, which is the main reason for using consistency signals in the first place, I would suggest switching this function to use an Action channel. The main reason is that it's going to be more onerous for the user to construct a `ConsistencySignal` than an `Action` for `expected_signal` here, because the consistency signal also has to include the pending event data, which is extraneous in this case, and if it doesn't match up perfectly then this match will be false.

hdk-rust/tests/integration_test.rs
```diff
@@ -302,10 +303,67 @@ fn example_valid_entry_address() -> Address {
...
+) -> Result<ConsistencyEvent, HolochainError> {
+    if max_tries > 0 {
+        signal_receiver
+            //THIS FIRST PART IS JUST FOR OPTIMIZATION PURPOSES WHILE FINDING A GOOD BALANCE OF SEPERATION
+            //============================================================================================
+            .try_iter()
+            //skip the first
+            .skip(up_till.unwrap_or(0))
+            //only iterate over the 10
+            .take(10)
```
maackle: You don't want this line in the final version do you?

hdk-rust/tests/integration_test.rs
```diff
@@ -302,10 +303,67 @@ fn example_valid_entry_address() -> Address {
...
+#[allow(dead_code)]
+fn wait_for_action(
+    signal_receiver: &SignalReceiver,
+    expected_signal: ConsistencyEvent,
+    max_tries: i8,
+    up_till: Option<usize>,
+) -> Result<ConsistencyEvent, HolochainError> {
+    if max_tries > 0 {
+        signal_receiver
+            //THIS FIRST PART IS JUST FOR OPTIMIZATION PURPOSES WHILE FINDING A GOOD BALANCE OF SEPERATION
```
maackle: What do you mean by balance of separation?

hdk-rust/tests/integration_test.rs
AshantiMutinta: The purpose is just to reduce the sample size that the find checks against, maybe a panic would fair better than a non

hdk-rust/tests/integration_test.rs
AshantiMutinta: Will switch this to use the action

hdk-rust/tests/integration_test.rs
AshantiMutinta: Seperation of concerns from a high order function perspective but will add that

AshantiMutinta:
Merging into entry intergration test branch and closing PR. makes more sense contextually but carrying over changes

PR #1713, opened 2019-09-17
Cause hc-test-fmt to error if any errors

maackle:
## PR summary

`hc-test-fmt` stopped failing. This makes it fail when there are fmt diffs.

However `hn-rust-fmt` is also broken, due to this error:

    thread 'main' panicked at 'bad span: `.`: `)`', src/tools/rustfmt/src/source_map.rs:52:13

Which is referenced here: https://github.com/rust-lang/rustfmt/issues/3666. So, we have no way to fix the fmt errors that are now being reported by CI

## changelog

thread 'main' panicked at 'bad span: `.`: `)`', src/tools/rustfmt/src/source_map.rs:52:13


maackle:
@thedavidmeister everything is was a manual rustfmt run, because the holonix fmt fix isn't working, and we already have a bunch of code that fails the fmt check

thedavidmeister:
@maackle oh sure, that makes sense, i still approve

maackle:
Fixed in #1822

PR #1718, opened 2019-09-22
Dont update if new entry hash is the same as old entry hash

AshantiMutinta:
## PR summary
Throws a workflow error if new entry hash is the same as old entry hash
## testing/benchmarking notes

## followups

## changelog


willemolding:
I'm glad we are starting to think about this problem of cyclic updates. This is a good solution to the problem of 1-entry loops (updating to an entry to the same hash) but I wonder if we could come up with a more general solution to resolve larger loops (e.g. updating A->B->C->A) and even to figure out what is the correct thing to return in this case.

Maybe we need to do something similar to with links and reference the top chain header of the publishing agent in the update/delete. This would unroll the graph into a tree (since two agents may update the same entry simultaneously or in partitions which then merge). We could provide a resolution method (e.g. highest timestamp wins) but also have the ability to return the entire tree in a call to get_entry so zome/consuming code can resolve it itself.

Whatcha think?

AshantiMutinta:
>  similar to with links and reference the top chain header of the publishing agent in the update/delete. This wo

I think that could definitely work and goes more into the section of Conflict Resolution which will be one of our big items going forward. 

> 
> 
> I'm glad we are starting to think about this problem of cyclic updates. This is a good solution to the problem of 1-entry loops (updating to an entry to the same hash) but I wonder if we could come up with a more general solution to resolve larger loops (e.g. updating A->B->C->A) and even to figure out what is the correct thing to return in this case.
> 
> Maybe we need to do something similar to with links and reference the top chain header of the publishing agent in the update/delete. This would unroll the graph into a tree (since two agents may update the same entry simultaneously or in partitions which then merge). We could provide a resolution method (e.g. highest timestamp wins) but also have the ability to return the entire tree in a call to get_entry so zome/consuming code can resolve it itself.
> 
> Whatcha think?

I think that definitely we would need a more general solution for the conflict resolution of bigger loops. I would love to do a more general solution in another branch since this is a quick fix for a problem which philip had but I have no doubt that we could generalize it.

Something to put in the tree once the networking dies down?

[also I just realized that I need to check this against the latest entry rather than the address passed in]

PR #1724, opened 2019-09-27
Add UUID to DnaConfig and disable integrity check if UUID is specified

maackle:
## PR summary

Allow dna.uuid to be specified as config. Previously the only way to set a UUID on a dna was by installing through the admin interface.

## changelog


conductor_api/src/conductor/base.rs
```diff
@@ -750,7 +750,11 @@ impl Conductor {
 
                 // Get DNA
 
-                let mut dna_config = self.config.dna_by_id(&instance_config.dna).unwrap();
+                // self.config.dnas.iter_mut().fing(|dna_config| dna_config.id == instance_config.dna)
+                // .map(|dna_config| {
+
+                // })
```
timotree3: Did you mean to leave this comment here?

conductor_api/src/conductor/base.rs
```diff
@@ -750,7 +750,11 @@ impl Conductor {
 
                 // Get DNA
 
-                let mut dna_config = self.config.dna_by_id(&instance_config.dna).unwrap();
```
timotree3: Was this a latent bug before? `dna_by_id` clones its result so I doubt that mutating it has the desired effect. If so, it might be good to add a regression test.

conductor_api/src/config.rs
```diff
@@ -478,6 +478,15 @@ impl Configuration {
         self.dnas.iter().find(|dc| &dc.id == id).cloned()
     }
 
+    /// Returns the DNA configuration with the given ID if present
```
timotree3: I think this comment got copy-pasted unchanged :)
```suggestion
    /// Updates the hash of the DNA with the given ID, returns true if it was present.
```

conductor_api/src/conductor/base.rs
```diff
@@ -758,58 +749,72 @@ impl Conductor {
...
-                                    &dna_hash_computed);
-                                    log_error!(context, "{}", msg);
-
-                                    return Err(e.to_string());
+                                    &dna_hash_computed,
+                                    &dna_hash_computed_from_file, &dna_file)?;
+                            },
+                            Err(_) => {
+                                let msg = format!("Conductor: Could not load DNA file {:?}.", &dna_file);
+                                log_error!(context, "{}", msg);
```
timotree3: No need to use `format!` and allocate a `String` just to pass it as a format-arg. Just format it in the log macro
```suggestion
                                log_error!(context, "Conductor: Could not load DNA file {:?}.", &dna_file);
```

conductor_api/src/conductor/base.rs
```diff
@@ -758,58 +749,72 @@ impl Conductor {
...
+                                    &dna_hash_from_conductor_config,
+                                    &dna_hash_computed) {
+                                    Ok(_) => (),
+                                    Err(e) => {
+                                        let msg = format!("\
+                                        Conductor: DNA hashes mismatch: 'Conductor config' != 'Conductor instance': \
+                                        '{}' != '{}'",
+                                        &dna_hash_from_conductor_config,
+                                        &dna_hash_computed);
+                                        log_error!(context, "{}", msg);
```
timotree3: No need to use format! and allocate a String just to pass it as a format-arg. Just format it in the log macro
```suggestion
                                        log_error!(
                                            context,
                                            "Conductor: DNA hashes mismatch: \
                                            'Conductor config' != 'Conductor instance': \
                                            '{}' != '{}'",
                                            &dna_hash_from_conductor_config,
                                            &dna_hash_computed
                                        );
```

conductor_api/src/conductor/admin.rs
```diff
@@ -209,12 +209,11 @@ impl ConductorAdmin for Conductor {
...
-        new_config.instances.push(new_instance_config);
-        new_config.check_consistency(&mut self.dna_loader)?;
-        let instance = self.instantiate_from_config(id, Some(&mut new_config))?;
+        self.config.instances.push(new_instance_config);
+        self.config.check_consistency(&mut self.dna_loader)?;
+        let instance = self.instantiate_from_config(id)?;
         self.instances
             .insert(id.clone(), Arc::new(RwLock::new(instance)));
-        self.config = new_config;
         self.save_config()?;
```
lucksus: Hm, this change seems problematic to me. The old code would jump out in line 213 if the new config is bogus and not change `self` at all. Now the new instance will always get added to `self.config.instances`.

I don't understand why we need to change these lines at all. What is wrong with the old code?

conductor_api/src/conductor/admin.rs
```diff
@@ -209,12 +209,11 @@ impl ConductorAdmin for Conductor {
                     .into(),
             },
         };
-        new_config.instances.push(new_instance_config);
-        new_config.check_consistency(&mut self.dna_loader)?;
-        let instance = self.instantiate_from_config(id, Some(&mut new_config))?;
+        self.config.instances.push(new_instance_config);
```
maackle:
```suggestion
        // TODO: use cloned config to check consistency first, before setting self.confg
        self.config.instances.push(new_instance_config);
```

lucksus:
This got merged with #1725

PR #1726, opened 2019-09-27
core: Rewrite `await!(future)` to `future.await`

### LINKED ISSUE #1723: Compilation fails on nightlies since 2019-7-30 due to old await syntax
https://github.com/rust-lang/rust/pull/62293 removed support for the `await!` syntax that we're currently using. The new `.await` syntax has worked since nightly-2019-05-08. It seems like it'd be safe to update the syntax we use.

timotree3:
##  PR summary

This resolves #1723 by migrating away from the old syntax.

The other change this commit makes is removing `#![feature(await_macro)]` invocations.

## changelog

The only "breaking change" this PR causes is change our minimum supported nightly to 2019-05-08. This doesn't seem changelog worthy to me.


timotree3:
I just found out that the make script sets an override for nightly-2019-07-14. That would explain why nobody is running into this.

timotree3:
[A cursory search](https://github.com/holochain/holochain-rust/search?q=nightly-2019-07-14&unscoped_q=nightly-2019-07-14) reveals that there are 8 different places that would need to be updated to a new version of rust to use after this change. 

It seems like this PR is then blocked on determining a new version of nightly to stick with that's after 2019-7-30. I imagine that's low priority for the team. Hmmm...

timotree3:
Oh never mind, 2019-07-14 supports the new syntax as well.

PR #1727, opened 2019-09-30
Treat header entry address as special case for aspect getting

maackle:
## PR summary

Properly get content aspect for header entry, when using an address that corresponds to a header

## changelog


core/src/network/handler/mod.rs
```diff
@@ -253,53 +262,73 @@ fn get_content_aspect(
...
+        .agent()
         .iter_chain()
-        .find(|ref chain_header| chain_header.entry_address() == entry_address);
+        .find(|ref chain_header| chain_header.address() == *entry_address)
+        .map(|h| (h, true))
+        .or_else(|| {
+            state
+                .agent()
+                .iter_chain()
+                .find(|ref chain_header| chain_header.entry_address() == entry_address)
```
timotree3: It might be nice to add a comment or something to point out that this is looking for `entry_address` not `address`. I had to squint for a while to figure out the difference between these two finds.

core/src/network/handler/mod.rs
```diff
@@ -253,53 +262,73 @@ fn get_content_aspect(
...
-        Some(EntryWithHeader {
-            entry: get_entry_from_cas(&state.agent().chain_store().content_storage(), header.entry_address())?
+    let maybe_entry_with_header = match maybe_chain_header {
+        Some((header, true)) => Some(create_entry_with_header_for_header(&*state, &header)?),
+        Some((header, false)) => {
+            // ... we can just get the content from the chain CAS
+            Some(EntryWithHeader {
+                entry: get_entry_from_cas(
+                    &state.agent().chain_store().content_storage(),
+                    header.entry_address(),
```
timotree3: Since `header.entry_address` and `entry_address` are equal, is it possible to fail faster by doing this O(1) check before searching the entire chain for the full header?

core/src/network/handler/mod.rs
```diff
@@ -253,53 +262,73 @@ fn get_content_aspect(
...
+        }
+        None => {
+            // ... but if we didn't author that entry, let's see if we have it in the DHT cas:
+            if let Some(entry) = get_entry_from_cas(&state.dht().content_storage(), entry_address)?
+            {
+                // If we have it in the DHT cas that's good,
+                // but then we have to get the header like this:
+                let headers = context
+                    .state()
+                    .expect("Could not get state for handle_fetch_entry")
```
timotree3: At the top of the function we call `context.state()` and store it to a variable. Can we re-use that here?

core/src/network/handler/mod.rs
```diff
@@ -253,53 +262,73 @@ fn get_content_aspect(
...
+                    .get_headers(entry_address.clone())
+                    .map_err(|error| {
+                        let err_message = format!(
+                            "net/fetch/get_content_aspect: Error trying to get headers {:?}",
+                            error
+                        );
+                        log_error!(context, "{}", err_message.clone());
```
timotree3: No need to clone here. Format should take by reference.
```suggestion
                        log_error!(context, "{}", err_message);
```

core/src/agent/state.rs
```diff
@@ -212,6 +213,21 @@ pub fn create_new_chain_header(
     ))
 }
 
+pub fn create_entry_with_header_for_header(
```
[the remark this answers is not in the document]
lucksus: It actually is the specific characteristic of this function that it creates the (virtual) header for a header. For convenience, it returns that header header together with the header (which here has the role of the entry) in the `EntryWithHeader` struct.

(I known that sounds confusing. The thing is that publishing headers means that they are treated as entries - and as such they need to also have a header.)

core/src/network/handler/mod.rs
willemolding: Yeah I'm a tiny bit confused about this part in general it might need a few explainer comments. @maackle what is the idea behind searching both using the header address and the headers entry address?

maackle:
Thanks for the review @timotree3, I incorporated your feedback

Issue #1733, opened 2019-10-03
Tracking issue for cleaning-up allocations


timotree3:
The current code-base does a lot of unnecessary allocations. Here are some examples of common things in the code-base that cause allocation:

- `format!`
- `.to_string()`
- `.clone()`
- `.collect()`
- `.to_path_buf()`
- `.to_string_lossy()`
- `.into_vec()`

This issue is intended to classify the different patterns in the code-base that allocate unnecessarily and track the progress in cleaning them up.

- [ ] Calling functions that take `&String`s (in progress: #1732)
- [ ] Calling functions that take `&PathBuf` (in progress: #1701)
- [ ] Converting between strings and paths (in progress: #1701)
- [ ] Calling `format!` only to use the result as a format argument
- [ ] Taking things by reference just to immediately clone them
- [ ] Allocating `PathBuf`s only to `join` them
- [ ] Calling `.collect()` only to iterate over the result
- [ ] Calling `.to_string()` only to use the result for comparison
- [ ] Converting `String`s to `JsonString`s
- [ ] Returning `String` errors
- [ ] Storing owned collections in a struct when giving the struct a lifetime and storing references would do

Feel free to check these boxes or add more to-dos if any come up! I believe maintainers are able to edit this post.

timotree3:
As I've started to do with these two "roundup" PRs. I intend to slowly check off each box on this list.

thedavidmeister:
sounds good :)

my request would be to keep PRs as easy to review and focussed as possible

timotree3:
I'm totally with you! :-) I've been making a serious effort to do just that.

I've been going for the strategy of each commit having one well explained transformation, applied across the code base. The only other option that I've thought of is to apply all the transformations to a single module and then make a PR for each module. 

Is there some approach that you have in mind?

thedavidmeister:
@timotree3 nothing specific, just keep it as easy to review as possible :)

PR #1736, opened 2019-10-03
Allocate fewer `PathBuf`s only to join them

timotree3:
## PR summary

Motivated by #1733

This PR mainly rewrites two patterns in the code:
- "`[PathBuf]` collecting"
  ```rust
  let path: PathBuf = [
      PathBuf::from("/tmp"),
      PathBuf::from("holochain"),
      some_path_buf_arg,
  ].iter().collect();
  ```
  =>
  ```rust
  let path = {
      let mut path = PathBuf::new();
      path.push("/tmp");
      path.push("holochain");
      path.push(&some_path_buf_arg);
      path
  };
  ```
- "Serial `join`ing"
  ```rust
  let path = self.dir.join(foo).join(bar).join(baz);
  ```
  =>
  ```rust
  let mut path = self.dir.join(foo);
  path.push(bar);
  path.push(baz);
  ```
## testing/benchmarking notes

## followups

## changelog


AshantiMutinta:
I would actually prefer the join when possible, only because pushing introduces a mutable state on variables that might not need it.

timotree3:
In many cases I used I block here to limit mutable state. I can go through and update the rest

timotree3:
For example, like this: https://github.com/holochain/holochain-rust/blob/f294ed3e001644448e1a816044311cc52ae3c8f8/cli/src/cli/chain_log.rs#L19-L24

Would doing them like that address your concern?

timotree3:
@AshantiMutinta I updated this PR, now the only remaining cases where this PR introduces mutability are a few lines before the end of a test. I figure that in blocks of code that small, its easy enough to reason about either way. Does that address your feedback?

AshantiMutinta:
> @AshantiMutinta I updated this PR, now the only remaining cases where this PR introduces mutability are a few lines before the end of a test. I figure that in blocks of code that small, its easy enough to reason about either way. Does that address your feedback?

Woops yes I am fine with this sorry for the delay

zippy:
Hey @timotree3, this looks great.  Sorry that our recent re-arrangement of the the repo into crates sub-dir requires a complex merge...

timotree3:
@zippy thanks! I believe this is once again ready.

PR #1737, opened 2019-10-04
updates to latest lib3h

zippy:
## PR summary

This PR points core to use a recent lib3h branch and makes updates for protocol changes, namely removing HoldEntry and Lib3hClientProtocol::FailureResult, as well as additions for Libh3Uri.

This is important because of a couple dependent crate upgrades needed by holofuel

## testing/benchmarking notes

## followups

## changelog


zippy:
This branch is now blocking the way the pre-deadlock fix was.  AND I don't see any "Immortal" errors showing up.

net/src/p2p_config.rs
```diff
@@ -178,9 +179,9 @@ impl P2pConfig {
 
     pub fn new_with_memory_lib3h_backend(
         server_name: &str,
-        bootstrap_nodes: Vec<url::Url>,
+        bootstrap_nodes: Vec<Url>,
     ) -> Self {
-        let host_name = server_name
+        let _host_name = server_name
```
willemolding: drop this as well if we don't need it or is it going to be used later?

net/src/in_memory/memory_server.rs
```diff
@@ -466,12 +466,12 @@ impl InMemoryServer {
         // Sender must be tracking
         let sender_info = Some((from_agent_id.clone(), Some(msg.request_id.clone())));
         let is_tracking =
-            self.priv_check_or_fail(&dna_address, &from_agent_id, sender_info.clone())?;
+            self.priv_check_or_fail(&dna_address.clone().into(), &from_agent_id, sender_info.clone())?;
```
willemolding: If this is never called with anything other than `dna_address.clone().into()` it might be worth changing the function so it doesn't need the `into` and/or `clone`

timotree3: I see the problem somewhere else. These newtype wrappers are causing use to call `.clone().into()` when we just want to pass references to data that's represented the same way. If we could convert references from one type to another, then we wouldn't need this boilerplate anyway. (Especially if we implemented `Deref`)

PR #1742, opened 2019-10-04
reinstate tests

AshantiMutinta:
## testing/benchmarking notes

## followups

## changelog


zippy:
Hmmm looks like there is a deadlock?

maackle:
Yeah there is still occasionally deadlock when conductors are shutting down, which will cause a process to hang which will make the test never end. I am working on a new try-o-rama that will shut down a conductor if it hasn't been active for a few minutes.

Or I could just fix the deadlock :thinking:

PR #1744, opened 2019-10-05
Sim2h Integration

zippy:
## PR summary

* [x] Sim2hWorker
* [x] Sim2h server as nix command
* [x] app-spec running against Sim2h
==== parity with sim1h ^^ =====
* [x] Reconnect to sim2h node if connection breaks up
* [x] Offline-mode: mirror publish messages back to core as store request
* [x] Wire encryption
* [x] Signed `WireMessages`

## changelog


app_spec/test/default.nix
```diff
@@ -26,7 +26,24 @@ let
...
+  hc-conductor-install
+  (cd app_spec && APP_SPEC_NETWORK_TYPE="sim2h" ./build_and_test.sh);
+  '';
+
+  name-sim2h-server = "hc-sim2h-server";
+
+  script-sim2h-server = pkgs.writeShellScriptBin name-sim2h-server ''
+  set -euo pipefail
+  cargo install sim2h_server -f --git https://github.com/holochain/sim2h
+  RUST_LOG=debug sim2h_server
```
Connoropolous: It would be more flexible to allow the port to be passed in optionally from the outside for sim2h, I guess it's likely been categorized as a 'thing for later'

app_spec/test/index.js
```diff
@@ -57,6 +57,21 @@ if (process.env.APP_SPEC_NETWORK_TYPE === "sim1h")
   );
 }
 
+if (process.env.APP_SPEC_NETWORK_TYPE === "sim2h")
+{
+    transport_config = {
+        type: 'sim2h',
+        sim2h_url: "http://localhost:9000",
```
Connoropolous: how is it that using `http` works here when the websocket is running at `wss://...`

net/src/sim2h_worker.rs
```diff
@@ -0,0 +1,374 @@
...
+                                    error,
+                                    payload
+                                )
+                        }
+
+
+                    }
+                }
+                RequestToParent::IncomingConnection {uri} =>
+                    warn!("Got incomming connection from {:?} in Sim2hWorker - This should not happen and is ignored.", uri),
```
Connoropolous:
```suggestion
                    warn!("Got incoming connection from {:?} in Sim2hWorker - This should not happen and is ignored.", uri),
```

net/src/sim2h_worker.rs
```diff
@@ -0,0 +1,374 @@
...
+                RequestToParent::ErrorOccured {uri, error} =>
+                    error!("Transport error occured on connection to {:?}: {:?}", uri, error),
```
Connoropolous:
```suggestion
                    error!("Transport error occurred on connection to {:?}: {:?}", uri, error),
```

net/src/sim2h_worker.rs
```diff
@@ -0,0 +1,374 @@
...
+        let mut transport: TransportActorParentWrapper<Sim2hWorker, GhostTransportWebsocket> =
+            GhostParentWrapper::new(
+                transport_raw,
+                "t1_requests", // prefix for request ids in the tracker
+            );
+
+
+        // bind to some port:
+        // channel for making an async call sync
+        debug!("Trying to bind to nework...");
```
Connoropolous:
```suggestion
        debug!("Trying to bind to network...");
```

.circleci/config.yml
```diff
@@ -225,7 +238,7 @@ workflows:
   jobs:
    - build
    - fmt
-   - app-spec-tests-sim1h
+   - app-spec-tests-sim2h
```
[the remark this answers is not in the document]
zippy: yep fixed

app_spec/test/index.js
zippy: heh!  good question.  I guess the protocol type is being ignored by the our transport code.

app_spec/test/default.nix
zippy: actually this is worth adding over on the sim2h repo.  But I will make it so if you don't put anything on the command line it will default to port 9000 so we don't have to touch this code.

zippy: done:  if you want now you can pass in the port with --port=xxx

Connoropolous: Nice, yeah, that's what I was thinking... Do It on the sim2h side.

app_spec/test/index.js
Connoropolous: May as well switch it here, for correctness

Connoropolous:
```suggestion
        sim2h_url: "wss://localhost:9000",
```

net/src/sim2h_worker.rs
```diff
@@ -0,0 +1,374 @@
...
+    pub fn advertise(self) -> url::Url {
+        Url::parse("ws://example.com").unwrap()
+    }
+
+    /// Create a new worker connected to the sim2h instance
+    pub fn new(handler: NetHandler, config: Sim2hConfig) -> NetResult<Self> {
+        let transport_raw = GhostTransportWebsocket::new(
+            Address::from("sim2h-worker-transport"),
+            TlsConfig::Unencrypted,
+            Address::from("sim2h-network"),
```
Connoropolous: personally I know that these are hardcoded to match settings over in sim2h, but that might deserve some code comments

Connoropolous: magic strings and all

lucksus: Those strings are not used at all, definitely not in sim2h. I’m not completely clear on what the websocket ghost actor does with them. I had set them like this to give a hint should they appear somewhere when debugging.

Connoropolous: I guess I just meant these values match these values over in sim2h and I got the impression things wouldn't work if those values didn't match. Could be very easily wrong tho. https://github.com/holochain/sim2h/blob/b2453a222830c2dcf2d2f8b3bba5acc59f85d65b/crates/sim2h_server/src/main.rs#L24

conductor_api_api/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
```
lucksus: @AshantiMutinta, take note of this new crate. It only includes this `ConductorApi` struct which is a proxy for the conductor api that does not depend on anything in the `conductor_api` crate - which is good because we can't depend on that in the `net` crate without creating cyclic dependencies. `ConductorApi` was in `core` before we can't depend on that either.

With the repository clean-up that we discussed yesterday, I would suggest with rename this crate to `conductor_api` since the current `conductor_api` will be `conductor_lib`, right?

net/src/sim2h_worker.rs
```diff
@@ -0,0 +1,407 @@
...
+
+
+        // bind to some port:
+        // channel for making an async call sync
+        debug!("Trying to bind to network...");
+        let (tx, rx) = crossbeam_channel::unbounded();
+        transport.request(
+            Span::todo("Find out how to use spans the right way"),
+            RequestToChild::Bind {
+                spec: Url::parse("wss://localhost:0").expect("can parse url").into(),
```
neonphog: Didn't we decide we need `wss://127.0.0.1:0` to ensure we use ipv4 for now?

net/src/sim2h_worker.rs
```diff
@@ -0,0 +1,407 @@
...
+    /// We got a message from core
+    /// -> forward it to the NetworkEngine
+    fn receive(&mut self, data: Lib3hClientProtocol) -> NetResult<()> {
+        self.inbox.push(data);
+        Ok(())
+    }
+
+    /// Check for messages from our NetworkEngine
+    fn tick(&mut self) -> NetResult<bool> {
+        self.num_ticks += 1;
```
neonphog: as a `u32`, if tick gets called once per millisecond (on a fast computer hopefully it'll get called more often than this even) - `num_ticks` will overflow in ~50 days. I think it should be a `u64` : )

net/src/sim2h_worker.rs
```diff
@@ -0,0 +1,407 @@
...
+        //if self.num_ticks % 100 == 0 {
+        //    io::stdout().flush()?;
+        //}
+        if let Err(transport_error) = detach_run!(&mut self.transport, |t| t.process(self)) {
+            error!("Transport error: {:?}", transport_error);
+            // This most likely means we have connection issues.
+            // Send ping to reestablish a potentially lost connection.
+            let _ = self.send_wire_message(WireMessage::Ping);
```
neonphog: not a big fan of ignoring results in this manner... It hides problems when debugging.

```suggestion
            if let Err(e) = self.send_wire_message(WireMessage::Ping) {
                debug!("send ping failure: {:?}", e);
            }
```

net/src/sim2h_worker.rs
lucksus: Hm, good point. Interesting. It seems this was only a problem with the server. This code worked on @zippy's machine with IPv6 interfaces... will change it anyway to be sure.

net/src/sim2h_worker.rs
lucksus: Yes. I've removed `num_ticks` now since it wasn't used anyway.

lucksus:
> In sim2h you changed this to 0.0.0.0 right... How come not the same here?

@Connoropolous, the sim2h_server needs to be accessible from the outside so it has to bind to real network interfaces. The sim2h worker is a client that only calls out, we don't even want somebody to connect on that port.

PR #1747, opened 2019-10-07
implemented boostrap nodes set up env

AshantiMutinta:
## PR summary
-Bootstrap nodes configured for runnning tests
## testing/benchmarking notes

## followups

## changelog

### DIFF
--- a/test_utils/src/lib.rs
+++ b/test_utils/src/lib.rs
@@ -476,7 +476,15 @@ pub fn start_holochain_instance<T: Into<String>>(
     let test_config = env::var("INTEGRATION_TEST_CONFIG").map(|test_config|{
             if test_config =="lib3h"
             {
-                TestNodeConfig::MemoryGhostEngine(vec![])
+                let bootstrap_nodes = env::var("BOOTSTRAP_NODES").map(|bootstrap_config|{
+                    
+                    //https://bootstrap1,http://boostrap2 in the env config to set up before tests
+                     bootstrap_config.split(",").map(|s|{
+                         url::Url::parse(s).expect(&format!("Could not parse thus could not set up test for run {:?}",&s))
+                     }).collect()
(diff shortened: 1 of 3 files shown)


test_utils/src/lib.rs
```diff
@@ -476,7 +476,15 @@ pub fn start_holochain_instance<T: Into<String>>(
     let test_config = env::var("INTEGRATION_TEST_CONFIG").map(|test_config|{
             if test_config =="lib3h"
             {
-                TestNodeConfig::MemoryGhostEngine(vec![])
+                let bootstrap_nodes = env::var("BOOTSTRAP_NODES").map(|bootstrap_config|{
+                    
+                    //https://bootstrap1,http://boostrap2 in the env config to set up before tests
```
timotree3: Is it intentional that one of these links is `https` and the other isn't?

test_utils/src/lib.rs
```diff
@@ -476,7 +476,15 @@ pub fn start_holochain_instance<T: Into<String>>(
...
+                     bootstrap_config.split(",").map(|s|{
+                         url::Url::parse(s).expect(&format!("Could not parse thus could not set up test for run {:?}",&s))
```
timotree3: `.expect` evaluates its argument regardless of it uses it. To avoid computing this `format!` in the happy path, I suggest using `.unwrap_or_else(|_| panic!(...))` instead. Also might as well include the parse error.
```suggestion
                         url::Url::parse(s).unwrap_or_else(|e| panic!("failed to set up for test run: failed to parse as url: (input: {:?}, error: {:?})",&s, e))
```

test_utils/src/lib.rs
AshantiMutinta: oh it was just a problem in the commenting. Mostly just parses a string to a url, doesn't really matter :)

PR #1751, opened 2019-10-08
Use try-o-rama flag for strict conductor timeout

maackle:
## PR summary

#1750 adds a no-activity timeout for conductors, allowing tests to complete even when a conductor locks up while it's shutting down, which is happening often. For 3rd party devs, this is acceptable, since they just want to see their test pass, and it's not their concern to have their test fail because holochain is failing. For us, we want to see a conductor locking up as a test failure. This PR passes an env var to try-o-rama which will raise an exception when the conductor timeout occurs.

## testing/benchmarking notes

We should see this branch fail CI intermittently until the deadlock is fixed. I am working on fixing the actual deadlock separately in #1752. Once that's merged, this should pass again.

## changelog


lucksus:
Got merged as part of another PR

PR #1752, opened 2019-10-08
Fix shutdown deadlock

maackle:
## PR summary

We have frequent deadlocks when conductors shut down

## changelog


maackle:
I spent a long time today trying to fix the deadlock that frequently occurs when shutting down conductors. I found what is essentially causing it but don't know of an easy fix yet, and haven't tracked down the exact cause. This much I know:

- The following three guards are immortal every time:
  1. `Conductor::stop_all_instances()`: Write on `Holochain`
  2. `Instance::process_action()`: Write on `State`
  3. `reduce_shutdown()`: Lock on `NetworkState::P2pNetwork`
- `handle_send_message_result` is getting blocked because it can't get a Read lock on State
  + It is always a message that comes in *after* the conductor has started to shutdown, but before the net thread has fully stopped
  + It always seems to be a ValidationPackage that is being returned as a message
- `ShutdownFuture` sometimes gets blocked because it can't get a Read lock on State (it was one of the few futures that did not get converted to use context.try_state, so it gave an HcLockError)
  - At first I thought it may actually be a good idea to revert this change and use plain locks that can time out on Futures so we can see their HcLockErrors and get backtraces to know that there is where the blocking is occurring. But I tried running it a few times this way, and I didn't get any lock timeouts, and I also didn't get an immortal guards detected. I think it just locks the entire system up so we don't even get this output. So, bad idea.
  - Maybe we can take a hybrid approach where we let futures time out, and return some useful info if they timed out because they were waiting for a mutex lock the whole time.
- Sometimes I get `"Context::block_on() waiting for future but instance is not alive anymore => we gotta let this thread panic!"`
- I noticed that `Instance::drop` stops the action loop and also shuts down the network, but these same things happen in `Holochain::stop`, so they may be happening twice. I commented one out and it didn't seem to have any effect, though.

So in summary, the one clear thing is that the Write lock on State in `process_action` seems to be causing a lot of things to hang. But it is not clear what is causing `process_action` itself to hang so that it won't give up this write lock. I haven't figured out which Action it's reducing when it hangs, which would probably be a good next thing to look into.

maackle:
@lucksus and I fixed it! :tada: 

We had to take out the joining of the `NetConnectionThread`, which was causing the deadlock. When stopping the thread, it does not wait for join, it just assumes that the thread loop will be shutting down soon, as soon as the loop is broken. We also removed the Mutex in NetworkState, instead making `P2pNetwork: Clone`.

This has the undesirable consequence that there is not a guaranteed 1:1 correspondence with a `NetConnectionThread` and an actual thread, since it is cloneable. The field struct for the underlying thread becomes `Arc<Mutex<Option<JoinHandle<()>>>>`, and a `None` in the Option would signify that this `NetConnectionThread` is dangling and points to nothing, and can be discarded.

But by golly, it's working.

core_types/Cargo.toml
```diff
@@ -50,3 +50,6 @@ holochain_logging = { version="=0.0.1", path = "../logging" }
 test_utils = { path = "../test_utils"}
 holochain_core = { path = "../core"}
 maplit = "=1.0.1"
+
+[features]
+deadlock_detection = []
```
lucksus: This was activating parkin-lot‘s detection code, right @maackle? Guess it can go now?

net/src/connection/net_connection_thread.rs
```diff
@@ -124,23 +128,34 @@ impl NetConnectionThread {
...
+        debug!("Telling NetWorker to stop");
         self.can_keep_running.store(false, Ordering::Relaxed);
-        if self.thread.join().is_err() {
-            bail!("NetConnectionThread failed to join on stop() call");
-        }
-        // Call shutdown closure if any
-        if let Some(mut done) = self.done {
-            done();
+        Ok(())
+    }
```
lucksus: This function can‘t fail anymore, shouldn‘t return a result.

core_types/src/error/mod.rs
```diff
@@ -118,6 +118,7 @@ pub enum HolochainError {
     DnaHashMismatch(HashString, HashString),
     EntryNotFoundLocally,
     EntryIsPrivate,
+    // MutexWouldBlock,
```
zippy: still need this commented variant?

core_types/Cargo.toml
maackle: I was thinking that this would be helpful for selectively activating our own deadlock detection. But it is really just uselessly dangling right now.

core_types/src/error/mod.rs
maackle: Guess I'll remove that, it was a half-implemented idea

PR #1755, opened 2019-10-09
try-o-rama stress tests

maackle:
## PR summary

Basic implementation of some stress tests. There are two tests which do the same thing, one in series, one in parallel. Both are parameterized by two variables, the number of conductors (N) and the number of instances per conductor (M)

- spin up N conductors with M instances each
- have each instance commit a unique entry
- have one instance commit another entry and link all other entries onto it
- have each instance read links and verify that they see all M * N entries

This is similar to the existing cluster tests, but leveraging try-o-rama.

## followups

Once we know what kinds of stress tests we'll want to run for various hApps, it will be time to build a bit of a framework for making it easy to stress test a new hApp.

## changelog


maackle:
These might still be flaky. It might be helpful to be more clear on what we want in stress tests before merging this. Once we merge this we should be committed to keeping it green, for whatever tests are in here, we should choose wisely.

app_spec/cluster_test/default.nix
```diff
@@ -1,16 +1,25 @@
...
+  script-cluster = pkgs.writeShellScriptBin name-cluster
   ''
   hc-cli-install
   hc-conductor-install
    ( cd hc_cluster_test && npm install && ./node_modules/.bin/tsc)
    ( cd app_spec && mkdir -p dist && hc package --output dist/app_spec.dna.json --strip-meta )
    ( EMULATION_HOLOCHAIN_BIN_PATH=./.cargo/bin/holochain node ./app_spec/cluster_test/index.js 2)
   '';
+
+  name-stress = "hc-stress-test-sim1h";
```
lucksus: Should this rather be in `stress-test/default.nix`?

maackle: Yeah good call.

maackle:
Stress tests are now parameterized from the command line, so you can do for instance:

    hc-stress-test-sim2h 50 2

which will run stress tests with 50 conductors, 2 instances per conductor

PR #1759, opened 2019-10-14
Fixes to CLI

willemolding:
## PR summary

A few changes to the CLI which is not working well at the moment.

- Use relative paths for specifying the build asset output (and don't write it to .cargo)
- Fixes for using try-o-rama

## changelog


cli/src/cli/scaffold/rust.rs
```diff
@@ -79,7 +78,7 @@ fn interpolate_cargo_template(
 
 impl RustScaffold {
     pub fn new(package_name: &str, macro_style: HdkMacroStyle) -> RustScaffold {
-        let target_dir = wasm_target_dir(&package_name.into(), &String::new().into());
+        let target_dir = PathBuf::from("../target");
```
willemolding: I don't quite understand why this was the way it was. Using `wasm_target_dir` would resulted in the build artifacts being given absolute path (non-portable). Does it not run on CI this way perhaps?

Connoropolous:
On a related note to this PR, I think the --networked flag on `hc run` is way broken.

PR #1701, opened 2019-09-08
Path string roundup

timotree3:
## PR summary

The primary change of this PR is to convert `String` and `&str` to `PathBuf` and `&Path` when applicable to resolve #660.

This change is valuable because you may obtain a path that contains non-utf8 data and try to use it with these APIs, but if these APIs take Strings, you have to lossily convert it and you can't operate on those files.

The main non-obvious change as a result of this effort are changes like:

```rust
format!("using path: \'{}\'", some_string_path)
```
to
```rust
format!("using path: {:?}", some_pathbuf)
```

these result in a change from single quotes to double quotes.

The alternative substitution is

```rust
format!("using path: \'{}\'", some_pathbuf.to_string_lossy())
```
this has the advantage of preserving single-quotes, but has the
disadvantage of requiring an extra allocation over the more
efficient Debug-impl of PathBuf.

There are existing instances in the code where the latter approach is used, it's probably worth changing all instances to use the same choice. Which one seems better?

The other change that this PR makes is to change all `&PathBuf`s in signatures to `&Path`s. This is liking changing `&String` to `&str`, no downside. Read further in the first commit message.

## testing/benchmarking notes

The test suite fails for me on current develop and it also fails for me on my branch.

## followups

holochain/lib3h#309

## changelog

Please check one of the following, relating to the [CHANGELOG-UNRELEASED.md](https://github.com/holochain/holochain-rust/blob/develop/CHANGELOG-UNRELEASED.md)

- [ ] this is a code change that effects some consumer (e.g. zome developers) of holochain core so it is added to the CHANGELOG-UNRELEASED.md (linked above), with the format `- summary of change [PR#1234](https://github.com/holochain/holochain-rust/pull/1234)`
- [ ] this is not a code change, or doesn't effect anyone outside holochain core development
- [x] I'm not sure which parts of the code are outward facing so I don't know

PR #1765, opened 2019-10-16
Crates reorganization

AshantiMutinta:
## PR summary
Reorganizing crates and making sure they are compatible with crates.io.

## testing/benchmarking notes

## followups

## changelog


crates/cli/Cargo.toml
```diff
@@ -6,8 +6,8 @@ authors = ["Holochain Core Dev Team <devcore@holochain.org>"]
 [dependencies]
 holochain_core_types = { path = "../core_types" }
 holochain_core = { path = "../core" }
-holochain_common = { path = "../common" }
-holochain_conductor_api = { path = "../conductor_api" }
+holochain_common = { path = "../../common" }
+holochain_conductor_api = { path = "../conductor_lib" }
```
lucksus: We should also change the crate name to `holochain_conductor_lib`.

AshantiMutinta: That is a good idea, I will change that right now

AshantiMutinta:
> There is still a core_types dir in the root with a README, probably from a merge?
> 
> Also, are we not moving `logging` and `common` crates into `crates/`?
> 
> @zippy noticed that the root level Cargo.toml went away, maybe we can have that back?
> 
> Otherwise this looks good!

Oh yes I had it in  my stash instead, restored that,

I have moved the logging but I am not so sure about the common, it seems like it is just a supporting cast like test_utils?

AshantiMutinta:
> 
> 
> Looks like root-level Cargo.toml is missing "crates/core_types"?
> 
> It's such a huge PR that it may be hard to ensure that everything is exactly correct before merging, but as long as tests pass and basic cargo/nix commands work I feel we can iron out any remaining details we find when using it more extensively. If it works seamlessly as is then that's a bonus!

Checked and it seems it was that way in develop as well

willemolding:
Is that also a good time to clear some cruft from the repo? I think the `qt/` dir is no longer needed and I moved the `benchmarks/` to the holochain-persistence repo so we can remove that.

maackle:
We might need to pull out the large amount of fmt applied in this PR if possible, and do it at a more opportune time. In particular, we have https://github.com/holochain/holochain-rust/pull/1701 waiting in the wings, which is another PR that makes a huge number of changes, and it would be a shame to have a messy merge just because of fmt. @AshantiMutinta think you can revert the commit that did the fmt?

Issue #1303, opened 2019-04-19
Option to also clear instance storage when using `admin/instance/remove`


maackle:
`admin/instance/remove` could take an extra boolean `clean` parameter, which if true, also deletes the instance's storage.

Alternatively, add the option to `admin/instance/add` (perhaps calling it `replace` in this case), which will cause storage to be cleared if it already exists.

Use case:

During development, let's say I have been testing my DNA with a particular instance. Now I want to update the DNA. If I keep the same instance ID, it will attempt to use the old storage (since storage directories are named by the instance ID.

Issue #1047, opened 2019-02-22
fail to compile holochian-rust on ubuntu16.04 LTS. due to "error: failed to run custom build command for `rust_sodium-sys v0.10.3"


luomijie:
I compile holochian-rust ,I'm getting an error now on rust_sodium-sys.And I have no idea really what this is .Please help me.

error: failed to run custom build command for `rust_sodium-sys v0.10.3 (/home/roger/projects/holochain/src/holochain/rust_sodium-sys)`
process didn't exit successfully: `/home/roger/projects/holochain/src/holochain/nodejs_conductor/native/target/release/build/rust_sodium-sys-77c181f75143ae12/build-script-build` (exit code: 101)
--- stdout
cargo:rerun-if-env-changed=RUST_SODIUM_LIB_DIR
cargo:rerun-if-env-changed=RUST_SODIUM_USE_PKG_CONFIG
cargo:rerun-if-env-changed=RUST_SODIUM_SHARED

--- stderr
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `"ecf1f540da63099d5335581b288593f66a5531bd634c0549363dcc6ef52f23eb"`,
 right: `"eeadc7e1e1bcef09680fb4837d448fbdf57224978f865ac1c16745868fbd0533"`: 

Downloaded libsodium file failed hash check.

', /home/roger/projects/holochain/src/holochain/rust_sodium-sys/build.rs:85:5
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

warning: build failed, waiting for other jobs to finish...
error: build failed
neon ERR! cargo build failed

Error: cargo build failed
    at Target.<anonymous> (/usr/local/lib/node_modules/neon-cli/lib/target.js:121:35)
    at step (/usr/local/lib/node_modules/neon-cli/lib/target.js:32:23)
    at Object.next (/usr/local/lib/node_modules/neon-cli/lib/target.js:13:53)
    at fulfilled (/usr/local/lib/node_modules/neon-cli/lib/target.js:4:58)
    at processTicksAndRejections (internal/process/next_tick.js:81:5)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @holochain/holochain-nodejs@0.4.2-alpha compile: `neon build --release`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the @holochain/holochain-nodejs@0.4.2-alpha compile script.
[...]

PR #1775, opened 2019-10-18
[WIP] add a clean option to the admin/instance/remove method

jamesray1:
## PR summary

Work on #1303, going with the first option to add a clean option to the "admin/instance/remove" method, and if this option is true, remove the instance's storage. These changes are adapted to the latest reorg, as compared to #1047 

## testing/benchmarking notes

Adds a test `test_remove_instance_clean_true`, which currently is failing due to 89 errors (E0432, E0599, E0658) from futures-io-preview.

## followups

## changelog

### DIFF
--- a/crates/conductor_lib/src/conductor/admin.rs
+++ b/crates/conductor_lib/src/conductor/admin.rs
@@ -40,7 +40,7 @@ pub trait ConductorAdmin {
         dna_id: &String,
         agent_id: &String,
     ) -> Result<(), HolochainError>;
-    fn remove_instance(&mut self, id: &String) -> Result<(), HolochainError>;
+    fn remove_instance(&mut self, id: &String, clean: bool) -> Result<(), HolochainError>;
     fn add_interface(&mut self, new_instance: InterfaceConfiguration)
         -> Result<(), HolochainError>;
     fn remove_interface(&mut self, id: &String) -> Result<(), HolochainError>;
@@ -244,6 +244,9 @@ impl ConductorAdmin for Conductor {
         }
         if let Some(instance) = self.instances.remove(id) {
             instance.write().unwrap().kill();
+            if clean == true {
+                remove_dir_all(instance.storage.path)?;
+            }
         }
         let _ = self.start_signal_multiplexer();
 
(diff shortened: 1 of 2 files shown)


jamesray1:
The build is failing with many "Tried to use droppped state" panics from https://github.com/holochain/holochain-rust/pull/1569.

jamesray1:
Not sure how to get the storage path of the instance correctly when passing it to `remove_dir_all`. So the `remove_instance_` tests are failing.

jamesray1:
I can't see whether there is a clear way to get the storage configuration path from an instance. If the `StorageConfiguration` type is memory, there won't be a persistent file path, while if it it [File or Pickle](https://developer.holochain.org/api/v0.0.32-alpha2/holochain_conductor_api/config/enum.storageconfiguration) then there will be a path. Perhaps I am missing something, but I think there needs to be a new function to get a Some of this storage configuration path, if it exists, or return None if not?

jamesray1:
After looking through the code and API I found this through a search in the codebase. So to get a better understanding of Holochain, I'm going to finish reading the guidebook, as it looks like I missed all of the completed articles when I got up to the (E) Going Live with Holochain Apps section.

jamesray1:
I'm still struggling to figure out how to best get the storage path from an instance.

`instance.state().agent().chain_store().content_storage()` won't work because I don't know what the address is for the storage path.

jamesray1:
One of the stress tests is failing (details [here](https://circleci.com/gh/holochain/holochain-rust/44729?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link), although I'm not sure why.

jamesray1:
Just need to figure out a way to test that the storage of the instancehas been removed without using what I've added.

jamesray1:
For future extensibility with testing it seems like it would be helpful to generate more customizable tomls, by passing fields or using defaults. However, that would be best be done in a separate PR, perhaps before continuing with this one.

jamesray1:
More details:

`create_test_conductor` is used in `test_remove_instance_clean_true`. `create_test_conductor` uses `test_toml`, which in turn uses `instance1()`. And, as explained above, `instance1()` uses the memory variant of storage, without a persistent path. So in order to create a proper `test_remove_instance_clean_true`, that will actually be able to check whether the storage directory is removed, I need a different toml. To do that, it seems like it would be best to be able to pass a different instnace block to what are currently created with `test_toml` and `instance1`. For just this PR, the easiest way seems like to just add an option to select whether to use a memory type, with no path, or a file type, with some path relative to the config path, and then adjust wherever `instance1()` is called.

```
pub fn test_toml(test_name: &str, port: u32) -> String {
        let mut toml = header_block(test_name);

        toml = add_block(toml, agent1());
        toml = add_block(toml, agent2());
        toml = add_block(toml, dna());
        toml = add_block(toml, instance1());
        toml = add_block(toml, instance2());
        toml = add_block(toml, interface(port));
        toml = add_block(toml, logger());
        toml
    }
```

```
    pub fn instance1() -> String {
        r#"[[instances]]
agent = 'test-agent-1'
dna = 'test-dna'
id = 'test-instance-1'

[instances.storage]
type = 'memory'"#
            .to_string()
    }
```

jamesray1:
Hmm, wasn't expecting this fail:

```
thread 'conductor::admin::tests::test_remove_instance_clean_true' panicked at 'The storage directory for the instance doesn't exist after creating the test conductor!', crates/conductor_lib/src/conductor/admin.rs:1267:9
```

jamesray1:
I'm not sure why the fmt test is required, but it looks like a bunch of diffs related to the changes I've made. Would people like to review? @AshantiMutinta, @maackle, @thedavidmeister, @zippy, @ddd-mtl, @lucksus

jamesray1:
Okay, all checks are passing, feel free to review!

jamesray1:
Need to write app spec tests.

thedavidmeister:
i think i missed some of the original context around this, @maackle ?

maackle:
So, for context: this is based on a pretty old issue #1303 which I wrote as an idea for fixing the problem one might encounter as a DNA developer, where you modify your DNA but try to load it in your conductor with the same instance ID as before, and if you're using file storage, now you've got a collision with your old storage because the directory name is based on the instance ID.

It is now 6 months later. @jamesray1, this is good work, but I think a good first step would have been to have a discussion in the issue about how to approach this. Since the issue was created, this problem was tackled in another direction by https://github.com/holochain/holochain-rust/pull/1335, which instead of clearing out the stale storage, makes it impossible to load DNA from a storage that does not match the intended hash.

Unfortunately we have been pretty bad about using GH issues for visibility of what we're working on. I feel personally responsible for this one since I created the issue and even marked it as a good first issue, so, sorry for this being misleading. We have some leveling up to do around making it easier for others to contribute!

I think this could still be a useful feature, but it might look a little different now, so I wouldn't want to merge it in right now. We are pretty focused on other priorities right now in preparation for Holo closed alpha launch, so we don't even have the bandwidth to come back to discuss this issue right now -- so this PR will have to sit for a while. I appreciate you jumping in to work on this, though, and doing a thorough job of it. Beginning of next year we will be able to broaden our focus, and maybe even have a better practice about doing triage on issues for contributors to take on.

jamesray1:
Sorry that I didn't try to get more clarity and have a discussion about what would be the right approach. Please feel free to work with me and let me know if there is anything I can do to help with the closed alpha launch.

thedavidmeister:
@jamesray1 @maackle i think a really good thing to do would be to triage issues and open these discussions with the goal to either:

- close the issue if not relevant any more
- keep the issue open as a low priority, but add information that can help us meaningfully progress without re-treading the same discussions/investigations on the next pass
- write some code now that resolves an issue

just sort the issue queue by oldest and find things to knock off :)

i used to do the same thing with the old drupal issue queues and ended up triaging literally thousands of issues that were 6-9+ years old at the time!

i've found that code is often the _last_ thing that needs doing on an active open source issue queue :sweat_smile:

jamesray1:
I'm happy to work with you all on clearing out stale issues and resolving ones that are still relevant, but as @maackle said, that is not high on the list of priorities right now, with the Holo closed alpha launch, so I don't want to distract you from those efforts. I did start looking for a job yesterday, as I got the impression based on @maackle's comment above that there was lack of capacity to onboard new developers at the moment.

jamesray1:
I can have a look into this again and see how it can be updated after looking at #1335, while trying to work on it independently, because as mentioned, bandwidth to discuss it is limited right now.

PR #1780, opened 2019-10-19
trycp

zippy:
## PR summary

conductor manager for spinning up remote nodes; includes:

- [x] bash script.
- [x] json rpc deamon wrapper for spawning and killing conductors that try-o-rama can use to orchestrate tests
- [x] readme and better tests
- [x] mock node launcher for testing
- [x] nixify trycp_server install

## testing/benchmarking notes

## followups

## changelog


crates/trycp_server/src/main.rs
```diff
@@ -0,0 +1,153 @@
+extern crate structopt;
+extern crate tempfile;
+
+//use log::error;
+//use std::process::exit;
+use self::tempfile::tempdir;
+use jsonrpc_core::{IoHandler, Params, Value};
+use jsonrpc_ws_server::ServerBuilder;
```
maackle: This may be a lot easier to work with as a `jsonrpc-http-server` instead. There's no need for a persistent socket connection if we're not getting notifications back.

zippy: I tried that at first, but it didn't work easily in my testing for some weird reason.  And then, I realized that we actually wanted this because we wanted the closing of the websocket to be the indicator of how to shut down the node right?

maackle: Yeah I am OK with this

maackle:
FYI I will have one more small commit coming in tomorrow, adding a parameter to the kill method (for the signal, e.g. SIGINT)

thedavidmeister:
FYI i'm throwing docker related stuff in here but don't let that stop anyone merging

maackle:
@struktured this is the new method I was talking about on Fri, @zippy I can fill you in next week: https://github.com/holochain/holochain-rust/pull/1780/commits/e14f86f2471d4fdcc01bede458d1688ef07bd648

PR #1783, opened 2019-10-21
Add checkers for storage configuration variant

jamesray1:
## PR summary
Working on #1775, it seems like in order to get the `storageConfigurationPath` for an instance, a function could be written to check the given instance's storage field, then if it is a field that has a persistent file path (currently `File` or `Pickle`, not `Memory`), then get the path from this value (returned in a Some), otherwise return None. This PR adds checkers to see whether a storage value is a particular `StorageConfiguration` variant, and also then uses these to get the path, if applicable.

## testing/benchmarking notes

## followups

## changelog


crates/conductor_lib/src/config.rs
```diff
@@ -707,6 +707,41 @@ pub enum StorageConfiguration {
     Pickle { path: String },
 }
 
+impl StorageConfiguration {
+    fn is_memory(&self) -> bool {
+        match self {
+            StorageConfiguration::Memory => true,
```
AshantiMutinta: You could probably just self == StorageConfiguration::Memory here

crates/conductor_lib/src/config.rs
```diff
@@ -707,6 +707,41 @@ pub enum StorageConfiguration {
...
+    fn is_file(&self) -> bool {
+        match self {
+            StorageConfiguration::File => true,
+            _ => false,
+        }
+    }
+
+    fn is_pickle(&self) -> bool {
+        match self {
+            StorageConfiguration::Pickle => true,
```
AshantiMutinta: same as here

crates/conductor_lib/src/config.rs
jamesray1: Thanks for the tip, will do! Currently I'm just trying to sort out getting next to persistently work across terminal sessions so that I can test locally.

crates/conductor_lib/src/config.rs
```diff
@@ -707,6 +707,41 @@ pub enum StorageConfiguration {
...
+    fn is_memory(&self) -> bool {
+        match self {
+            StorageConfiguration::Memory => true,
+            _ => false,
+        }
+    }
+
+    fn is_file(&self) -> bool {
+        match self {
+            StorageConfiguration::File => true,
```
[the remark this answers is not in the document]
jamesray1: == requires a fully constructed value to compare to, not a pattern. So a match statement is needed, or more succintly the match crate. Advice from https://discordapp.com/channels/442252698964721669/448238009733742612.

jamesray1:
OK so I've commented out the checkers for the variants rather than deleting them as they may be useful in future?

`get_path` is also unused, so I will use it in https://github.com/holochain/holochain-rust/pull/1775.

willemolding:
If you merge latest develop branch you will now also have the LmdbStorage variant!

jamesray1:
Great, thanks Willem!

jamesray1:
I'm going to close this because by itself it is an unused feature and needs to be a part of #1775.

PR #1784, opened 2019-10-21
Sim2h: remove GhostActor around websocket transport & reconnect after connection break ups

struktured:
## PR summary

1. Replaces the use of a GhostActor around the websocket transport in sim2h to directly use the underlying `StreamManager`.
2. Copies over the `websocket` module from lib3h to iterate on the stream manager code without forcing releases on lib3h for trivial changes (making functions public, ...)
3. Uses `StreamManager`s API to handle connections correctly

Having direct access to `StreamManager` makes information about the state of the connection available to sim2h, which is used in `Sim2hWorker` to handle connection issues and reconnect after a connection loss.

This has a high chance of fixing the flaky failures we have seen with sim2h in app-spec and in the stress-test.

## testing/benchmarking notes

@lucksus ran the first stress-test scenario locally with 100 agents and without noticing connection issues.

## followups

The second scenario of the stress tests creates a lot of load on the zome interface of a single instance which caused the tests to fail when @lucksus tried to go beyond 50 nodes locally (on the same device). This test was deactivated to go up to 100 nodes to create more stress on the network. We might want to be more intentional with our stress tests to test different distinct aspects of the whole integrated system in different scenarios. Currently, the two scenarios we have test a mixed profile which doesn't give an easy and clear insight (without further analysis) into what is the limiting factor. 

## changelog


crates/net/src/sim2h_worker.rs
```diff
@@ -4,42 +4,35 @@ use crate::connection::{
...
-    protocol::{
-        RequestToChild, RequestToChildResponse, RequestToParent, TransportActorParentWrapper,
-    },
-    websocket::{
-        actor::GhostTransportWebsocket,
-        tls::{TlsCertificate, TlsConfig},
-    },
-};
+
+use lib3h_zombie_actor::WorkWasDone;
```
lucksus: I guess we could now do without this, no?

crates/net/src/sim2h_worker.rs
```diff
@@ -389,15 +357,12 @@ impl NetWorker for Sim2hWorker {
...
     /// Check for messages from our NetworkEngine
     fn tick(&mut self) -> NetResult<bool> {
-        if Instant::now().duration_since(self.time_of_last_sent) > Duration::from_secs(PING_DURATION_SECS) {
-            self.send_ping();
-        }
-        if let Err(transport_error) = detach_run!(&mut self.transport, |t| t.process(self)) {
-            error!("Transport error: {:?}", transport_error);
-            // This most likely means we have connection issues.
-            // Send ping to reestablish a potentially lost connection.
+        if Instant::now().duration_since(self.time_of_last_sent)
```
lucksus: Without the websocket actor, this ping doesn't make sense any more and should either just go away, or better: we actually check if the connection is still there and try to reconnect immediately if we lost it..

crates/net/src/sim2h_worker.rs
```diff
@@ -89,120 +79,95 @@ impl Sim2hWorker {
...
         let payload = wire_message_into_escaped_string(&message);
         let signature = self
             .conductor_api
-            .execute(
-                payload.clone(),
-                CryptoMethod::Sign,
-            )
-            .expect(&format!("Couldn't sign wire message in sim2h worker: {}", payload));
+            .execute(payload.clone(), CryptoMethod::Sign)
+            .expect(&format!(
```
AshantiMutinta: better to use unwrap_or_else(|| panic!("error msg here")) rather than expect(format because expect is not lazily evaluated

crates/net/src/sim2h_worker.rs
```diff
@@ -371,6 +336,7 @@ impl Sim2hWorker {
         Ok(())
     }
 
+    #[allow(dead_code)]
     fn send_ping(&mut self) {
```
AshantiMutinta: is send_ping not being used anywhere? Should we keep it?

crates/net/src/sim2h_worker.rs
```diff
@@ -389,24 +355,25 @@ impl NetWorker for Sim2hWorker {
...
-            }
-            did_something = WorkWasDone::from(true);
+        match self.stream_manager.connection_status(&self.server_url.clone().into()) {
+            ConnectionStatus::None => {
+                warn!("No connection to sim2h server. Trying to reconnect...");
+                self.stream_manager.connect(&self.server_url.clone().into())?;
+            },
+            ConnectionStatus::Initializing => debug!("connecting..."),
+            ConnectionStatus::Ready => {
+                let client_messages = self.inbox.drain(..).collect::<Vec<_>>();
```
AshantiMutinta: do we have to collect here or can we just use a for_each() to get the same results on the iterator? Just asking cause we are collecting and then iterating again

crates/sim2h/src/websocket/tls.rs
```diff
@@ -0,0 +1,386 @@
...
+}
+
+impl TlsCertificate {
+    /// Creates a self-signed certificate with an entropy key and passphrase.
+    /// This makes it possible to use a TLS encrypted connection securely between two
+    /// peers using the lib3h websockt actor.
+    pub fn build_from_entropy() -> Self {
+        let (key, cert) = generate_pair();
+
+        let random_passphrase: String = thread_rng().sample_iter(&Alphanumeric).take(30).collect();
```
AshantiMutinta: it would be nice to not have remove the magic number and put it in a variable just to give a little bit of context on why it is taking 30, or even a comment would suffice

crates/sim2h/src/websocket/tls.rs
```diff
@@ -0,0 +1,386 @@
...
+        pub fn process(&mut self) {
+            for _ in 0..10 {
+                self.priv_process();
+                std::thread::sleep(std::time::Duration::from_millis(1));
+            }
+        }
+
+        pub fn write_all(&mut self, is_srv: bool, mut data: &[u8]) {
+            for _ in 0..10 {
+                match match is_srv {
```
AshantiMutinta: I think even a single line if statement would suffice here. match if is_srv { &mut self.srv} else {&mut self.cli}

crates/net/src/sim2h_worker.rs
struktured: @AshantiMutinta Not sure if @lucksus intends to bring it back soon, otherwise I agree w/you.

crates/net/src/sim2h_worker.rs
struktured: I thought same thing but you end up mutably borrowing `self` twice using that pattern and the compiler complains. Must be some way to do that without a full copy of the collections though.

crates/net/src/sim2h_worker.rs
lucksus: Yeah, I thought we might want to use it again. That's why I've marked it with this instead of removing it..

crates/sim2h/src/websocket/tls.rs
neonphog: The normal case is a user-entered passphrase... the length of the passphrase is irrelevant.

I'm not convinced we even need a real passphrase for this ephemeral case. We could set it to an empty string.

crates/sim2h/src/websocket/streams.rs
```diff
@@ -0,0 +1,565 @@
...
+                self.event_queue
+                    .push(StreamEvent::ConnectionClosed(info.url));
+                info.stateful_socket = WebsocketStreamState::None;
+                continue;
+            }
+            self.stream_sockets.insert(id, info);
+            //match info.stateful_socket {
+            //    WebsocketStreamState::None => {None},
+            //    _ => self.stream_sockets.insert(id, info),
+            //};
```
neonphog: comment why we're keeping this or remove?

crates/sim2h/src/websocket/tls.rs
struktured: @neonphog Is there a one line suggested change you can provide?

struktured:
> Conditional Approval--Are we abandoning the effort to share code between sim2h and lib3h?

Please correct me If I am wrong @lucksus, but I believe the motivation was to iterate faster to work around the connection issues with sim2h. We can probably get rid of this code after some decided point of stability, I would hope. Maybe even use ghost actor 2.0 in some idealized future. At very least we should merge any critical changes back into lib3h.

lucksus:
> > Conditional Approval--Are we abandoning the effort to share code between sim2h and lib3h?
> 
> Please correct me If I am wrong @lucksus, but I believe the motivation was to iterate faster to work around the connection issues with sim2h. We can probably get rid of this code after some decided point of stability, I would hope. Maybe even use ghost actor 2.0 in some idealized future. At very least we should merge any critical changes back into lib3h.

Yes, exactly @struktured, @neonphog. We had to go through a few iterations of finding a bug or needing a trivial change that became apparent on the level of integration, which basically was either app-spec or the stress test in this repo. That then needed a change in the lib3h repo, followed by a version bump, which always had to happen in all crates (and repositories) between lib3h and holochain-rust (esp. sim1h and sim2h). By moving sim2h and this code into holochain-rust we can have a change set validated by CI and merged to one develop branch BEFORE bumping any version and having to coordinate between repos. Even though the changes of this version compared to lib3h's develop are minimal at this point, I would like to be able to respond quickly if more changes are needed between now and upcoming releases - which I believe is reasonable to expect.

lucksus:
But once things are settled, we should definitely bring things back together.

PR #1797, opened 2019-10-23
Add redux crate to isolate parts of `State` and `Instance` logic

timotree3:
## PR summary

Motivation: a lot of the tests in `instance.rs` have a lot of specific logic for the listener-thread and the DNA initialization process. Separating out this logic and making it generic upon `State` type facilitates the creation of simple and easy to verify tests.

I now have version one of the `redux` API set up. It will take a lot of work to integrate this API into the rest of the code-base. I'm making a draft PR in the meantime so I can get feedback on the crate design.

## testing/benchmarking notes

None yet

## followups

None yet

## changelog


AshantiMutinta:
I think this could be beneficial, definitely opens the door to more deterministic tests, I had a idea of borrowing from the functional school but I think it wouldn't be a bad idea to introduce immutable data structures https://docs.rs/im/13.0.0/im/ , track our history then and eliminate some of the mutability, 

That way we have pure deterministic tests in the redux loop.

What do you think?

timotree3:
> I had a idea of borrowing from the functional school but I think it wouldn't be a bad idea to introduce immutable data structures https://docs.rs/im/13.0.0/im/ , track our history then and eliminate some of the mutability,

That's an interesting idea. If we were committed to a reduce interface without mutability, I think that would definitely be the way to go. Something like this:

```rust
fn reduce(&self, &Self::Action) -> Self;
```

However I'm not so sure I want that interface. In my view, the goal of eliminating mutability, while very relevant in languages where it's easy to abuse it, like Javascript (the originator of the Redux pattern), isn't relevant in Rust. Even as the docs for that crate you linked say

> While immutable data structures can be a game changer for other programming languages, the most obvious benefit - avoiding the accidental mutation of data - is already handled so well by Rust's type system that it's just not something a Rust programmer needs to worry about even when using data structures that would send a conscientious Clojure programmer into a panic.

thedavidmeister:
does rust really remove all the problems of modelling identity and value, therefore making mutability somehow "obviously safe"?

since they mention clojure programmers specifically, here is the writeup from the clojure docs https://clojure.org/about/state

we know that we need a very clear "as at" model where all our wasm callbacks and other logic see a consistent view of data (e.g. the value of the source chain cannot change underneath two callbacks that are ostensibly part of the "same workflow") - could immutability crates help us model that better?

AshantiMutinta:
After investigating deadlocks this week and looking at multiple state locations, I think it would be beneficial to have an immutable state with immutable data structures. I think in theory what they say is true, the rust type system is very strong but how strong it is is dependent on how complex the problem becomes. Holochain is a huge codebase with multiple asynchronous operations, having an immutable state would be beneficial because no matter how you model it, you are assured that the operations for the most part have no side effects.

I think it is not only a type issue but it has become a design issue as well.

thedavidmeister:
@AshantiMutinta i agree, type systems and immutability solve totally different things

i'm a bit lost as to how types are supposed to fix the thing we want immutability to fix in our redux model

PR #1798, opened 2019-10-23
Sim2h - RRDHT "R-Value" Stub Calculation

neonphog:
## PR summary

Adds rrdht "R-Value" calculation for storage arc radius to sim2h in a stub manner
- the calculation is performed, but just gives a warning if the recommended radius < ARC_RADIUS_MAX
- we have no uptime data for clients, so uptime is assumed to be 1/2 in all cases

## testing/benchmarking notes

## followups

## changelog


crates/sim2h/src/lib.rs
```diff
@@ -34,22 +37,37 @@ use parking_lot::RwLock;
...
 pub struct Sim2h {
+    crypto: Box<dyn CryptoSystem>,
     pub bound_uri: Option<Lib3hUri>,
     connection_states: RwLock<HashMap<Lib3hUri, ConnectionState>>,
     spaces: HashMap<SpaceHash, RwLock<Space>>,
     stream_manager: StreamManager<std::net::TcpStream>,
-    num_ticks: u32,
+    num_ticks: u64,
+    /// sim2h currently uses the same radius for all connections
+    rrdht_arc_radius: u32,
```
lucksus: Hm, I think this is the wrong place for this variable to be. The sim2h server can server many spaces, and each space will have a different aggregated arc radius, no?

crates/sim2h/src/lib.rs
```diff
@@ -65,6 +83,44 @@ impl Sim2h {
...
 
+    /// recalculate arc radius for our connections
+    fn recalc_rrdht_arc_radius(&mut self) {
+        let mut peer_record_set = RValuePeerRecordSet::default()
+            // sim2h is currently omniscient
+            .arc_of_included_peer_records(Arc::new(0.into(), ARC_LENGTH_MAX));
+        for (_k, c_state) in self.connection_states.read().iter() {
+            match c_state {
+                ConnectionState::Limbo(_) => (),
+                ConnectionState::Joined(_, _, dht_data) => {
```
lucksus: Here we are neglecting the space that agent is in and assuming all agents to be in the space, right? I think we should calculate r-values per space.

PR #1807, opened 2019-10-26
hc hash --property

lucksus:
## PR summary

Add option "property" to hc hash that sets DNA properties for hash calculation.

```
USAGE:
    hc hash [OPTIONS]

OPTIONS:
    -p, --path <path>               Path to .dna.json file [default: dist/<dna-name>.dna.json]
    -x, --property <property>...    Property (in the from 'name=value') that gets set/overwritten before calculating
                                    hash
```
Like so:

```
[nix-shell:~]$ hc hash --path identity-manager.dna.json --property agent_id=HcSc82gf923hd
Reading DNA from QmZUhssczL8RdBUjKLrtFyV9co2BCHubmJTdj4oeMaUYki.dna.json
DNA Hash: QmWPhCUU3Gib9kTHs2gtiD1VFnwrvqX1zcKFf1WNNnD6mk
```
## followups

Use this in Holoscape to be able to detect installed instances even if they were installed with specific/unique properties (such as identity manager).

## changelog

## documentation

### DIFF
--- a/crates/cli/src/cli/hash_dna.rs
+++ b/crates/cli/src/cli/hash_dna.rs
@@ -2,10 +2,13 @@ use error::DefaultResult;
 use failure::err_msg;
 use holochain_conductor_lib::conductor::Conductor;
 use holochain_persistence_api::cas::content::{Address, AddressableContent};
-use std::path::PathBuf;
 use serde_json::Map;
+use std::path::PathBuf;
 
-pub fn hash_dna(dna_file_path: &PathBuf, properties: Option<Vec<String>>) -> DefaultResult<Address> {
+pub fn hash_dna(
+    dna_file_path: &PathBuf,
+    properties: Option<Vec<String>>,
+) -> DefaultResult<Address> {
     let mut dna = Conductor::load_dna(dna_file_path)?;
     let mut map = if let serde_json::Value::Object(map) = dna.properties {
         map.clone()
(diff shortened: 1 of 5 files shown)


crates/cli/src/cli/hash_dna.rs
```diff
@@ -1,9 +1,39 @@
...
+                .map(String::from)
+                .collect::<Vec<String>>();
+            if parts.len() != 2 {
+                return Err(err_msg(format!(
+                    "Can't parse property: {}",
+                    property_string
+                )));
+            }
+            let value = parts.pop().unwrap();
+            let name = parts.pop().unwrap();
```
AshantiMutinta: You might be able to shorten this by :

let equal_sign= property_string.find("=").unwrap_or_else(|| Err(err_msg(format!(
                    "Can't parse property: {}",
                    property_string
                ))))?
let (name,value) = property_string.split_at(break_off);
and then map.insert(name,serde_json::Value::String(value))

crates/cli/src/main.rs
```diff
@@ -192,6 +192,12 @@ enum Cli {
             help = "Path to .dna.json file [default: dist/<dna-name>.dna.json]"
         )]
         path: Option<PathBuf>,
+        #[structopt(
+            long,
+            short = "x",
+            help = "Property (in the from 'name=value') that gets set/overwritten before calculating hash"
```
maackle:
```suggestion
            help = "Property (in the form 'name=value') that gets set/overwritten before calculating hash"
```

maackle: You would intuitively think it would be "p". How do you feel about making this a breaking change and making "path" a positional argument so we can use "p" for this?

PR #1810, opened 2019-10-28
Instrument dispatcher to publish latencies

struktured:
## PR summary

This PR instruments holochain so the dispatching agent publishes the latency of zome function call. A `MetricPublisher` trait is introduced which is implemented for cloudwatch via `risoto_cloudwatch` and a logger based one for local testing etc.

## testing/benchmarking notes

## followups

## changelog

## documentation


crates/metrics/src/metrics.rs
```diff
@@ -0,0 +1,49 @@
...
+    pub fn new(name: &str, value: f64) -> Self {
+        Self {
+            name: name.to_string(),
+            value,
+        }
+    }
+}
+
+pub trait MetricPublisher: Sync + Send {
+    fn publish(&mut self, metric: &Metric);
```
maackle: Does this really have to be a mutable self reference? What other kinds of MetricPublishers are you envisioning in the future? The logger one clearly doesn't need the mutable self.

crates/core/src/context.rs
```diff
@@ -86,6 +89,7 @@ pub struct Context {
     pub(crate) signal_tx: Option<Sender<Signal>>,
     pub(crate) instance_is_alive: Arc<AtomicBool>,
     pub state_dump_logging: bool,
+    pub metric_publisher: Arc<RwLock<dyn MetricPublisher>>,
```
maackle: It seems that the only reason this needs to be in a mutex is because of the mutable reference to self in `publish`. So can we talk about if we really need that? Any logging/reporting should be as lightweight as possible.

crates/metrics/src/metrics.rs
struktured: @maackle I assumed that mutation is required to send something to aws. I can double check that though.

struktured:
> I'm highly skeptical of this `&mut self` in `publish`. I'm curious about what a non-logging case would look like. What would happen when you publish a Metric for real?
> 
> I'm wondering if the production implementation of MetricPublisher wouldn't simply be something like:
> 
> ```rust
> impl MetricPublisher for crossbeam_channel::Sender<Metric> {
>     fn publish(&self, metric: &Metric) {
>         self.send(metric.clone()).unwrap()
>     }
> }
> ```
> 
> with publish just a wrapper around send. Then we would just have to set up a reporter thread with the channel receiver. This publish would also not need mutability.

I like the idea of using an async channel to avoid any expensiveness. Will look into that.

crates/conductor_lib/src/context_builder.rs
```diff
@@ -175,6 +184,9 @@ impl ContextBuilder {
         let eav_storage = self
             .eav_storage
             .unwrap_or_else(|| Arc::new(RwLock::new(EavMemoryStorage::new())));
+        let metric_publisher = self
+            .metric_publisher
+            .unwrap_or_else(|| Arc::new(RwLock::new(DefaultMetricPublisher::new())));
```
AshantiMutinta: if you derive a default implementation for DefaultMetricPublisher you might be able to get away with unwrap_or_default()

PR #1812, opened 2019-10-28
core_types::sync gets its own crate

maackle:
## PR summary

In preparation for updating HcMutex et al. to use some of the same improvements used in GhostMutex. After experimenting with this mutex diagnosis stuff for a while, it seems useful enough to put in its own crate. It has also grown to be a little unwieldy for a single a file, so it was a good time to refactor anyway. Also, one day this may become a general open source tool.

- moves core_types::sync to a new crate called `holochain_locksmith`, living at `crates/locksmith`
    - if it ever becomes a solid general-purpose tool, we can just rename the crate "locksmith"
- drops the `Hc` prefix for all the wrapped structs, i.e. it just directly exports things called `Mutex`, `RwLock`, etc., as drop-in replacements
- renames a few other things:
    - `spawn_hc_guard_reporter` becomes `spawn_locksmith_guard_reporter`
    - `HcLockError` becomes `LocksmithError`
- updates entire codebase with the above in mind

## followups

- Better documentation for this crate
- Implement GhostMutex improvements

## changelog

## documentation


AshantiMutinta:
Do we want to also publish this to crates.io? In case multiple repos want to use it? Should we add a hook for it?

maackle:
@AshantiMutinta that would be great. I think after #1813 starts working would be a good time. I'll merge this in for now though.

PR #1813, opened 2019-10-28
Incorporate ideas from GhostMutex into HcMutex

maackle:
## PR summary

- Does away with the polling behavior of `try_[lock|read|write]_until` now that we are using parking_lot we can use the built-in functionality
- Removes the PENDING_LOCKS idea which we never wound up using
- Adds self-consuming `unlock_fair` function to all guard types, for op-in fair unlocking
- Updates Borrow/Deref impls for Option-backed structs

I decided not to make all unlocking fair, because 1. life isn't always fair, and 2. fair unlocking ain't free, it requires a context switch for every unlock, interrupting the current thread. We have a lot of little mutexes that we don't particularly need to worry about, so I think we should apply the fair unlocking judiciously in tight loops like futures and other hotspots. [Read the parking_lot docs for more](https://amanieu.github.io/parking_lot/parking_lot/struct.MutexGuard.html#method.unlock_fair)

## changelog

## documentation


maackle:
Re-opened this as #1826, because the branch name was misleading

PR #1816, opened 2019-10-30
Feature/hdk prelude

willemolding:
## PR summary

A nice easy review. Adds a new module called `prelude` to the HDK. This is a common pattern among Rust crates and it contains the macros and types that are required to effectively use the crate. In many cases this simplifies the `use` statements in a zome to a single line.

As part of this PR I also migrated the zomes not currently using Rust edition 2018 to use it (only the blog zome). This means that `extern crate` is no longer required further simplifying the experience of using the HDK.

It also catches a few cases in the HDK macros where it assumed certain types were imported. The imports are now explicit in the macro.

The changes to the HDK are 100% backward compatible!!! So this won't break any existing zome code or examples.

## testing/benchmarking notes

## changelog

## documentation


crates/hdk/src/prelude.rs
```diff
@@ -0,0 +1,46 @@
+
+/// Types required by all but the most trivial zomes. 
+/// This can greatly simplify imports for the majority of developers
+/// by simply adding use hdk::prelude::*;
+
```
willemolding: This is really the only file needed to be reviewed. Please check if you think these types are sufficient or if they are too broad. I made a judgement call to include these ones in particular based on the app spec.

app_spec/zomes/simple/code/src/lib.rs
```diff
@@ -1,41 +1,10 @@
-#![warn(unused_extern_crates)]
```
willemolding: A great example of why this is awesome

crates/hdk/src/prelude.rs
```diff
@@ -0,0 +1,33 @@
+/// Types required by all but the most trivial zomes.
+/// This can greatly simplify imports for the majority of developers
+/// by simply adding use hdk::prelude::*;
```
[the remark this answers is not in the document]
willemolding: Nice, thanks I didn't know you could do that with doc comments

PR #1819, opened 2019-10-30
Telephone-game stress tests

lucksus:
## PR summary

Adds stress tests that setup Holochain nodes to play the "telephone game" .
Out of the N nodes, only 2 are online at the same time. The assertions proof that later nodes can see all the entries and links that were created previously before they entered the network.

There are different scenarios that all apply the same basic logic of the telephone game but with different entry/link topologies that are shared and checked between the nodes.

This was added because there was a hypothesis that some bug would prevent nodes from "seeing" (either receiving or validating) entries/links that were created before they were online. That would have explained some high-level flakiness the hApps that we field-tested recently with Holoscape.

But these new tests and the fact that they are passing shows that this is not the issue.

## followups

Stress test in CI are configured to run with N of 10 which is a reasonable number for these telephone tests. While it makes sense to scale up N for the other stress tests, I don't see these scenarios providing more insight when run with greater N, as only two nodes are only at the same time anyway. We might want to split out stress tests on the nix-command level to select which ones are run in CI and how often.

## changelog

## documentation


willemolding:
Hey Nico note I have a similar but less general test in [this branch](https://github.com/holochain/holochain-rust/pull/1620/files#diff-39c6390defd84cfde93e1611204a268d). I'll switch that over to using this when we merge

PR #1822, opened 2019-10-31
fix hc-test

neonphog:
## PR summary

- fixes `hc-test-fmt` - it was failing to return an error code on format diff (also ran `cargo fmt`)
- `hc-rust-test` was running sim1h tests which don't work without dynamo - added `--exclude sim1h`
- `hc-rust-wasm-compile` was breaking due to unset HC_TARGET_PREFIX environment variable
- removed call to `hc-app-spec-test` which no longer exists

## testing/benchmarking notes

## followups

## changelog

## documentation

### DIFF
--- a/rust/wasm/compile/default.nix
+++ b/rust/wasm/compile/default.nix
@@ -14,7 +14,7 @@ let
  ''
  set -euxo pipefail
  export WASM_PATH=${path}/
- cargo build --release --target wasm32-unknown-unknown --manifest-path "$WASM_PATH"Cargo.toml --target-dir "$HC_TARGET_PREFIX"/"$WASM_PATH"target;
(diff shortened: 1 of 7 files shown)


rust/wasm/compile/default.nix
```diff
@@ -14,7 +14,7 @@ let
  ''
  set -euxo pipefail
  export WASM_PATH=${path}/
- cargo build --release --target wasm32-unknown-unknown --manifest-path "$WASM_PATH"Cargo.toml --target-dir "$HC_TARGET_PREFIX"/"$WASM_PATH"target;
+ cargo build --release --target wasm32-unknown-unknown --manifest-path "$WASM_PATH"Cargo.toml --target-dir "''${HC_TARGET_PREFIX:-.}"/"$WASM_PATH"target;
```
maackle: what the hey is this? how does adding a kodama emoticon `:-.` help?

neonphog: haha : ) - `:-` is bash variable expansion default value, so `${VAR:-DEF}` will result in `DEF` if `VAR` is not set.

PR #1826, opened 2019-11-01
Locksmith improvements

maackle:
## PR summary

Most of this is the noise of splitting one big file into several, and cleaning up, but there are also some new features embedded in here:

- Does away with the polling behavior of `try_[lock|read|write]_until` now that we are using parking_lot we can use the built-in functionality
- Removes the PENDING_LOCKS idea which we never wound up using
- Adds self-consuming `unlock_fair` function to all guard types, for op-in fair unlocking
- Updates Borrow/Deref impls for Option-backed structs

I decided not to make all unlocking fair automatically, because 1. life isn't always fair, and 2. fair unlocking ain't free, it requires a context switch for every unlock, interrupting the current thread. We have a lot of little mutexes that we don't particularly need to worry about, so I think we should apply the fair unlocking judiciously in tight loops like futures and other hotspots. [Read the parking_lot docs for more](https://amanieu.github.io/parking_lot/parking_lot/struct.MutexGuard.html#method.unlock_fair)

~Note that this is causing even more locks than in develop. It seems that unfair locking doesn't make much difference in this regard.~ Now that `threadpool` has been merged in, tests pass :tada:

## changelog

```markdown
```

## documentation


crates/locksmith/src/error.rs
```diff
@@ -0,0 +1,27 @@
...
+    Lock,
+    Read,
+    Write,
+}
+
+#[derive(Debug)]
+pub struct LocksmithError {
+    lock_type: LockType,
+    kind: LocksmithErrorKind,
+}
```
neonphog: just for rust idiom, even if you don't override any of the provided functions:

```suggestion
}

impl std::error::Error for LocksmithError {
}
```

crates/locksmith/src/guard.rs
```diff
@@ -0,0 +1,172 @@
...
+            puid: ProcessUniqueId,
+            inner: Option<$Guard<'a, T>>,
+            fair_unlocking: bool,
+        }
+
+        impl<'a, T: ?Sized> $HcGuard<'a, T> {
+            pub fn new(inner: $Guard<'a, T>) -> Self {
+                let puid = ProcessUniqueId::new();
+                GUARDS
+                    .lock()
```
neonphog: I know we shouldn't go too crazy fixing potential problems in our code that is designed to fix other problems... but it's totally possible we could end up with a deadlock in this GUARDS mutex... do you think we should do a `try_lock_for(so_long_time_value)` that can just panic so we know there's a problem?

crates/locksmith/src/guard.rs
```diff
@@ -0,0 +1,172 @@
...
+                    inner: Some(inner),
+                    fair_unlocking: false,
+                }
+            }
+
+            /// Add some context which will output in the case that this guard
+            /// lives to be an immortal
+            pub fn annotate<S: Into<String>>(self, annotation: S) -> Self {
+                GUARDS
+                    .lock()
```
neonphog: ditto try_lock_for?

crates/locksmith/src/guard.rs
```diff
@@ -0,0 +1,172 @@
...
+            }
+
+            /// Explicitly consume and unlock this mutex fairly, regardless
+            /// of what kind of unlocking was specified at initialization
+            pub fn unlock_fair(mut self) {
+                self._unlock_fair();
+            }
+
+            fn _unlock_fair(&mut self) {
+                if let Some(inner) = std::mem::replace(&mut self.inner, None) {
```
neonphog: I liked your suggestion to start using `take()` instead of std::mem::replace

maackle:
Addressed comments :+1:

Issue #1824, opened 2019-11-01
Link from AGENT_ADDRESS not working


freesig:
## Issue
Linking to an agents `AGENT_ADDRESS` doesn't seem to propagate to another conductor.
## Steps to reproduce
Alice and Bob are separate agents running on separate `holochain` conductors pointing to the same local sim2h. (Although the same behavior happens with n3h). 
1. Alice commits an entry and links it `from` her `AGENT_ADDRESS`.
2. Alice is able to retrieve this entry by getting the link.
3. Bob tries to get the same entry using Alice's `AGENT_ADDRESS`.
4. Bob gets back an empty list even after waiting for 15 minutes.

## Source
- DNA https://gist.github.com/freesig/23e15c7e41167fc3d353e4eaafd98de1
- Alice's config https://gist.github.com/freesig/89046a1641db4e983a6f65b70cb03301
- Bob's config https://gist.github.com/freesig/4370676c0d22c11694656f7dc0cbf099

## Debug
I think this might be the issue. __could not be validated due to unresolved dependencies.__
This happens when Bob tries to get the links in step 3.
```DEBUG 2019-11-01 15:22:20 [holochain::test-instance] retry_validation/puid-d-ed crates/core/src/workflows/hold_link.rs:78 workflow/hold_link: Link could not be validated due to unresolved dependencies and will be tried later. List of missing dependencies: [HashString("HcScjdwyq86W3w5y3935jKTcs4x9H9Pev898Ui5J36Sr7TUzoRjMhoNb9fikqez"), HashString("QmQHBpnU5f9zccWx7vNRnWVMvzd91aFinXPdis7dCspcT6")]
DEBUG 2019-11-01 15:22:20 [holochain::test-instance] retry_validation/puid-d-ee crates/core/src/workflows/hold_entry.rs:71 workflow/hold_entry: QmfTXMcaHgfTZ9pfMJWCsDrtpHqu4d2v7ED29y5wWxaJ7C could not be validated due to unresolved dependencies and will be tried later. List of missing dependencies: [HashString("HcScjdwyq86W3w5y3935jKTcs4x9H9Pev898Ui5J36Sr7TUzoRjMhoNb9fikqez"), HashString("QmQHBpnU5f9zccWx7vNRnWVMvzd91aFinXPdis7dCspcT6")]
```
This might be related. It happens at the beginning.
```
[...]

PR #1834, opened 2019-11-05
Sim2h Gossip Refetch

lucksus:
## PR summary

Currently, sim2h only requests gossip lists from nodes (i.e. the list of hashes a node is holding) only once - right after the node connected. If that node is missing something, sim2h picks a random other node to get that missing data from, to then delegate it back to the incomplete node.

So far, this process was quite fragile and might contribute for a good portion of the issues we encounter in the wild: if anything goes wrong (like that random node not having those entries neither, or a connection failure), there is nothing that would fix that. This node will never got to hold those missing entries, unless it reconnects again. Which is also what matches our experience of real-world tests.

These changes improve the handling of missing data at nodes.
We now cache data about which node is missing what. And then we periodically check-in with that node, re-request it's gossip list and trigger the gossip process again, if something is still missing

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -289,10 +289,13 @@ jobs:
   docker:
    - image: holochain/holonix:ubuntu
   steps:
+   - run: sudo apt-get update && sudo apt-get install -y git ssh
    - checkout
    - run:
       name: Deploy linux binaries
-      command: nix-shell --run hc-release-github-binaries
+      command: |
+       . /home/docker/.nix-profile/etc/profile.d/nix.sh
+       nix-shell --run hc-release-github-binaries
 
 workflows:
  version: 2
--- a/CHANGELOG-UNRELEASED.md
+++ b/CHANGELOG-UNRELEASED.md
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 {{ version-heading }}
 
 ### Added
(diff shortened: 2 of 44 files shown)


Connoropolous:
this sounds fantastic @lucksus ... I've been finding in my own testing that something along these lines has definitely been occurring, during real world tests I've done so far.

Connoropolous:
I wonder if, in order to test this type of functionality, you would actually need to mock unreliable/intermittently connected nodes?

lucksus:
@Connoropolous, I'm afraid this doesn't fix (at least all) of these problems. Tried yesterday with @zippy and I could see how this code gets exercised - but we could confirm that @zippy's node got entry data that is then later missing in its gossip list. So there seems to be a problem in core, which we believe is related to this issue: https://github.com/holochain/holochain-rust/issues/1824

But yeah, this change is still something that will catch many other problems, so we should get it in :)

lucksus:
Regarding flaking network testing: yes!
We are also working on setting up test scenarios with many nodes and real Internet between them.

But programmatically creating package and connection drops in tests would be a really good thing. Is there a tool you could recommend?

lucksus:
There might have been discussion around the name `arc_radius` somewhere here, @maackle: https://github.com/holochain/lib3h/pull/487.

Otherwise, mentioning @neonphog for a potential clarification :)

PR #1836, opened 2019-11-06
Fix loading

lucksus:
## PR summary

Loading of instances was broken.

Two reasons why we didn't notice this right away:
1. The test for this was not really testing loading and got deactivated. Which then allowed a change to happen that broken the loading code in every case.
2. The conductor always first tries to load from the given storage. If that fails, it creates a new instance. This second branch obviously doesn't erase everything in the CAS and was ending up with a partially loaded state. Source chain was loaded correctly. Agent keys are stored in the keystore anyway. So we only missed some things, especially the holding list.

## changelog

```markdown
```

## documentation

### DIFF
--- a/crates/conductor_lib/src/holochain.rs
+++ b/crates/conductor_lib/src/holochain.rs
@@ -392,7 +392,6 @@ mod tests {
         let temp_filestorage_dir = temp.path().to_str().unwrap();
         let agent = registered_test_agent("persister");
         let (signal_tx, _signal_rx) = signal_channel();
-        //let logger = test_logger();
         let mut dna = create_arbitrary_test_dna();
         dna.name = "TestApp".to_string();
 
--- a/crates/conductor_lib/src/holochain.rs
+++ b/crates/conductor_lib/src/holochain.rs
@@ -431,6 +431,9 @@ mod tests {
         );
 
         let result = Holochain::load(context_load.clone());
+        if let Err(e) = result {
+            panic!("Error during Holochain::load: {:?}", e);
+        }
         assert!(result.is_ok());
         let hc = result.unwrap();
         let instance = hc.instance.as_ref().unwrap();
@@ -442,7 +445,7 @@ mod tests {
         assert_eq!(network_state.agent_id.is_some(), true);
         assert_eq!(network_state.dna_address.is_some(), true);
     }
-    
+
     #[test]
     fn fails_instantiate_if_init_fails() {
         let dna = create_test_dna_with_wat(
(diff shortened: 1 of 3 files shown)


crates/core/src/workflows/application.rs
```diff
@@ -8,15 +8,24 @@ use holochain_core_types::{
...
 };
+use log::error;
 use std::sync::Arc;
 
 pub async fn initialize(
     instance: &Instance,
-    dna: Option<Dna>,
+    maybe_dna: Option<Dna>,
     context: Arc<Context>,
 ) -> HcResult<Arc<Context>> {
```
maackle: What happens if we have a Dna in the context/state, but then call initialize with Some(dna)? If that is invalid behavior, maybe that is an argument for splitting this into two different functions so we can't do something invalid.

lucksus: Please have a look again. I've left it as one function for now, but added comments and a panic in case `maybe_dna` is `Some` AND there is a DNA already in the nucleus state.

crates/conductor_lib/src/holochain.rs
```diff
@@ -387,29 +387,63 @@ mod tests {
...
     #[test]
-    // TODO: This test is not really testing if loading works. But we need a test for that.
-    // Persistence relies completely on the CAS, so the path would need to be used by
-    // creating a FileStorage CAS in the context that is passed to Holochain::load:
+    fn can_persistant_and_load() {
+        let temp = tempdir().unwrap();
+        let temp_filestorage_dir = temp.path().to_str().unwrap();
+        let agent = registered_test_agent("persister");
+        let (signal_tx, _signal_rx) = signal_channel();
+        //let logger = test_logger();
```
willemolding:
```suggestion
```

crates/core/src/workflows/application.rs
```diff
@@ -8,15 +8,39 @@ use holochain_core_types::{
...
+        // to check here if we really deal with a fresh state and no DNA in the Nucleus already:
+        if context.get_dna().is_some() {
+            panic!("Tried to initialize instance that already has a DNA in its Nucleus");
+        }
+        dna
+    } else {
+        // No DNA provided as parameter.
+        // This is the loading-case - we assume to find a DNA in the Nucleus state:
+        context.get_dna().ok_or_else(|| {
+            error!("No DNA provided during loading and none found in state");
```
willemolding: Should this use proper logging with the context?
```suggestion
            log_error!(context, "No DNA provided during loading and none found in state");
```

PR #1841, opened 2019-11-07
[WIP]Futures diagnostics

AshantiMutinta:
## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation

### DIFF
--- a/crates/core/src/instance.rs
+++ b/crates/core/src/instance.rs
@@ -20,6 +20,7 @@ use holochain_core_types::{
     dna::Dna,
     error::{HcResult, HolochainError},
     ugly::lax_send_sync,
+    diagnostic::FuturesDiagnosticTrace
 };
 use holochain_locksmith::{RwLock, RwLockReadGuard};
 #[cfg(test)]
@@ -157,6 +158,7 @@ impl Instance {
         sub_context.set_state(self.state.clone());
         sub_context.action_channel = self.action_channel.clone();
         sub_context.observer_channel = self.observer_channel.clone();
+        FuturesDiagnosticTrace::run(sub_context.future_trace.clone());
         Arc::new(sub_context)
     }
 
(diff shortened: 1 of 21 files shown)


crates/core/src/instance.rs
```diff
@@ -158,6 +159,7 @@ impl Instance {
         sub_context.set_state(self.state.clone());
         sub_context.action_channel = self.action_channel.clone();
         sub_context.observer_channel = self.observer_channel.clone();
+        FuturesDiagnosticTrace::run(sub_context.future_trace.clone());
```
lucksus: Can we make this conditional so we can decide through the conductor config if we want to run this?

crates/core/src/network/actions/custom_send.rs
```diff
@@ -50,15 +51,25 @@ pub async fn custom_send(
...
+    running_time: Instant,
 }
 
 impl Future for SendResponseFuture {
     type Output = Result<String, HolochainError>;
 
     fn poll(self: Pin<&mut Self>, cx: &mut std::task::Context) -> Poll<Self::Output> {
+        self.context
+            .future_trace
+            .write()
```
lucksus: I am wary about the performance hit if we add (another) lock acquisition to each future. Would be good to either pull this in as per config setting as well (which I guess is hard to reference from here) or only acquire the lock and write elapsed time once every second or so..

crates/core/src/network/actions/shutdown.rs
```diff
@@ -30,12 +38,18 @@ pub async fn shutdown(
...
     state: Arc<RwLock<StateWrapper>>,
+    running_time: Instant,
 }
 
 impl Future for ShutdownFuture {
     type Output = HcResult<()>;
 
     fn poll(self: Pin<&mut Self>, cx: &mut std::task::Context) -> Poll<Self::Output> {
+        if self.running_time.elapsed() > Duration::from_secs(70) {
+            panic!("future has been running for too long")
```
lucksus: This is a different style of future diagnostics. I would suggest we harmonize and have at least all be the same. This one might be less performance problematic..

PR #1843, opened 2019-11-07
Sim2h Stress Ping-Pong

neonphog:
## PR summary

```
$ cargo run --bin sim2h_stress

FINAL STATS: StressStats {
    master_tick_count: 7501,
    log_stats: {
        "received_pong_count": StressLogStats {
            count: 17309,
            min: 1.0,
            max: 1.0,
            avg: 1.0,
        },
        "tick_elapsed_ms": StressLogStats {
            count: 279728,
            min: 0.0,
            max: 30.0,
            avg: 0.04367814448321181,
        },
    },
}
```

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation


crates/stress/src/bin/sim2h_stress.rs
```diff
@@ -0,0 +1,357 @@
...
+impl StressSuite for Suite {
+    fn start(&mut self) {}
+
+    fn progress(&mut self, stats: &StressStats) {
+        println!("PROGRESS: {:#?}", stats);
+    }
+
+    fn stop(&mut self, stats: StressStats) {
+        *self.sim2h_cont.lock().unwrap() = false;
+        std::mem::replace(&mut self.sim2h_join, None)
```
maackle: I have seen us use this pattern a lot, but I just learned that `self.sim2h_join.take()` accomplishes the same thing. I think.

crates/stress/src/lib.rs
```diff
@@ -0,0 +1,329 @@
...
+    /// give the stress runner some processor time
+    fn tick(&mut self) -> bool {
+        if std::time::Instant::now() > self.run_until {
+            *self.should_continue.lock().unwrap() = false;
+            return false;
+        }
+        {
+            let mut cur_job_count = self.job_count.lock().unwrap();
+            while *cur_job_count < self.config.job_count {
+                (*self.job_queue.lock().unwrap()).push_front(StressJobInfo {
```
maackle: Is it necessary to acquire and release the lock every loop iteration or can the acquisition go outside the loop?

crates/stress/src/lib.rs
```diff
@@ -0,0 +1,329 @@
...
+                    job_index: self.job_last_index,
+                    job: (self.config.job_factory)(),
+                });
+                self.job_last_index += 1;
+                *cur_job_count += 1
+            }
+        }
+        for _ in 0..1000 {
```
maackle: Why 1000? Since the runner.tick is inside a loop itself, would it be enough to just drain the log_recv() and then move on, rather than trying to receive a fixed number of times? Maybe with a small delay and then a retry if the first try_recv() comes up empty?

A comment at least would help. And maybe putting 1000 into a const?

crates/stress/src/bin/sim2h_stress.rs
neonphog: Oh, sweet, good to know :+1:

crates/stress/src/lib.rs
neonphog: This is tricky - It would probably be more straight forward to get rid of the separate count variable and lock the whole job queue outside the loop. But the job factory could have some overhead, and the pool can't pick off jobs while the job_queue is locked.

Also we don't want the count to drop while a thread is running a job, because then this loop might fill in additional jobs, and we'll go over our target... so that's why the count is separate and locked the whole time.

I thought this solution might be a happy medium.

crates/stress/src/lib.rs
neonphog: lol, originally this did indeed loop until there was nothing to try_recv... but I had a println! in there, and the IO overhead was more that the log generation, so it was generating logs faster than it could print them, so the loop never terminated.

Will add a comment : )

PR #1849, opened 2019-11-08
Lmdb large entry test

willemolding:
## PR summary

A bug in holochain_persistence_lmdb caused it to fail to expand the MMAP if a single entry larger than the total MMAP size was written (adding a bunch of smaller ones was fine).

A fix for this is available in holochain/holochain-persistence#13

This PR includes a test for writing a large entry to LMDB. This is currently failing but switching to use the branch of holochain_persistence_lmdb allows it to pass. Once another release of Holochain persistence is made we can bump the version here and switch back to lmdb as default.


lucksus:
Anything that keeps us from just cutting that new release on holochain-persistence, @willemolding?

willemolding:
> Anything that keeps us from just cutting that new release on holochain-persistence, @willemolding?

That has already been done. And for lib3h so this should be ready to merge as soon as CI passes

PR #1852, opened 2019-11-08
Fix lots of deadlocks by managing threads and encapsulating locks

lucksus:
## PR summary

1. Use a thread pool to throttle number of threads that get spawned dynamically
2. Throttle number of simultaneously run zome calls by implementing a queue within `NuceleusState`
3. Queue all holding workflows in `DhtStore` and consume/run only one at a time
4. Simplify `Instance::process_action` and replace panic with an error which fixes shutdown deadlocks and makes Rust tests pass again
5. Don't pass around lock guards with `Context::state()` and instead return copies
6. Don't pass around lock guards of the CAS/EAV storages through state getters and handle thread synchronization locally inside those getters

### Limiting threads and workflows (points 1-3)
This started with the idea to use a thread pool in order to not have the conductor spawn way to many threads under heavy load. Quickly realized that threads are the wrong level of granularity to put a limit on, since we have high-level workflows (such as calling zome functions and holding entries) that consist of one root thread, but that also might spawn a number of sub-threads that the root thread then might wait for completion on. Not limiting those workflows but limiting the threads they spawn creates another source for deadlocks. Hence adding points 2 and 3.

### try-lock in action loop
After adding those limits on workflows, I could get much further into our stress tests, but Rust tests started failing. This seemed to be a problem only happening during shutdown. The redux action loop would panic after not being able to get a lock on the state, while the instance was already shutting down. [Here](https://github.com/holochain/holochain-rust/pull/1852/commits/fe9a76305010d0c29be9fae55f43b0f81b0c764d) I've changed the signature of `Instance::process_action()` not deal with observers any more (handled in the loop that calls it) and instead returning a result - and transforming the panic when not being able to get a lock, to an error. This way we can have the redux action loop run again if can't get a lock. It might realize that we already got a kill signal and can stop trying to get a lock on something that isn't there any more. This fixed the failing Rust test.

### Not passing around lock-guards
With [this change](https://github.com/holochain/holochain-rust/pull/1852/commits/87c284f681ae2a143b4a249a3e5309f2eb465050) `Context::state()` is not handing out lock guards (which is asking for deadlocks) and instead returns a copy. This is actually not expensive at all since the `State` consists almost only of `Arcs` on the top level anyway.

This removed several deadlocks we kept seeing in the stress tests.

We were left with deadlocks involving the CAS/EAV storages.
The remaining commits are doing the same (hiding locks locally and not passing around lock guards) for the CAS locks too.

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation


crates/core/src/network/mod.rs
```diff
@@ -128,11 +128,13 @@ pub mod tests {
             .block_on(commit_entry(entry.clone(), None, &context1))
             .unwrap();
         {
-            let dht1 = context1.state().unwrap().dht();
-            {
-                dht1.content_storage().write().unwrap().add(&entry).unwrap();
-                dht1.add_header_for_entry(&entry, &header2).unwrap();
-            }
+            // Writing to the storage case now requires (as it should be) a mutable reference
```
zippy: comments?  Can we delete this, or is there a followup?

crates/core/src/context.rs
```diff
@@ -329,6 +346,16 @@ impl Context {
         }
     }
 
+    pub fn spawn_thread<F>(&self, f: F)
```
dymayday: Might be a misleading name as the threads are already spawned and in the pool, so we only execute the task.

lucksus: Hm, good point. So maybe `spawn_task`?

crates/core/src/instance.rs
```diff
@@ -213,26 +229,27 @@ impl Instance {
...
         // Mutate state
         {
             let new_state: StateWrapper;
 
             // Get write lock
             let mut state = self
                 .state
-                .write()
-                .expect("owners of the state RwLock shouldn't panic");
+                .try_write_until(Instant::now().checked_add(Duration::from_secs(10)).unwrap())
```
maackle: can we make this a const?

PR #1854, opened 2019-11-08
Don't filter out locksmith logs! And make them a little quieter

maackle:
## PR summary

An exclusion rule snuck into our LogRules and has been there for a while which filters out logs from our locksmith mutex watcher. This includes the IMMORTAL lock guard notice. This PR removes that filter, and makes the locksmith logs a little less noisy, only reporting on active locks every 5 seconds rather than every 500 ms.

## changelog

```markdown
```

## documentation


maackle:
I'm just gonna merge this with the 1 review, it's so minor and so needed, and we have been losing so much context around deadlocks without this.

PR #1857, opened 2019-11-10
Just use parking_lot

maackle:
## PR summary

Strips back locksmith to be nothing more than a thin parking_lot wrapper. Essentially the same as just using parking_lot.

This is part of an experiment to see why #1826 is failing. If this PR fails, then the weird locking logic present in develop is actually making a difference, mitigating real problems.

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation


maackle:
This was an experiment whose purpose has been served.

It illustrated that holochain_locksmith was not adding anything extra that was causing tests to fail

PR #1866, opened 2019-11-12
Fix consistency model for links

lucksus:
# Makes hachiko actually wait for links

This should fix many flaky failures where hachiko/try-o-rama is involved (looking esp. at app-spec and stress tests with sim1h).

Discovered in https://github.com/holochain/holochain-rust/pull/1852, where all holding workflows were made sequential, that the test wouldn't wait for links (the meta part) to be held. Adding sleeps fixed that. That led me to look into the consistency model...

`AddLink` entries don't have a `crud_link` set in their header which is why the old version was wrong. The reason it was working most of the time must have been that both holding workflows (for the link entry itself and for the meta aspect) were validated in parallel such that the meta holding was done roughly at the same time the link entry was added.

But I'm pretty sure that we had many cases in which that was not case and that led to a flaky test failure.


lucksus:
> yes and can we map some unit tests to expected signals for actions?

yes and can this be a follow, @thedavidmeister?

thedavidmeister:
@lucksus yes and tarpaulin is potentially coming back Soon™

PR #1869, opened 2019-11-12
Substitute HDK_VERSION for GIT_HASH

pjkundert:
## PR summary

Assuming that the cargo build source is an actual Git repository is problematic.  Instead of using GIT_HASH, use HDK_VERSION for various HDK version checks.  Also, obtain HDK_VERSION from either the HDK_VERSION environment variable or (preferably) the CARGO_PKG_VERSION variable.

Replaces GIT_HASH with HDK_HASH, derived from Nix derivation MD5 hash (or, from HDK_HASH, for non-standard build environments, defaulting to "00..." if unknown/unspecifed).

This supports the building of holochain-rust in pure Nix environments, with no `git ...` or assumption of source code being held in a Git repository.

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation


pjkundert:
The problem with the Git hash, is that `build.rs` assumes that the source is being built from a Git repo, which is not always the case.

What about changing this to `SRC_HASH`, and obtaining the SHA256 of the holochain-conductor source code being built by Nix?

pjkundert:
OK, I'm deducing the `HDK_HASH` from the Nix build output directory hash.  This encompasses both the full source code, *and* the full build environment plus all dependencies.

pjkundert:
The Nix build hash, in the present incarnation of `holonix`, does *not* reflect the holochain-rust source code or dependencies.

Once we transition to a clean, pure Nix build environment (eg. `holo-nixpkgs`), this hash will reflect the build environment, source code and all dependencies.

So, `HDK_HASH` now replaces `GIT_HASH`, and will eventually reflect the source-code and dependencies, as well as just the build environment.

pjkundert:
I just discovered that only the HDK_VERSION metadata is available to Zome writers, via the `hdk::version` API.  There is no method to access the HDK_HASH (or, previously, GIT_HASH).

Is there any reason not to return both the HDK_VERSION *and* HDK_HASH from `hdk::version`?  They're both realistically described as the "version" of the HDK, from different perspectives...

pjkundert:
I have added `hdk::version_hash`, returning the MD5 hash from `HDK_HASH` or Nix's `out` environment variable.  Previously, there was no `hdk::...` API that made `GIT_HASH` available to Zome code available.  The `hdk::version` API is also tested, to ensure that it returns the `HDK_VERSION` or `CARGO_PKG_VERSION` value supplied during compilation.

This completes the transition from using `git ...` during the build process to obtain version and source-code hash metadata from the source-code environment.  Now, `hdk::version` comes from the holochain conductor's Cargo.toml, and `hdk::version_hash` comes from the Nix build environment's `out` environment variable.

Both of these can be overridden by supplying `HDK_VERSION` or `HDK_HASH` to the build environment, if desired.  If neither `HDK_HASH` or `out` are available, a default MD5 hash consisting of all zeros is supplied for `hdk::version_hash`.  This ensures that an operational HDK is produced, even in a non-standard (ie. non-Nix) build environment.

Finally -- this supports building holochain-rust in a standard Nix build environment (such as for inclusion in the standard Nix pkgs) -- where no `git ...` build dependencies are allowed.

pjkundert:
One outstanding question I have, is why we have hdk::meta, at all?  It only knows about HDK_VERSION and HDK_HASH, both of which are compile-time constants.

Why not get rid of hc__meta(), and hdk::version...() entirely, and just make `HDK_VERSION` and `HDK_HASH` into hdk contants `hdk::VERSION` and `hdk::HASH` constants (like `hdk::DNA_ADDRESS`, `hdk::AGENT_ADDRESS`, etc.)?

Is there something else planned for hc_meta in the future?

Issue #1768, opened 2019-10-16
On restart of a persisted instance, getting "Failed to load instance test-instance from storage: DnaMissing"


Connoropolous:
Then it goes to 
`Initializing new chain...`

I might be missing something, but I think this is a bug wherein you can't actually reload a whole chain from storage :s

Connoropolous:
I think the problem is that it doesn't exit when it encounters this kind of error...
https://github.com/holochain/holochain-rust/blob/d83246669ba565ce659cb036a5353ab1fae96d02/conductor_api/src/conductor/base.rs#L880-L888

Connoropolous:
@lucksus @AshantiMutinta @zippy can you provide thoughts on this?

Connoropolous:
In general this makes sense cause I hear that it doesn't actually write the DNA into storage

jamesray1:
May you please add more repro steps?

thedavidmeister:
@jamesray1 yes you couldn't safely save any state that is only unique to a given process (e.g. as opposed to UUIDs) - that said i haven't looked around to see if there are still snowflakes in the state (those comments are pretty old)

jamesray1:
LTM like there are still snowflakes, e.g.:

In the network state:
https://github.com/holochain/holochain-rust/blob/aa53657ee25bd35304c177a8d5560318644b0ccf/crates/core/src/network/state.rs#L68

In ActionWrapper
https://github.com/holochain/holochain-rust/blob/master/crates/core/src/action.rs#L40-L55

etc:
https://github.com/holochain/holochain-rust/search?q=snowflake&unscoped_q=snowflake

thedavidmeister:
ok, if these are being saved/loaded it makes sense to me to use UUIDs instead

PR #1876, opened 2019-11-14
Control direct CAS access

willemolding:
## PR summary

A bunch of places in the codebase access the CAS (chain or DHT) directly without any kind of state control. This made a recent bug (where loading from persistence was broken and entries would not be gossiped) difficult to detect. It also is a potential security risk.

There were also multiple implementations of a `get_entry_from_cas` style function scattered around the codebase.

This tidies up all CAS access through three traits: `ContentStore`, `AddContent` and `GetContent`. These are designed to be implemented by ChainStore and DhtStore.

`ContentStore` provides a function to expose the underlying raw store. Both `AddContent` and `GetContent` have default implementations for the simple case or can be overridden in the future to include extra protection and checks. This has the added bonus that by only bringing one of the traits into scope (e.g. only GetContent) you can be sure that no mutations to the store are possible.


Connoropolous:
@willemolding does this relate to this issue? James Ray recently brought it back up with me .. 
https://github.com/holochain/holochain-rust/issues/1768

willemolding:
> @willemolding does this relate to this issue? James Ray recently brought it back up with me ..
> #1768

I believe there was issues with loading from persistence but that has already been fixed. This is really just a refactor.

crates/core/src/agent/chain_store.rs
```diff
@@ -217,6 +184,16 @@ impl ChainStore {
     }
 }
 
+impl ContentStore for ChainStore {
+    fn content_storage(&self) -> Arc<RwLock<dyn ContentAddressableStorage>> {
```
lucksus: I don't see it as an improvement that we re-add a function that returns a reference to the lock. It enables creation of lock guards in arbitrary scopes which is what led to those dreaded deadlocks. Also, it enables calling contexts to write to the CAS with an immutable reference to the state.

crates/core/src/content_store.rs
```diff
@@ -0,0 +1,40 @@
...
+        Ok((*self.content_storage().read().unwrap()).fetch(address)?)
+    }
+
+    fn contains(&self, address: &Address) -> Result<bool, HolochainError> {
+        Ok(self.get_raw(address)?.is_some())
+    }
+}
+
+pub trait AddContent: ContentStore {
+    fn add<T: AddressableContent>(&self, content: &T) -> Result<(), HolochainError> {
```
lucksus: At the very least, this function should take a `&mut self` so we can't write to the case without a mutable reference. But with the function `content_storage()` present, it doesn't do much since anybody could just get a write lock on the CAS without using this subtrait.

crates/core/src/dht/dht_store.rs
```diff
@@ -276,16 +260,10 @@ impl DhtStore {
...
-    ) -> Result<Option<Entry>, HolochainError> {
-        if let Some(json) = self.cas_fetch(&address)? {
-            let entry = Entry::try_from_content(&json)?;
-            Ok(Some(entry))
-        } else {
-            Ok(None) // no errors but entry is not in CAS
-        }
+    // Getters (for reducers)
+    // =======
+    pub(crate) fn meta_storage(&self) -> Arc<RwLock<dyn EntityAttributeValueStorage<Attribute>>> {
```
lucksus: One of the main fixes of https://github.com/holochain/holochain-rust/pull/1852 was to remove this function (and `content_storage()`) to encapsulate the lock inside the state and not pass around references to the lock or it's lock guards. This reverts that.

crates/core/src/agent/chain_store.rs
maackle: This function is private, so it's only available in this module, in the two trait implementations. So the guard cannot leak out. Does that change your outlook?

crates/core/src/dht/dht_store.rs
maackle: Agree that this should be removed.

crates/core/src/agent/chain_store.rs
maackle: Actually I want to look into if it is possible to access this by importing ChainStore...if this is publicly accessible then I think this is a problem.

maackle: OK yeah, I misunderstood how trait methods work. It turns out that by importing `core::content_store::ContentStore`, you can then access this method, which would leak out the Mutex, which indeed is something we don't want to risk happening again.

One suggestion would be to remove ContentState, and have structs implement `GetContent` and `AddContent` explicitly. A little extra boilerplate, but offers proper encapsulation.

willemolding: @maackle Yeah I think that is the best option. It doesn't take much more code really but means we lose the awesome default implementations

crates/core/src/content_store.rs
willemolding: Ok I got rid of content_storage to keep all that stuff internal. It didn't add much code duplication at all really thanks to `get_raw`. And add now takes a `&mut self`

crates/core/src/dht/dht_store.rs
willemolding: Turns out this was only used once internally so it was easy to drop it completely

willemolding:
I couldn't figure out how to make it work using a `&mut self` for add. I am also not sure if this is actually what we want. Thoughts @lucksus @maackle ?

lucksus:
So you're removing this function with a mutable self reference:
```rust
pub(crate) fn cas_add<T: AddressableContent>(&mut self, content: &T) -> HcResult<()>
```
and you replace it with this trait implementation:
```rust
impl AddContent for ChainStore {
    fn add<T: AddressableContent>(&self, content: &T) -> Result<(), HolochainError> {
        (*self.content_storage.write().unwrap())
            .add(content)
            .map_err(|e| e.into())
    }
}
```

Why couldn't it be `AddContent::add(&mut self, content: &T)`? If it used exactly in those places where we currently use `cas_add()` it should be fine since we already have mutable references there.

The point is that adding something to the case mutates it. The reason the compiler isn't complaining with `&self` is that we access the CAS through a `RwLock` anyway. But we don't want random code outside of reducers to be able to mutate the State. And we decided to model CAS/EAV as part of the State. Requiring a mutable reference for add is just making a bit harder for somebody to do something that very likely is the wrong approach.

crates/core/src/content_store.rs
```diff
@@ -0,0 +1,25 @@
...
+        }
+    }
+
+    fn contains(&self, address: &Address) -> Result<bool, HolochainError> {
+        Ok(self.get_raw(address)?.is_some())
+    }
+}
+
+pub trait AddContent {
+    fn add<T: AddressableContent>(&self, content: &T) -> Result<(), HolochainError>;
```
lucksus: I still would like to see this change:
```suggestion
    fn add<T: AddressableContent>(&mut self, content: &T) -> Result<(), HolochainError>;
```
(plus the according changes in the two implementaions of this trait)

crates/core/src/agent/chain_store.rs
```diff
@@ -217,6 +187,20 @@ impl ChainStore {
     }
 }
 
+impl GetContent for ChainStore {
+    fn get_raw(&self, address: &Address) -> Result<Option<Content>, HolochainError> {
+        Ok((*self.content_storage.read().unwrap()).fetch(address)?)
```
struktured: Maybe use `HcResult<Option<Content>>` instead of plain `Result`?

crates/core/src/agent/chain_store.rs
```diff
@@ -217,6 +187,20 @@ impl ChainStore {
...
+    }
+}
+
+impl AddContent for ChainStore {
```
struktured: Same- can use `HcResult` here.

crates/core/src/content_store.rs
```diff
@@ -0,0 +1,25 @@
+use holochain_core_types::{entry::Entry, error::HolochainError};
+use holochain_persistence_api::cas::content::{Address, AddressableContent, Content};
+
+pub trait GetContent {
+    /// Return the content at this addres, do not attempt to convert to an entry
```
struktured:
```suggestion
    /// Return the content at this address, do not attempt to convert to an entry
```

crates/core/src/dht/dht_store.rs
```diff
@@ -199,18 +201,23 @@ impl DhtStore {
...
-                    .map(|content| ChainHeader::try_from_content(&content))
+                    .map(|entry| match entry {
+                        Entry::ChainHeader(chain_header) => Ok(chain_header),
+                        _ => Err(HolochainError::ErrorGeneric(
+                            "Unexpected non-chain_header entry".to_string(),
+                        )),
+                    })
                     .collect::<Result<Vec<_>, _>>()
             })?
             .map_err(|err| {
```
struktured: Can probably delete this entire `map_err` block.

willemolding:
> So you're removing this function with a mutable self reference:
> 
> ```rust
> pub(crate) fn cas_add<T: AddressableContent>(&mut self, content: &T) -> HcResult<()>
> ```
> 
> and you replace it with this trait implementation:
> 
> ```rust
> impl AddContent for ChainStore {
>     fn add<T: AddressableContent>(&self, content: &T) -> Result<(), HolochainError> {
>         (*self.content_storage.write().unwrap())
>             .add(content)
>             .map_err(|e| e.into())
>     }
> }
> ```
> 
> Why couldn't it be `AddContent::add(&mut self, content: &T)`? If it used exactly in those places where we currently use `cas_add()` it should be fine since we already have mutable references there.
> 
> The point is that adding something to the CAS means to mutates it. The reason the compiler isn't complaining with `&self` is that we access the CAS through a `RwLock` anyway. But we don't want random code outside of reducers to be able to mutate the State. And we decided to model CAS/EAV as part of the State. Requiring a mutable reference for add is just making a bit harder for somebody to do something that very likely is the wrong approach.

I don't think we do have a mutable reverence to the store in the places where it is used (which seems to be only in tests). Since it is inside an `Arc` we can't get a mutable reference right? Happy to be proven wrong but I can't figure out the right way to do it.

lucksus:
> > So you're removing this function with a mutable self reference:
> > ```rust
> > pub(crate) fn cas_add<T: AddressableContent>(&mut self, content: &T) -> HcResult<()>
> > ```
> > 
> > 
> > and you replace it with this trait implementation:
> > ```rust
> > impl AddContent for ChainStore {
> >     fn add<T: AddressableContent>(&self, content: &T) -> Result<(), HolochainError> {
> >         (*self.content_storage.write().unwrap())
> >             .add(content)
> >             .map_err(|e| e.into())
> >     }
> > }
> > ```
> > 
> > 
> > Why couldn't it be `AddContent::add(&mut self, content: &T)`? If it used exactly in those places where we currently use `cas_add()` it should be fine since we already have mutable references there.
> > The point is that adding something to the CAS means to mutates it. The reason the compiler isn't complaining with `&self` is that we access the CAS through a `RwLock` anyway. But we don't want random code outside of reducers to be able to mutate the State. And we decided to model CAS/EAV as part of the State. Requiring a mutable reference for add is just making a bit harder for somebody to do something that very likely is the wrong approach.
> 
> I don't think we do have a mutable reverence to the store in the places where it is used (which seems to be only in tests). Since it is inside an `Arc` we can't get a mutable reference right? Happy to be proven wrong but I can't figure out the right way to do it.

The code currently in develop only offers above `cas_add(&mut self,..)` function as API for writing to the CAS. Except tests, there is no (and should be no) write to the case outside of reducers where we do have a mutable reference. This is what came in new last week and what this PR would revert - hence my change-request.

Those few places in the tests where we don't have direct, mutable access to the state (which could be refactored and is an old style of writing the tests that is a remainder from before some refactorings) I cloned the `DhtStore` out of its `Arc` into a mutable local stack variable and then called `cas_add` on that clone - which is totally a hack and just a quick way of getting this to work without refactoring those tests (which is what we should do eventually). I would love to make this impossible and then actually do refactor those tests.

```rust
let _ = (*store.dht()).clone().cas_add(&entry);
```

You actually have changed those lines within this branch, I copied from the diff view :)

PR #1878, opened 2019-11-15
Several improvements and fixes during investigating zome call lock

lucksus:
## PR summary

Started off of looking into this lock up: https://gist.github.com/lucksus/ac5fcaf24993954bc75878786641becf

1. Pulled the driving of the zome call future out of the scope of the conductor lock which was accidentally only allowing one zome call per conductor (!) at a time.
2. Next thing that showed up was a previously masked [deadlock when persisting instance states](https://github.com/holochain/holochain-rust/pull/1878/commits/e684eb29a2f8d1f0193065ffd38f9df8e15c8318)
3. Some fixes of the state dump that I needed to use Holoscape's debug view for further debugging
4. With that it became apparent that there was a bug in the new holding queue code which resulted in entries being held many times - the last three commits here fix that

## changelog

```markdown
```

## documentation


crates/core/src/dht/actions/pop_next_holding_workflow.rs
```diff
@@ -4,10 +4,43 @@ use crate::{
...
+
+    fn poll(self: Pin<&mut Self>, cx: &mut std::task::Context) -> Poll<Self::Output> {
+        //
+        // TODO: connect the waker to state updates for performance reasons
+        // See: https://github.com/holochain/holochain-rust/issues/314
+        //
+        cx.waker().clone().wake();
+
+        if let Some(state) = self.context.try_state() {
+            match state.dht().next_queued_holding_workflow() {
```
maackle: Would be great to make `next_queued_holding_workflow` return an address, because this is going to clone a bunch of data on every poll, which could add up

crates/core/src/instance.rs
```diff
@@ -267,6 +273,60 @@ impl Instance {
...
+                            .state()
+                            .expect("Couldn't get state in run_pending_validations")
+                            .dht()
+                            .next_queued_holding_workflow();
+                        if let Some(pending) = maybe_holding_workflow {
+                            log_debug!(context, "Found queued validation: {:?}", pending);
+                            context.block_on(pop_next_holding_workflow(
+                                pending.clone(),
+                                context.clone(),
+                            ));
```
maackle: Can't we just pop in one step, rather than checking the next queued item and then popping? Having two steps creates the possibility of a very slim race condition.

crates/core/src/state_dump.rs
```diff
@@ -41,8 +43,9 @@ impl From<Arc<Context>> for StateDump {
         let source_chain: Vec<ChainHeader> = source_chain.into_iter().rev().collect();
 
         let queued_calls: Vec<ZomeFnCall> = nucleus.queued_zome_calls.into_iter().collect();
-
         let running_calls: Vec<ZomeFnCall> = nucleus.running_zome_calls.into_iter().collect();
+        let call_results: Vec<(ZomeFnCall, Result<_, _>)> =
+            nucleus.zome_call_results.into_iter().collect();
```
maackle: I just noticed that we're now storing all zome call results in the state. I'm concerned about this growing huge for zome calls that return get_links. Imagine a DNA with hundreds of users with a UI that polls a get_all_messages zome function periodically...

crates/conductor_lib/src/holochain.rs
```diff
@@ -240,53 +239,37 @@ impl Holochain {
...
         params: &str,
     ) -> HolochainResult<JsonString> {
-        let context = self.context.as_ref().unwrap();
         let metric_name = format!("{}.{}", zome, fn_name);
         with_latency_publishing!(
             metric_name,
             context.metric_publisher,
             Self::call_inner,
-            self,
+            context.clone(),
```
maackle: a field of the context and the context itself are both passed into this macro, could be unified

lucksus: The 3rd macro argument is a function to call and all following arguments are passed to that function. So the cloned context is actually the first argument to `Self::call_inner`.

crates/core/src/state_dump.rs
lucksus: Yep, but this is not new. So far this wasn't a problem (as the zome call results do net get persisted in state snapshots) but I agree we should add a way of clearing that Vec, or transform it into a ring buffer - but in a separate PR.

crates/core/src/instance.rs
lucksus: It is hard to make pop an atomic action like you would if the queue would be owned by the local thread, but the queue lives in the instance's state and can only get mutated by reducers run in the redux action thread. The race condition is taken care of by having the item you think you're popping be an argument of the pop action. If the head is not that item anymore the reducer will not pop the next one.

crates/conductor_lib/src/holochain.rs
maackle: @lucksus I'm talking about the second and fourth arguments, `context.metric_publisher` and `context`

maackle: Oh I see, it's kind of a funky macro. I guess this works.

PR #1884, opened 2019-11-19
Sim2h gossip improvements

lucksus:
## PR summary

1. Don't try to fill missing aspects by asking an agent that also doesn't have those aspects (which is bound to create an `EntryNotFoundLocally` error on that node)
2. Merge authoring list into gossip list so sim2h has authoring nodes as sources for gossiping newly authored data (as it wouldn't ask those nodes otherwise with 1.)
3. Clear cached missing entry data when agent leaves and drop all cached data about space when the last agents leaves the space.

## testing/benchmarking notes

If a large number of nodes enter into a network at the same time (as tested in our stress-tests), the assumption that was underlying the current gossip-mock code in sim2h doesn't apply: if a node is missing an entry aspect, sim2h assumes all other nodes to have it and picks one randomly. But if 50 nodes come on at the same time, all of them is missing entry aspects (namely agent_id and chain headers of the other 49 nodes). But only one of those other nodes actually has that data - drawing one from the whole pool with uniform distribution leads to a gossip failure in 98% of cases.

The changes in this branch might reduce the network load in our stress-tests substantially.

## followups

## changelog

```markdown
```

## documentation


crates/sim2h/src/cache.rs
```diff
@@ -19,7 +19,7 @@ pub struct Space {
     crypto: Box<dyn CryptoSystem>,
     agents: HashMap<AgentId, AgentInfo>,
     all_aspects_hashes: AspectList,
-    missing_aspects: HashMap<AgentId, HashSet<AspectHash>>,
+    missing_aspects: HashMap<AgentId, HashMap<EntryHash, Vec<AspectHash>>>,
```
willemolding: Can this be `HashSet<AspectHash>` rather than `Vec<AspectHash>` to make it explicit there can be no duplicates?

crates/sim2h/src/lib.rs
```diff
@@ -503,14 +507,14 @@ impl Sim2h {
...
-                if aspect_hashes.len() > 0 {
+                let missing_hashes: HashSet<(EntryHash, AspectHash)> = (&aspects_missing_at_node).into();
+                if missing_hashes.len() > 0 {
                     let mut space = self.spaces
                         .get(space_address)
                         .expect("This function should not get called if we don't have this space")
                         .write();
-                    for aspect_hash in aspect_hashes.clone() {
-                        space.add_missing_aspect(agent_id.clone(), aspect_hash);
+                    for pair in missing_hashes {
```
willemolding: You could de-structure here to make it more readable e.g.
```
for (entry_hash, aspect_hash) in missing_hashes
```
and this avoids having to use `pair.0, pair.1`

crates/sim2h/src/cache.rs
zippy: :+1:

crates/sim2h/src/cache.rs
```diff
@@ -58,6 +74,41 @@ impl Space {
         self.missing_aspects.keys().cloned().collect()
     }
 
+    /// Returns true if the given agent agent is missing all of the given aspects for the given entry.
```
zippy:
```suggestion
    /// Returns true if the given agent is missing all of the given aspects for the given entry.
```

crates/sim2h/src/cache.rs
```diff
@@ -58,6 +74,41 @@ impl Space {
...
+    /// That is: if all of the aspects are stored as missing for that agent.
```
zippy:
```suggestion
    /// That is: if all of the aspects are recorded as missing for that agent.
```

crates/sim2h/src/cache.rs
```diff
@@ -58,6 +74,41 @@ impl Space {
...
+    /// If one of the given aspects is not in that vector of missing entries, the agent is supposed
+    /// to have it and this function returns fals.
```
zippy:
```suggestion
    /// to have it and this function returns false.
```

crates/core/src/network/handler/lists.rs
```diff
@@ -19,44 +19,47 @@ use std::{collections::HashMap, sync::Arc};
...
+            space_address: get_list_data.space_address,
+            provider_agent_id: get_list_data.provider_agent_id,
+            request_id: get_list_data.request_id,
+            address_map,
         });
+        dispatch_action(context.action_channel(), ActionWrapper::new(action));
+    });
+}
+
+fn create_authoring_map(context: Arc<Context>) -> HashMap<EntryHash, Vec<AspectHash>> {
```
maackle: Do we want this to be `HashSet<AspectHash>` as well?

crates/core/src/network/handler/lists.rs
```diff
@@ -81,26 +84,34 @@ fn get_all_aspect_addresses(entry: &Address, context: Arc<Context>) -> HcResult<
     Ok(address_list)
 }
 
+fn create_holding_map(context: Arc<Context>) -> HashMap<EntryHash, Vec<AspectHash>> {
```
maackle: HashSet?

crates/sim2h/src/cache.rs
```diff
@@ -58,6 +74,41 @@ impl Space {
...
 
+    /// Returns true if the given agent is missing all of the given aspects for the given entry.
+    /// That is: if all of the aspects are recorded as missing for that agent.
+    /// If one of the given aspects is not in that vector of missing entries, the agent is supposed
+    /// to have it and this function returns false.
+    pub fn agent_is_missing_all_aspects(
+        &self,
+        agent_id: &AgentId,
+        entry_hash: &EntryHash,
+        aspects: &Vec<AspectHash>,
```
maackle: This should also be a HashSet, and then we can use set operations to answer the question this function asks

crates/sim2h/src/cache.rs
```diff
@@ -58,6 +74,41 @@ impl Space {
...
+        let missing_aspects_for_entry = maybe_vec_of_missing_aspects_for_entry.unwrap();
+
+        // We check that every of the given aspects is the missing list.
+        // If one is missing from the missing list this block returns some
+        // and the whole function returns false.
+        for aspect in aspects {
+            if !missing_aspects_for_entry.contains(aspect) {
+                return false;
+            }
+        }
```
maackle: FYI we could use [`HashSet::difference()`](https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.difference) for this, but I'm not sure if the speed gain is worth it

crates/sim2h/src/lib.rs
```diff
@@ -578,40 +587,75 @@ impl Sim2h {
...
+    ) -> Option<AgentId> {
+        let agents_not_missing_entry: Vec<&AgentId> = {
+            let space_lock = self.spaces.get(space_address)?.read();
+            agent_pool
+                .iter()
+                // We ignore all agents that are missing all of the same aspects as well since
+                // they can't help us.
+                .filter(|a| !space_lock.agent_is_missing_all_aspects(*a, entry_hash, aspects))
+                .collect()
+        };
```
maackle: This could get slow as we scale up the number of agents, since we are iterating over all agents and calculating which are missing aspects. Since we ultimately only want one agent that meets the criteria, a faster way would be to generate a random permutation of vec indices so we can iterate over the agents randomly, and return the first one that doesn't have missing aspects.

Can use: https://docs.rs/rand/0.6.5/rand/seq/trait.SliceRandom.html#tymethod.shuffle

crates/core/src/network/handler/lists.rs
```diff
@@ -25,21 +28,76 @@ pub fn handle_get_authoring_list(get_list_data: GetListData, context: Arc<Contex
...
+                EntryAspect::Deletion(header),
+            )),
+            _ => None,
+        };
+
+        if let Some((base_address, meta_aspect)) = maybe_meta_aspect {
+            address_map
+                .entry(EntryHash::from(base_address.clone()))
+                .or_insert_with(|| HashSet::new())
+                .insert(AspectHash::from(meta_aspect.address()));
```
maackle: We should hold on to the entry from before, when we added the content aspect. We are certain that the entry exists because we created it there. Then we can just insert. There should be no chance that the aspect list will only hold a meta entry and not the content entry, which this code currently implies.

maackle: Nevermind, read this wrong, didn't see it's for a different entry

crates/core/src/network/handler/lists.rs
```diff
@@ -57,7 +115,9 @@ fn create_authoring_map(context: Arc<Context>) -> HashMap<EntryHash, Vec<AspectH
             &Vec::new(),
         ).expect("Must be able to create dummy header header when responding to HandleGetAuthoringEntryList");
         let content_aspect = EntryAspect::Content(chain_header_entry, header_entry_header);
-        address_map.insert(entry_hash, vec![AspectHash::from(content_aspect.address())]);
+        let mut set = HashSet::new();
+        set.insert(AspectHash::from(content_aspect.address()));
+        address_map.insert(entry_hash, set);
```
maackle: Is there a chance that this loop could visit the same entry_hash multiple times? If so then we will overwrite the aspect set for that entry on each loop..

crates/core/src/network/handler/lists.rs
```diff
@@ -108,16 +168,20 @@ fn create_holding_map(context: Arc<Context>) -> HashMap<EntryHash, Vec<AspectHas
...
+        let mut address_map = create_authoring_map(context.clone());
         let holding_map = create_holding_map(context.clone());
-
-        let address_map = authoring_map.into_iter().chain(holding_map).collect();
+        holding_map.into_iter().for_each(|(entry, set)| {
+            address_map
+                .entry(entry)
+                .or_insert_with(|| set.clone())
+                .union(&set);
+        });
```
maackle: I don't think this does anything, because `union` does not modify a set but returns a new one! I can fix this one.

PR #1890, opened 2019-11-19
NucleusState::pending_validations -> DhtStore::queued_holding_workflows & fix too fast validation retry

lucksus:
## PR summary

In https://github.com/holochain/holochain-rust/pull/1852 I already added `DhtStore::queued_holding_workflows` to manage the execution of entry holding workflows instead of just spawning an unlimited number of threads for each. That already made the good old `PendingValidations` become redundant.

This PR cleans things up and removes all old redundant pending validation code that now gets handled more elegantly since every entry validation is added to a queue - validations that fail due to missing dependencies just get readded to that queue.

Also this adds an exponential back-off for the first time, doubling the delay every time a holding workflows fails again (capped by at least once per hour).

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation


crates/core/src/dht/dht_store.rs
```diff
@@ -297,12 +302,37 @@ impl DhtStore {
...
+                if let Ok(time_elapsed) = maybe_time_elapsed {
+                    if time_elapsed < *delay {
+                        continue;
+                    }
+                }
+                return Some((pending, Some(*delay)));
+            }
+            return Some((pending, None));
+        }
+        None
```
AshantiMutinta: Can we get away with a skip_while here, or will that make it more complex? It would be nice to see some comments here on what is happening either way

self.queued_holding_workflows
.iter()
.skip_while(|(pending,maybe_delay)|{/// comparison on time_elapsed})
.map(|(pending,maybe_delay)|(pending,maybe_delay.map(|f,s|Some(s)).unwrap_or(None)
next()

crates/core/src/dht/dht_store.rs
```diff
@@ -297,12 +302,37 @@ impl DhtStore {
...
     }
 
     pub(crate) fn has_queued_holding_workflow(&self, pending: &PendingValidation) -> bool {
-        self.queued_holding_workflows.contains(pending)
+        for (current, _) in self.queued_holding_workflows.iter() {
+            if current == pending {
+                return true;
+            }
+        }
+        false
```
AshantiMutinta: Could be a one liner

self.queued_holding_workflows.iter().any(|(current,_)|current==pending)

crates/core/src/instance.rs
```diff
@@ -307,9 +307,28 @@ impl Instance {
...
-                                ),
+                                Err(HolochainError::ValidationPending) => {
+                                    // And with a delay so we are not trying to re-validate many times per second.
+                                    let mut delay = if let Some(old_delay) = maybe_delay {
+                                        // Exponential back-off:
+                                        // If this was delayed before we double the delay.
+                                        old_delay * 2
+                                    } else {
+                                        Duration::from_secs(10)
+                                    };
```
AshantiMutinta: Could also be a one liner let mut delay = maybe_delay.map(|old_delay| old_delay*2).unwrap_or_else(|| Duation::from_secs(10)) but not necessary

crates/core/src/action.rs
```diff
@@ -112,11 +113,12 @@ pub enum Action {
     // DHT actions:
     // -------------
     /// Adds a holding workflow to the queue.
-    QueueHoldingWorkflow(PendingValidation),
+    /// With optional delay where the SystemTime is the time when the action got dispatched
```
zippy: can we add a line of comment here explaining the use of the PendingValidation struct is for here?

PR #1893, opened 2019-11-21
Ensure query only responds holding list

willemolding:
## PR summary

Follow on from #1876 

Using the traits introduced above there is a single place we can add management of the DHT holding list. It is now self managing and the `mark_entry_as_held` can be made private.

Entries added to the DHT are now added to the holding list automatically and reads will return `None` if an entry is not in the holding list rather than checking the CAS. This means less CAS reads and less chance of leaking private entries!

## changelog

```markdown
```

## documentation

### DIFF
--- a/crates/core/src/dht/dht_reducers.rs
+++ b/crates/core/src/dht/dht_reducers.rs
@@ -199,23 +199,24 @@ pub mod tests {
 
     use crate::{
         action::{Action, ActionWrapper},
-        content_store::{AddContent, GetContent},
+        content_store::{GetContent},
         dht::{
             dht_reducers::{reduce, reduce_hold_entry},
             dht_store::create_get_links_eavi_query,
         },
         instance::tests::test_context,
-        network::entry_with_header::EntryWithHeader,
+        network::entry_with_header::{test_entry_with_header},
         state::test_store,
     };
     use holochain_core_types::{
         agent::{test_agent_id, test_agent_id_with_name},
         chain_header::test_chain_header,
         eav::Attribute,
-        entry::{test_entry, test_sys_entry, Entry},
+        entry::{test_entry, Entry},
         link::{link_data::LinkData, Link, LinkActionKind},
     };
     use holochain_persistence_api::cas::content::AddressableContent;
+    use std::sync::Arc;
 
     #[test]
     fn reduce_hold_entry_test() {
@@ -223,11 +224,8 @@ pub mod tests {
         let store = test_store(context);
 
         // test_entry is not sys so should do nothing
-        let sys_entry = test_sys_entry();
-        let entry_wh = EntryWithHeader {
-            entry: sys_entry.clone(),
-            header: test_chain_header(),
-        };
+        let entry_wh = test_entry_with_header();
+        let entry = entry_wh.entry.clone();
 
         let new_dht_store =
             reduce_hold_entry(&store.dht(), &ActionWrapper::new(Action::Hold(entry_wh)))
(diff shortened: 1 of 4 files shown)


crates/core/src/dht/dht_reducers.rs
```diff
@@ -251,9 +246,12 @@ pub mod tests {
     fn can_add_links() {
         let context = test_context("bob", None);
         let store = test_store(context.clone());
-        let entry = test_entry();
+        let entry_wh = test_entry_with_header();
+        let entry = entry_wh.entry.clone();
+
+        let store = reduce_hold_entry(&store.dht(), &ActionWrapper::new(Action::Hold(entry_wh)))
```
willemolding: Had to rewrite this test to use proper mutations of the DHT store. This is a better way of doing things anyway.

crates/core/src/nucleus/actions/get_entry.rs
```diff
@@ -126,20 +126,14 @@ pub fn get_entry_with_meta(
...
     #[test]
     fn test_get_entry_from_dht_cas() {
-        let entry = test_entry();
-        let context = test_context_with_state(None);
-        let result = super::get_entry_from_dht(&context, &entry.address());
-        assert_eq!(Ok(None), result);
-        let _ = (*context.state().unwrap().dht()).clone().add(&entry);
-        let result = super::get_entry_from_dht(&context, &entry.address());
-        assert_eq!(Ok(Some(entry.clone())), result);
+    // write this test when its easier to get a mutable dht state
```
willemolding: This test is no longer an easy thing to write as it is impossible to get a mutable reference to the DhtStore and then put it back in the state. Rewriting this would require setting up the context to process actions and it doesn't seem worth it to test a one-line function that is tested higher up anyway.

crates/core/src/dht/dht_store.rs
```diff
@@ -317,15 +317,19 @@ impl DhtStore {
 
 impl GetContent for DhtStore {
     fn get_raw(&self, address: &Address) -> HcResult<Option<Content>> {
-        Ok((*self.content_storage.read().unwrap()).fetch(address)?)
+        if self.holding_list.contains(address) {
```
lucksus: I said yesterday in voice chat that I would be surprised (in a good way) if we can actually have this check here on this general level, as I was suspecting the query and fetch code to be assuming direct CAS access. 

I still haven't checked this but I see both app-spec and stress-test fail and have this error message in the logs: `Error getting entry aspects of authoring list: EntryIsPrivate`

willemolding: > I said yesterday in voice chat that I would be surprised (in a good way) if we can actually have this check here on this general level, as I was suspecting the query and fetch code to be assuming direct CAS access.
> 
> I still haven't checked this but I see both app-spec and stress-test fail and have this error message in the logs: `Error getting entry aspects of authoring list: EntryIsPrivate`

Ah interesting I think this might need a small revisit. I saw the app spec pass locally and assumed it was just flakyness on the CI side. Its really close though!
```
# tests 179
# pass  177
# fail  2
```

willemolding: and passes on a rerun... Still no luck with the stress test though

willemolding: And this was repeatable locally as well. I didn't think to run the stress tests (I assumed if app spec passed so would they. Any idea what could be causing stress tests to fail but not app spec?

crates/core/src/dht/dht_store.rs
```diff
@@ -317,15 +317,19 @@ impl DhtStore {
...
     }
 }
 
 impl AddContent for DhtStore {
     fn add<T: AddressableContent>(&mut self, content: &T) -> HcResult<()> {
-        (*self.content_storage.write().unwrap())
-            .add(content)
-            .map_err(|e| e.into())
+        (*self.content_storage.write().unwrap()).add(content)?;
+        self.mark_entry_as_held(content);
```
willemolding: Ok further investigations show the error is caused by this line. The assumption is that everything added to the DHT belongs on the hold list but there must be some exceptions to this (likely a Dna | CapTokenGrant | CapTokenClaim )

I don't think this code is at fault but rather it has revealed a fault in another part of the code base where something was added to the CAS via the DHT store that was not meant to be shared

zippy:
Hey @willemolding, the PR is going to need to be harmonized with the work we've done over here in: https://github.com/holochain/holochain-rust/pull/1904

PR #1896, opened 2019-11-21
Use FakeServer for stream_manager to avoid binding to port, for now

maackle:
## PR summary

I heard that Sim2hWorker is binding to a port that it never uses. This switches the stream manager to not do that. The reason is that the port binding is random and interferes with the conductor interface port which must be chosen explicitly and passed as config. The worker port gets bound before the interface port does, so there is a chance of collisions, and we see this often in CI tests.

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation

### DIFF
--- a/crates/sim2h/src/websocket/streams.rs
+++ b/crates/sim2h/src/websocket/streams.rs
@@ -223,8 +223,9 @@ impl<T: Read + Write + std::fmt::Debug> StreamManager<T> {
 
     fn priv_process_accept(&mut self) -> DidWork {
         match &mut self.acceptor {
-            Err(err) => {
-                warn!("acceptor in error state: {:?}", err);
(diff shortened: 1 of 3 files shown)


crates/sim2h/src/websocket/streams.rs
```diff
@@ -223,8 +223,9 @@ impl<T: Read + Write + std::fmt::Debug> StreamManager<T> {
 
     fn priv_process_accept(&mut self) -> DidWork {
         match &mut self.acceptor {
-            Err(err) => {
-                warn!("acceptor in error state: {:?}", err);
+            Err(_err) => {
+                // TODO: turn back on
+                // warn!("acceptor in error state: {:?}", err);
```
neonphog: :+1:

PR #1898, opened 2019-11-21
Add separate proc-macro-attribute definition crate for metrics

dymayday:
## PR summary

This PR add the definition of proc-macro-attribute as a separate crate in order to be used by the [metric crate](https://github.com/holochain/holochain-rust/tree/develop/crates/metrics).
It needs to be put in a separate crate because currently, the same crate cannot export rule macros and attribute macros in the same time.
```bash
error: `proc-macro` crate types cannot export any items other than functions tagged with `#[proc_macro_derive]` currently
```

Help is welcome to find a way to properly instantiate a metrics's publisher.

## changelog

```markdown
```

## documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,26 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.0.40-alpha1] - 2019-12-01
+
+### Added
+
+- Pruning the State of old/stale/history data to prevent it from using up a slowly but infinitely growing amount of memory. [#1920](https://github.com/holochain/holochain-rust/pull/1920)
+
+### Changed
+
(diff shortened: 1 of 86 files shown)


crates/metrics/tests/proc_macro_attr.rs
```diff
@@ -0,0 +1,16 @@
...
+
+    #[latency]
+    pub fn test_latency() {
+        ::std::thread::sleep(::std::time::Duration::from_millis(1001));
+    }
+
+    #[test]
+    fn main_test() {
+        eprintln!("Ahoy !");
+    }
```
maackle: Is there any way to make this into an actual test? Or is the metric publisher not configurable to specify where things get published to? @struktured

crates/attribute-macros/src/lib.rs
```diff
@@ -0,0 +1,92 @@
...
+
+extern crate proc_macro;
+
+use proc_macro::TokenStream;
+use proc_macro2::Span;
+use quote::quote;
+use syn::{parse_macro_input, parse_quote, Attribute, FnArg, Ident, ItemFn, PatType, ReturnType};
+
+#[proc_macro_attribute]
+pub fn latency(args: TokenStream, input_function: TokenStream) -> TokenStream {
```
maackle: I kind of agree with @struktured that this could be more descriptive. I also like publish_latency.

```suggestion
pub fn publish_latency(args: TokenStream, input_function: TokenStream) -> TokenStream {
```

PR #1902, opened 2019-11-22
--info flag for conductor

zippy:
## PR summary

adds a --info flag to the conductor that simply prints out info (HDK_HASH/VERSION and GIT_HASH if available)

This is important as a way to check during development that one is using the exact version of the conductor that one expects because the -V option isn't always enough.

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation


docker/Dockerfile.trycp_server
```diff
@@ -1,4 +1,4 @@
-FROM holochain/holochain-rust:minimal as build
+FROM holochain/holochain-rust:minimal.holochain-info as build
```
maackle: probably don't want to commit this, right?

crates/core_types/build.rs
```diff
@@ -41,4 +41,14 @@ fn main() {
...
+
+    if Path::new("../../.git/HEAD").exists() {
+        println!("cargo:rerun-if-changed=../../.git/HEAD");
+    }
+    let maybe_output = Command::new("git").args(&["rev-parse", "HEAD"]).output();
+    if maybe_output.is_ok() {
+        let output = maybe_output.unwrap();
+        let git_hash = String::from_utf8(output.stdout).unwrap();
+        println!("cargo:rustc-env=GIT_HASH={}", git_hash);
+    }
```
maackle: There's no reason to not use an `if let Some(output) = ...` here

PR #1905, opened 2019-11-22
in_stream websocket stream stack & traits

neonphog:
## PR summary

- Separates concerns of client (connection) and server (listener) so sim2h worker will no longer have to have a listening socket
- Separates code into layers, so TCP, TLS, and WebSocket code no longer have to be maintained in one monolithic stream manager
- Normalized dealing with handshakes - accepting connection from listeners, or creating new connections give you "partial" streams. you periodically "process" these partials, and they will eventually error or give you a complete stream.

## followups

- not super happy with the FrameType associated type, and FROM/INTO on the read_frame/write_frame functions... they sometimes make you turbofish.
- probably need to make the FrameType for websockets be a new type around Vec<u8> that also lets you do things like Close/Ping/Pong/utf8 frames. - added in https://github.com/holochain/holochain-rust/pull/1913

## example as included in the lib.rs doc test

```rust
use url2::prelude::*;
use in_stream::*;

let (send_binding, recv_binding) = crossbeam_channel::unbounded();

let server_thread = std::thread::spawn(move || {
    let mut listener: InStreamListenerWssType = InStreamListenerWss::bind(
        &url2!("wss://0.0.0.0:0"),
        WssBindConfig::default()
            // for unit test we want to use a fake certificate
            // because it's slow to generate an RSA keypair
            .sub_bind_config(TlsBindConfig::with_fake_certificate()),
    ).unwrap();

    send_binding.send(listener.binding()).unwrap();

    // we have a partial websocket that still needs to handshake
    let mut srv_wss_partial = listener.accept_blocking().unwrap();

    // we process it to get a full stream
    let mut srv_wss = srv_wss_partial.process_blocking().unwrap();

    srv_wss.write_frame(b"hello from server".to_vec()).unwrap();

    loop {
        std::thread::sleep(std::time::Duration::from_millis(1));

        match srv_wss.read_frame::<Vec<u8>>() {
            Ok(frame) => {
                assert_eq!(
                    "hello from client",
                    &String::from_utf8_lossy(&frame),
                );
                break;
            }
            Err(e) if e.would_block() => (),
            Err(e) => panic!("{:?}", e),
        }
    }
});

let client_thread = std::thread::spawn(move || {
    let binding = recv_binding.recv().unwrap();

    let mut cli_wss_partial: InStreamPartialWssType = InStreamPartialWss::connect(
        &binding,
        Default::default(),
    ).unwrap();

    let mut cli_wss = cli_wss_partial.process_blocking().unwrap();

    cli_wss.write_frame(b"hello from client".to_vec()).unwrap();

    loop {
        std::thread::sleep(std::time::Duration::from_millis(1));

        match cli_wss.read_frame::<Vec<u8>>() {
            Ok(frame) => {
                assert_eq!(
                    "hello from server",
                    &String::from_utf8_lossy(&frame),
                );
                break;
            }
            Err(e) if e.would_block() => (),
            Err(e) => panic!("{:?}", e),
        }
    }
});

server_thread.join().unwrap();
client_thread.join().unwrap();
```


crates/in_stream/src/mem.rs
```diff
@@ -0,0 +1,348 @@
...
+            recv_buf: Vec::new(),
+        }
+    }
+}
+
+impl Read for InStreamMem {
+    fn read(&mut self, buf: &mut [u8]) -> Result<usize> {
+        let mut disconnected = false;
+        loop {
+            // first, drain everything from our channel
```
struktured: Is the _drain everything_ idiom performant enough here? I'm not suggesting it isn't just want to make sure it's thought about. I suppose it depends on the how often `read` is called and the size / number of messages to pull from the channel.

crates/in_stream/src/mem.rs
```diff
@@ -0,0 +1,348 @@
...
+    /// create a new singleton
+    fn new() -> Self {
+        Self {
+            listeners: HashMap::new(),
+        }
+    }
+
+    /// manage binding a new MemListener interface
+    fn bind(&mut self, url: &Url2) -> Result<InStreamListenerMem> {
+        if "mem" != url.scheme() {
```
struktured: Maybe a make a constant for url scheme "mem".

crates/in_stream/src/mem.rs
```diff
@@ -0,0 +1,348 @@
...
+            return Err(Error::new(
+                ErrorKind::InvalidInput,
+                "mem bind: url scheme must be mem",
+            ));
+        }
+        match url.port() {
+            None | Some(0) | Some(4242) => (),
```
struktured: Same here, constant for port 4242

crates/in_stream/src/mem.rs
```diff
@@ -0,0 +1,348 @@
...
+            _ => {
+                return Err(Error::new(
+                    ErrorKind::InvalidInput,
+                    "mem bind: url port must be None, 0, or 4242",
```
struktured: Use constant here..

crates/in_stream/src/mem.rs
```diff
@@ -0,0 +1,348 @@
...
+                ));
+            }
+        }
+        if url.host_str().is_none() {
+            return Err(Error::new(
+                ErrorKind::InvalidInput,
+                "mem bind: host_str must be set",
+            ));
+        }
+        let new_url = Url2::parse(&format!("mem://{}:4242", url.host_str().unwrap(),));
```
struktured: And use constant for port here.

crates/in_stream/src/trait.rs
```diff
@@ -0,0 +1,94 @@
...
+make_partial!(InStreamPartial, InStream);
+
+/// a stream implementing the core std::io::Read and std::io::Write traits
+pub trait InStream: Sized + Debug + Read + Write + Send + Sync {}
+
+make_listener!(InStreamFramedListener, InStreamFramedPartial);
+make_partial!(InStreamFramedPartial, InStreamFramed);
+
+/// a framed stream designed to read and write whole messages at once
+pub trait InStreamFramed: Sized + Debug + Send + Sync {
```
struktured: Why does this have to be its own trait? I thought there would be something similar to an existing read/write trait?

crates/in_stream/src/ws.rs
```diff
@@ -0,0 +1,382 @@
...
+            .write_message(tungstenite::Message::Binary(data.into()));
+        match &res {
+            Ok(()) => Ok(()),
+            Err(tungstenite::error::Error::Io(e)) if e.would_block() => {
+                // ignore would-block errors on write
+                // tungstenite queues them in pending, they'll get sent
+                Ok(())
+            }
+            Err(tungstenite::error::Error::Io(_)) => {
+                if let Err(tungstenite::error::Error::Io(e)) = res {
```
struktured: This looks odd. What are you trying to achieve here?

neonphog: In our version of nightly, you can't do conditional match arms with owned data... you can in stable.

so... I have to use a ref for the top match, but then I need to own the Io error to return it from the function.

crates/in_stream/src/mem.rs
neonphog: added `for _ in 0.100` instead of the `loop`

neonphog:
closing to refactor

crates/in_stream/src/ws.rs
lucksus: Second reason I saw today for bumping our rust version to > 1.39.

PR #1908, opened 2019-11-24
Thread tuning

lucksus:
## PR summary

* Not spawning threads when the parent thread blocks for completion of the child but instead just run it in the parent - was the case for getting validation packages and running validation
* Not using threads from the thread pool for timeouts as this just binds a thread to have it sleep for a minute => this would eat up the thread pool quickly and made zome calls lock up, waiting for a new thread to become available

## testing/benchmarking notes

This makes stress tests much faster and solves the zome call lock-up issue.

## followups

We should implement timeouts with just one thread that polls the state looking for timeouts to trigger instead of creating a thread per task just to have it sleep.

## changelog

```markdown
```

## documentation


willemolding:
@freesig does this touch on any of the things you were seeing in the profiler?

lucksus:
@zippy, @willemolding, please note I've added another commit that removes what has become dead code by making validation and validation package calls synchronous and not mediated by actions and the State.

crates/core/src/nucleus/actions/build_validation_package.rs
```diff
@@ -9,25 +8,20 @@ use crate::{
...
     error::HolochainError,
     signature::Provenance,
-    ugly::lax_send_sync,
     validation::{ValidationPackage, ValidationPackageDefinition::*},
 };
-use snowflake;
-use std::{pin::Pin, sync::Arc, vec::Vec};
+use std::{sync::Arc, vec::Vec};
 
 pub async fn build_validation_package<'a>(
```
willemolding: I believe you can drop the `async` from this function now right?

crates/core/src/nucleus/actions/run_validation_callback.rs
```diff
@@ -21,95 +18,45 @@ use holochain_metrics::Metric;
 /// Dispatches an `Action::ReturnValidationResult` after completion of the WASM call.
 /// Returns a future that waits for the result to appear in the nucleus state.
 pub async fn run_validation_callback(
```
willemolding: and same here?

freesig:
> @freesig does this touch on any of the things you were seeing in the profiler?

It's a little hard to tell as my knowledge of how the whole code base plays together is very limited but this isn't the stuff I was looking at in our call. 

I can profile this PR if you like though

freesig:
This is looking faster but still getting this pattern 
image

PR #1910, opened 2019-11-25
Draft - Rollback header publishing

willemolding:
## PR summary

In preparation for a possible alternative the header entries have been rolled back. This should reduce storage and validation load during testing and make things much easier to debug.

The plan is to re-introduce a similar concept but with a slightly different implementation to allow offline agent full chain validation.

## changelog

```markdown
```

## documentation


willemolding:
Ok actually we don't want to remove this right now. Good to know if we want to it is super easy

PR #1911, opened 2019-11-25
WIP: 2019 11 25 publish dry run

thedavidmeister:
## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation


thedavidmeister:
```
so this is a bit of a PITA... i wanted to have a dry run of publishing crates so that we wouldn't find out that the publish fails at the last minute...

19:20
but when we bump dependencies in things that causes the dry run to fail, because it still points to the published crate 

19:21
so for example we have a new version of native-tls that works when everything is using local paths to compile, but when the publish dry run happens it goes back to the published 0.0.40-alpha1 which has the old version of native-tls which is incompatible with our local crates and then  

19:22
honestly i don't know what to do about that, seems like a total blocker for being able to do publishing dry runs in general 
```

PR #1914, opened 2019-11-25
Strangely, hdk::sign cannot sign any String containing JSON escapes

pjkundert:
## PR summary

We need to be able to sign arbitrary data; presently, the `hdk::sign` API fails on any String containing any symbols that need to be escaped.

The `crates/wasm-test` test named `test_signing` illustrates the failure.

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation


pjkundert:
I unsuccessfully added some `println!`s to various places, in a failed attempt to try to isolate where this was breaking.  See branch `2019-11-25-fix-signing-printing`.

I am wondering if we perhaps should consider (eventually) converting the arguments to hdk::sign from a String to a Vec<u8>?  It seems to make more sense -- we need to be able to sign arbitrary data.  Is String at any time assumed to be UTF-8?  And, if so, are there any restrictions on shipping arbitrary binary data via this serialization?

thedavidmeister:
the issue was that the json rpc logic in the conductor api wasn't using the standard `JsonString` so it wasn't serializing things, it was just string interpolating them, which meant that `\"` breaks stuff (maybe other things broke too).

i threw `DefaultJson` at it and things seem to work, a few hashes changed tho

atm i think that (because everything is handled via json rpc) binary would need to be handled as base64, so yes it can be arbitrary but it needs encoding first

pjkundert:
Sending a Rust String containing arbitrary data should be OK, so long as JSON de/serialization doesn't try to interpret it as UTF-8 or anything like that...  We should test it with some obviously non-UTF-8 Strings.

thedavidmeister:
@pjkundert sure, if you want to stick a test up to see what happens we can try it out :)

pjkundert:
I think that we need to revisit the approach to fixing this...

The problem is, at its core -- that any API that accepts a Rust `String` as its argument *cannot* accept binary data.  It can *only* accept UTF-8 data.  That is all that a Rust `String` can hold.

`hdk::sign` (and probably other APIs) *must* be able to accept binary data; they must, therefore, use a `Vec<u8>` argument.

Now, they can easily handle UTF-8 data: `String::as_bytes()`.  So, to sign a UTF-8 String is almost as easy as it is, now.

crates/conductor_api/src/conductor_api.rs
```diff
@@ -53,8 +54,10 @@ impl ConductorApi {
             CryptoMethod::Encrypt => (String::from("encrypt"), String::from("message")),
             CryptoMethod::Decrypt => (String::from("decrypt"), String::from("message")),
         };
-
-        send_json_rpc(self.0.clone(), payload, request_response)
+        // all crypto payloads are base64 encoded as we need to support arbitrary data and JSON
+        // handling is painful without some kind of encoding
+        let encoded_payload = base64::encode(&payload);
+        send_json_rpc(self.0.clone(), encoded_payload, request_response)
```
pjkundert: This will *not* handle arbitrary binary data; it will only handle the *subset* of possible binary data that also corresponds to a valid UTF-8 `String`.

thedavidmeister: @pjkundert you're correct, my comment is wrong

pjkundert:
Since we need to inform the wormhole counterparty (ie. `holo-envoy`) of the encoding of the data payload: "utf-8" (present data), or "base-64".

For Entry signing, there is no need to encode it as base-64, since its already a UTF-8 String, and can be signed as-is.  No need for additional encoding (or the addition 4/3 increase in bandwidth).

For `hdk::sign` signing, which (eventually) will carry arbitrary data, we need to indicate that we are shipping "base-64" encoded data, signalling the downstream "wormhole" processors that they nee 
d to appropriately decode the data to obtain the raw bytes buffer.

Since we're now re-defining the API between Holochain conductor and the "wormhole" processors, we should add this indicator to the JSON-encoded API call.

thedavidmeister:
i got an approval from @pjkundert in chat so let's do it!

PR #1915, opened 2019-11-26
M:N Thread Pool for Futures

lucksus:
## PR summary

Currently, we execute futures mostly in their own threads.
But we do actually have quite some logical dependencies which leads to threads spawning threads just to wait on the children to finish, which can be seen in @freesig's profiling result:

Here we (finally) use a standard (!= our `context.block_on()`) future executor namely the [ThreadPool](https://docs.rs/futures/0.3.1/futures/executor/struct.ThreadPool.html) from the futures lib that does [M:N task scheduling](https://docs.rs/futures/0.3.1/futures/executor/index.html#using-a-thread-pool-mn-task-scheduling):
> Most of the time tasks should be executed on a thread pool. A small set of worker threads can handle a very large set of spawned tasks (which are much lighter weight than threads). 

## testing/benchmarking notes

## changelog

```markdown
```

## documentation

### DIFF
--- a/crates/core/src/context.rs
+++ b/crates/core/src/context.rs
@@ -48,8 +48,6 @@ use futures::executor::ThreadPool;
 #[cfg(test)]
 use test_utils::mock_signing::mock_conductor_api;
 
-//const NUM_WORKER_THREADS: usize = 20;
-
 pub struct P2pNetworkWrapper(Arc<Mutex<Option<P2pNetwork>>>);
 
 impl P2pNetworkWrapper {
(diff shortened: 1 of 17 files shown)


freesig:
I'll give this a profile now

crates/core/src/context.rs
```diff
@@ -44,11 +44,11 @@ use std::{
     time::Duration,
 };
 
+use futures::executor::ThreadPool;
 #[cfg(test)]
 use test_utils::mock_signing::mock_conductor_api;
-use threadpool::ThreadPool;
 
-const NUM_WORKER_THREADS: usize = 20;
+//const NUM_WORKER_THREADS: usize = 20;
```
zippy: Can we either use or delete this?

lucksus: :+1:

PR #1916, opened 2019-11-26
Timeout refactoring

lucksus:
## PR summary

We had a very naive implementation for query and DM timeouts: just spawning a thread, letting it sleep for the timeout and then dispatch an action.

This PR replaces that with a proper timeout handling where queries and direct messages store the time of creation and a duration in the state and a *scheduled job* polls this info every second and dispatches actions if a timeouts get hit.

## testing/benchmarking notes

## changelog

```markdown
```

## documentation

### DIFF
--- a/crates/core/src/action.rs
+++ b/crates/core/src/action.rs
@@ -144,10 +144,11 @@ pub enum Action {
     /// Note that the given address is that of the entry NOT the address of the header itself
     PublishHeaderEntry(Address),
 
-    ///Performs a Network Query Action based on the key and payload, used for links and Entries
-    Query((QueryKey, QueryPayload)),
+    /// Performs a Network Query Action based on the key and payload, used for links and Entries.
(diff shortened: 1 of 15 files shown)


crates/core/src/action.rs
```diff
@@ -163,7 +164,8 @@ pub enum Action {
 
     /// Makes the network module send a direct (node-to-node) message
     /// to the address given in [DirectMessageData](struct.DirectMessageData.html)
-    SendDirectMessage(DirectMessageData),
+    /// Includes the timeout information: system time of dispatch and duration until it timeouts.
+    SendDirectMessage((DirectMessageData, Option<(SystemTime, Duration)>)),
```
maackle: @lucksus this pattern keeps showing up, maybe we can wrap this up in a struct with some methods rather than having these complex tuples everywhere

Issue #166, opened 2018-07-27
state history should not use unbounded memory


thedavidmeister:
see examples:

- `agent::AgentState::actions`
- `::instance::State::history`
- `nucleus::NucleusState::ribosome_calls`

all hash maps that never drop k/v pairs

these should have limits, like a FIFO or LRU cache that would drop stale values eventually

a quick google presents several options for structures designed for caches implementing both strategies

PR #1920, opened 2019-11-27
State pruning

### LINKED ISSUE #196: standard `ActionResponse` trait
while `Action` is global, `ActionResponse` is specific to the thing responding to the action, e.g. agent/nucleus

however, a standard trait for each `ActionResponse` implementation probably makes sense

lucksus:
## PR summary

We saw this coming a long time ago, but so far there was no immediate reason to do this - until now :)
There were these fields in the state that never got cleaned up. It was no problem for a long time as they stored only small objects. But if an instance keeps running for hours/days, this would still accumulate to high amounts eventually.

1. Get rid of `State::history` which was only really used in tests (and one other place)
2. Have futures dispatch a clear-action after they read the state field they were polling for.
3. Prune `AgentState::actions`, `NetworkState::actions` after 60 seconds to make sure that eventually dropped futures don't leave anything back.

Closes #196 and #166 

## testing/benchmarking notes

## changelog

```markdown
```

## documentation


crates/core/src/instance.rs
```diff
@@ -76,6 +76,11 @@ impl Instance {
         scheduler
             .every(10.seconds())
             .run(scheduled_jobs::create_state_dump_callback(context.clone()));
+        scheduler
+            .every(5.seconds())
```
zippy: Shouldn't there be a direct relationship between ACTION_PRUNE_MS and this number?  I.e. why have it at 3000ms and then only run the task every 5 seconds?

crates/core/src/agent/state.rs
```diff
@@ -159,13 +161,29 @@ impl AddressableContent for AgentStateSnapshot {
 // @TODO abstract this to a standard trait
```
zippy: If this really closes #196 then can remove this comment?

zippy:
Why is build failing?

crates/core/src/instance.rs
lucksus: Yeah, good point. Also I'm not sure if these numbers are ok anyway. I guess it could make nodes lock up if under stress, if the future that is polling the state for an action doesn't get run before the action result is pruned already - the future would keep waiting for ever.

So I guess we should at least increase `ACTION_PRUNE_MS` to something like 30 or 60 seconds..

maackle: Shouldn't we just remove the stale data as soon as we don't need it? At soon as the future gets what it needs, shouldn't it just clear the data from the state at that moment? I think we're bound to run into trouble like you mentioned if we don't separate the concerns of actual application state with the concerns of visibility and diagnostics.

crates/core/src/agent/state.rs
maackle: Yeah I could use some background on this, I never even knew we had ActionResponses. Sounds GhostEndpoint-y

crates/core/src/agent/state.rs
```diff
@@ -34,7 +35,7 @@ pub struct AgentState {
     /// every action and the result of that action
     // @TODO this will blow up memory, implement as some kind of dropping/FIFO with a limit?
     // @see https://github.com/holochain/holochain-rust/issues/166
```
jamesray1: If this PR closes #166 then we should also remove this comment?

crates/core/src/agent/state.rs
```diff
@@ -270,15 +287,39 @@ fn reduce_commit_entry(
...
     });
 
+    agent_state.actions.insert(
+        action_wrapper.clone(),
+        Response::from(AgentActionResponse::Commit(result)),
+    );
+}
+
+fn reduce_prune(agent_state: &mut AgentState, _root_state: &State, action_wrapper: &ActionWrapper) {
+    assert_eq!(action_wrapper.action(), &Action::Prune);
```
jamesray1: Since `action_wrapper` is only used here in this function, would it make sense to pass `prune: &Action::Prune` directly, instead of `action_wrapper: &ActionWrapper`?

crates/core/src/instance.rs
```diff
@@ -77,6 +77,11 @@ impl Instance {
         scheduler
             .every(1.second())
             .run(scheduled_jobs::create_timeout_callback(context.clone()));
+        scheduler
+            .every(5.seconds())
```
jamesray1: Should we use constants here, rather than literal numbers?

crates/core/src/instance.rs
jamesray1: @maackle good point, but can we guarantee that the data in the state won't be needed after a particular future gets the data? For instance, what if there is a future running in a concurrent thread that is also trying to get the data? There might be a delay between when the future gets the data and it is then stored (if applicable), so if the data needs to be accessed between that time, should that scenario be handled, or prevented from occurring in the first place if it can currently hypothetically occur?

lucksus: Ok, I've added actions and reducers for removing responses from the state separately and dispatch according actions from the futures when they resolve. I have left the general prune in too but set the timings to 60 seconds timeout for responses and a 30 second poll interval for the scheduled job, which I would consider to be safe.

@jamesray1, we don't have any futures share actions. We explicitly make actions unique with a `ProcessUniqueId` in the `ActionWrapper`. So cleaning up the response from the future when it resolves seems like the right way to go here.

crates/core/src/agent/state.rs
lucksus: We have `ActionWrapper` as the type that gets processed by the action loop. As stated above, the difference is a `ProcessUniqueId` to make actions unique. We can't (easily) change that for just one reducer.

crates/core/src/instance.rs
lucksus: I started doing this, but I actually think it makes it less readable in this case. Those numbers are only used here and seeing them here gives much more context as they are next to the function that gets scheduled. What we could do (in another PR) is to pull all these constants out and into the config so we can alter them without rebuild. But there wasn't a strong pull for this yet.

crates/core/src/network/reducers/prune.rs
```diff
@@ -0,0 +1,32 @@
...
+    network_state
+        .actions
+        .iter()
+        .filter_map(|(action, response)| {
+            if let Ok(elapsed) = response.created_at.elapsed() {
+                if elapsed > Duration::from_millis(ACTION_PRUNE_MS) {
+                    return Some(action);
+                }
+            }
+            None
```
jamesray1: Should we try to handle/log the error from `elapsed()` here and the other places that it is used?

lucksus: According to [the documentation](https://doc.rust-lang.org/std/time/struct.SystemTime.html), `SystemTime::elapsed()` can fail if the system clock drifts or went backwards (i.e. got (re-)set in between). I don't know how we should handle this case better than just ignoring the drift and in the worst-case waiting a bit longer. Do you have a suggestion how to handle the error?

maackle: We might consider using `Instant` for timeout checks rather than `SystemTime`, which is less accurate, but is guaranteed to be monotonic unlike `SystemTime`. Not a big problem either way.

maackle: BTW I pushed a commit that implements this reducer a little more straightforwardly, please review @lucksus

jamesray1: Yes it also occurred to me to use [`Instant`](https://doc.rust-lang.org/std/time/struct.Instant.html), however, as you say it is less accurate, so I didn't suggest it. Alternative solutions could get complicated, and doesn't seem worth thinking about further, since the elapsed time should eventually exceed the duration for pruning.

PR #1921, opened 2019-11-27
All performance

lucksus:
## PR summary

This is the combination of #1915, #1916, #1917, #1920 (and thus #1923). Each of these focuses on a particular performance issue which is easier to review separate from the rest, but all together are intended to mitigate the problems @freesig found in his profiling runs.

## testing/benchmarking notes

( if any manual testing or benchmarking was/should be done, add notes and/or screenshots here )

## followups

( any new tickets/concerns that were discovered or created during this work but aren't in scope for review here )

## changelog

- [ ] if this is a code change that effects some consumer (e.g. zome developers) of holochain core,  then it has been added to [our between-release changelog](https://github.com/holochain/holochain-rust/blob/develop/CHANGELOG-UNRELEASED.md) with the format 

```markdown
- summary of change [PR#1234](https://github.com/holochain/holochain-rust/pull/1234)
```

## documentation

- [ ] this code has been documented according to our [docs checklist](https://hackmd.io/@freesig/Hk9AmKJNS)

PR #1923, opened 2019-11-27
Use immutable data-structures to make State cloning much cheaper

lucksus:
## PR summary

The [im crate](https://docs.rs/im/14.0.0/im/) provides drop-in replacements for `HashMap` and `HashSet`. Using those inside our core State makes cloning it much cheaper (which happens a lot).

This was a lot easier than expected :)

## testing/benchmarking notes

## changelog

```markdown
```

## documentation

### DIFF
--- a/crates/core/src/dht/aspect_map.rs
+++ b/crates/core/src/dht/aspect_map.rs
@@ -1,8 +1,8 @@
 use crate::holochain_wasm_utils::holochain_persistence_api::cas::content::AddressableContent;
 use holochain_core_types::network::entry_aspect::EntryAspect;
 use lib3h_protocol::types::{AspectHash, EntryHash};
-use std::collections::{HashMap, HashSet};
-
+use std::collections::HashMap as StdHashMap;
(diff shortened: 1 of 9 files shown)


crates/core/src/dht/aspect_map.rs
```diff
@@ -81,7 +76,7 @@ impl AspectMap {
             .join("\n")
     }
 
-    pub fn merge(map1: &AspectMap, map2: &AspectMap) -> AspectMap {
+    pub fn merge(map1: AspectMap, map2: AspectMap) -> AspectMap {
```
maackle:
```suggestion
    pub fn merge(map1: &AspectMap, map2: &AspectMap) -> AspectMap {
```

Is there a reason to not leave these as references? It turns out that in our current codebase we always have freshly created maps to pass to this function, but that alone isn't enough reason to make this require owned values.

maackle: In conversation I found that the `im` methods require owned values

PR #1924, opened 2019-11-28
Add smarter pending validation resolution

willemolding:
## PR summary

Adds a sorting stage to the process of selecting the next pending validation to reduce the probability of blocking indefinitely.

This sort must take into account the dependency graph of entries (if entry A depends on entry B we should always try and validate B first). To build the dependency graph this adds a new trait `ValidationDependencies` which is implemented by `EntryWithHeader`. Links depend on their target and base, header entries depend on the previous header, entries also depend on the previous header.

The dependencies are stored with the pending validations and only leaf nodes/sinks are considered when selecting the next entry to validate. It also keeps the delay logic from before so it won't thrash trying to validate something impossible.

## Remaining tasks

- [x] Write tests for `get_free_dependencies` and `next_queued_holding_workflow` 
- [x] There are still some serious optimizations to do. The `get_free_dependencies` function makes no attempt at all to be efficient. Cloning everywhere
- [x] Refactor with types rather than tuples `(PendingValidation, Option<(SystemTime, Duration)>)` 
- [x] It is still possible to state something to be validated even though we can tell before it will fail (because there are sleeping un-validated dependencies). Fixed by selecting only leaves

## changelog

```markdown
```

## documentation

### DIFF
--- a/crates/core/src/dht/dht_store.rs
+++ b/crates/core/src/dht/dht_store.rs
@@ -449,13 +449,4 @@ pub mod tests {
             vec![b, c]
         );
     }
-
-    #[test]
-    #[should_panic(expected = "Cyclic validation dependencies detected!!")]
-    fn test_dependency_resolution_cycle() {
-        // A depends on B, B depends on A. Uh Oh we have a cycle!
-        let a = pending_validation_for_entry(test_entry_a(), vec![test_entry_b().address()]);
-        let b = pending_validation_for_entry(test_entry_b(), vec![test_entry_a().address()]);
-        get_free_dependencies(&vec![a.clone(), b.clone()]);
-    }
 }
--- a/crates/core/src/dht/dht_store.rs
+++ b/crates/core/src/dht/dht_store.rs
@@ -265,12 +265,16 @@ impl DhtStore {
 
     pub(crate) fn next_queued_holding_workflow(
         &self,
-    ) -> Option<(&PendingValidation, Option<Duration>)> {
-        // WTODO: This is where we should toposort to find the next thing to validate
+    ) -> Option<(PendingValidation, Option<Duration>)> {
+        // calculate the leaf dependencies (the things we can validate right now)
+        let sorted_holding_workflows =
+            get_sorted_dependencies(self.queued_holding_workflows.clone());
 
-        self.queued_holding_workflows
-            .iter()
+        // respect the delays on the leaf nodes
+        sorted_holding_workflows
+            .into_iter()
             .skip_while(|(_pending, maybe_delay)| {
+                // skip pending validation with an unelapsed delay
                 if let Some((time_of_dispatch, delay)) = maybe_delay {
                     let maybe_time_elapsed = time_of_dispatch.elapsed();
                     if let Ok(time_elapsed) = maybe_time_elapsed {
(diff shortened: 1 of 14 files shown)


crates/core/src/entry/validation_dependencies.rs
```diff
@@ -0,0 +1,114 @@
...
+
+pub trait ValidationDependencies {
+    fn get_validation_dependencies(&self) -> Vec<Address>;
+}
+
+impl ValidationDependencies for EntryWithHeader {
+    fn get_validation_dependencies(&self) -> Vec<Address> {
+        match &self.entry {
+            Entry::App(_, _) => {
+                // an app entry is dependent on its previous header being validated
```
lucksus: Why? I don‘t see why this would be true in the general case..

willemolding: Right, so without doing a call into the zome to retrieve the ValidationPackageDefinition it is impossible to know if it needs just the entry or the full chain. This assumes it is always full chain to support all cases but possibly at the expense of validating entries in a different, less efficient order

crates/core/src/entry/validation_dependencies.rs
```diff
@@ -0,0 +1,115 @@
...
+    fn get_validation_dependencies(&self) -> Vec<Address>;
+}
+
+impl ValidationDependencies for EntryWithHeader {
+    fn get_validation_dependencies(&self) -> Vec<Address> {
+        match &self.entry {
+            Entry::App(_, _) => {
+                // in the future an entry should be dependent on its header but 
+                // for now it can require nothing
+                Vec::new()
```
lucksus: If the entry is an update, which we can tell from the header (`self.header.crud_link.is_some()`) that should be the dependency.

willemolding: Good catch, thanks. It is kind of nice at the moment that these are only used for improving the ordering. If we don't include the correct dependencies it just falls back to how it worked originally

lucksus:
I wonder if we should also hook-up the validation dependencies that are returned by `validate_entry()` as `ValidationError::UnresolvedDependencies(dependencies)`. They used to get put into the `PendingValidationStruct` but now it only gets logged. See: https://github.com/holochain/holochain-rust/blob/5d5b05234c7a204743a20b724eb003b1248f2882/crates/core/src/workflows/hold_entry.rs#L52

willemolding:
> I wonder if we should also hook-up the validation dependencies that are returned by `validate_entry()` as `ValidationError::UnresolvedDependencies(dependencies)`. They used to get put into the `PendingValidationStruct` but now it only gets logged. See:
> 
> https://github.com/holochain/holochain-rust/blob/5d5b05234c7a204743a20b724eb003b1248f2882/crates/core/src/workflows/hold_entry.rs#L52

I don't think we need that right? Provided we have successfully captured everything in `validation_dependencies` which I think we have. This will just return the same things right? Or is there something I am missing?

crates/core/src/dht/dht_store.rs
```diff
@@ -279,30 +290,69 @@ impl DhtStore {
...
+            // only add the dependencies that are also in the pending validation list
+            if let Some(to) = index_map.get(&to_addr) {
+                graph.add_edge(*from, *to, ());
+            }
+        }
+    }
+
+    if is_cyclic_directed(&graph) {
+        // this might be expensive..
+        panic!("Cyclic validation dependencies detected!!")
```
willemolding: I'm really not sure what we should do here. If there are cyclic validation dependencies the DNA is cooked. I don't think there is any way to recover from this.

zippy: This isn't about the DNA it's about the data.  We can't panic because that means if someone malicious finds a way to send people fake headers and entries that create a cycle it kills their conductor.   It think what we have to do here is declare those entries invalid and reject them.  This then become part of validation.   I'm ok with adding this as follow-up, but it's pretty high priority to get right before it fades in our memory.

crates/core/src/dht/pending_validations.rs
```diff
@@ -157,3 +166,36 @@ impl From<PendingValidationStruct> for EntryAspect {
...
     }
 }
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub struct ValidationTimeout {
+    pub time_of_dispatch: SystemTime,
+    pub delay: Duration,
+}
+
+impl ValidationTimeout {
```
zippy: I am seeing a couple other places where we are using an `Option<SystemTime, Duration>` as a timeout:
- https://github.com/holochain/holochain-rust/blob/f725e8f741e12ee966dc2d08cf0a3f0c97a8d3b7/crates/core/src/action.rs#L120
- https://github.com/holochain/holochain-rust/blob/f725e8f741e12ee966dc2d08cf0a3f0c97a8d3b7/crates/core/src/action.rs#L153
- https://github.com/holochain/holochain-rust/blob/f725e8f741e12ee966dc2d08cf0a3f0c97a8d3b7/crates/core/src/action.rs#L175
- https://github.com/holochain/holochain-rust/blob/f725e8f741e12ee966dc2d08cf0a3f0c97a8d3b7/crates/core/src/network/state.rs#L37
- https://github.com/holochain/holochain-rust/blob/f725e8f741e12ee966dc2d08cf0a3f0c97a8d3b7/crates/core/src/network/state.rs#L46

Would it make sense to have this refactor be a bit more general and work for all those cases?
.  Instead of creating this special case `ValidationTimeout`

crates/core/src/dht/dht_store.rs
lucksus: Agree with @zippy. I would actually want to have this panic be an error already before merging.

lucksus: Hm, ok, I see that it is not easy to deal with this here. So I, too, don't want to block merging this, but we need to handle this somehow. Panicking will might kill the whole conductor, also other DNAs/hApps that would work. Maybe we can detect the cycle and at least continue validating the other entries and if there are no other entries just trying to validate any of the entries in the cycle ?

crates/core/src/dht/pending_validations.rs
maackle: This struct could even become a more general Timeout struct. We already have a Timeout struct which is just a newtype wrapper around a Duration so this would be something else for internal use. Then we could use this struct in all those other places, which could DRY up a lot of repeated logic around checking if the timeout has occurred.

crates/core/src/entry/validation_dependencies.rs
```diff
@@ -0,0 +1,115 @@
...
+
+pub trait ValidationDependencies {
+    fn get_validation_dependencies(&self) -> Vec<Address>;
+}
+
+impl ValidationDependencies for EntryWithHeader {
+    fn get_validation_dependencies(&self) -> Vec<Address> {
+        match &self.entry {
+            Entry::App(_, _) => {
+                // in the future an entry should be dependent on its header but
```
maackle: Just a note, in the future it could also include a user-defined pure function of the entry which returns extra dependencies to require for validation

crates/core/src/dht/pending_validations.rs
willemolding: > This struct could even become a more general Timeout struct. We already have a Timeout struct which is just a newtype wrapper around a Duration so this would be something else for internal use. Then we could use this struct in all those other places, which could DRY up a lot of repeated logic around checking if the timeout has occurred.

I'm going to save this for another PR I think

Issue #395, opened 2018-09-29
Add advanced Ribosome debug info


ddd-mtl:
Like a stacktrace and backtrace

Connoropolous:
@ddd-mtl seems you've tackled this one

maackle:
Connor says this was tackled a year ago, and I'm not sure what was in mind anyway, so, closing.

PR #1925, opened 2019-11-28
Tracing pt. 1

maackle:
## PR summary

Note to self: see [tracing-1-michael](https://github.com/holochain/holochain-rust/tree/tracing-1-michael) for local crate paths for local dev on this

Adding basic tracing to core, using newly added updates to holochain-tracing.

- Sprinkles `#[autotrace]` macro in many places to build up full thread-local stack traces
- Liberally uses SpanWrap and EncodedSpanWrap to send Spans with data across thread and process boundaries, respectively.
    - action channel now uses SpanWrap
    - WireMessage lib3h communication uses EncodedSpanWrap
- Adds a Tracer and a long-running Reporting thread to both `holochain` and `cli` crates, which reports spans to a locally running Jaeger instance.
    - The conductor reporting thread dynamically creates a Reporter for each instance, and stores them all in a big RwLock<HashMap<_,_>>, similar to the signal receivers

### How to run tests locally:

Before running a test, run a Jaeger server (put this in a script)

```
docker run -d --rm --name jaeger \
  -p 5775:5775/udp \
  -p 6831:6831/udp \
  -p 6832:6832/udp \
  -p 5778:5778 \
  -p 16686:16686 \
  -p 14267:14267 \
  -p 14268:14268 \
  -p 9411:9411 \
  jaegertracing/all-in-one:1.16
```

- Make sure the test you want includes a "tracing" config in tryorama. See [app_spec/test/config.js](https://github.com/holochain/holochain-rust/pull/1925/files#diff-cf440d1063f2ba6dcb7fdb98f6157756R69) for an example
- Use the `--tracing-name` flag for sim2h_server, e.g. `hc-sim2h-server --tracing-name sim2h`
- Run the test, and go to http://localhost:16686 to see the Jaeger UI and search for traces

### Final exam / TryCP integration notes:

Run all-in-one on final exam:

```
docker run -d --rm --name jaeger \
  -p 5775:5775/udp \
  -p 6831:6831/udp \
  -p 6832:6832/udp \
  -p 5778:5778 \
  -p 16686:16686 \
  -p 14267:14267 \
  -p 14268:14268 \
  -p 9411:9411 \
  jaegertracing/all-in-one:1.16
```

Run jaeger-agent on each trycp machine

```
HOST=final-exam.holo.host

docker run \
  --rm \
  -p5775:5775/udp \
  -p6831:6831/udp \
  -p6832:6832/udp \
  -p5778:5778/tcp \
  jaegertracing/jaeger-agent:1.16 \
  --reporter.tchannel.host-port "$HOST:14267"
```

## followups

* Configurable sampling strategy
* More fine-tuned traces -- current ones are noisy and incomplete, in particular the discontinuity in queued_pending_validations, which is documented with TODOs in the code
* More use of tags
* Send traces over the network!

## changelog

```markdown
```

## documentation


jamesray1:
Good to see this! This may possibly help to close or actually close #395?

maackle:
@jamesray1 maybe so, but that issue was more like a note-to-self than a unit of work, so I closed it anyway

crates/core/src/dht/dht_reducers.rs
```diff
@@ -158,15 +158,18 @@ pub fn reduce_queue_holding_workflow(
 ) -> Option<DhtStore> {
     let action = action_wrapper.action();
     let (pending, maybe_delay) = unwrap_to!(action => Action::QueueHoldingWorkflow);
+
+    // TODO: TRACING: this is where we would include a Span, so that we can resume
+    // the trace when the workflow gets popped (see instance.rs), but we can't do that
+    // until we stop cloning the State, because Spans are not Cloneable.
```
lucksus: I'm not sure if we ever will stop cloning the state. As a matter of fact, we recently added more State cloning (instead of tossing around lock guards which lead to deadlocks) and have made that cloning cheaper since the State is comprised almost only of `Arc`s and `im` collections..

Could spans live in `Arc`s inside the state?

crates/core/src/instance.rs
```diff
@@ -190,23 +192,31 @@ impl Instance {
...
                 ProcessUniqueId::new().to_string()
             ))
             .spawn(move || {
                 let mut state_observers: Vec<Observer> = Vec::new();
-                let mut unprocessed_action: Option<ActionWrapper> = None;
+                let mut unprocessed_action: Option<ht::SpanWrap<ActionWrapper>> = None;
                 while kill_receiver.try_recv().is_err() {
-                    if let Some(action_wrapper) = unprocessed_action.clone().or_else(|| rx_action.recv_timeout(Duration::from_secs(1)).ok()) {
+                    // TODO: I changed a clone() to a take() because I couldn't see how this would work otherwise. Was it right??
+                    if let Some(action_wrapper) = unprocessed_action.take().or_else(|| rx_action.recv_timeout(Duration::from_secs(1)).ok()) {
```
lucksus: Hm, I guess that is right.

maackle:
Currently waiting on the ability to publish a new release of holochain_tracing (and a brand new one for holochain_tracing_macros) to crates.io, so that holochain-rust and lib3h crates can be properly updated to use the new tracing versions

crates/core/src/dht/dht_reducers.rs
```diff
@@ -186,21 +190,7 @@ pub fn reduce_remove_queued_holding_workflow(
...
-            // The holding thread will still try to pop that next item, so we need
-            // this else case where we just remove an item from some position inside the queue:
-            new_store
-                .queued_holding_workflows
-                .retain(|PendingValidationWithTimeout { pending: item, .. }| item != pending);
-        }
-    } else {
+    if let None = new_store.remove_holding_workflow(pending) {
         error!("Got Action::PopNextHoldingWorkflow on an empty holding queue!");
     }
```
maackle: I replaced this big chunk of code with a simpler method on DhtStore. It seemed unnecessary to have two separate cases for whether the item is at the front of the queue or not, either way we just want to remove it. Can someone make sure I did this replacement right though? @lucksus?

lucksus: Looks good!

maackle:
I think the problem is that suddenly some dependencies of rustracing_jaeger are suddenly not compilable to wasm. That's not a problem, since we don't actually need tracing in WASM, but we have to do some conditional compilation to make sure it doesn't attempt to build rustracing for wasm target. This remains on ice.

freesig:
Does holochain_tracing_macros need to be bumped? This is failing because it can't find the new relic stuff in the macros crate

crates/conductor_lib/src/conductor/base.rs
```diff
@@ -437,6 +450,22 @@ impl Conductor {
             .expect("Must be able to spawn thread")
     }
 
+    pub fn spawn_trace_reporter_thread(&mut self) -> thread::JoinHandle<()> {
+        let reporters = self.trace_reporters.clone();
+        thread::Builder::new()
+            .name("signal_multiplexer".to_string())
```
neonphog: do we want a whole thread here? could be a lighter weight futures task - then use a futures channel allowing a more efficient `.await` rather than thread::sleep - just a thought.

if this stays a thread maybe mention jaeger or tracing in the thread name.

maackle:
keep the feedback coming, we can deal with it tomorrow :+1:

freesig:
I think we will want to put this under a feature flag or something. I don't think we want it when running performance tests.
Also the tracing isn't working for me locally right now so I must have messed something up.

maackle:
My theory is that when we use a NullTracer, the performance hit will be negligible. By default, that's actually what gets used -- actual trace sampling only occurs when you set the `[tracing]` section of the conductor config, or use `sim2h_server -- tracing-name foo`

crates/common/src/lib.rs
```diff
@@ -1,7 +1,18 @@
+#[macro_use]
```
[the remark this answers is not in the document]
maackle: Ah I didn't things changed in 2018 for #[macro_use] too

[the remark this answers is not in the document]
freesig: Yep you don't need macro use anymore, Instead you just import macros like any other type

maackle:
I just realized one shortcoming of this branch: Reporters for each Instance are only created when instantiating from Config, but not when adding Instances from the admin interface. This means e.g. Holoscape wouldn't be able to do tracing until you restart your conductor (though Holoscape doesn't have a way to specify that tracing should be added to the config yet anyway).

crates/sim2h/src/lib.rs
```diff
@@ -411,15 +416,18 @@ pub struct Sim2h {
     sim2h_handle: Sim2hHandle,
     connection_count: usize,
     metric_gen: MetricsTimerGenerator,
+    tracer: ht::Tracer,
```
freesig: This not being in an option means there is no way to tell if tracing is enabled. This means if there's anything that does have a cost that is done in tracing it will happen all the time

maackle: feel free to make it Option as we discussed, to avoid unwrapping the opaque message earlier than we need to for times when we're not even doing tracing

app_spec/test/files/test.js
```diff
@@ -7,7 +7,7 @@ module.exports = scenario => {
     t.notEqual(alice.info('app').agentAddress, bob.info('app').agentAddress)
   })
 
-  scenario('send ping', async (s, t) => {
+  scenario.only('send ping', async (s, t) => {
```
lucksus: This needs to be removed before merging:
```suggestion
  scenario('send ping', async (s, t) => {
```

crates/conductor_lib/src/conductor/base.rs
```diff
@@ -557,6 +586,8 @@ impl Conductor {
     /// Starts all instances
     pub fn start_all_instances(&mut self) -> Result<(), HolochainInstanceError> {
         notify("Start all instances".to_string());
+        // TODO: how to stop?
```
lucksus: Killswitch channel?

crates/core/src/instance.rs
```diff
@@ -297,6 +316,11 @@ impl Instance {
                 while kill_receiver.try_recv().is_err() {
                     log_trace!(context, "Checking holding queue...");
                     loop {
+                        // TODO: TRACING: it would be ideal to be able to associate a tracing Span with each queued holding workflow.
+                        // To do this, we'd need to store a Span in each item of the DhtStore::queued_holding_workflows.
+                        // However, Span is not Clone, and the entire DhtStore needs to be Cloned.
+                        // So, the span has to be cut short here until we stop cloning the state.
+
```
lucksus: Again, what about putting `Arc` wrapped spans into the state? We are doing this with the state slices already to not clone the whole state every time - just the root reference.

crates/net/src/p2p_network.rs
```diff
@@ -59,56 +60,60 @@ impl P2pNetwork {
...
             }),
             // Create an Sim1hWorker
-            P2pBackendKind::SIM1H => Box::new(move |h| {
-                let backend_config = match &p2p_config.clone().backend_config {
-                    BackendConfig::Sim1h(config) => config.clone(),
-                    _ => return Err(format_err!("mismatch backend type, expecting sim1h")),
-                };
-                Ok(Box::new(Sim1hWorker::new(h, backend_config)?) as Box<dyn NetWorker>)
+            P2pBackendKind::SIM1H => Box::new(move |_h| {
+                unimplemented!()
```
lucksus: I know we removed the sim1h test suite. Are we (did we?) officially and fully deprecate sim1h?

maackle: Let's discuss that today, and what that should look like

crates/core/src/instance.rs
maackle: @lucksus The problem is that we rely on the Span being dropped to mark the end of the span and to actually send it for reporting. If we allow it inside an `Arc`, we lose control over when it goes out of scope if we don't keep track of every clone. If all clones don't drop at the same time, we get an inaccurate span. We could solve this with explicit coordination, but that's just more complexity. Unless you see a simple way that I'm not seeing?

crates/net/src/p2p_network.rs
maackle: Oops, forgot to bring it up today. Tomorrow!

crates/core/src/network/reducers/init.rs
```diff
@@ -57,6 +58,8 @@ pub fn reduce_init(state: &mut NetworkState, root_state: &State, action_wrapper:
         p2p_config,
         Some(Address::from(network_settings.agent_id.clone())),
         Some(root_state.conductor_api.clone()),
+        // TODO: This probably shouldn't be none but I can't figure out how to get the tracer
+        None,
```
freesig: I couldn't figure out how to get a tracer here @maackle It's needed to pass to the Sim2hWorker

maackle: Well I thought we could get it in through the ActionWrapper. But then we would have to make Tracer `Serialize` :see_no_evil:

crates/core/src/instance.rs
lucksus: I see. Well, that explicit coordination could just be an `Arc<Option<..>>`, right?
(Or `Arc<Mutex<Option<..>>>`)

PR #1931, opened 2019-12-01
changes to support nightly 2019-11-25

thedavidmeister:
## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation


config.nix
```diff
@@ -15,7 +15,7 @@ rec {
 
   # true = use a github repository as the holonix base (recommended)
   # false = use a local copy of holonix (useful for debugging)
-  use-github = true;
+  use-github = false;
```
zippy: Are you sure you want to be changing this flag?  Otherwise looks good.

thedavidmeister: @zippy nope, this was only temporary while i was cross testing things

PR #1932, opened 2019-12-02
Add ability to validate entries with full chain validation when author is offline

willemolding:
## PR summary

Finally adds ability to validate ChainFull entries when the author is offline (with tests to prove it). This is essential for correct operation of HoloFuel.

- Adds timeouts for GetValidationPackage actions. Without this it is unable to fallback to restoring the authors chain from the headers
- Adds a three staged approach in `validation_package`. First tries to build locally, then tries to message the author to get their chain directly. Finally rebuilds the authors chain from the DHT and validates against that. 
- Adds that entries are dependent on the previous header entry already being valid. Technically this is only required in the ChainFull validation case but it doesn't make the distinction at this time

## changelog

```markdown
```

## documentation


crates/core/src/nucleus/validation/build_from_dht.rs
```diff
@@ -0,0 +1,251 @@
...
+                },
+            ..
+        })) = get_entry_result
+        {
+            headers.push(chain_header.clone());
+            current_header = chain_header;
+        } else {
+            log_debug!(context, "When building validation package from DHT, Could not retrieve a header entry at address: {:?}. Got {:?}", next_header_addr, get_entry_result);
+            return Err(HolochainError::ErrorGeneric(
+                format!("When building validation package from DHT, Could not retrieve a header entry at address: {:?}", next_header_addr))
```
zippy: Can we add the Error reason in here too, because this is what may be reported back out of zome calls and pouring through the DEBUG logs would be difficult.

crates/core/src/nucleus/validation/build_from_dht.rs
```diff
@@ -0,0 +1,251 @@
...
+
+        if let Some(EntryWithMetaAndHeader {
+            entry_with_meta: EntryWithMeta { entry, .. },
+            ..
+        }) = get_entry_result
+        {
+            entries.push(entry.clone());
+        } else {
+            return Err(HolochainError::ErrorGeneric(
+                format!("When building validation package from DHT, Could not retrieve entry at address: {:?}", header.entry_address()))
```
zippy: ditto, get actual error into the error message

crates/core/src/workflows/application.rs
```diff
@@ -66,8 +68,24 @@ pub async fn initialize(
...
+    //     and publish the headers of the initial chain entries
     if first_initialization {
         call_init(dna, &instance_context).await?;
+
+        log_debug!(
+            context,
+            "dna/initialize: publishing headers of initial chain entries...",
+        );
+
+        for header in context.state().unwrap().agent().iter_chain() {
```
zippy: @maackle, can you confirm that this change here doesn't require us updating the consistency model?

crates/core/src/workflows/mod.rs
```diff
@@ -89,21 +78,160 @@ async fn try_make_local_validation_package(
...
+    }
+
+    // 2. Try and get it from the author
+    log_debug!(
+        context,
+        "validation_package:{} - Could not build locally. Trying to retrieve from author",
+        entry_with_header.entry.address()
+    );
+
+    // this will block until the network goes down! It needs to timeout
```
zippy: Is this comment true?  If so then this is a real problem.

app_spec/test/files/offline-validation.js
```diff
@@ -0,0 +1,107 @@
...
+        // be regenerated from the published headers (which bob should hold)
+        t.comment('waiting for Carol to come online')
+        await carol.spawn()
+        t.comment('Carol is online')
+
+        t.comment('Waiting for Carol to get all data via gossip')
+        await s.consistency()
+        t.comment('consistency has been reached')
+
+        await delay(50000) // keep this until consistency works with dynamic starting agents
```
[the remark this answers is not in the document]
willemolding: The reason it has to be quite long is because it waits a full 10 seconds for the author to reply with their chain before falling back to rebuilding it. This delay just needs to be large relative to that. From my experiments locally we can't make it much shorter without introducing flakyness

crates/core/src/workflows/mod.rs
willemolding: Ah nope this was the issue I was having that is fixed now

app_spec/test/files/offline-validation.js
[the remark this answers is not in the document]
maackle: As soon as it becomes high enough priority, we can update hachiko's NetworkModel to account for dynamically changing network membership. The change can probably be done in half a day. Maybe it's time?

app_spec/zomes/blog/code/src/post.rs
```diff
@@ -97,7 +97,7 @@ pub fn definition() -> ValidatingEntryType {
                 "%agent_id",
                 link_type: "recommended_posts",
                 validation_package: || {
-                    hdk::ValidationPackageDefinition::ChainFull
+                    hdk::ValidationPackageDefinition::Entry
```
maackle: Just curious, did you change these because there was no reason to have them set to ChainFull, or does this actually have an impact?

crates/core/src/network/reducers/get_validation_package.rs
```diff
@@ -34,5 +39,35 @@ pub fn reduce_get_validation_package(
...
+) {
+    let action = action_wrapper.action();
+    let address = unwrap_to!(action => crate::action::Action::GetValidationPackageTimeout);
+
+    network_state
+        .get_validation_package_timeouts
+        .remove(address);
+
+    if let Some(Some(_)) = network_state.get_validation_package_results.get(address) {
+        // A result alreay came back from the network so don't overwrite it
```
maackle:
```suggestion
        // A result already came back from the network so don't overwrite it
```

crates/core/src/nucleus/validation/build_from_dht.rs
```diff
@@ -0,0 +1,251 @@
...
+    let mut current_header = header.clone();
+    let mut headers = Vec::new();
+
+    while let Some(next_header_addr) = current_header.link() {
+        log_debug!(context, "About to try and get header: {}", next_header_addr);
+
+        let timeout = Timeout::new(GET_TIMEOUT_MS);
+        // when trying to run the line below there is a panic because the state is dropped
+        // this does not happen when running the tests below. Only when triggered as part of a
+        // validate entry workflow
```
maackle: is this still the case?

crates/core/src/workflows/application.rs
maackle: As long as everyone holds these header entries like any other entry (via the `Hold` action reducer), there should be no change needed. In fact I think we could simplify the model by taking out the InitNetwork event which we added to handle the special case of the initial entries (not that we actually should do that)

crates/core/src/workflows/mod.rs
```diff
@@ -89,21 +78,160 @@ async fn try_make_local_validation_package(
...
     entry_with_header: &EntryWithHeader,
     context: Arc<Context>,
 ) -> Result<Option<ValidationPackage>, HolochainError> {
-    // 1. Try to construct it locally:
-    if let Ok(package) =
-        try_make_local_validation_package(&entry_with_header, context.clone()).await
+    // 0. Call into the DNA to get the validation package definition for this entry
+    // e.g. what data is needed to validate it (chain, entry, headers, etc)
+    let entry = &entry_with_header.entry;
+    let validation_package_definition = get_validation_package_definition(entry, context.clone())
```
maackle: Not related to this PR, but why on earth does `get_validation_package_definition` return a `CallbackResult`?

crates/core/src/workflows/mod.rs
maackle: The comment is still there FYI

app_spec/zomes/blog/code/src/post.rs
willemolding: Basically because there is no reason to. I think it makes the tests easier to understand as well

crates/core/src/network/reducers/mod.rs
```diff
@@ -69,6 +71,7 @@ fn resolve_reducer(action_wrapper: &ActionWrapper) -> Option<NetworkReduceFn> {
         Action::Query(_) => Some(reduce_query),
         Action::QueryTimeout(_) => Some(reduce_query_timeout),
         Action::GetValidationPackage(_) => Some(reduce_get_validation_package),
+        Action::GetValidationPackageTimeout(_) => Some(reduce_get_validation_package_timeout),
```
AshantiMutinta: There is a question of Is there anyway we can combine QueryTimeout and GetValidationPackageTimeout? I would imagine the code is similar? I imagine just Action::Timeout and maybe with QueryTimeout and :GetValidationPackageTimeout as variants?

crates/core/src/nucleus/validation/build_from_dht.rs
```diff
@@ -0,0 +1,248 @@
...
+
+async fn all_chain_headers_before_header_dht(
+    context: Arc<Context>,
+    header: &ChainHeader,
+) -> Result<Vec<ChainHeader>, HolochainError> {
+    let mut current_header = header.clone();
+    let mut headers = Vec::new();
+
+    while let Some(next_header_addr) = current_header.link() {
+        log_debug!(context, "About to try and get header: {}", next_header_addr);
```
AshantiMutinta: If ChainHeader had a method to follow the link by link chain we could be able to to fold this without changing current_header state?

current_header.follow_links().fold(Ok(Vec::new()),|headers,chain_header|{})

crates/core/src/workflows/application.rs
lucksus: I don't think we can rely on this explicit publishing of anything during initialization. We rely on the authoring and gossip lists being correct. If tests fail without this explicit publish it means that the lists are not correct (missing these addresses). Making the tests work with this is just faking the tests and not fixing the problem for the case where we have complex network topologies and nodes coming and going, which we are not completely testing right now. So I think I will change my approval to requesting this to be removed and fixed by looking to why the authoring/gossip lists are not enough..

zippy: @lucksus also had a concern about publishing here only fixing the problem in one case and masking validation problems.

willemolding:
> Requesting to remove the explicit publishing of headers during initialization and making it work without this.

@lucksus if you check out the author list this actually will gossip all headers (including initial chain ones) without this initialization publish stage. Putting them here is a way to trick the consistency model into waiting for them.

I agree we can probably come up with a better solution

lucksus:
I wanted to try out removing that explicit publish to see what happens, but I can't get the offline-validation test to pass on my local machine, even without any change :thinking:

zippy:
Agree with Lucksus. Ok to merge

maackle:
So are we saying that the explicit publish should have no effect, and that it's only there for the sake of the consistency model?

lucksus:
I have reran CI now several times. There is one failing app-spec assertion with sim1h which fails consistently. It is neither the CI timeout-flakyness nor the port issue. I think we have to investigate this. Could be the header publishing. This is the failure log:
```
  not ok 96 should be equivalent
⨯ g...    ---
      operator: deepEqual
      expected: |-
        [ 'HcSCJ965aADATg7ksqHTea4R5crA9evmv6k8PPaD6t5zfhzj6f87tb8ZfHAdmbz', 'HcSCJ965aADATg7ksqHTea4R5crA9evmv6k8PPaD6t5zfhzj6f87tb8ZfHAdmbz', 'HcSCj545ecJRQDnbstxD4tYrMRByo8d5xeKb8Z6o94Pn9TconEWgG94iXUvf9bz', 'HcScIeB7Qf6ik9pp33iUc6DM5CzyhnppfMU4Q4zejpm637ivOi43H9kYmSSgftz' ]
      actual: |-
        [ 'HcSCJ965aADATg7ksqHTea4R5crA9evmv6k8PPaD6t5zfhzj6f87tb8ZfHAdmbz', 'HcSCJ965aADATg7ksqHTea4R5crA9evmv6k8PPaD6t5zfhzj6f87tb8ZfHAdmbz', 'HcSCj545ecJRQDnbstxD4tYrMRByo8d5xeKb8Z6o94Pn9TconEWgG94iXUvf9bz' ]
      at: <anonymous> (/root/project/app_spec/test/files/entry.js:235:7)
      stack: |-
        Error: should be equivalent
            at Test.assert [as _assert] (/root/project/app_spec/test/node_modules/tape/lib/test.js:225:54)
            at Test.bound [as _assert] (/root/project/app_spec/test/node_modules/tape/lib/test.js:77:32)
            at Test.tapeDeepEqual (/root/project/app_spec/test/node_modules/tape/lib/test.js:422:10)
            at Test.bound [as deepEqual] (/root/project/app_spec/test/node_modules/tape/lib/test.js:77:32)
            at /root/project/app_spec/test/files/entry.js:235:7
            at runMicrotasks (<anonymous>)
            at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

maackle:
Hmm, I ran it several times locally and it passed. I wonder why a change to core would cause sim1h to fail and not sim2h. I'll try restarting CI one more time.

maackle:
I got it to pass twice in CI, but most recently it failed again. So it is not a unilateral failure, but more flakiness. I suppose an increase in flakyness is a problem. But could this just be some artifact of sim1h? Rather than a problem with core?

maackle:
passthrough-dna's tests are also failing with sim1h. Seems like something is up with sim1h.

app_spec/test/files/offline-validation.js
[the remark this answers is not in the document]
thedavidmeister: does https://github.com/holochain/holochain-rust/pull/1957 allow us to remove this giant delay?

app_spec/zomes/blog/code/src/memo.rs
```diff
@@ -47,7 +47,7 @@ pub fn definition() -> ValidatingEntryType {
         sharing: Sharing::Private,
 
         validation_package: || {
-            hdk::ValidationPackageDefinition::ChainFull
```
thedavidmeister: @willemolding i'm kind of skim reading but don't these need to be `ChainFull` if we're testing full chain validation?

willemolding: These are private entries anyway (see line 47) so these are not actually being used to test the validation. These never needed to be ChainFull to begin with so I changed them to Entry to simplify debugging.

It is the `blog` entry types that are used for testing offline validation and these have the `ChainFull` validation package requirement.

willemolding:
I have removed the header publishing on `init` and everything is still working on the author list alone!

On the flaky sim1h test that test isn't one that I would have expected to be touched by this PR. I guess everything is so connected there is no way of knowing for sure but could this be unrelated flakyness?

PR #1934, opened 2019-12-02
New Websocket Stack

neonphog:
## PR summary

```rust
use url2::prelude::*;
use in_stream::*;

let (send_binding, recv_binding) = crossbeam_channel::unbounded();

let server_thread = std::thread::spawn(move || {
    let config = TcpBindConfig::default();
    let config = TlsBindConfig::new(config).fake_certificate();
    let config = WssBindConfig::new(config);
    let mut listener:
        InStreamListenerWss<InStreamListenerTls<InStreamListenerTcp>> =
        InStreamListenerWss::bind(
            &url2!("wss://127.0.0.1:0"),
            config
        ).unwrap();

    println!("bound to: {}", listener.binding());
    send_binding.send(listener.binding()).unwrap();

    let mut srv = loop {
        match listener.accept() {
            Ok(srv) => break srv,
            Err(e) if e.would_block() => {
                std::thread::sleep(std::time::Duration::from_millis(1));
            }
            Err(e) => panic!("{:?}", e),
        }
    };

    srv.write("hello from server".into()).unwrap();
    srv.flush().unwrap();

    std::thread::sleep(std::time::Duration::from_millis(100));

    let mut res = WsFrame::default();
    srv.read(&mut res).unwrap();
    assert_eq!("hello from client", res.as_str());
});

let client_thread = std::thread::spawn(move || {
    let binding = recv_binding.recv().unwrap();
    println!("connect to: {}", binding);

    let mut cli: InStreamWss<InStreamTls<InStreamTcp>> =
        InStreamWss::connect(
            &binding,
            WssConnectConfig::new(
                TlsConnectConfig::new(
                    TcpConnectConfig::default()))).unwrap();

    cli.write("hello from client".into()).unwrap();
    cli.flush().unwrap();

    std::thread::sleep(std::time::Duration::from_millis(100));

    let mut res = WsFrame::default();
    cli.read(&mut res).unwrap();
    assert_eq!("hello from server", res.as_str());
});

server_thread.join().unwrap();
client_thread.join().unwrap();
```


crates/in_stream/src/mem.rs
```diff
@@ -0,0 +1,433 @@
...
+        loop {
+            // first, drain all pending connections from our recv channel
+            match self.recv.try_recv() {
+                Ok(stream) => {
+                    self.accept_queue.push(stream);
+                }
+                Err(crossbeam_channel::TryRecvError::Empty) => break,
+                Err(crossbeam_channel::TryRecvError::Disconnected) => {
+                    // wait until our user has accepted all pending connections
+                    // before letting them know the channel is broken
```
struktured: Why wait? Is this to suppress it pipe breaks on purpose if a "reconnect" occurs beforehand? This is very subtle, just want to make sure I'm grasping the intended behavior.

crates/in_stream/src/mem.rs
```diff
@@ -0,0 +1,433 @@
...
+        }
+        if self.accept_queue.is_empty() {
+            // acceptor is non-blocking we have nothing to return
+            return Err(Error::with_would_block());
+        }
+        // pull the next item off the queue
+
+        Ok(self.accept_queue.remove(0))
+    }
+}
```
struktured: I'm not in love this with non atomic pattern of accessing the vector (`is_empty / remove(0)`) but as it's currently a single threaded engine there's nothing technically wrong with it.

crates/in_stream/src/mem.rs
```diff
@@ -0,0 +1,433 @@
...
+        InStreamMem::raw_connect(url, config)
+    }
+}
+
+impl InStream<&mut [u8], &[u8]> for InStreamMem {
+    /// we want a url like mem://
+    const URL_SCHEME: &'static str = SCHEME;
+
+    fn raw_connect<C: InStreamConfig>(url: &Url2, config: C) -> Result<Self> {
+        let _ = MemConnectConfig::from_gen(config)?;
```
struktured: Confused by this pattern. Why is the return value ignored? How does it get associated with the memory manager? Is there a sneaky singleton underneath the hood?

crates/in_stream/src/mem.rs
```diff
@@ -0,0 +1,433 @@
...
+        Ok(get_mem_manager().connect(url)?)
```
struktured: That's annoying but I assume necessary for the error converter?

crates/in_stream/src/mem.rs
```diff
@@ -0,0 +1,433 @@
...
+    }
+
+    fn read(&mut self, buf: &mut [u8]) -> Result<usize> {
+        let mut disconnected = false;
+        if let Some(recv) = &mut self.recv {
+            for _ in 0..100 {
```
struktured: At least define a constant for the chunk size. Could be config option eventually.

crates/in_stream/src/mem.rs
```diff
@@ -0,0 +1,433 @@
...
+    /// manage binding a new MemListener interface
+    fn bind(&mut self, url: &Url2) -> Result<InStreamListenerMem> {
+        if SCHEME != url.scheme() {
+            return Err(Error::new(
+                ErrorKind::InvalidInput,
+                format!("mem bind: url scheme must be '{}'", SCHEME),
+            ));
+        }
+        match url.port() {
+            None | Some(0) | Some(PORT) => (),
```
struktured: You can match on static variable patterns? Awesome- even ocaml can't do that.

crates/in_stream/src/mem.rs
neonphog: No, we just have a buffer of data here... If we send the error first, they'll never receive what's in our buffer.

crates/in_stream/src/mem.rs
neonphog: yeah... alas `remove_item` on vec isn't stable... and rust's borrow rules make this safe even if we were using multiple threads... if accept_queue was an Arc<Mutex<_>> we'd just have to make sure the lock guard was in scope for both operations, right?

crates/in_stream/src/mem.rs
neonphog: Don't actually need any configuration for memory connects, so it's literally ignored. Just wanted the placeholder / error case to be in there in case we add configuration elements in the future.

crates/in_stream/src/mem.rs
neonphog: Nope! I can totally just return that result directly... will fix :+1:

crates/in_stream/src/mem.rs
neonphog: :v:

crates/in_stream/src/tcp.rs
```diff
@@ -0,0 +1,323 @@
...
+        let addr = tcp_url_to_socket_addr(url)?;
+        let listener = net2::TcpBuilder::new_v4()?
+            .bind(addr)?
+            .listen(config.backlog)?;
+        listener.set_nonblocking(true)?;
+        Ok(Self(listener))
+    }
+
+    fn binding(&self) -> Url2 {
+        let local = self.0.local_addr().unwrap();
```
struktured: How sensitive is this unwrap()?

crates/in_stream/src/tcp.rs
```diff
@@ -0,0 +1,323 @@
...
+    connect_timeout: Option<std::time::Instant>,
+}
+
+/// basic tcp socket stream
+#[derive(Shrinkwrap, Debug)]
+#[shrinkwrap(mutable)]
+pub struct InStreamTcp {
+    #[shrinkwrap(main_field)]
+    pub stream: std::net::TcpStream,
+    connecting: Option<TcpConnectingData>,
```
struktured: I'm surprised an optional enum is enough here, opposed to a custom enum with lots of various state transitions.

crates/in_stream/src/tcp.rs
neonphog: hard to know for sure... it looks like it's mainly this [underlying inet family match](https://github.com/rust-lang/rust/blob/master/src/libstd/sys_common/net.rs#L127-L141):

```rust
    match storage.ss_family as c_int {
        c::AF_INET => {
            assert!(len as usize >= mem::size_of::<c::sockaddr_in>());
            Ok(SocketAddr::V4(FromInner::from_inner(unsafe {
                *(storage as *const _ as *const c::sockaddr_in)
            })))
        }
        c::AF_INET6 => {
            assert!(len as usize >= mem::size_of::<c::sockaddr_in6>());
            Ok(SocketAddr::V6(FromInner::from_inner(unsafe {
                *(storage as *const _ as *const c::sockaddr_in6)
            })))
        }
        _ => Err(Error::new(ErrorKind::InvalidInput, "invalid argument")),
    }
```

crates/in_stream/src/tcp.rs
```diff
@@ -0,0 +1,323 @@
...
+                Ok(data.len())
+            }
+        } else {
+            self.write_buf.extend_from_slice(data);
+            Ok(data.len())
+        }
+    }
+
+    fn flush(&mut self) -> Result<()> {
+        loop {
```
struktured: I can't prove easily that this will terminate. What if the buffer refills at the start of each loop?

crates/in_stream/src/trait.rs
```diff
@@ -0,0 +1,143 @@
+use crate::*;
+use std::{
+    fmt::Debug,
+    io::{Error, ErrorKind, Result},
+};
+use url2::prelude::*;
+
+/// dynamic type for passing around in_stream configuration
+/// this is a stopgap until rust allows better constraints on associated types
+pub type InStreamConfigAny = Box<dyn std::any::Any + 'static + Send>;
```
struktured: Would like a deeper explanation of this issue, can be done outside of the PR review however.

crates/in_stream/src/tcp.rs
neonphog: yeah, the [net2](https://crates.io/crates/net2) non-blocking connect api is a little odd. It specifies just calling the same connect call repeatedly on the same stream until it gives something other than a WouldBlock response.

crates/in_stream/src/trait.rs
```diff
@@ -0,0 +1,143 @@
...
+/// `write` will buffer any data if the stream is not ready to write
+/// `flush` will block until pending data has been written
+pub trait InStream<R: Sized + Debug + Send + Sync, W: Sized + Debug + Send + Sync>:
+    Sized + Debug + Send + Sync
+{
+    /// your implementation should work with a single url scheme/protocol
+    /// if you need to support multiple (i.e. ws:// vs wss://) consider
+    /// using macros/code generation to DRY
+    const URL_SCHEME: &'static str;
+
```
struktured: Traits can define constants? Good to know!

crates/in_stream/src/ws.rs
```diff
@@ -0,0 +1,412 @@
...
+impl<Sub: InStreamListenerStd> InStreamListener<&mut WsFrame, WsFrame>
+    for InStreamListenerWss<Sub>
+{
+    type Stream = InStreamWss<Sub::StreamStd>;
+
+    fn raw_bind<C: InStreamConfig>(url: &Url2, config: C) -> Result<Self> {
+        let config = WssBindConfig::from_gen(config)?;
+        validate_url_scheme(url)?;
+        let mut url = url.clone();
+        url.set_scheme(Sub::StreamStd::URL_SCHEME).unwrap();
```
struktured: Why not use `?` here instead?

crates/in_stream/src/ws.rs
```diff
@@ -0,0 +1,412 @@
...
+            >,
+        >,
+    ),
+    Ready(tungstenite::WebSocket<StdStreamAdapter<Sub>>),
+}
+
+/// websocket stream
+#[derive(Debug)]
+pub struct InStreamWss<Sub: InStreamStd> {
+    state: Option<WssState<Sub>>,
```
struktured: Could consider making `WssState::None` to simplify this type a little.

crates/in_stream/src/ws.rs
```diff
@@ -0,0 +1,412 @@
...
+                _ => Err(Error::with_would_block()),
+            },
+        }
+    }
+
+    fn write(&mut self, data: WsFrame) -> Result<usize> {
+        self.priv_process()?;
+        self.write_buf.push_back(data);
+        self.priv_write_pending()?;
+        Ok(1)
```
struktured: Feels odd to me that this will always be one. Is this the consequence of forcing it to a frame and not having a `write_all(...)` batch equivalent?

crates/in_stream/src/tcp.rs
lucksus:
```suggestion
        let local = self.0.local_addr().expect("Couldn't unwrap local_addr() of TcpListener when trying to get binding URL");
```
?

crates/in_stream/src/tcp.rs
```diff
@@ -0,0 +1,323 @@
...
+        let addr = tcp_url_to_socket_addr(url)?;
+        let stream = net2::TcpBuilder::new_v4()?.to_tcp_stream()?;
+        stream.set_nonblocking(true)?;
+        match stream.connect(addr) {
+            Err(_) => Self::priv_new(
+                stream,
+                Some(TcpConnectingData {
+                    addr,
+                    connect_timeout: match config.connect_timeout_ms {
+                        None => None,
```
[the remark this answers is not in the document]
neonphog: exactly, None is keep trying forever

crates/in_stream/src/tcp.rs
neonphog: I suppose it may not... will add a TODO for potential future timeout flush config

crates/in_stream/src/trait.rs
neonphog: :v:

crates/in_stream/src/ws.rs
neonphog: can't fail unless scheme is mal-formed - and scheme is a constant... will add a comment.

crates/in_stream/src/ws.rs
neonphog: more like, this was me not wanting to introduce more generics to my traits...

Right now, we have R for the read type, W for the write type -- we could also add RR and WR for read-return and write-return so we could specify the return types as well.

The 1 to me *almost* makes sense... since the framing isn't an array type, yeah, you're just always dealing with individual elements.

neonphog:
@struktured 

> Also, what's the future of lib3huri and how does it relate to url2 crate?

lib3huri and url2 are sort of unrelated.

- url2 is meant to make the crates.io url crate more ergonomic to work with
- lib3huri is mainly for distinguishing the lib3h scheme types - it uses the url crate under the hood for parsing / encoding - we could switch it to use url2 : )

crates/in_stream/src/ws.rs
neonphog: I started with that... but I ran into borrowing problems... the `tungstenite` handshake api is an owned `self` method... so I need to be able to extract it...

Definitely one of the things I'm most frustrated with about rust... haven't settled into a good pattern I'm super happy with.

Issue #1600, opened 2019-07-17
Make DNA hash optional again


nmrshll:
Since v0.0.23, in the `holochain` conductor's config file, in `[[dnas]]`, the value `hash` is required again.

For developers, this seems to mean that we'll have to update the hash for every change in the zome code (manually or in automated fashion).

Is it possible to make that value optional again ? Or what is the justification behind that change ?
Thanks in advance

tjdforb:
See https://github.com/holochain/holochain-rust/pull/1335

lucksus:
I was about to respond that the way to run DNAs when developing is `hc run`. But then I first checked and found that we missed to update `hc` to actually set the right hash. I've just pushed a [PR to fix that](https://github.com/holochain/holochain-rust/pull/1603) which unfortunately missed the code freeze for version 0.0.25 this morning and will go into 0.0.26 next week.

For context:
This DNA hash check was introduced to help exactly with your use-case: running the conductor during development. There was a pitfall that had cost both the Holo as well as the Holochain core team several hours of debugging. If you run a production conductor (=`holochain`) for the first time it will read your DNA file and create an instance from that. That instance will initialize the source chain with that DNA as its first entry. The source chain is stored in the CAS (the storage that you configure in the conductor config with the instance). If you would now change your DNA code, recompile the DNA and replace the DNA file and then restart the conductor, you would still run the old DNA version since that is what is stored in the source chain. You would actually have to clean the storage to have `holochain` recreate the chain from the new file.

Making the hash mandatory and thus forcing the developer to explicitly say which DNA to run makes it possible that the conductor checks the version in the storage against that hash as well as the file against that hash. And complain if you have changed the file without changing the other two.

You should use `hc run`. If you want to use `holochain` for development you can easily get the hash of your DNA file with `hc hash -p <DNA file path>`.

pythagorean:
> You should use `hc run`. If you want to use `holochain` for development you can easily get the hash of your DNA file with `hc hash -p <DNA file path>`.

Am I understanding that #1594 won't be fixed until 0.0.26?

nmrshll:
[...]

PR #1939, opened 2019-12-03
Run hApp-bundles during develpment with `hc run`

lucksus:
## PR summary

Currently hApp developers can only run very simple, one DNA hApps with `hc run`. And they even have to host their own local HTTP server for the UI.

Every hApp that is a bit more complex (multiple DNAs, bridges, etc.) can't be easily run during development. This lead devs to run the production conductor with a real conductor config. That in turn required them to couple the UI to this development setup for instance by injecting interface URLs/ports via ENVs. These patterns are not compatible with production deployments.

What hApp devs should do is start their hApp by writing a hApp-bundle as the hApps manifest. That conveys all information to setup any run-time to run the hApp (be it dev, production Holoscape or production Holo).

This PR adds the needed capability to `hc` to read hApp-bundles and start a development conductor accordingly.

`hc run` will now first look for a file `bundle.toml` in the current working directory and if it is present will read it and configure the internal conductor to implement that hApp bundle, i.e. add all the instances and UIs (via its static HTTP server) and run that. If there is no `bundle.toml` file present it will proceed as before.

`hc run` will not follow HTTP links to download DNAs or UIs and will also not unzip UI archives. It expects the URIs to be of type `file:` for DNAs and of new type `dir:` for the UIs. The latter has to point to a directory and will be served by the static web server of the conductor at the given port. It should be exactly that UI directory that will be packaged into a zip file when deploying the bundle.

## Main benefit

This makes it possible to run complex hApp setups with multiple DNAs and bridges without having to write and maintain a conductor config. With that it enables a hApp development workflow that is centered around hApp-bundles from start and thus future-proof for deployment. It solves the problems discussed in https://github.com/holochain/holochain-rust/issues/1600.

## followups

Add functionality to `hc` for creating hApp bundles and updating them with the right DNA hash and UI directory path.

## changelog

```markdown
```

## documentation


philipbeadle:
Looks good Nico except that as a web dev I really want to use the React tools to run the front end, so could we make it run an npm command for the web server step? That way it addresses Zippy's issue as well.

crates/conductor_lib/src/happ_bundle.rs
```diff
@@ -0,0 +1,189 @@
...
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct HappBundleInstanceReference {
+    pub ui_handle: String,
+    pub instance_id: String,
+}
+
+impl HappBundle {
+    pub fn id_references_are_consistent(&self) -> Result<(), String> {
```
willemolding: This is great. It would be really easy to wasm-bindgen this integrity check code for holoscape so it isn't duped in Rust and Javascript

lucksus:
@zippy, the static web server works. I used the Community bundle for testing and had it running in my browser pointing to localhost. Agree that tests should be added. I don't fully understand how the npm step could help with that though, @philipbeadle?

philipbeadle:
Because I'd like to be able to make changes to my web code and see those changes. Rather than have to make a change, build, and rerun hcrun.

thedavidmeister:
@philipbeadle @lucksus we can include npm in the wrapped hc in holonix if we need...

lucksus:
@philipbeadle, the static web server of the conductor does not cache files. If you change a file and reload the page you will get the new content without having to restart `hc run`. (just confirmed locally)

PR #1940, opened 2019-12-03
Proposal: Naive Sharding Algorithm

neonphog:
Ha. Sharding algorithm that boils down to 2 lines of code:

if there are more than 50 nodes:
```rust
    // divide up the space so on average data will be stored by 50 nodes
    let dist: f64 = 4294967295.0 / (node_count as f64 / 50.0);

    // determine if this specific piece of data should be stored by this node
    agent_loc.forward_distance_to(data_addr_loc) < dist as u32
```

if there are less than 50 nodes, everybody stores everything.

The test adds 10 nodes at a time up to 10,000, making sure that the < 50 case is true, and > 50 is sane.

Every iteration it tests 100 random data points, counting the number of nodes that store it in the network. We get results like:

```
count: 99600 (because we track tests with < 50 nodes)
min: 25
max: 78
mean: 49.99037148594384
```


crates/sim2h/src/naive_sharding.rs
```diff
@@ -0,0 +1,131 @@
...
+                }
+
+                if nodes.len() < 50 {
+                    // if we have less than 50 nodes
+                    // make sure all nodes store all data
+                    assert_eq!(store_count, nodes.len());
+                } else {
+                    // if we have > 50 nodes,
+                    // assert that a reasonable number of nodes store the data
+                    assert!(store_count > 15);
```
willemolding: Bonus marks, what is the probability of a failure here? :P

crates/sim2h/src/naive_sharding.rs
```diff
@@ -0,0 +1,131 @@
...
+
+/// implement a super simple sharding algorithm
+/// to distribute data when node counts go > 50
+pub fn naive_sharding_should_store(
+    agent_loc: Location,
+    data_addr_loc: Location,
+    node_count: u64,
+) -> bool {
+    // if there are < 50 nodes, everyone should store everything
+    if node_count < 50 {
```
willemolding: Maybe pull out the magic 50 as a param and the magic space size as a const

crates/sim2h/src/naive_sharding.rs
jmday: ((n/50)-1)^n

jmday: (1-(50/n))^n ?

PR #1944, opened 2019-12-05
fixes for having reasonable path for trycyp loging

zippy:
## PR summary

For gathering log information from many remote servers being managed by trycp, the log files need to be at a known path.  This PR updates it so logs can be pulled from:

`/tmp/trycp/conductors/\.tmp[a-zA-Z0-9]*/[0-9]+/stderr.txt`

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation


crates/trycp_server/src/main.rs
```diff
@@ -80,11 +82,22 @@ struct TrycpServer {
     port_range: PortRange,
 }
 
+fn make_conductor_dir() -> PathBuf {
+    let conductor_path = PathBuf::new().join(TRYCP_DIRNAME).join(CONDUCTORS_DIRNAME);
+    std::fs::create_dir_all(conductor_path.clone()).expect("can create conductors dir");
```
struktured: This part should probably be a _create if it doesn't exist already_ pattern.

crates/trycp_server/src/main.rs
```diff
@@ -80,11 +82,22 @@ struct TrycpServer {
...
+    let dir = Builder::new().tempdir_in(conductor_path).expect("should create conductor tmp dir").into_path();
```
struktured: This is always random so shouldn't ever fail due to uniqueness issues, unlike the previous step.

crates/trycp_server/src/main.rs
```diff
@@ -80,11 +82,22 @@ struct TrycpServer {
...
+    dir
+}
+
 impl TrycpServer {
     pub fn new(port_range: PortRange) -> Self {
+        let root_path = PathBuf::new().join(TRYCP_DIRNAME);
+        std::fs::create_dir_all(root_path.clone()).expect("can create trycp dir");
+        let dna_dir = root_path.join(DNAS_DIRNAME);
```
maackle: It would feel cozier if there were also a `make_dna_dir()` function alongside `make_conductor_dir()`

zippy: done

crates/trycp_server/src/main.rs
zippy: I think that's exactly what create_dir_all does.  If it's already there it won't panic

PR #1945, opened 2019-12-05
State invariants

maackle:
## PR summary

Adds new failing tests, which make assertions on conductor state dumps at certain moments in time. I'd like to add various tests here to enforce our basic assumptions of key concepts like authoring and holding lists.

Tests:

* After starting up N instances, we expect each instance to hold aspects for N * 6 entries

## testing/benchmarking notes

This branch is about adding tests which fail. We can start new branches based on this branch with the work to make the tests pass, then merge them in when they pass.

## changelog

```markdown
```

## documentation

### DIFF
--- a/app_spec/test/package.json
+++ b/app_spec/test/package.json
@@ -3,7 +3,7 @@
     "standard": "^14.3.1"
   },
   "dependencies": {
-    "@holochain/tryorama": "^0.2.0-rc.4",
+    "@holochain/tryorama": "^0.2.1-rc.1",
     "faucet": "0.0.1",
     "json3": "^3.3.3",
     "sleep": "^6.1.0",
--- a/stress-test/package.json
+++ b/stress-test/package.json
@@ -15,8 +15,8 @@
     "typescript": "^3.6.3"
   },
   "dependencies": {
-    "@holochain/tryorama": "^0.2.0-rc.4",
-    "@holochain/tryorama-stress-utils": "0.0.2",
+    "@holochain/tryorama": "^0.2.1-rc.1",
+    "@holochain/tryorama-stress-utils": "0.0.3",
     "ramda": "^0.26.1",
     "tape": "^4.11.0",
     "sleep": "^6.1.0"
--- a/stress-test/src/all-on.ts
+++ b/stress-test/src/all-on.ts
@@ -12,7 +12,7 @@ module.exports = (scenario, N, M) => {
 
     // Make every instance of every conductor commit an entry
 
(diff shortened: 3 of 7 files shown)


maackle:
Here are results of running the new state-invariants test in this branch with various numbers of conductors:

`hc-stress-test-sim2h 1 1` : 0 aspects held (6 expected)
`hc-stress-test-sim2h 2 1` : 4 aspects held (12 expected)
`hc-stress-test-sim2h 3 1` : 8 aspects held (18 expected)
`hc-stress-test-sim2h 4 1` : 12 aspects held (24 expected)

i.e. `4(n - 1)`

When I merge this branch with `consistency-model-headers` (#1974), which does explicit header publishing of the first three chain entries, here are the results:

`hc-stress-test-sim2h 1 1` : 3 aspects held (6 expected)
`hc-stress-test-sim2h 2 1` : 7 aspects held (12 expected)
`hc-stress-test-sim2h 3 1` : 11 aspects held (18 expected)
`hc-stress-test-sim2h 4 1` : 15 aspects held (24 expected)
`hc-stress-test-sim2h 5 1` : 19 aspects held (30 expected)

i.e. `4(n - 1) + 3`

The expected formula is `6n`.

The values are taken by spawning the conductors, doing `await s.consistency()`, and then waiting 10 seconds. The pattern continues linearly like this for any number of nodes.

It's especially interesting to that we see zero aspects held when a single conductor boots up alone.

lucksus:
@maackle, I think `6n` is actually incorrect as it would include the DNA and we are not holding the DNA in the DHT role as everybody should have it in their source chain. But we are holding the DNA entry's header, so it would be `5n`.

So that is:
* 1 for DNA header
* 2 for AgentID + header
* 2 for zome call cap-token + header

With the current sim2 implementation we are not holding any of these at the source.

There is one "hack" in the current sim2h_worker that makes the local node hold everything that gets explicitly published. Because of this hack, the changes in #1974 make it hold 3 more (the headers). I call it hack because it makes the worker assume a full-sync DHT - but everything else assumes sim2h server to make the decision who should hold what.

With sharding, we will have to take out that line and rely on sim2h telling us what to hold.

If we say we always want to hold our own entries + headers, I would suggest we don't `hold` explicitly (through a `StoreEntryAspect` message) but we factor our authored (=source chain) entries into the holding list. I believe the query code already checks local storage which is why it works with local data as seen from hApp UIs. But that would make it harder for hachiko to tell if a node is holding something, unless we would make the formula be:
```
5(n-1)
```

PR #1946, opened 2019-12-05
Dynamic Interface Ports

lucksus:
## PR summary

This is to combat the `Address already in use` issue we see throughout all our tests in a random way.

Adds a new optional config flag `choose_free_port` to the `InterfaceConfiguration` which, if set, will still try to bind to the given port, but if that is used will iterate through following port numbers until an unused port is found.

Either way, the conductor will print out the port it bound to for each interface including the interface id.

## testing/benchmarking notes

Verify that this fixes our port issues in stress tests and CI in general.

## followups

As @maackle mentioned, this is brittle. We might want to remove this code again when we have other options. One promising might be domain sockets for local connections.

## changelog

```markdown
```

## documentation


maackle:
@lucksus I'm concerned about how this will play in conjunction with `admin/instance_add` or any other method that causes the interface to restart. In that case, the interface will restart and potentially bind to an altogether different port. For any client to realistically support a conductor with this flag set, it would have to maintain a dynamic mapping of instance IDs to websocket ports, constantly monitoring stdout from the conductor to find changes in port mappings, and manage its own websocket connections when a change occurs, tearing down the old one and starting a new one...

Tryorama doesn't use these admin functions, so it only has to update its port mappings once when the conductor starts up, so it will work for that case. But as a general feature, I think this is too brittle for general use -- unless we also update hc-web-client to gracefully deal with these port changes.

Do you think this is a pattern we want to use and encourage more generally?

AshantiMutinta:
How do we feel about reserved ports?

https://www.ietf.org/rfc/rfc1700.txt?number=1700

I'm thinking a use-case were we bind to a port 22 that is usually reserved for SSH services? Is it also round robin?

maackle:
Maybe we should be using "ephemeral ports" then (49152 and up) https://en.wikipedia.org/wiki/Ephemeral_port

maackle:
Also, this will not work with docker-based TryCP remote machines, since the docker images have only 20 ports exposed, so we really need to use the correct port there. So, if we encounter port allocation issues on a TryCP box, this solution will not help.

If this solves the problem of local and other CI tests, it's still useful, as long as TryCP / final exam are not seeing the port allocation issue.

maackle:
I changed the base of this PR because we should merge this at the same time as #1952 -- in particular, if we don't use this approach, we should probably merge neither

lucksus:
@maackle, I was thinking 3.
I was also thinking we want domain sockets (or named pipes) in most cases anyway. This just might be (hopefully) the fastest way to get our test frameworks and esp. stress tests going without tipping over ports..

maackle:
Let's get some "here be dragons" comments in here, which will also let us run through CI one more time and see if tests really passed, then I approve.

PR #1951, opened 2019-12-06
Added test to check if sim2h_url is set

freesig:
## PR summary
This adds a test to check that the dna property `sim2_url` is actually set by the DNA config file and affects the sim2h url that is pointed to.

## testing/benchmarking notes

This test is run by `cargo test should_set_sim2h_url`.
However it will hang if there is no sim2h_server running.
Maybe this should be an Tryorama test instead?

## followups

The only way I could figure out how to actually check that the property was set was to return it through the p2p_endpoint(). Although I don't think this is correct. Might be better to just check using Tryorama that it actually connects to the server of the right address and doesn't on the wrong address. Although currently it just hangs if it hits the wrong address.

Do I need to mark anything in the Miro board?

## changelog

The functionality is already there so I'm not sure if this PR should affect the changelog.

## documentation


crates/core/src/network/reducers/init.rs
```diff
@@ -156,4 +158,38 @@ pub mod test {
...
+        root_state = root_state.reduce(ActionWrapper::new(Action::InitializeChain(dna)));
+
+        let result = reduce_init(&mut network_state, &root_state, &action_wrapper);
+
+        assert_eq!(result, ());
+
+        if let Some(network) = network_state.network {
+            assert_eq!(network.p2p_endpoint().as_str(), "wss://localhost:9000/")
+        } else {
+        }
```
lucksus: I think this can be simplified to the following since we actually expect `network` to be `Some`:
```rust
let network = network_state.network.unwrap();
assert_eq!(network.p2p_endpoint().as_str(), "wss://localhost:9000/");
```
Also, I would suggest to make the changed URL differ more - I almost overlooked that the port is different.

freesig:
So it looks like it's this loop https://github.com/holochain/holochain-rust/blob/bf9a155d0834af9843638ac75255d262f6b3d0a0/crates/net/src/connection/net_connection_thread.rs#L63 that just loops with a 5 second delay waiting for sim2h. What is the proper behaviour here? What should happen if no connection is found after the timeout?

lucksus:
Yes, it should keep trying and I think it (=`Sim2hWorker`) would keep trying in a different thread.

The worker factoring is calling `Sim2hWorker::new` which will then fail in this line I guess:
https://github.com/holochain/holochain-rust/blob/bf9a155d0834af9843638ac75255d262f6b3d0a0/crates/net/src/sim2h_worker.rs#L110

Removing that `?` and setting `instance.connection_status` to `ConnectionStatus::None` seems like the right thing to do here.

freesig:
Ok I have changed it to not error and just set the connection to `None` which works and the test passes. However it now retries to connect on every tick here:
https://github.com/holochain/holochain-rust/blob/70ed96c339fd9fa2bbf1a0b32aeddc43095ed5f6/crates/net/src/sim2h_worker.rs#L130
Perhaps there should also be a small sleep here as well?

PR #1955, opened 2019-12-06
Sim2h - New Socket Stack

neonphog:
### Results of running `sim2h_stress` on `develop` branch

```
== RUN COMPLETE - Results ==
 - master_tick_count: 6392
+-------------------------------+--------+------+-------+----------------------+
|                          STAT | COUNT  | MIN  | MAX   | MEAN                 |
+-------------------------------+--------+------+-------+----------------------+
|               dm_handle_count | 201    | 1    | 1     | 1                    |
|               dm_result_in_ms | 21     | 5663 | 13235 | 10981.952380952382   |
|                 dm_send_count | 1000   | 1    | 1     | 1                    |
|          ping_recv_pong_in_ms | 246    | 3009 | 10943 | 7337.150406504065    |
|               ping_send_count | 1000   | 1    | 1     | 1                    |
| publish_received_aspect_in_ms | 20764  | 2959 | 11091 | 7355.958630321717    |
|            publish_send_count | 1000   | 1    | 1     | 1                    |
|           tick_job_elapsed_ms | 812251 | 0    | 18    | 0.006220983415225202 |
|         tick_sim2h_elapsed_ms | 6      | 228  | 2884  | 1556.5               |
+-------------------------------+--------+------+-------+----------------------+
```

### Results of running `sim2h_stress` on this `sim2h-ws-switch` branch

```
== RUN COMPLETE - Results ==
 - master_tick_count: 7109
+-------------------------------+--------+------+-------+---------------------+
|                          STAT | COUNT  | MIN  | MAX   | MEAN                |
+-------------------------------+--------+------+-------+---------------------+
|               dm_handle_count | 332    | 1    | 1     | 1                   |
|               dm_result_in_ms | 121    | 4918 | 10649 | 7438.231404958678   |
|                 dm_send_count | 1000   | 1    | 1     | 1                   |
|          ping_recv_pong_in_ms | 328    | 2325 | 9153  | 5273.381097560976   |
|               ping_send_count | 1000   | 1    | 1     | 1                   |
| publish_received_aspect_in_ms | 33683  | 2211 | 9187  | 5332.009114390069   |
|            publish_send_count | 1000   | 1    | 1     | 1                   |
|           tick_job_elapsed_ms | 600343 | 0    | 21    | 0.02151769904870945 |
|         tick_sim2h_elapsed_ms | 1123   | 0    | 58    | 7.061442564559217   |
+-------------------------------+--------+------+-------+---------------------+
```
Note the major fix in `tick_sim2h_elapsed_ms` - mean goes from 1556 ms to 7 ms. Everything else is across the board better as well, just not quite by so many orders of magnitude.


crates/sim2h/src/job/listen.rs
```diff
@@ -0,0 +1,36 @@
...
+    wss_send: crossbeam_channel::Sender<TcpWss>,
+}
+
+impl ListenJob {
+    pub(crate) fn new(listen: TcpWssServer, wss_send: crossbeam_channel::Sender<TcpWss>) -> Self {
+        Self { listen, wss_send }
+    }
+
+    fn run(&mut self) -> JobContinue {
+        match self.listen.accept() {
```
zippy: does this need the same check of `self.cont` as in the `ConnectionJob` run function?

crates/in_stream/src/mem.rs
```diff
@@ -383,6 +386,10 @@ mod tests {
             }
             .into_std_stream();
 
+            let rurl = srv.remote_url();
+            assert_ne!(listener.binding(), rurl);
+            assert_eq!(SCHEME, rurl.scheme());
```
AshantiMutinta: any reason for assert rather than returning an error? Is it unrecoverable?

crates/in_stream/src/tcp.rs
```diff
@@ -122,10 +124,12 @@ impl InStreamTcp {
 
     fn priv_new(
         stream: std::net::TcpStream,
+        url: Url2,
         connecting: Option<TcpConnectingData>,
     ) -> Result<Self> {
         Ok(Self {
             stream,
+            url,
```
AshantiMutinta: priv_new looks liike it will always returns OK, do we need the result here?

crates/sim2h/src/job/listen.rs
neonphog: Right now the listening socket just lives for the whole life of the process (or panic!s on error : ) - we can totally make this more robust in the future if we have need to start / stop listening on network interfaces at run-time.

crates/in_stream/src/tcp.rs
```diff
@@ -175,6 +179,7 @@ impl InStream<&mut [u8], &[u8]> for InStreamTcp {
         match stream.connect(addr) {
             Err(_) => Self::priv_new(
                 stream,
+                url.clone(),
                 Some(TcpConnectingData {
                     addr,
                     connect_timeout: match config.connect_timeout_ms {
```
AshantiMutinta: I know this was already written by someone else but we could get rid of the match and just do a config.connect_timeout_ms.map(|ms|{//add goes here})

Just reading through to gain better insights on how all this cones together

crates/in_stream/src/ws.rs
```diff
@@ -270,6 +270,16 @@ impl<Sub: InStreamStd> InStream<&mut WsFrame, WsFrame> for InStreamWss<Sub> {
         }
     }
 
+    fn remote_url(&self) -> Url2 {
+        let mut url = match self.state.as_ref().unwrap() {
+            WssState::MidCliHandshake(s) => s.get_ref().get_ref().remote_url(),
+            WssState::MidSrvHandshake(s) => s.get_ref().get_ref().remote_url(),
+            WssState::Ready(s) => s.get_ref().remote_url(),
```
AshantiMutinta: if they are the same type we can probably get away with a 
WssState::MidCliHandshake(s) | WssState::MidSrvHandshake(s) | WssState::Ready(s) => s.get_ref().get_ref().remote_url(),

crates/in_stream/src/tls.rs
```diff
@@ -248,6 +253,15 @@ impl<Sub: InStreamStd> InStream<&mut [u8], &[u8]> for InStreamTls<Sub> {
         }
     }
 
+    fn remote_url(&self) -> Url2 {
+        let mut url = match self.state.as_ref().unwrap() {
+            TlsState::MidHandshake(s) => s.get_ref().remote_url(),
+            TlsState::Ready(s) => s.get_ref().remote_url(),
```
AshantiMutinta: Same here 

TlsState::MidHandshake(s) | TlsState::Ready(s) => s.get_ref().remote_url

crates/sim2h/src/job/listen.rs
zippy: Yah I saw that panic.  What kind of error could there be?  I.e. what happens to the listening socket if the user changes the network interface?   I guess for the sim2h_server this shouldn't happen, so we can ignore.

crates/sim2h/Cargo.toml
```diff
@@ -24,13 +24,15 @@ lib3h_zombie_actor = "=0.0.25"
...
 holochain_core_types = { version = "=0.0.40-alpha1", path = "../core_types" }
 holochain_locksmith = { version = "=0.0.40-alpha1", path = "../locksmith" }
+in_stream = { version = "=0.0.40-alpha1", path = "../in_stream" }
 uuid = { version = "0.4", features = ["v4"] }
 log = "=0.4.8"
 env_logger = "=0.6.1"
 lazy_static = "=1.2.0"
 nanoid = "=0.2.0"
+num_cpus = "=1.11.1"
 url = "=2.1.0"
```
struktured: Why do you still need the `url` crate explicitly? Shouldn't all `url::Url`s be `Url2` by now?

crates/sim2h/src/job.rs
```diff
@@ -0,0 +1,20 @@
+/// typedef for code clarity
+pub(crate) type JobContinue = bool;
+
+/// an item that can be executed via thread pool
+pub(crate) trait Job: 'static + Send {
```
struktured: Where does the `'static` constraint come in from?

crates/sim2h/src/lib.rs
```diff
@@ -25,35 +24,69 @@ pub use crate::message_log::MESSAGE_LOGGER;
...
 use rand::{seq::SliceRandom, thread_rng};
 use std::{
     collections::{HashMap, HashSet},
     convert::TryFrom,
+    sync::Arc,
 };
 
+use holochain_locksmith::Mutex;
+
+pub(crate) trait MutexExt<T> {
```
struktured: Can you document this trait? And, briefly, why is traditional mutex not good enough?

crates/sim2h/src/lib.rs
```diff
@@ -25,35 +24,69 @@ pub use crate::message_log::MESSAGE_LOGGER;
...
+    fn f_lock(&self) -> holochain_locksmith::MutexGuard<T>;
+}
+
+impl<T> MutexExt<T> for Mutex<T> {
+    fn f_lock(&self) -> holochain_locksmith::MutexGuard<T> {
+        self.try_lock_for(std::time::Duration::from_millis(1000))
+            .expect("failed to obtain mutex lock")
```
struktured: No need for a result oriented api here?

crates/sim2h/src/lib.rs
```diff
@@ -62,35 +95,114 @@ pub struct Sim2h {
...
-        debug!("Trying to bind to {}...", bind_spec);
-        let url: url::Url = bind_spec.clone().into();
-        sim2h.bound_uri = Some(
-            sim2h
-                .stream_manager
-                .bind(&url)
-                .unwrap_or_else(|e| panic!("Error binding to url {}: {:?}", bind_spec, e))
-                .into(),
-        );
+        sim2h.priv_bind_listening_socket(url::Url::from(bind_spec).into(), wss_send);
```
struktured: Use url2 directly?

crates/sim2h/src/lib.rs
```diff
@@ -62,35 +95,114 @@ pub struct Sim2h {
...
+    /// bind a listening socket, and set up the polling job to accept connections
+    fn priv_bind_listening_socket(
+        &mut self,
+        url: Url2,
+        wss_send: crossbeam_channel::Sender<TcpWss>,
+    ) {
+        let config = TcpBindConfig::default();
+        let config = TlsBindConfig::new(config).dev_certificate();
+        let config = WssBindConfig::new(config);
+        let listen: TcpWssServer = InStreamListenerWss::bind(&url, config).unwrap();
```
struktured: We should simplify this config chain into a higher level function inside the socket api itself (think I mentioned that in the original socket PR). Not urgent...

crates/sim2h/src/lib.rs
```diff
@@ -62,35 +95,114 @@ pub struct Sim2h {
...
+        self.pool
+            .push_job(Box::new(Arc::new(Mutex::new(ListenJob::new(
+                listen, wss_send,
+            )))));
+    }
+
+    /// if our listening socket has accepted any new connections, set them up
+    fn priv_check_incoming_connections(&mut self) {
+        if let Ok(wss) = self.wss_recv.try_recv() {
+            let url: Lib3hUri = url::Url::from(wss.remote_url()).into();
```
struktured: Still concerned over the use of lib3huri, url, and url2. Maybe it deserves a follow up issue in the tree.

crates/sim2h/src/lib.rs
```diff
@@ -62,35 +95,114 @@ pub struct Sim2h {
...
+            let job = Arc::new(Mutex::new(ConnectionJob::new(wss, self.msg_send.clone())));
+            if let Err(error) = self.handle_incoming_connect(url.clone()) {
+                error!("Error handling incoming connection: {:?}", error);
```
struktured: Is the best we can do in terms of error handling? I assume yes because you can't communicate back to a broken socket.

crates/sim2h/src/lib.rs
```diff
@@ -62,35 +95,114 @@ pub struct Sim2h {
...
+        self.disconnect(&uri);
+    }
+
+    /// if our connections sent us any data, process it
+    fn priv_check_incoming_messages(&mut self) {
+        if let Ok((url, msg)) = self.msg_recv.try_recv() {
+            let url: Lib3hUri = url::Url::from(url).into();
+            match msg {
+                Ok(frame) => match frame {
+                    WsFrame::Text(s) => self.priv_drop_connection_for_error(
```
struktured: Would this custom protocol be replaced by something like capnproto in the future?

crates/sim2h/src/job/pool.rs
```diff
@@ -0,0 +1,55 @@
...
+                    }
+                }
+
+                if let Ok(mut job) = recv.try_recv() {
+                    if job.run() {
+                        send.send(job).expect("failed to send job");
+                    }
+                }
+
+                std::thread::yield_now();
```
struktured: Interesting optimization..

struktured: Now that i study it, this often might be an intelligence alternative to `thread:sleep(Duration::from_milllis(1))`.

crates/sim2h/src/job/pool.rs
```diff
@@ -0,0 +1,55 @@
...
+impl Pool {
+    pub(crate) fn new() -> Self {
+        let (job_send, job_recv) = crossbeam_channel::unbounded::<Box<dyn Job>>();
+        let job_cont = Arc::new(Mutex::new(true));
+        let mut job_threads = Vec::new();
+        for _ in 0..num_cpus::get() {
+            let cont = job_cont.clone();
+            let send = job_send.clone();
+            let recv = job_recv.clone();
+            job_threads.push(std::thread::spawn(move || loop {
```
struktured: Would be cool to record the cpu number in the thread id, not essential though.

crates/sim2h/src/job/pool.rs
```diff
@@ -0,0 +1,55 @@
...
+                {
+                    if !*cont.f_lock() {
+                        return;
+                    }
+                }
+
+                if let Ok(mut job) = recv.try_recv() {
+                    if job.run() {
+                        send.send(job).expect("failed to send job");
```
struktured: What happens if the thread dies? Are you tracking this somehow? The pool might need to recreate threads to the match the number of cores.

crates/stress/src/bin/sim2h_stress.rs
```diff
@@ -45,6 +45,10 @@ struct OptStressRunConfig {
     /// total runtime for the test
     run_time_ms: u64,
 
+    #[structopt(short, long, env = "STRESS_WARM_TIME_MS", default_value = "5000")]
```
struktured: An `env attribute`? I did not know about that!

crates/in_stream/src/tcp.rs
neonphog: The places `priv_new` is called from expect results - this saves me putting `Ok()`s in those spots : )

crates/sim2h/Cargo.toml
neonphog: : ( - alas, there's a lot of places that still expect the `url` crate type. e.g. Lib3hUri takes it, which is why i need to explicitly do a `Url::from(my_url2_instance).into()`

crates/sim2h/src/job.rs
neonphog: Makes things sooo much easier, because I don't have to specify lifetimes on the trait-object job instances - just means we can't hold non-static references in the closures / structs for the jobs, which I'm totally cool with.

crates/sim2h/src/lib.rs
neonphog: This is a case where I totally prefer a hard panic! - If we haven't been able to get a mutex lock in this time frame, it probably means we have a dead-lock.

crates/sim2h/src/lib.rs
neonphog: sim2h was built with Lib3hUri - there's no direct `From` for url2

crates/sim2h/src/lib.rs
neonphog: Yeah, in some ways this is just to be expected - We may even want to lower this to a trace! or debug! at some point.

crates/sim2h/src/lib.rs
neonphog: No, this is the websocket protocol. It's 100% analog to the tungstenite enum - just abstracts it so we can swap out our underlying websocket lib in the future with less trouble if we need to for some reason.

crates/in_stream/src/tcp.rs
neonphog: good suggestion, will update :+1:

crates/in_stream/src/ws.rs
neonphog: alas, they are not, tungstenite has different structs for each of these cases.

crates/sim2h/src/job/pool.rs
neonphog: I like that idea, will do :+1:

crates/sim2h/src/job/pool.rs
neonphog: Good point. For now, I'll make it so if a thread panics it will bring down the whole process - I almost feel like rust should behave that way by default anyways... We can look into more robust thread handling in the future.

PR #1956, opened 2019-12-06
Stat percentage change check

struktured:
## PR summary

This PR adds metadata to the metrics so we can group by a particular scenario.

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation


crates/metrics/src/main.rs
```diff
@@ -145,26 +115,57 @@ fn print_cloudwatch_stats(
...
 
 /// Prints to stdout human readonly pass/fail info
 /// Saves to `result_csv_file` gradient info
 fn print_stat_check(
-    _expected_csv_file: String, // StatsByMetric
-    _actual_csv_file: String,   // StatsByMetric
-    _result_csv_file: String,   // A collection of CheckedStatRecords
+    expected_csv_file: String, // StatsByMetric
+    actual_csv_file: String,   // StatsByMetric
+    result_csv_file: String,   // A collection of CheckedStatRecords
```
neonphog: Should these be `std::path::PathBuf` instead of `String`?

PR #1957, opened 2019-12-06
Use latest tryorama with improved hachiko network model

maackle:
## PR summary

hachiko 0.5.0 fixes the network model :crossed_fingers: to properly deal with nodes dynamically joining the network. Once we also update the consistency model to take headers into account, we should be able to remove the hard delays in the telephone game stress test.


maackle:
Note: we still have to fix the consistency model in core to deal with headers. This just fixes hachiko.

PR #1962, opened 2019-12-10
fix macos tls in-stream dev certificate generation

neonphog:
macos apparently doesn't like empty string passphrases for tls certificates : )


neonphog:
@struktured 

> Anything else need to change for this auth to work?

nope, it returns the passphrase from that function and in-tern uses it to decrypt the private key. I tested the fix in-place via circle-ci ssh.

PR #1963, opened 2019-12-10
address sim2h ws switchout code review comments

neonphog:
- use Option#map instead of match
- make thread panics take down the whole process
- name the threads by the cpu index
- add some extension trait comments

### DIFF
--- a/crates/in_stream/src/tcp.rs
+++ b/crates/in_stream/src/tcp.rs
@@ -182,14 +182,11 @@ impl InStream<&mut [u8], &[u8]> for InStreamTcp {
                 url.clone(),
                 Some(TcpConnectingData {
                     addr,
-                    connect_timeout: match config.connect_timeout_ms {
-                        None => None,
-                        Some(ms) => Some(
-                            std::time::Instant::now()
-                                .checked_add(std::time::Duration::from_millis(ms))
-                                .unwrap(),
-                        ),
-                    },
+                    connect_timeout: config.connect_timeout_ms.map(|ms| {
+                        std::time::Instant::now()
+                            .checked_add(std::time::Duration::from_millis(ms))
+                            .unwrap()
+                    }),
                 }),
             ),
             Ok(_) => Self::priv_new(stream, url.clone(), None),
--- a/crates/sim2h/src/job/pool.rs
+++ b/crates/sim2h/src/job/pool.rs
@@ -7,30 +7,50 @@ pub(crate) struct Pool {
     job_send: crossbeam_channel::Sender<Box<dyn Job>>,
 }
 
+lazy_static! {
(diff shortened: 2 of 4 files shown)


crates/in_stream/src/tcp.rs
```diff
@@ -182,14 +182,11 @@ impl InStream<&mut [u8], &[u8]> for InStreamTcp {
...
                     addr,
-                    connect_timeout: match config.connect_timeout_ms {
-                        None => None,
-                        Some(ms) => Some(
-                            std::time::Instant::now()
-                                .checked_add(std::time::Duration::from_millis(ms))
-                                .unwrap(),
-                        ),
-                    },
+                    connect_timeout: config.connect_timeout_ms.map(|ms| {
```
struktured: Wooh functional programming!

crates/sim2h/src/job/pool.rs
```diff
@@ -7,30 +7,50 @@ pub(crate) struct Pool {
...
+            std::process::exit(1);
+        }));
+        true
+    };
+}
+
 impl Pool {
     pub(crate) fn new() -> Self {
+        // make sure if a thread panics, the whole process exits
+        assert!(*SET_THREAD_PANIC_FATAL);
```
struktured: Would this get optimized away with a compiler flag? If so, intentional?

crates/sim2h/src/job/pool.rs
```diff
@@ -7,30 +7,50 @@ pub(crate) struct Pool {
...
-                    if job.run() {
-                        send.send(job).expect("failed to send job");
-                    }
-                }
-
-                std::thread::yield_now();
-            }));
+            job_threads.push(
+                std::thread::Builder::new()
+                    .name(format!("sim2h-pool-thread-{}", cpu_index))
```
struktured: :+1:

crates/sim2h/src/lib.rs
```diff
@@ -47,17 +47,29 @@ use std::{
...
             .expect("failed to obtain mutex lock")
     }
 }
 
+/// if a channel send fails, it means it is disconnected
+/// this extension trait simplifies panic!ing in that case
+/// in a lot of places, we expect the channel to always be open
+/// and don't have the infrustructure to deal with degenerate cases
+/// this trait makes sending more readable when we want to panic! on disconnects
 pub(crate) trait SendExt<T> {
```
struktured: :+1:

crates/sim2h/src/job/pool.rs
neonphog: I hope not! - the dereference to trigger the hook update is the whole point : )

struktured:
Just researched it, it's safe:
https://doc.rust-lang.org/std/macro.assert.html

PR #1964, opened 2019-12-10
Cache and retry failed wire messages in Sim2hWorker

lucksus:
## PR summary

What happened before:
If we are offline (i.e. don't have connection to the sim2h server) the sim2h worker would not process its inbox and thus would not generate any new messages to be sent.

But:
if the sending of a wire message failed we would have lost that message.
And if a connection would break up, we most likely would realise that by trying to send and getting an error. So there was a high-chance that connection issues would actually lead to message/data loss which could lead to all sorts of problems.

Now:
If we get a `TransportError` during send, we are caching the message and once we have a connection again, we are processing all cached messages.

We are now also processing the inbox even if offline.
* This makes sense because we currently still process queries locally anyway and also create a store message for every publish locally. Before we wouldn't do this when offline which would have resulted in not seeing local data until we reconnect again
* If the processing of those inbox messages leads to messages to be sent to sim2h, that is not a problem any more since those message will get chached.

## testing/benchmarking notes

There is no test for this in this PR. The only way to proper test this is to simulate connection drops, for which we have another ticket. Maybe that should be done within this PR?

## followups

## changelog

```markdown
```

## documentation

### DIFF
--- a/crates/net/src/sim2h_worker.rs
+++ b/crates/net/src/sim2h_worker.rs
@@ -55,6 +55,7 @@ pub struct Sim2hWorker {
     connection_status: ConnectionStatus,
     time_of_last_connection_attempt: Instant,
     metric_publisher: std::sync::Arc<std::sync::RwLock<dyn MetricPublisher>>,
+    outgoing_failed_messages: Vec<WireMessage>,
 }
 
 fn wire_message_into_escaped_string(message: &WireMessage) -> String {
@@ -109,6 +110,7 @@ impl Sim2hWorker {
             metric_publisher: std::sync::Arc::new(std::sync::RwLock::new(
                 DefaultMetricPublisher::default(),
             )),
+            outgoing_failed_messages: Vec::new(),
         };
 
         instance.connection_status = instance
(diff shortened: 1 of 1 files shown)


crates/net/src/sim2h_worker.rs
```diff
@@ -374,23 +382,32 @@ impl NetWorker for Sim2hWorker {
...
+                    // Ignore error here since send_wire_message logs TransportErrors during send
+                    // and puts back the failed message into self.outgoing_failed_messages
+                    let _ = self.send_wire_message(message);
                 }
             }
         }
 
+        let client_messages = self.inbox.drain(..).collect::<Vec<_>>();
+        for data in client_messages {
+            debug!("CORE >> Sim2h: {:?}", data);
```
neonphog: took me a minute to figure out what was going on here, perhaps a comment like:

```suggestion
            debug!("CORE >> Sim2h: {:?}", data);
            // outgoing messages triggered by `self.hand_client_message` that fail because of
            // connection status, will automatically be re-sent via `self.outgoing_failed_messages`
```

crates/net/src/sim2h_worker.rs
```diff
@@ -374,23 +382,34 @@ impl NetWorker for Sim2hWorker {
...
-                    debug!("CORE >> Sim2h: {:?}", data);
-                    if let Err(error) = self.handle_client_message(data) {
-                        error!("Error handling client message in Sim2hWorker: {:?}", error);
-                    }
-                    did_something = true;
+                let previously_failed_messages =
+                    self.outgoing_failed_messages.drain(..).collect::<Vec<_>>();
+                for message in previously_failed_messages {
+                    // Ignore error here since send_wire_message logs TransportErrors during send
+                    // and puts back the failed message into self.outgoing_failed_messages
```
AshantiMutinta: does this comment still apply?

lucksus: yes

PR #1969, opened 2019-12-11
make sim2h not use all the cpu

neonphog:
This isn't 100% ideal

We could switch the underlying in_stream sockets to be based on mio and get the epoll/iocp benefits - which we may want to do anyway, at least to get named sockets on windows as the analog on that platform for unix domain sockets.

However, in the short-term, this lowers the idle cpu usage of sim2h_server from ~800% on my 8 core system to ~40% without sacrificing any of the fully loaded stress test performance.

### DIFF
--- a/crates/sim2h/src/job/listen.rs
+++ b/crates/sim2h/src/job/listen.rs
@@ -12,10 +12,12 @@ impl ListenJob {
         Self { listen, wss_send }
     }
 
-    fn run(&mut self) -> JobContinue {
+    fn run(&mut self) -> JobResult {
         match self.listen.accept() {
             Ok(wss) => {
                 self.wss_send.f_send(wss);
+                // we got data this time, check again right away
+                return JobResult::default();
             }
             Err(e) if e.would_block() => (),
             Err(e) => {
@@ -25,12 +27,13 @@ impl ListenJob {
                 panic!(e);
             }
         }
-        true
+        // no data this round, wait 5ms before checking again
+        JobResult::default().wait_ms(5)
     }
 }
 
 impl Job for Arc<Mutex<ListenJob>> {
-    fn run(&mut self) -> JobContinue {
+    fn run(&mut self) -> JobResult {
         self.f_lock().run()
     }
 }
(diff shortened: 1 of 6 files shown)


maackle:
What is going on here, intuitively, that leads to such a speedup? Hard to tell just from the changes.

neonphog:
@maackle 

> What is going on here, intuitively, that leads to such a speedup? Hard to tell just from the changes.

I had cpu-count threads in loops with thread yields, but they were still checking for work every time they got cpu time, so were maxing out the system (although they would have yielded to other processes/threads had they any real work to do).

Now the jobs can choose to return a wait value, and the scheduler will park them until that wait expires.

crates/sim2h/src/job/listen.rs
```diff
@@ -12,10 +12,12 @@ impl ListenJob {
...
 
-    fn run(&mut self) -> JobContinue {
+    fn run(&mut self) -> JobResult {
         match self.listen.accept() {
             Ok(wss) => {
                 self.wss_send.f_send(wss);
+                // we got data this time, check again right away
+                return JobResult::default();
             }
             Err(e) if e.would_block() => (),
```
struktured: I have mixed feelings about using this pattern guard, but it's premature optimization to get much further into it. In OCaml I know this can have a severe effect on the pattern matching speed, for instance. https://wiki.c2.com/?PatternMatching is a nice abstract treatment on the topic.

crates/sim2h/src/job/listen.rs
```diff
@@ -25,12 +27,13 @@ impl ListenJob {
                 panic!(e);
             }
         }
-        true
+        // no data this round, wait 5ms before checking again
+        JobResult::default().wait_ms(5)
```
zippy: Make the `5` be a constant and match with 5 above?  and why 5 and not 1 2 or 3?

crates/sim2h/src/job/pool.rs
```diff
@@ -34,20 +34,47 @@ impl Pool {
...
+                                    send.f_send(job);
+                                } else {
+                                    parked_jobs.push((t, job));
+                                }
                             }
-                        }
 
-                        std::thread::yield_now();
+                            if let Ok(mut job) =
+                                recv.recv_timeout(std::time::Duration::from_millis(5))
```
zippy: another 5?

crates/sim2h/src/job/listen.rs
neonphog: Yeah, the problem is, anything else wouldn't be more performant, I still need to do that if there... if only you could make tri-state results in rust : )

PR #1972, opened 2019-12-12
Metric publishing related updates

struktured:
## PR summary

 - Creates a channel based version of a metric publisher to avoid processing on the main thread
 - Extends `holochain_metrics` to print the raw (non aggregated) statistics of a performance run
 - Changes manual latency calculation in `run_validation` to instead use the `with_latency_publishing` macro
 - Provides ability to aggregate any csv file of raw metrics using a regular expression pattern.
## testing/benchmarking notes

## followups

- [x] Finish printing raw metrics
- [ ] Actually test/use channel based publisher
- [x] finish aggregation pattern feature for local logs

## changelog

```markdown
```

## documentation


crates/conductor_lib/src/context_builder.rs
```diff
@@ -166,6 +166,26 @@ impl ContextBuilder {
...
     pub fn with_metric_publisher(mut self, config: &MetricPublisherConfig) -> Self {
+        let config = match &config {
+            MetricPublisherConfig::CloudWatchLogs(config) => {
+                let log_stream_name = config.log_stream_name.clone().map(|log_stream_name| {
+                    format!(
+                        "{}.{}",
+                        log_stream_name,
+                        self.instance_name
+                            .clone()
+                            .unwrap_or_else(|| format!("{}", snowflake::ProcessUniqueId::new()))
```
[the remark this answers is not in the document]
maackle: Yeah @struktured this catches my attention too, should it be a UUID?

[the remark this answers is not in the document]
struktured: I will probably drop this snowflake code all together- it doesn't end up being used and the snowflake id wasn't unique enough either for some reason.

crates/core/src/nucleus/actions/run_validation_callback.rs
```diff
@@ -10,56 +10,57 @@ use holochain_core_types::error::HolochainError;
...
+                        // silently failing validation
+                        panic!(error_string)
+                    }
+                }
+                Err(error) => panic!(error.to_string()), // same here
+            }
+        },
+        address,
+        call,
+        context
```
maackle: Sorry, not relevant to this PR, but I'm just seeing `with_latency_publishing!` in action for the first time so I have a question. @struktured what is the reason for supplying a closure that takes args and then passing them into the closure in the macro? Why not just have a zero arity closure and move the values needed into it?

struktured: @maackle It's a good question. The idea behind the macro is you usually pass a defined function and its arguments to the macro and it invokes the function for you. In this case I didn't bother defining a separate function so it appears more awkward. Will take your advice and make the closure simpler.

crates/metrics/src/main.rs
```diff
@@ -31,35 +31,56 @@ enum Command {
...
+    )]
+    PrintMetricStats {
+        #[structopt(name = "csv_file", short = "f")]
+        csv_file: PathBuf,
         #[structopt(
-            name = "assume_role_arn",
-            short = "a",
-            help = "Optional override for the amazon role to assume when querying"
+            name = "aggregation_pattern",
+            long = "aggregation_pattern",
```
maackle:
```suggestion
            long = "aggregation-pattern",
```

I think this would be a more idiomatic cli flag in kebab-case

crates/metrics/src/main.rs
```diff
@@ -31,35 +31,56 @@ enum Command {
...
         name = "print-log-stats",
         about = "Prints descriptive stats in csv format over a log file"
     )]
     PrintLogStats {
         #[structopt(name = "log_file", short = "f")]
-        log_file: String,
+        log_file: PathBuf,
+        #[structopt(
+            name = "aggregation_pattern",
+            long = "aggregation_pattern",
```
maackle:
```suggestion
            long = "aggregation-pattern",
```

same

crates/metrics/src/main.rs
```diff
@@ -125,12 +179,51 @@ fn print_cloudwatch_stats(
...
+    let file = BufWriter::new(std::io::stdout());
+    let mut writer = csv::Writer::from_writer(file);
+
+    for m in metrics {
+        writer.serialize(m).unwrap();
+    }
+    writer.flush().unwrap();
+}
+
+// TODO use aggregation patern
```
maackle: TODO for this PR, or later?

struktured: Later is fine, this is for generating metrics from just local log files, not cloudwatch.

crates/conductor_lib/src/context_builder.rs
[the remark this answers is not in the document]
struktured: Removed snowflake.

crates/core/src/nucleus/actions/run_validation_callback.rs
struktured: Simplified the closure to take just a unit arg. My macro would have to slightly change to support the zero argument case (can do in a future PR though).

PR #1974, opened 2019-12-13
Consistency model headers

maackle:
## PR summary

Make consistency model aware of publishing headers

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation


crates/core/src/workflows/application.rs
```diff
@@ -66,8 +68,23 @@ pub async fn initialize(
...
     if first_initialization {
         call_init(dna, &instance_context).await?;
+
+        log_debug!(
+            context,
+            "dna/initialize: publishing headers of initial chain entries...",
+        );
+
+        for header in context.state().unwrap().agent().iter_chain() {
+            publish_header_entry(header.entry_address().clone(), &context).await?;
```
lucksus: Just for the record, did I understand this correct: you're adding this explicit header publish here because it makes scenario tests of some hApps work, which otherwise are flaky?

maackle: @lucksus yes, I'm pretty sure that the rule we have in the consistency model for waiting for chain initialization is not working properly, so any scenario that relies on AgentID is not actually able to reliably wait for that entry to be held.

The proof is that without this branch, the telephone games which fetch AgentID as the check step fail without a hard timeout, but with this branch, they pass without a hard timeout, relying only on the consistency model.

maackle:
I think there is still one thing missing, and that is for an instance to Hold all of its own initial chain entries and headers, which is the point of this PR which we still need to make tests pass for: https://github.com/holochain/holochain-rust/pull/1945

Without that, it is possible for hachiko to wait until it hard-timeouts, since e.g. the source of the AgentId entry may never actually Hold it. Which is not such a big deal, but it makes our tests run longer and indicates an incomplete model of what's going on.

crates/core/src/workflows/application.rs
willemolding: Hmm it is a shame to have to add this back at least we know we are not relying on it for things to work (the holding list does the actual sharing), but it is only for tests

thedavidmeister:
@maackle can we close this in light of https://github.com/holochain/holochain-rust/pull/1990 and https://github.com/holochain/holochain-rust/pull/1993?

maackle:
Let's leave it open until we really know what's going on @thedavidmeister

maackle:
OK let's close this, assuming that the hachiko timeouts in #1990 are solely due to some CI problem

PR #1975, opened 2019-12-13
Instream drop on demand

willemolding:
## PR summary

To be able to test the behavior of various tests a way to artificially induce failures at different layers of the network stack is required.

We propose using env vars which are picked up inside the in_stream crate to drop connections, messages or packets depending on the test requirement. The currently implementation only drops ws connections by ignoring incoming and outgoing messages for the duration of an error burst.

As this crate is to be used by both the sim2h worker and sim2h server we can simulate failures at either end with no extra work.

### Dropping ws Connections

The env var `WS_FAILURE_MODEL` is a tuple (S, $\frac{1}{\lambda}$, $\mu$) where $\frac{1}{\lambda}$ is the mean time between failures, $\mu$ is the expected drop duration (in ms) and $S$ is a seed for the random number generator to ensure repeatability. 

The failure model follows a telegraph process which switched between being in an error state (burst) or an ok state. The time between errors and the error burst duration are exponentially distributed random variables independent of each other.

## followups

## changelog

```markdown
```

## documentation


crates/in_stream/src/ws.rs
```diff
@@ -281,6 +298,10 @@ impl<Sub: InStreamStd> InStream<&mut WsFrame, WsFrame> for InStreamWss<Sub> {
     }
 
     fn read(&mut self, data: &mut WsFrame) -> Result<usize> {
+        if self.is_in_failure() {
```
willemolding: @neonphog These are the bits I'm not 100% on. I guess this will definitely make everything fail for the duration but do we maybe want to do this another way?

neonphog: @willemolding - A couple other options:

- One would be to move this code to a lower level, and in Tcp and Mem streams, call [shutdown](https://doc.rust-lang.org/std/net/struct.TcpStream.html#method.shutdown) when your failure timing is triggered.
- Alternately, if we wanted to go for more encapsulation, you could create `InStreamFailure` and `InStreamListenerFailure` structs that would mostly be pass-through like Tls, but instead of encrypting / decrypting things it would randomly drop its sub-connection object, letting rust close the underlying streams.

zippy:
As per stand-up today, from an activating from testing point of view, adding being able to specify the values from trycp_server "spawn" command is needed.  Will review as soon as that's ready.

crates/in_stream/src/failure_model.rs
```diff
@@ -0,0 +1,102 @@
...
+        }
+    }
+}
+
+#[cfg(test)]
+pub mod tests {
+    // use super::*;
+    // use std::thread::sleep;
+
+    // #[test]
```
thedavidmeister: did you mean to leave this out @willemolding ?

PR #1976, opened 2019-12-13
Tiny - Drop dep on petgraph and simplify logic for computing free pending validation

willemolding:
## PR summary

Since we were not actually using the topological sort functionality of petgraph this drops it as a dependency and uses a HashSet to calculate the leaves of the dependency graph.


crates/core/src/dht/dht_store.rs
```diff
@@ -306,44 +306,30 @@ impl DhtStore {
     }
 }
 
-use petgraph::{graph::DiGraph, prelude::NodeIndex, Direction::Outgoing};
-use std::collections::HashMap;
+use std::collections::HashSet;
```
AshantiMutinta: Could we use immutable crate here? it is much faster imo

crates/core/src/dht/dht_store.rs
```diff
@@ -306,44 +306,30 @@ impl DhtStore {
...
+    // only return those that don't have anything also pending as a dependency
+    // as we know these will always fail
+    pending
+        .clone()
+        .into_iter()
+        .filter(|p| {
+            p.pending
+                .dependencies
+                .iter()
+                .all(|dep_addr| !unique_pending.contains(dep_addr))
```
AshantiMutinta: If we are only returning tings that don't have a dependency could we get away with using a any here? that way it would short circuit at any sign of a dependancy, all would require O(n) operations

crates/core/src/dht/dht_store.rs
```diff
@@ -306,44 +306,30 @@ impl DhtStore {
...
+    let unique_pending: HashSet<Address> = pending
+        .clone()
+        .into_iter()
+        .map(|p| p.pending.entry_with_header.entry.address())
+        .collect();
+
+    // only return those that don't have anything also pending as a dependency
+    // as we know these will always fail
+    pending
+        .clone()
```
AshantiMutinta: We are cloning these twice, can we get away without doing this?

crates/core/src/dht/dht_store.rs
maackle: `all` is also short-circuiting

crates/core/src/dht/dht_store.rs
maackle: Yeah I wonder if you can just get away with using addresses?

crates/core/src/dht/dht_store.rs
maackle: [citation needed] ;)

crates/core/src/dht/dht_store.rs
AshantiMutinta: Yes that is true!

crates/core/src/dht/dht_store.rs
AshantiMutinta: https://docs.rs/im/14.0.0/im/

"The maps - HashMap and OrdMap - generally perform similarly to their equivalents in the standard library, but tend to run a bit slower on the basic operations (HashMap is almost neck and neck with its counterpart, while OrdMap currently tends to run 2-3x slower). On the other hand, they offer the cheap copy and structural sharing between copies that you'd expect from immutable data structures."

-having the hashmap be almost neck and neck but benefiting from structural sharing if we ever want to do a clone makes it look more efficient from my perspective but I could be wrong.

We could also just say no more cloning but use of the std hashmap will be slower if a clone is ever introduced

crates/core/src/dht/dht_store.rs
willemolding: I rejigged it a bit. Not sure if is actually better but I don't think it is worse...

willemolding:
@lucksus I was unable to reproduce your issue but I'm fairly sure it was related to losing the sorted-ness of the queue. I have rewritten it to just use a filter to eliminate pending validations with pending dependencies so it should be ok.
I want to run trial to check before merging though.

lucksus:
@willemolding I'd suggest to merge this now and get it into 0.0.41 as it is definitely an improvement and no regression judging from the passing tests.

PR #1978, opened 2019-12-13
Have "choose_free_port" set it 0 so OS picks port

lucksus:
## PR summary

This deactivates the [previously added](https://github.com/holochain/holochain-rust/pull/1946) code that manually sweeps through ports to find a open one to bind an interface to and rather relies on the OS finding a free port when binding to port 0.

We can easily get the bound address from JSON-RPC's `Server` and print that out for try-o-rama to use.

The problem with the previous approach was that there still was a split-second in which we had to let go of the port in order to have the JSON-RPC server bind to it.

This hopefully eliminates the last `Address already in use` problems that create false negatives in stress tests.

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation


crates/conductor_lib/src/conductor/base.rs
```diff
@@ -1538,29 +1539,44 @@ fn run_interface(
...
-            || {
-                let r = HttpInterface::new(port).run(handler, kill_switch);
-                println!("{}", magic_port_binding_string(&interface_config.id, port));
-                r
-            },
-        ),
 
+    match interface_config.driver {
+        InterfaceDriver::Websocket { port } => {
+            let port = if interface_config.choose_free_port.unwrap_or(false) {
```
AshantiMutinta: I was wondering if choose_free_port needs to be an option? I'm thinking it is either you choose the free port or you don't right?

willemolding: Its nicer if its an option right because then Serde can still deserialize it if the field isn't there?

AshantiMutinta: We could use #[serde(default)] that way it default to false already

crates/conductor_lib/src/conductor/base.rs
```diff
@@ -1517,6 +1517,7 @@ fn _make_interface(interface_config: &InterfaceConfiguration) -> Box<dyn Interfa
     }
 }
 
+#[allow(dead_code)]
 fn with_port_heuristic<T, F: FnOnce() -> T>(
```
AshantiMutinta: Planning to use this in function in the future I guess? :)

crates/conductor_lib/src/conductor/base.rs
willemolding: oh yeah nice

PR #1988, opened 2019-12-17
Sim2h Worker using InStream WebSocket Stack

neonphog:


crates/net/src/sim2h_worker.rs
```diff
@@ -426,36 +434,59 @@ impl NetWorker for Sim2hWorker {
...
-                    self.time_of_last_connection_attempt = Instant::now();
-                    warn!("No connection to sim2h server. Trying to reconnect...");
-                    if let Err(e) = self.stream_manager.connect(&self.server_url.clone().into()) {
-                        error!("TransportError trying to connect to sim2h server: {:?}", e);
+        if self.ws_frame.is_none() {
+            self.ws_frame = Some(WsFrame::default());
+        }
+
+        if self.connection_ready() {
+            self.try_send_from_outgoing_buffer();
```
zippy: Should this be returning a `did_something` boolean too?  i.e. did_something=true if we sent something?

PR #1998, opened 2019-12-19
Fix gossip issue - Have FetchEntry include meta aspects from source chain

lucksus:
## PR summary

While diagnosing the issue of pending validations with unresolved links, I discovered something interesting when looking at sim2h logs and comparing hashes with the debug view of my local Holoscape instance:

Sim2h reported several missing aspects for all my running instances. Including the Personas&Profiles instance which is made to be unique: currently there can only be exactly one instance of the same DNA - an installation on another device would create a different random hash.

So every aspect hash Sim2h knows about can only come from my instance. But then how can there be something missing? With [these changes](https://github.com/holochain/holochain-rust/pull/1990) we have sim2h_worker make core store/hold everything that is in the authoring list. That is, sim2h_worker first fetches every entry/aspect via a `HandleFetchEntry` message, and then translates the response into `StoreEntryAspect`s.

The problem was that the `HandleFetchEntry` handler *only* asked the DHT/EAV for data to construct all aspects of the requested entry. But if we are relying on FetchEntry to get the meta aspects from the chain into the DHT/EAV, those aspects will never get there.

The latter was the case, and both explains:
1. why sim2h server was stuck finding missing aspects and had no way to get those
2. why we could get into a situation where a node gets incomplete data through gossip -> being stuck with entries in the validation queue for which it never gets the dependencies (the node that gossiped those entry/aspects might have had the needed dependencies in *its* source chain and used those to validate the bogus entries, but nobody else can if those source chain entries never get gossiped)

## testing/benchmarking notes

I could verify that point 1 above is fixed with these changes. I am confident that it also fixes point 2, but in order to be sure we need to run according experiments. If CI proves that there are no regressions I would suggest to merge this without the proof for 2. It is still an important improvement and remaining issues with dependent validations could get fixed as a follow up.

## followups

## changelog

```markdown
```

## documentation


crates/core/src/network/handler/fetch.rs
```diff
@@ -2,30 +2,40 @@ use crate::{
...
-                Ok(mut meta_aspects) => aspects.append(&mut meta_aspects),
-                Err(get_meta_error) => {
-                    log_error!(context, "net/handle_fetch_entry: Error getting meta aspects for entry ({:?}), error: {:?}",
-                        address,
-                        get_meta_error,
-                    );
+            aspects.insert(content_aspect);
+            for result in &[
+                get_meta_aspects_from_chain(&address, context.clone()),
+                get_meta_aspects_from_dht_eav(&address, context.clone()),
```
lucksus: This is where we used to only get meta aspects from the DHT/EAV (the old fn name `get_meta_aspects()` implies that we were not thinking about getting it from the chain at all). Now we ask both sources for meta aspects for the given entry address and merge the results.

crates/core/src/network/handler/fetch.rs
```diff
@@ -2,30 +2,40 @@ use crate::{
...
 use lib3h_protocol::data_types::FetchEntryData;
-use std::sync::Arc;
+use std::{collections::HashSet, sync::Arc};
 
 /// The network has requested a DHT entry from us.
 /// Lets try to get it and trigger a response.
 pub fn handle_fetch_entry(get_dht_data: FetchEntryData, context: Arc<Context>) {
     let address = get_dht_data.entry_address.clone();
-    let mut aspects = vec![];
+    let mut aspects: HashSet<EntryAspect> = HashSet::new();
```
lucksus: Because of this we need the `Eq` derivations and the `Hash` implementation below.

crates/net/src/sim2h_worker.rs
```diff
@@ -333,6 +335,42 @@ impl Sim2hWorker {
         }
     }
 
+    fn self_store_authored_aspects(&mut self) {
```
lucksus: The changes in this file make the booting of a conductor faster in the case of already holding/authoring a lot. With yesterday's naive implementation we would always send store (and first fetch-) requests for everything in the authoring list, no matter if we are holding already. This waits until we have both lists so we can actually compare if there is anything in the source chain that needs to be held.
(Which needs to be here in sim2h because it is a full-sync DHT assumption and should go away or get deactivated once we activate sharding)

maackle: If there is code here that needs to go away can you make that visible in the comments with a TODO?

crates/core/src/network/handler/mod.rs
```diff
@@ -362,7 +364,69 @@ fn get_content_aspect(
     ))
 }
 
-fn get_meta_aspects(
+/// This function converts an entry into the right "meta" EntryAspect and the according
+/// base address to which it is meta, if the entry is the source entry of a meta aspect,
+/// i.e. a CRUD or link entry.
+/// If the entry is not that it returns None.
+fn entry_to_meta_aspect(entry: Entry, header: ChainHeader) -> Option<(Address, EntryAspect)> {
```
thedavidmeister: @lucksus when i see things like this my first reaction is that we want a `From` or `TryFrom` somewhere so that the rust compiler can help us more, but it's not worth blocking this PR at all

crates/net/src/sim2h_worker.rs
thedavidmeister: @maackle @lucksus i don't think this is worth blocking a merge here as we are getting fairly overdue for a release and @lucksus is heading off for the year...

a TODO would be nice but also it sounds like sharding simply won't work until this is upgraded so i doubt we'd miss it :man_shrugging:

crates/net/src/sim2h_worker.rs
```diff
@@ -333,6 +335,42 @@ impl Sim2hWorker {
...
+        if !self.has_self_stored_authored_aspects
+            && self.initial_gossiping_list.is_some()
+            && self.initial_authoring_list.is_some()
+        {
+            let authoring_list = self.initial_authoring_list.take().unwrap();
```
thedavidmeister: minor point, but this is weird and the kind of thing clippy doesn't like, what would it mean for a list to be `None` vs. `Some`-yet-empty, and then why do we do an `is_some()` in the condition but then need to `unwrap()` in the body? @lucksus

crates/net/src/sim2h_worker.rs
maackle: To me it sounds like sharding will work just fine, but we'll be potentially Holding stuff that the sharded network doesn't need us to hold. Either way I guess it's enough that at least the three of us have taken note.

crates/net/src/sim2h_worker.rs
maackle: Yeah this could have been written where the lists are `take`n and bound to local vars before the condition, to avoid the unwrap. But, minor.

The some-ness indicates whether or not the list got returned from the network, but if it's empty, that's just what was fetched.

thedavidmeister: @maackle ah, i see, yeah that does make sense "unknown" vs. "known and empty"

PR #2001, opened 2019-12-19
Set hachiko to strict mode

maackle:
## PR summary

This will cause a test failure whenever hachiko times out. Since we have been having lots of uncertainty around hachiko and our consistency model, which are cornerstones of our scenario testing strategy, it's pretty important that have confidence in them. So, let's make failures red.

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation


maackle:
Curiously, we get lots of hachiko timeouts in CI, but none locally. Also, when I run a CI build with SSH and then run the test again on the same machine, it passes without timeouts.

When it does timeout, it causes the whole test to hang indefinitely because a conductor is not getting killed. And then the CI test can run for hours until it's manually canceled...

maackle:
I think there's just something wrong with the way conductors connect to sim2h in CircleCI. I see similar failures when I run the tests locally without sim2h running...

We'll just have to keep an eye on this until we figure out what might be wrong :man_shrugging:

PR #2002, opened 2019-12-20
Handle unseen aspects from gossip list not just authoring list

zippy:
## PR summary

There's a bug where sim2h isn't keeping track of unseen aspects in it's model of which aspects exist in the space from gossiping data only from authoring data.  This means that in this situation:  1) all the nodes that have some aspect data go off line, 2) a new node comes on-line,  3) an old node which had data but was not the author, then sim2h would not know to send the data to the new node.

This PR adds a test and refactors the handle_unseen code for use in both situations.

## testing/benchmarking notes

It also identified that tryorama was using the in-memory storage for stress tests, which would cause this case to fail because the old non-authoring node coming back on line would have lost it's holding list.

The new `gossip.ts` in the stress-tests creates a regression test.

## followups

## changelog

```markdown
```

## documentation

### DIFF
--- a/crates/sim2h/src/lib.rs
+++ b/crates/sim2h/src/lib.rs
@@ -26,7 +26,10 @@ use cache::*;
 use connection_state::*;
 use lib3h_crypto_api::CryptoSystem;
 use lib3h_protocol::{
-    data_types::{EntryData, FetchEntryData, GetListData, Opaque, SpaceData, StoreEntryAspectData},
+    data_types::{
+        EntryData, EntryListData, FetchEntryData, GetListData, Opaque, SpaceData,
+        StoreEntryAspectData,
+    },
     protocol::*,
     types::SpaceHash,
     uri::Lib3hUri,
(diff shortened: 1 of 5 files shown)


zippy:
stress test won't pass until tryorama updates get released

crates/sim2h/src/lib.rs
```diff
@@ -558,34 +586,15 @@ impl Sim2h {
...
-                }
+                self.handle_unseen_aspects(uri, space_address, agent_id, &list_data);
                 Ok(())
             }
             WireMessage::Lib3hToClientResponse(Lib3hToClientResponse::HandleGetGossipingEntryListResult(list_data)) => {
                 debug!("GOT GOSSIPING LIST from {}", agent_id);
                 if (list_data.provider_agent_id != *agent_id) || (list_data.space_address != *space_address) {
                     return Err(SPACE_MISMATCH_ERR_STR.into());
                 }
+                self.handle_unseen_aspects(uri, space_address, agent_id, &list_data);
```
struktured: This is the line that actually fixes it, correct?

stress-test/src/config.ts
```diff
@@ -52,7 +52,7 @@ const network =
   : networkType === 'sim2h'
   ? {
     type: 'sim2h',
-    sim2h_url: 'ws://localhost:9002'
+    sim2h_url: 'wss://localhost:9002'
```
struktured: :+1:

crates/sim2h/src/lib.rs
zippy: Yep

PR #2012, opened 2019-12-22
core_types: Update to 2018 edition

timotree3:
## PR summary

This PR
- Updates `holochain_core_types`'s `Cargo.toml` to use the 2018 edition of Rust (796c9b5)
- Fixes the broken `use` declarations caused by the changes to namespacing in the 2018 edition (796c9b5)
- Removes most `extern crate`s because the 2018 edition makes them unneeded (796c9b5)
- Deletes the remaining `#[macro_use] extern crate`s and adds `use` declarations for their macros where needed (ecc6d49)
- Updates some lifetime parameters to use the new `Foo<'_>` notation in accordance with 2018 edition idioms (9e95295)

## testing/benchmarking notes

N/A

## followups

I plan to make more of these for all crates that aren't 2018 in this mono-repo.

## changelog

```markdown
```

## documentation


timotree3:
CI is failing because the `use` declarations that I added for the `serde` derives aren't working on that machine. I added `use serde::{Deserialize, Serialize};`, which works on my machine, but it seems that in general for that to work you need to enable the `derive` feature of `serde`.

That leaves two options:
- Replace these instances of `serde::{Deserialize, Serialize}` with `serde_derive::{Deserialize, Serialize}`
- Enable the `derive` feature of `serde`, removing the need for the `serde_derive` dependency altogether.

Personally, I prefer the second approach, as it turns code that looks like

```rust
use serde::{Deserialize, Serialize};
use serde_derive::{Deserialize, Serialize};

#[derive(Deserialize, Serialize)]
struct Foo { /* ... */ }
```

into just

```rust
use serde::{Deserialize, Serialize};

#[derive(Deserialize, Serialize)]
struct Foo { /* ... */ }
```

I'll be pushing a commit that uses said approach.

timotree3:
The current CI failure seems to be cause by a regression in `structopt-derive v0.3.6`. I'm looking into it now.

timotree3:
I opened https://github.com/TeXitoi/structopt/issues/315. In the meantime, I suggest we mitigate this by adding an unused `structopt-derive` dependency with the same version as our `structopt` dependency. I'll push a commit that does this now.

timotree3:
As a matter of fact, the breakage in `structopt-derive v0.3.6` should be affecting the `develop` branch too. I'll make a PR applying the same mitigation I did here.

PR #2029, opened 2019-12-31
Dedicated thread pools for sim2h connection handling and message processing

struktured:
## PR summary

 - Makes all the shared fields of `Sim2h` struct into concurrent, thread safe structures (either a `CHashMap` or a `RwLock` strategy with `Arc<_>` guards.
 - Makes a separate thread pool to processing incoming connections (thus *not* in the processing loop any longer). 

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation


crates/sim2h/src/lib.rs
```diff
@@ -482,15 +601,26 @@ impl Sim2h {
...
+    fn process_internal(&mut self) -> Sim2hResult<()> {
+        trace!("process");
         self.num_ticks += 1;
         if self.num_ticks % 60000 == 0 {
             debug!(".");
             self.num_ticks = 0;
         }
 
-        self.priv_check_incoming_connections();
+        // Now down in separate thread!
```
struktured:
```suggestion
        // Now done in a separate thread!
```

crates/sim2h_server/src/main.rs
```diff
@@ -54,16 +55,18 @@ fn main() {
...
 
-    loop {
-        let result = sim2h.process();
-        if let Err(e) = result {
-            if e.to_string().contains("Bind error:") {
-                println!("{:?}", e);
-                exit(1)
-            } else {
-                error!("{}", e.to_string())
+    for _i in 0..num_cpus::get() {
```
neonphog: ... this doesn't start any threads or anything, and the infinite loop inside cannot exit. `_i` will never advance beyond `0`.

crates/sim2h/src/lib.rs
```diff
@@ -103,27 +108,34 @@ pub enum DhtAlgorithm {
...
+    connection_states: ConnectionStates,
+    spaces: Spaces,
+    pool: Arc<Pool>,
+    wss_recv: crossbeam_channel::Receiver<TcpWss>,
+    msg_send: crossbeam_channel::Sender<(Url2, FrameResult)>,
+    msg_recv: crossbeam_channel::Receiver<(Url2, FrameResult)>,
+    open_connections: OpenConnections,
     num_ticks: u64,
-    /// when should we recalculated the rrdht_arc_radius
+    /// when should we eecalculated the rrdht_arc_radius
```
neonphog:
```suggestion
    /// when we should recalculate the rrdht_arc_radius
```

crates/sim2h/src/lib.rs
```diff
@@ -179,17 +193,29 @@ impl Sim2h {
...
-                .insert(url, (job.clone(), outgoing_send));
-            self.pool.push_job(Box::new(job));
+        for _i in 0..NUM_CONNECTION_THREADS {
+            let connection_states = self.connection_states.clone();
+            let open_connections = self.open_connections.clone();
+            let msg_send = self.msg_send.clone();
+            let pool = self.pool.clone();
+            let wss_recv = self.wss_recv.clone();
+
+            std::thread::spawn(move || loop {
```
neonphog: I guess sim2h is still using that code that panics if a thread panics... but we still might want to capture the join handle off this spawn / clean it up if / when the sim2h instance gets dropped?

crates/sim2h/src/lib.rs
```diff
@@ -998,16 +1181,26 @@ impl Sim2h {
 
         let payload: Opaque = msg.clone().into();
 
-        match self.open_connections.get_mut(&uri) {
-            None => {
-                error!("FAILED TO SEND, NO ROUTE: {}", uri);
-                return;
-            }
-            Some((_con, outgoing_send)) => {
+        let get_result = open_connections.get_mut(&uri);
```
neonphog: I don't think this is going to do what you want... [get_mut](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.get_mut) returns a None if somebody else currently has a lock, in which case it'll print out the NO ROUTE error message below and exit, which isn't right.

neonphog: n/m it calling the get_mut on the map, not the arc : )

crates/sim2h_server/src/main.rs
struktured: Woops, I blame lack of sleep? Fixed.

crates/sim2h/src/lib.rs
struktured: Yes, definitely. Will do in a follow up commit if this code ends up useful in the first place.

struktured:
Important results from this branch, if any, have been incorporated into develop already.

PR #2030, opened 2019-12-31
Sim2h recording & playback of connections and messages

maackle:
## PR summary

Initial hookup for recording key Sim2h events in log files, and then parsing those logs and playing them back using sim2h_client rather than real conductors.

Recording:

- [ ] connect
- [ ] disconnect
- [ ] send wire message

Playback:

- [ ] create sim2h client for connect
- [ ] send message
- [ ] disconnect client

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation


freesig:
This is panicking on `crates/sim2h_client/src/lib.rs:125:19` now. It's getting an error about the connection being closed. Must be something to do with how I pulled in the new connection manager

zippy:
looks like status_check test still needs fixing here.

freesig:
I think this is still failing to connect because it's getting back a closed connection from sim2h and then panicking on `crates/sim2h_client/src/lib.rs:125`. I will try to fix this today.

crates/sim2h_client/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+        Ok(_) => {
+            if let WsFrame::Binary(b) = frame {
+                let msg: WireMessage = serde_json::from_slice(&b).unwrap();
+                Some(msg)
+            } else {
+                panic!("unexpected {:?}", frame);
+            }
+        }
+        Err(e) if e.would_block() => None,
+        Err(e) => panic!(e),
```
freesig: This is the line that is panicking.
I think this whole file needs to be matched to sim2h_client.rs in develop

freesig:
I'm not sure what the next move is with this. Basically the socket is busy from walkman's side then it timesout after 10 seconds. We could raise the timeout but I'm not sure why it's saying it's busy

freesig:
Actually it should probably just try connecting to a new socket if it times out

PR #2033, opened 2020-01-02
Profiling tools

AshantiMutinta:
## PR summary
-Added profiling tools for running perf under conductor and sim2h_server
-Added tools for generating flame graph
## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation

### DIFF
--- a/profiling/default.nix
+++ b/profiling/default.nix
@@ -6,11 +6,16 @@ let
   cd crates/sim2h_server && cargo build -p sim2h_server --release && perf record --call-graph dwarf sim2h_server
   '';
 
+  script-hc-conductor-perf = pkgs.writeShellScriptBin "hc-conductor-perf"
+  ''
+  cd crates/holochain && cargo build -p holochain --release && perf record --call-graph dwarf holochain
+  '';
+
   script-hc-generate-flame-graph = pkgs.writeShellScriptBin "hc-generate-flame-graph"
   ''
   rm -rf FlameGraph && git clone https://github.com/brendangregg/FlameGraph && perf script | perl FlameGraph/stackcollapse-perf.pl | perl FlameGraph/flamegraph.pl > generated-graph.svg
   '';
 in
 {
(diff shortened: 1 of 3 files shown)


thedavidmeister:
this doesn't work until the next version of holonix is cut and included

profiling/default.nix
```diff
@@ -0,0 +1,21 @@
...
+  '';
+
+  script-hc-conductor-perf = pkgs.writeShellScriptBin "hc-conductor-perf"
+  ''
+  cd crates/holochain && cargo build -p holochain --release && perf record --call-graph dwarf holochain
+  '';
+
+  script-hc-generate-flame-graph = pkgs.writeShellScriptBin "hc-generate-flame-graph"
+  ''
+  rm -rf FlameGraph && git clone https://github.com/brendangregg/FlameGraph && perf script | perl FlameGraph/stackcollapse-perf.pl | perl FlameGraph/flamegraph.pl > generated-graph.svg
```
zippy: Is there a reason we need to download the repo every time we generate a flame graph?  This also means we can run this command when we we aren't online.

profiling/default.nix
```diff
@@ -0,0 +1,21 @@
+{ pkgs,config }:
+let
+ name = "hc-sim2h-server-perf";
+  script-sim2h-server-perf = pkgs.writeShellScriptBin name
+  ''
+  cd crates/sim2h_server && cargo build -p sim2h_server --release && perf record --call-graph dwarf sim2h_server
```
zippy: this command needs to pass arbitrary parameters through to sim2h_server because I almost never start it just plain.

profiling/default.nix
AshantiMutinta: was looking at getting the latest version of the repo everytime but will change this around

profiling/default.nix
AshantiMutinta: Will change this

AshantiMutinta:
Wont work without this release 

https://github.com/holochain/holonix/pull/131

PR #2035, opened 2020-01-02
sim2h remove old rrdht calc

neonphog:


neonphog:
@struktured 

> Looks fine, I think there's a dead constant you can remove too at the top of `lib.rs`.

I deleted `RECALC_RRDHT_ARC_RADIUS_INTERVAL_MS` was there another?

PR #2036, opened 2020-01-02
Sim2h use IM crate for state

neonphog:
Switch to using IM for sim2h state in preparation for cloning the state to hand off to parallel handle_message jobs


crates/sim2h/src/cache.rs
```diff
@@ -25,6 +25,27 @@ pub struct Space {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Space")
+            .field("agents", &self.agents)
+            .field("all_aspect_hashes", &self.all_aspects_hashes)
+            .field("missing_aspects", &self.missing_aspects)
+            .finish()
+    }
+}
+
+impl Clone for Space {
```
struktured: What is the difference between this and deriving `Clone`?

crates/sim2h/src/lib.rs
```diff
@@ -349,11 +356,11 @@ impl Sim2h {
...
 
         if let Some(ConnectionState::Joined(space_address, agent_id)) =
-            self.connection_states.write().remove(uri)
+            self.state.connections.remove(uri)
         {
-            if let Some(space_lock) = self.spaces.get(&space_address) {
-                if space_lock.write().remove_agent(&agent_id) == 0 {
-                    self.spaces.remove(&space_address);
+            if let Some(space) = self.state.spaces.get_mut(&space_address) {
+                if space.remove_agent(&agent_id) == 0 {
```
struktured: Is this a modification of the original or a cloned copy?

struktured: Nevermind - I see that there's a clone here at all.

crates/sim2h/src/cache.rs
neonphog: the derive generates code very similar to this... doing it manually lets me do the `box_clone` on the crypto system

crates/sim2h/src/lib.rs
neonphog: I haven't put any code in to clone yet, so all the mut calls modify the original

crates/sim2h/src/cache.rs
AshantiMutinta: There is a rust crate  if we ever wanted to go that route :

https://github.com/dtolnay/dyn-clone

neonphog:
closing - these changes will be included in a later PR

PR #2038, opened 2020-01-03
Backtrace information is only resolved with feature flag

AshantiMutinta:
## PR summary
-Added a feature flag for backtracing during sync
## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation


crates/locksmith/src/tracker.rs
```diff
@@ -87,6 +92,20 @@ Backtrace at the moment of guard creation follows:
...
+        !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+        !!! IMMORTAL LOCK GUARD FOUND !!!
+        !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+{type:?} guard {puid} lived for > {time} seconds.{annotation}",
+            type=self.lock_type,
+            puid=self.puid,
+            time=IMMORTAL_TIMEOUT.as_secs(),
+            annotation=annotation
+        );
```
struktured: What is purpose of this new `error!` block here?

AshantiMutinta: Conditionally compiled, wanted to provide an error block that doesn't mention the backtrace and just having two blocks of code didn't seem too bad the macro + flag could catch them but could always clean this up

crates/locksmith/Cargo.toml
```diff
@@ -18,3 +18,13 @@ lazy_static = "=1.2.0"
 log = "=0.4.8"
 parking_lot ="=0.9.0"
 snowflake = "=1.3.0"
+
+[features]
+#default brings in no features
+default =[]
+
+#this feature allows the compilation with backtraces involved
+sync_backtrace = []
```
neonphog: in tracker.rs, it's `sync_backtrace_capture`

AshantiMutinta: good catch error during renaming :)

AshantiMutinta:
State - not going to be used so closing this

maackle:
I think we should give this a try!

One question @AshantiMutinta why is the feature called `sync_backtrace`? Can we just call it `backtrace` or something?

AshantiMutinta:
> 
> 
> I think we should give this a try!
> 
> One question @AshantiMutinta why is the feature called `sync_backtrace`? Can we just call it `backtrace` or something?

awesome! I think we should call it backtrace I was hoping maybe even "Debug Mode" but backtrace is much better

maackle:
> I feel like we will need to make sure our CI exercises both versions, something like:

@neonphog I feel like CI only needs to exercise the one we'll be using in production (default), since the backtraces one is one we'd only be using for diagnosis, and that is the case already. Thoughts?

neonphog:
@maackle 

> I feel like CI only needs to exercise the one we'll be using in production (default), since the backtraces one is one we'd only be using for diagnosis, and that is the case already. Thoughts?

Yeah, for a debug feature that makes sense to me :+1:

PR #2039, opened 2020-01-05
Sim2h Process State Enhancements -- Mutable/Immutable Call Separation

maackle:
## PR summary

Associates nanoids with entries in the two state HashMaps, and prints out debug info whenever something notable happens to them, so we can track the lifecycle of connections

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation


crates/sim2h/src/lib.rs
```diff
@@ -107,17 +118,18 @@ pub enum DhtAlgorithm {
...
+    open_connections: HashMap<Lib3hUri, OpenConnectionItem>,
     spaces: HashMap<SpaceHash, Space>,
 }
 
+type ConnectionStateItem = (String, ConnectionState);
+type OpenConnectionItem = (
+    String, // uuid
+    Arc<Mutex<ConnectionJob>>,
+    crossbeam_channel::Sender<WsFrame>,
+);
```
neonphog: There's an odd implicit assumption in these structs (that also existed before your changes) that one connection will only ever join one space.

- If we're going to continue with that assumption, should we use the same uuid for both of these? + maybe throw an error if we get a second JoinSpace?
- If we want to allow multiple spaces per connection, should we make ConnectionState actually support that?

zippy: Well, currently that optimization would have to happen across core and sim2h, and would require the conductor to hand the sim2h_worker a sim2h connection, instead of the sim2h_worker initiating that connection.  Given that total connections seems to be one of our bottlenecks we probably ought to fix this, and I'll put it into the tree, but I think we can do that later.

PR #2044, opened 2020-01-06
Keystore default hash config

lucksus:
## PR summary

This makes the `Keystore` default to the same low-key testing passphrase hashing config the conductor defaults to. This is to render direct users of this struct (such as helpers in hpos) compatible with the default of the conductor.

## changelog

```markdown
```

## documentation


crates/conductor_lib/src/keystore.rs
```diff
@@ -119,8 +119,11 @@ impl Keystore {
     /// This will query `passphrase_manager` immediately to set a passphrase for the keystore.
     pub fn new(
         passphrase_manager: Arc<PassphraseManager>,
-        hash_config: Option<PwHashConfig>,
+        mut hash_config: Option<PwHashConfig>,
     ) -> HcResult<Self> {
+        if hash_config.is_none() {
+            hash_config = test_hash_config()
+        }
```
AshantiMutinta: I feel like we could just get rid of this line of code? Seems to be doing the same thing line 163 is doing. That way hash_config doesn't have to be mutable as well

AshantiMutinta: if we do let hash_config = hash_config.or_else(||test_hash_config) we could get rid of that mutable

AshantiMutinta: (also does this mean every invocation of this will result in using a test_hash if it is not provided?)

lucksus: Yes, but since we use `hash_config` twice, we would either need to call the .or_else twice or create another variable. I thought this would be the best option..

zippy: If we change the name it should be to `fast_unsecure_hash` not default.

PR #2051, opened 2020-01-07
Goodies from sim2h-process-enhance

zippy:
## PR summary

This PR mostly adds debugging that's useful but does add one substantive thing which looping on all pending messages in the recv channel instead of handling just one at a time.

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation

### DIFF
--- a/crates/sim2h/src/lib.rs
+++ b/crates/sim2h/src/lib.rs
@@ -86,7 +86,7 @@ impl<T> SendExt<T> for crossbeam_channel::Sender<T> {
     }
 }
 
-const RETRY_FETCH_MISSING_ASPECTS_INTERVAL_MS: u64 = 10000; // 10 seconds
+const RETRY_FETCH_MISSING_ASPECTS_INTERVAL_MS: u64 = 30000; // 30 seconds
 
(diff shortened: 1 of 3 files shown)


crates/sim2h/src/lib.rs
```diff
@@ -192,17 +192,21 @@ impl Sim2h {
...
     }
 
     /// if our connections sent us any data, process it
     fn priv_check_incoming_messages(&mut self) {
-        if let Ok((url, msg)) = self.msg_recv.try_recv() {
+        let len = self.msg_recv.len();
+        if len > 0 {
+            debug!("Handling {} incoming messages", len);
+        }
+        let v: Vec<_> = self.msg_recv.try_iter().collect();
```
neonphog: might it make sense to add a `take(100)` or so here, to make sure we're not starving other jobs on low cpu-count machines?

neonphog: i guess this is still our single thread bottleneck, so doesn't matter as much here

zippy: I was guessing this would go away in what you are doing, so I just left it how I had it in the process-enhance branch

PR #2054, opened 2020-01-08
Parked: Get links entry requests on requesting node

willemolding:
## PR summary

Previously when making a call to `get_links` the node holding the link base (with the addresses of the `LinkAdd` entries as aspects) was also responsible for retrieving the LinkAdd entries themselves. This is how the target and CRUD status is obtained.

This made sense in a non-sharding context since both nodes almost certainly hold the entries and no extra work is required. With sharding however there is a good chance getting all the LinkAdd entries will require network requests and take some time.

It therefore seems much more fair for the requesting node to do this work instead of the base holder. This is especially important for scaling the number of links. Even when the number of aspects on a base is large, retrieving them from the EAV should still be a cheap operation.

- Current core tests are passing but there are some untested edge cases (link updates/deletes) that I'm pretty confident won't work

## changelog

```markdown
```

## documentation


zippy:
Here's an error from a final exam run:

```
thread 'ZomeFnCall { id: ProcessUniqueId { prefix: 10, offset: 632 }, zome_name: "transactions", cap: CapabilityRequest { cap_token: Hash
String("QmV1NgkXFwromLvyAmASN7MbgLtgUaEYkozHPGUxcHAbSL"), provenance: Provenance(HashString("HcScjYAI4OQ8Gssm6tyNJHYNKktfo8sceEWU35NxDi4V
j579XX5HjAAICfdp4pr"), Signature("3fvvJ5a3WPysoCKf3F4hH7jwE9SgAIEoiaOJgXfy9+sd6ujI+tZdfe+EuBUnuWITZGN1rEVqxD+Gv9MVIhDgDw==")) }, fn_name:
 "list_pending", parameters: JsonString("{}") }' panicked at 'called `Result::unwrap()` on an `Err` value: ErrorGeneric("Could not find E
ntries for Address: QmQyhRn2dKqwUvye9fyxqjvSJxdKtgD2BeaxmjTytsgS4Q, tag: ")', src/libcore/result.rs:1187:5
stack backtrace:
   0:     0x560463377b84 - backtrace::backtrace::libunwind::trace::hf6fab806f99173ae
                               at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/libunwind.rs:88
   1:     0x560463377b84 - backtrace::backtrace::trace_unsynchronized::h9d880e161d3dd682
                               at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/mod.rs:66
   2:     0x560463377b84 - std::sys_common::backtrace::_print_fmt::h7fa7703959a17af0
                               at src/libstd/sys_common/backtrace.rs:84
   3:     0x560463377b84 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::he957ece4cedb167a
                               at src/libstd/sys_common/backtrace.rs:61
   4:     0x5604633a1e1c - core::fmt::write::h58c6b2373fe36ade
                               at src/libcore/fmt/mod.rs:1030
   5:     0x5604633703c7 - std::io::Write::write_fmt::h60db73a9329f6f82
                               at src/libstd/io/mod.rs:1412
   6:     0x56046337a4ae - std::sys_common::backtrace::_print::h229cbaa1a1c4ad99
                               at src/libstd/sys_common/backtrace.rs:65
   7:     0x56046337a4ae - std::sys_common::backtrace::print::h7f2c9bdc49f4ce67
                               at src/libstd/sys_common/backtrace.rs:50
   8:     0x56046337a4ae - std::panicking::default_hook::{{closure}}::h70a5919b862fb68b
                               at src/libstd/panicking.rs:188
   9:     0x56046337a1a1 - std::panicking::default_hook::hccd427c2c397d85b
                               at src/libstd/panicking.rs:205
  10:     0x56046337abab - std::panicking::rust_panic_with_hook::h5b585879e33d0e2b
                               at src/libstd/panicking.rs:464
  11:     0x56046337a74e - std::panicking::continue_panic_fmt::h01f40c7ab9f22e43
                               at src/libstd/panicking.rs:373
  12:     0x56046337a636 - rust_begin_unwind
                               at src/libstd/panicking.rs:302
  13:     0x56046339e4be - core::panicking::panic_fmt::hd65e208a6ca5e8e9
                               at src/libcore/panicking.rs:82
  14:     0x56046339e5b7 - core::result::unwrap_failed::h2dd6a14bf5b119db
                               at src/libcore/result.rs:1187
  15:     0x560462a43e37 - <core::iter::adapters::Map<I,F> as core::iter::traits::iterator::Iterator>::fold::hbd80ef60db3a7586
  16:     0x560462a43a7b - <core::iter::adapters::Map<I,F> as core::iter::traits::iterator::Iterator>::fold::hb842a7e4ee2746ff
  17:     0x560462a02976 - <alloc::vec::Vec<T> as alloc::vec::SpecExtend<T,I>>::spec_extend::hd206b7bb91aeb2ef
  18:     0x560462a031d4 - <alloc::vec::Vec<T> as alloc::vec::SpecExtend<T,I>>::from_iter::h141ebb9c6142cdda
  19:     0x560462a9e00e - <std::future::GenFuture<T> as core::future::future::Future>::poll::h125b17d9451eea9a
  20:     0x560462a4f796 - holochain_core::context::Context::block_on::h4e4d059959189680
  21:     0x560462b38116 - holochain_core::nucleus::ribosome::api::get_links::invoke_get_links::h2a32fb094ff2a812
```

crates/core/src/workflows/get_link_result.rs
```diff
@@ -36,6 +41,16 @@ pub async fn get_link_result_workflow<'a>(
...
                 .into_iter()
+                .map(|(link_add_address, tag)| {
+                    // make DHT calls to get the entries for the links
+                    get_link_data_from_link_addresses(
+                        context,
+                        &link_add_address,
+                        &tag,
+                        link_args.options.headers,
+                    )
+                    .unwrap()
```
zippy: This can't be an unwrap!  I'm pretty sure that's where the panic I was was comming from.

crates/core/src/workflows/get_link_result.rs
```diff
@@ -34,20 +39,101 @@ pub async fn get_link_result_workflow<'a>(
...
-            Ok(GetLinksResult::new(get_links_result))
+                .map(|maybe_get_entry_crud| {
+                    maybe_get_entry_crud.map(|get_entry_crud| LinksResult {
+                        address: get_entry_crud.target.clone(),
+                        headers: get_entry_crud.headers.unwrap_or_default(),
+                        status: get_entry_crud.crud_status,
+                        tag: get_entry_crud.tag.clone(),
+                    })
+                })
+                .collect::<Result<Vec<LinksResult>, HolochainError>>()
```
willemolding: Cool rust trick! If you collect an iterator of `Results` into a `Result<Vec<_>>` it will abort on the first `Err` and return that and only return `Ok` if none of the elements are `Err`

willemolding: This is where we can also decide how we want to handle failures to retrieve `LinkAdd` entries. Options are:

- Fail if any of them cannot be retrieved (current implementation, fits best with existing code)
- Filter out failures and only return successes
- get_links returns a vec of results (or possible separate vecs for errors and results)

pdaoust:
Holding for the app dev perspective: I agree that retrieving targets should be the requestor's responsibility, and now that I'm aware that only the _address_ of the `LinkAdd` is stored on the base, it seems to make sense by extension to get the requestor to retrieve the `LinkAdd`s too. Question: Does this affect the ability to do efficient filtering and sorting (and, by extension, pagination)?

app_spec/test/files/links.js
```diff
@@ -77,6 +77,7 @@ module.exports = scenario => {
     t.ok(error.line)
   })
 
+// this is where the failing tests are
```
willemolding:
```suggestion
```

crates/core/src/workflows/hold_link.rs
```diff
@@ -80,7 +80,7 @@ pub async fn hold_link_workflow(
 
     log_debug!(context, "workflow/hold_link: added! {:?}", link);
 
-    //4. store link_add entry so we have all we need to respond to get links queries without any other network look-up
+    //4. store link_add entry so we have all we need to respond to get links queries without any other network look-up <== THIS!
```
willemolding: So as Nico mentioned this morning there is some logic for nodes to hold the LinkAdd entries for links on bases they hold!

I guess this kinda renders this whole PR redundant. It will actually result in more networks requests overall (assuming this logic for holding the LinkAdd entries actually works...)

willemolding:
> Holding for the app dev perspective: I agree that retrieving targets should be the requestor's responsibility, and now that I'm aware that only the _address_ of the `LinkAdd` is stored on the base, it seems to make sense by extension to get the requestor to retrieve the `LinkAdd`s too. Question: Does this affect the ability to do efficient filtering and sorting (and, by extension, pagination)?

Yep from my current understanding it prevents an implementation of any of those things :(. This was raised in our standup this morning.

Solutions:
- We put it back to how it was and the remote node must load all the `LinkAdd` s, sort them, filter and possibly paginate. It seems like a lot of work but I'm not sure if it really is in the scheme of things. Plus some new info has come to light that actually nodes store the `LinkAdd` entries for links they hold so at least no network requests... If this is a good thing or not it is yet to be decided.
- Add more data to the aspects that are stored on the base so there enough to sort and paginate without loading the entries

AshantiMutinta:
I think we could also employ laziness in this in that, if the eavi is lazy and returns an iterator we could bubble that up to the get_links and filtering operations could be done on that iterator. Paginations at that point would just be a matter of iter().take(20)

pdaoust:
@willemolding thanks for taking the time to explain and give some of the larger context that isn't reflected in this PR :+1:

@AshantiMutinta you have my vote of support for that :100: as a .NET alumnus, I have a special fondness for [LINQ](https://en.wikipedia.org/wiki/Language_Integrated_Query), which was able to compile query expressions down to native data access code (SQL or what-have-you) and was kind of the ultimate in laziness.

crates/core_types/src/network/query.rs
```diff
@@ -13,6 +13,14 @@ pub enum GetLinksNetworkQuery {
     Links(GetLinksQueryConfiguration),
 }
 
+// The data returned by a remote node on responding to a get_links query
+#[derive(Debug, Serialize, Deserialize, PartialEq, DefaultJson, Clone)]
+pub struct GetLinkFromRemoteData {
+    pub link_add_address: Address,
```
zippy: From what I can tell this name is probably not correct, because this is either the address of the entry that added the link or the address of the entry that removed it in the case that the link was removed and the `LinksStatusRequestKind` was `All` or `Deleted`  Is that correct?

If so then we should rename this to `linking_action_address` or something else that makes it clear that this field can hold the address of either type of entry.

crates/core/src/workflows/get_link_result.rs
```diff
@@ -36,20 +44,110 @@ pub async fn get_link_result_workflow<'a>(
...
+                    .map(|single_entry| match single_entry {
+                        Entry::LinkAdd(link_add) => Ok(GetLinkData::new(
+                            link_add_address.clone(),
+                            crud,
+                            link_add.link().target().clone(),
+                            tag.clone(),
+                            maybe_entry_headers,
+                        )),
+                        Entry::LinkRemove(link_remove) => Ok(GetLinkData::new(
+                            link_add_address.clone(),
```
zippy: see comment below regarding the confusing nature of this variable name.

app_spec/test/files/links.js
zippy: I think this comment can be removed now?

crates/core/src/network/mod.rs
```diff
@@ -341,15 +341,15 @@ pub mod tests {
...
-            ((links[0].address.clone(), links[0].crud_status.clone())
-                == (entry_addresses[1].clone(), CrudStatus::Live)
-                || (links[0].address.clone(), links[0].crud_status.clone())
-                    == (entry_addresses[2].clone(), CrudStatus::Live))
-                && ((links[1].address.clone(), links[0].crud_status.clone())
-                    == (entry_addresses[1].clone(), CrudStatus::Live)
-                    || (links[1].address.clone(), links[0].crud_status.clone())
-                        == (entry_addresses[2].clone(), CrudStatus::Live))
-        );
+        // assert!( // cant check this right now because of the logic being moved
```
zippy: Can we clean up these comments and actually check what is correct to check here?

crates/core/src/workflows/hold_link.rs
zippy: This is a possible future enhancement, but there are other problems getting this to work correctly, so for now we will merge this PR so that we can finish pagination which needs to be built on top of it.

artbrock:
> Currently (=as implemented in develop), _holding a LinkAdd aspect_ means both:
> 
> * storing an [EAVi](https://github.com/holochain/holochain-rust/blob/06b1d29d04aa4c508d2a2b3da5972aee1c53a143/crates/core/src/workflows/hold_link.rs#L81) that connects the LinkAdd entry with the base, and
> * storing the [LinkAdd entry](https://github.com/holochain/holochain-rust/blob/06b1d29d04aa4c508d2a2b3da5972aee1c53a143/crates/core/src/workflows/hold_link.rs#L86) at the DHT itself.
> 
> This enables us to respond to a `get_links` query with just one network round-trip since the node holding the base also holds all needed information to render the response.

Why would the requesting node of a get_links ever need the LinkAdd entries (with headers). They just need need the list of type/tag/targets.

This seems to conflate storing the LinkAdd entry with storing the TARGET of link. We don't need to return the link add entry, we need to return either the list of targets, or the pre-loaded targets. Just because a base node also stores the LinkAdd entries, doesn't mean it stores all the link target entries. My understanding is that waiting for the node to assemble all the target entries is what is creating the performance issue, and it should not be a base nodes responsibility to do this for an unlimited number of targets... and this seems like an especially bad idea to do before we have caching.

pdaoust:
you know, something that comes to mind as I read this discussion is that the HTTP/2 spec has this thing whereby the server can say "hey dude, I see you're requesting resource A; I'm pretty sure you're also gonna need resources B, C, and D -- I'm gonna push them to you unless you say otherwise". That might be useful here, in that the spec wouldn't have to define whether the target or `LinkAdd` entries are required in the response; you just let the responding agent push them if (a) it has them and (b) the requesting agent doesn't object. It would become a matter of optimisation rather than protocol.

Note: as an app dev I would definitely want to receive not just the target address but also the link type and tag -- because you can filter links by regex, and the contents of the tag might be useful to the requestor.

lucksus:
> Why would the requesting node of a get_links ever need the LinkAdd entries (with headers). They just need need the list of type/tag/targets.

The LinkAdd entry is what contains type, tag and target.
The requesting node also (potentially) needs the header of the LinkAdd(s) because our HDK has a [get_links_with_options()](https://github.com/holochain/holochain-rust/blob/0faafde6bd59a4908e915066deb53a541602503a/crates/hdk/src/api/get_links.rs#L41) which can be configured to return the headers of the links. [hdk::get_links()](https://github.com/holochain/holochain-rust/blob/0faafde6bd59a4908e915066deb53a541602503a/crates/hdk/src/api/get_links.rs#L103) is actually just a short-cut with default options:

```rust
pub fn get_links(
    base: &Address,
    link_type: LinkMatch<&str>,
    tag: LinkMatch<&str>,
) -> ZomeApiResult<GetLinksResult> {
    get_links_with_options(base, link_type, tag, GetLinksOptions::default())
}

pub fn get_links_with_options(
    base: &Address,
    link_type: LinkMatch<&str>,
    tag: LinkMatch<&str>,
    options: GetLinksOptions,
) -> ZomeApiResult<GetLinksResult> {
...
}

pub struct GetLinksResult {
    links: Vec<LinksResult>,
}

pub struct LinksResult {
    pub address: Address,
    pub headers: Vec<ChainHeader>,
    pub tag: String,
    pub status: CrudStatus,
}
```
https://github.com/holochain/holochain-rust/blob/0faafde6bd59a4908e915066deb53a541602503a/crates/wasm_utils/src/api_serialization/get_links.rs#L42

> This seems to conflate storing the LinkAdd entry with storing the TARGET of link. We don't need to return the link add entry, we need to return either the list of targets, or the pre-loaded targets. Just because a base node also stores the LinkAdd entries, doesn't mean it stores all the link target entries. My understanding is that waiting for the node to assemble all the target entries is what is creating the performance issue, and it should not be a base nodes responsibility to do this for an unlimited number of targets... and this seems like an especially bad idea to do before we have caching.

**The base nodes are not fetching targets.** They always only return the contents of the (locally stored) LinkAdd, potentially with headers depending on query options. The struct that is sent back is [GetLinkData](https://github.com/holochain/holochain-rust/blob/0faafde6bd59a4908e915066deb53a541602503a/crates/core_types/src/network/query.rs#L17), which is almost identical to `LinksResult` above (with the only difference that `GetLinkData` includes `address` (here: base) *and* `target`. `LinksResult` *only includes* `address` which is the target since the base is a parameter to `hdk::get_links()`.

Looking at this code, I've spotted a stale comment that might be partially be responsible for this confusion. In the `get_links` handler, it says:
https://github.com/holochain/holochain-rust/blob/0faafde6bd59a4908e915066deb53a541602503a/crates/core/src/network/handler/query.rs#L55)

I believe, that was an adequate comment in the very first implementation of this code where we *didn't* store the LinkAdds and just stored EAV items that directly matched the link: i.e. storing the target address as *value* on the base *entity* with the tag as *attribute*. The problem we encountered with this naive approach was above mentioned headers. Creating an extra EAV between base and LinkAdd header creates more complexity than just falling back on the fact that links originate from LinkAdd entries that already ontologically have a header.

lucksus:
Just for completion: to my knowledge, pre-loading targets doesn't happen anywhere in core. For that purpose we added a convenience function to the HDK: [hdk::get_links_and_load()](https://github.com/holochain/holochain-rust/blob/0faafde6bd59a4908e915066deb53a541602503a/crates/hdk/src/api/get_links.rs#L150).

PR #2055, opened 2020-01-09
Sim2h dedup fix

zippy:
## PR summary

We were inserting multiple copies of aspects into the aspect list.  This PR fixes that and adds a test. 

## testing/benchmarking notes

## followups

We really should be merging the AspectList with the AspectMap and doing so with benchmarking, but it's not immediately trivial because EntryData that goes over the wire is serialied as a Vec not a map so there's some from munging to do.

## changelog

```markdown
```

## documentation

### DIFF
--- a/crates/sim2h/src/cache.rs
+++ b/crates/sim2h/src/cache.rs
@@ -306,6 +306,32 @@ mod tests {
     use std::convert::TryFrom;
 
     #[test]
+    fn aspect_list_holds_aspects() {
+        let mut list = AspectList::from(HashMap::new());
+        assert_eq!(list.pretty_string(), "");
+        let entry_hash = EntryHash::from("entry_hash_1");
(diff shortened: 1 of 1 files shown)


crates/sim2h/src/cache.rs
```diff
@@ -208,10 +208,10 @@ impl AspectList {
     }
 
     pub fn add(&mut self, entry_address: EntryHash, aspect_address: AspectHash) {
-        self.0
-            .entry(entry_address)
-            .or_insert_with(Vec::new)
-            .push(aspect_address);
+        let list = self.0.entry(entry_address).or_insert_with(Vec::new);
+        if !list.contains(&aspect_address) {
```
neonphog: this is going to get slower as the list grows, do we want to convert to HashSet? - cool if that's an optimization for later

struktured: Agreed- just make it a hashset- way more efficient.

PR #2057, opened 2020-01-09
Sim2h multi send

zippy:
## PR summary

adds aggregation of many Lib3h messages into a single sim2h message for efficiency.

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation

### DIFF
--- a/crates/net/src/sim2h_worker.rs
+++ b/crates/net/src/sim2h_worker.rs
@@ -456,7 +456,7 @@ impl Sim2hWorker {
             WireMessage::Pong => {}
             WireMessage::Lib3hToClient(m) => self.to_core.push(Lib3hServerProtocol::from(m)),
             WireMessage::MultiSend(messages) => {
-                for m in messages {
+                for m in messages.into_iter() {
                     self.to_core.push(Lib3hServerProtocol::from(m));
                 }
             }
--- a/crates/net/src/sim2h_worker.rs
+++ b/crates/net/src/sim2h_worker.rs
@@ -455,6 +455,11 @@ impl Sim2hWorker {
             WireMessage::Ping => self.send_wire_message(WireMessage::Pong)?,
             WireMessage::Pong => {}
             WireMessage::Lib3hToClient(m) => self.to_core.push(Lib3hServerProtocol::from(m)),
+            WireMessage::MultiSend(messages) => {
+                for m in messages {
+                    self.to_core.push(Lib3hServerProtocol::from(m));
+                }
+            }
             WireMessage::ClientToLib3hResponse(m) => {
                 self.to_core.push(Lib3hServerProtocol::from(m))
             }
--- a/crates/net/src/sim2h_worker.rs
+++ b/crates/net/src/sim2h_worker.rs
@@ -23,7 +23,7 @@ use lib3h_protocol::{
 use log::*;
 use sim2h::{
     crypto::{Provenance, SignedWireMessage},
-    TcpWss, WireError, WireMessage,
+    TcpWss, WireError, WireMessage, WIRE_VERSION,
 };
 use std::{convert::TryFrom, time::Instant};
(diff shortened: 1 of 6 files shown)


crates/net/src/sim2h_worker.rs
```diff
@@ -455,6 +455,11 @@ impl Sim2hWorker {
             WireMessage::Ping => self.send_wire_message(WireMessage::Pong)?,
             WireMessage::Pong => {}
             WireMessage::Lib3hToClient(m) => self.to_core.push(Lib3hServerProtocol::from(m)),
+            WireMessage::MultiSend(messages) => {
+                for m in messages {
```
struktured:
```suggestion
                for m in messages.into_iter() {
```

struktured: I think this would be a bit more efficient memory wise- @neonphog do you agree?

crates/sim2h/src/cache.rs
```diff
@@ -208,10 +208,10 @@ impl AspectList {
     }
 
     pub fn add(&mut self, entry_address: EntryHash, aspect_address: AspectHash) {
-        self.0
-            .entry(entry_address)
-            .or_insert_with(Vec::new)
-            .push(aspect_address);
+        let list = self.0.entry(entry_address).or_insert_with(Vec::new);
```
struktured: I still think we should just change this type to a hashset.

zippy: I would like to very much, but it's not just a trivial change because it flows through to a lot of things..

crates/sim2h/src/lib.rs
```diff
@@ -539,6 +555,25 @@ impl Sim2h {
                 );
                 return Ok(());
             }
+            if let WireMessage::Hello(version) = message {
+                debug!("Sending HelloResponse in response to Hello");
```
neonphog: print out the version they sent in the debug?

crates/sim2h/src/lib.rs
```diff
@@ -1123,17 +1166,42 @@ impl Sim2h {
...
+                    version,
+                    uuid,
+                    job: _,
+                    sender: outgoing_send,
+                }) => {
                     open_lifecycle("send", uuid, &uri);
-                    if let Err(_) = outgoing_send.send(payload.as_bytes().into()) {
-                        self.disconnect(&uri);
+
+                    if (version > &mut 1)
```
neonphog: I have pretty strong reservations about forking our version support while we're still in a highly iterative development cycle. The debugging complexity could easily / quickly become untenable if we have to rationalize how different versions are interacting.

My vote would be to:

- capture the Cargo.toml version as a constant within the sim2h lib.
- sim2h_worker and hc_cli can pass that constant in when connecting to sim2h_server
- sim2h_server will reject any connections from a version that doesn't match

neonphog: If we're worried about breaking alpha tester patterns, we could potentially include this version in the url scheme for sim2h servers, so we could keep one or two previous versions running as people upgrade / transition.

zippy: I agree with this in general, however right now it kills holoscape 0.0.7 without having this check.  Happy to drop the check and code quickly.   Also, the problem with the Cargo.toml version may be moving too slowly?

neonphog: Yeah, if this is an emergency hotfix to break something, all good :+1:

I'm mostly concerned with mitigating bugs reported from the wild via the alpha test group... in which case they should only ever get new code when a Cargo.toml version is bumped, right?

If we also want to make sure devs aren't confused for CI runs / etc... one thing we did at TWC in our publish script was have it commit an additional version bump to the develop branch that had an odd numbered patch version, so we knew those were dev versions, and only released versions with even patch versions.

PR #2058, opened 2020-01-09
Sim2h extra tests

maackle:
## PR summary

Attempting to uncover logical errors in sim2h routing to explain the get_links timeout we see.

This probably needs to be rebased onto the appropriate branch, whatever that may be.


maackle:
Totally was looking at this the wrong way; nevermind

PR #2068, opened 2020-01-15
add string to Timeout for debugging

zippy:
## PR summary

Adds a string to timeout errors locate where they originate.

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation


freesig:
You should pass in 
```rust
format!("{}:{}", file!(), line!())
```
instead so it's easy to find.
In fact what we could do that might be even better is to make a proc_macro that constructs the error with the location so it could be like:
```
timeout!(())
```
Which could use the Span of the `()` to automatically get the file and line number

PR #2072, opened 2020-01-20
Sim2h Futures Attempt #4

neonphog:
Use futures and the Tokio threadpool runtime executor for the message handling half of the Sim2h server.

The tokio runtime gives us additional benefits above the raw futures crate in terms of task delay / scheduling and dedicated blocking task handling.

The raw TCP/Websocket side is still using my hand-rolled threadpool - at some point I'd like to spike switching to [tokio-tungstenite](https://crates.io/crates/tokio-tungstenite).


crates/sim2h/src/job/connection.rs
```diff
@@ -73,22 +73,24 @@ impl ConnectionJob {
...
-                trace!("frame read from {} {:?}", self.wss.remote_url(), frame);
-                self.report_msg(Ok(frame));
-                // we got data this time, check again right away
-                return Ok(JobResult::default());
-            }
-            Err(e) if e.would_block() => (),
-            Err(e) => {
-                error!("error in read for {}: {:?}", self.wss.remote_url(), e,);
-                return Err(e.into());
+        {
```
zippy: Is this block here because something was not being dropped that should have been?  Do we still need it?  If so a comment would be good.

crates/sim2h/src/lib.rs
```diff
@@ -684,128 +746,330 @@ impl Sim2hState {
...
-                        return true; //did work despite error.
-                    }
                     let uuid = nanoid::simple();
                     open_lifecycle("adding conn job", &uuid, &url);
-                    self.state.write().open_connections.insert(
+
+                    let (job, outgoing_send) = ConnectionJob::new(wss, msg_send.clone());
+                    let job = Arc::new(Mutex::new(job));
+
+                    //job_send.send(Box::new(job.clone())).expect("send fail");
```
zippy: why this commented out code?

crates/sim2h/src/lib.rs
```diff
@@ -684,128 +746,330 @@ impl Sim2hState {
...
+        for _ in 0..100 {
+            if let Ok(wss) = self.wss_recv.try_recv() {
+                did_work = true;
+                wss_list.push(wss);
+            } else {
+                break;
+            }
+        }
+        if !wss_list.is_empty() {
+            //let job_send = self.pool.get_push_job_handle();
```
zippy: ditto to comment?

crates/sim2h/src/lib.rs
```diff
@@ -684,128 +746,330 @@ impl Sim2hState {
...
+                                .await;
+                            }
+                        }
+                    });
+
+                    state
+                        .connection_states
+                        .insert(url.clone(), (nanoid::simple(), ConnectionState::new()));
+
+                    state.open_connections.insert(
```
zippy: Is there a possibility that spawning the Job into a tokio task before it's been added the `open_connections` list could cause a problem?  I.e. could it complete and try to remove itself from the open_connections list before it was added, or could it require a check on the open_connections list also before it's been added?

crates/sim2h/src/job/connection.rs
```diff
@@ -73,22 +73,24 @@ impl ConnectionJob {
...
+            match self.wss.read(self.frame.as_mut().unwrap()) {
+                Ok(_) => {
+                    trace!(
+                        "run_result: read from {} in connection",
+                        self.wss.remote_url()
+                    );
+                    let frame = self.frame.take().unwrap();
```
struktured: I know this isn't part of your PR but why do we have to `unwrap()` the `take()`?

crates/sim2h/src/naive_sharding.rs
```diff
@@ -79,13 +81,17 @@ mod tests {
...
+        println!("starting test");
+
         let thread_cont = std::sync::Arc::new(std::sync::Mutex::new(true));
         let mut hash_threads = Vec::new();
 
         let (id_send, id_recv) = crossbeam_channel::bounded::<Location>(20);
         let (addr_send, addr_recv) = crossbeam_channel::bounded::<Location>(100);
 
         for _ in 0..8 {
+            println!("starting hash thread");
```
struktured: Would be easy to switching to a `holochain_logger` `debug!` statement, as well as ones below it. (Yes you need that annoying `enable_logging_for_test` hack).

crates/sim2h/src/lib.rs
```diff
@@ -1188,211 +1484,284 @@ impl Sim2h {
...
-                        let disconnects =
-                            state
-                            .read()
-                            .build_query(space_address,query_data,redundant_count);
-                        tx.send(PoolTask::Disconnect(disconnects))
-                            .expect("should send");
+                    let sim2h_handle = self.sim2h_handle.clone();
+                    tokio::task::spawn(async move {
+                        let disconnects = sim2h_handle
+                            .lock_state().await
```
struktured: Possibly good example here where a future rwlock implementation that @AshantiMutinta brought up in standup would benefit us.

crates/sim2h/Cargo.toml
```diff
@@ -17,6 +17,7 @@ hcid = "=0.0.6"
 holochain_persistence_api = "=0.0.13"
 holochain_json_api = "=0.0.17"
 detach = "=0.0.19"
```
struktured:
```suggestion
```

struktured: (Detach is no longer used by sim2h)

crates/sim2h/src/job/connection.rs
neonphog: the take api returns an option, because the value could already have been taken, and so it'll be `None` -- since we checked above if it was None add an instance to it, we can unwrap safely here

crates/sim2h/src/lib.rs
```diff
@@ -684,128 +746,330 @@ impl Sim2hState {
...
+        self.metric_gen.timer(tag)
+    }
+
+    /// get our current dht algorithm
+    pub fn dht_algorithm(&self) -> &DhtAlgorithm {
+        &self.dht_algorithm
+    }
+
+    /// acquire a mutex lock to our state data
+    pub async fn lock_state(&self) -> tokio::sync::MutexGuard<'_, Sim2hState> {
```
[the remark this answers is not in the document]
neonphog: this is a little weird syntactically - since this itself is an async function and we're awaiting within it, it works out that the calling function will await on the lock being acquired.

It's true you need to be careful that that guard goes out of scope before any more awaits in the using function, but that's just something you have to be careful about in any rust code involving futures...

crates/sim2h/src/job/connection.rs
neonphog: woops, debugging I forgot to remove :+1:

crates/sim2h/src/lib.rs
neonphog: The only thing the job does as a separate task (this was true before this PR as well) is publish messages and errors from reading / writing on the socket. Those messages don't get polled until another call to process happens...

I don't *think* it should make any difference... but it also shouldn't hurt to swap the ordering here, so I'll go ahead and do that.

crates/sim2h/src/lib.rs
neonphog: agreed :+1: - but I'm going to leave as-is for now, I think the IM swapout will obviate this

crates/sim2h/src/lib.rs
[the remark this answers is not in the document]
neonphog: sorry -- re-read your comment, your concern is about not doing this in the first place so you don't have to be careful about it.

Agree that would be preferable - I was super careful about all the current uses making sure they are dropped appropriately and no `await`s happen in the mean time.

If the IM state stuff works out how I'm hoping I'll be able to get rid of this function / these MutexGuards entirely.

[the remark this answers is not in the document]
lucksus: Yes, agree with the last statement. As per discussion the stand-up today, this function will go away when implementing the IM/actor/internal mutability stuff.

Ok, fine with merging this for now.

Issue #2069, opened 2020-01-16
acorn-hc hits a conductor panic, based on link_eavs.len()


Connoropolous:
https://github.com/holochain/holochain-rust/blob/45e3d89aa179c17105b91de8c72c175cbd512610/crates/core/src/nucleus/actions/get_entry.rs#L86-L90

we are running v0.0.42-alpha5. The Conductor of the second node to activate panics when we connect two nodes to a sim2h_server

<img width="570" alt="Screen Shot 2020-01-16 at 11 29 41 AM" src="https://user-images.githubusercontent.com/1409121/72558597-de3be980-3857-11ea-8afa-d86e97c8f074.png">

Connoropolous:
Here is a new dump of the log with the full backtrace, and the full log from launch.

https://gist.github.com/Connoropolous/16f27cb573330ce47ab9b9e197ddc929

zippy:
I think this has been fixed now no?

Connoropolous:
yeah no longer an issue

PR #2075, opened 2020-01-21
get_entry panic debugging

Connoropolous:
## PR summary

This assert created a panic in the Conductor, during recent real-world testing with the Acorn app. @thedavidmeister recommended that I open a PR with the assert removed, to open the topic for discussion. 

The issue was logged at #2069 

There is some interesting looking context around this, like:
https://github.com/holochain/holochain-rust/pull/2075/files#diff-01676d0abf6516ad0140d154d6e4ba20L51-L52
" // TODO waiting for update/remove_eav() assert!(status_eavs.len() <= 1);"
"// For now look for crud-status by life-cycle order: Deleted, Modified, Live"

and the fact that in the file tests it says: 
"/*
    /*
        #[test]
        fn test_get_entry_from_agent_chain() {
          // write this test when its easier to get a mutable agent state
        }
        }
    *
"

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation

### DIFF
--- a/crates/core/src/nucleus/actions/get_entry.rs
+++ b/crates/core/src/nucleus/actions/get_entry.rs
@@ -83,11 +83,6 @@ pub(crate) fn get_entry_crud_meta_from_dht(
         IndexFilter::LatestByAttribute,
         None,
     ))?;
-    assert!(
-        link_eavs.len() <= 1,
-        "link_eavs.len() = {}",
-        link_eavs.len()
-    );
     if link_eavs.len() == 1 {
         maybe_link_update_delete = Some(link_eavs.iter().next().unwrap().value());
     }


Connoropolous:
all the tests are passing, that's a good sign

Connoropolous:
thanks @maackle for getting in here on this. Yeah, obviously something is amiss in the code, because I got "link_eavs.len() = 2" as the error message, which is why it panicked, but that means that something is not as expected in the way that information was propagating or being held.

AshantiMutinta:
hai @maackle @Connoropolous I think for some strange reason it is stating that the address provided has two crud_links instead of 1 which means it doesn't know which one to get? The trail of crud_links doesn't change as soon as it is attatched so I hope we can fix this just by relying on that the lines 86 to 87 

This whole method is in need of a refactor imo

Connoropolous:
"I think for some strange reason it is stating that the address provided has two crud_links "

I think what is needed to be determined is what that "strange reason" is.
I will try once again to produce the scenario of this panic

Connoropolous:
Here is a new dump of the log with the full backtrace, and the full log from launch.

https://gist.github.com/Connoropolous/16f27cb573330ce47ab9b9e197ddc929

maackle:
@Connoropolous thanks for the full log. Can you describe what behavior led to the panic? That would help in creating a regression test.

> I hope we can fix this just by relying on that the lines 86 to 87

@AshantiMutinta do you mean relying on the assert that this PR removes? The assert is causing the panic...

AshantiMutinta:
> @Connoropolous thanks for the full log. Can you describe what behavior led to the panic? That would help in creating a regression test.
> 
> > I hope we can fix this just by relying on that the lines 86 to 87
> 
> @AshantiMutinta do you mean relying on the assert that this PR removes? The assert is causing the panic...

@maackle  No i mean with the line 86 to 87 without the asserts, but that would mean crud_links only return if one eav is returned which could put a dangling reference there so probably not a good idea

We need to figure out what is going on before we can make any decision though

Connoropolous:
Hey!
So without being able to pin down the specifics yet, the general outline of the pattern seemed to be this: 
run one node of acorn-hc/acorn-ui
perform a couple actions that would trigger  commit_entry and update_entry 
spin up another node, networked over a sim2h server
let some gossip occur
shutdown that second node 
restart that second node
panic

AshantiMutinta:
> Hey!
> So without being able to pin down the specifics yet, the general outline of the pattern seemed to be this:
> run one node of acorn-hc/acorn-ui
> perform a couple actions that would trigger commit_entry and update_entry
> spin up another node, networked over a sim2h server
> let some gossip occur
> shutdown that second node
> restart that second node
> panic

hey @Connoropolous is it possible to get this log with the state dump set to true?

Connoropolous:
Yes ok I will try that. I was just looking into that yesterday

AshantiMutinta:
> g to think of what distinguishes between our 'production' and 'development' versions of Acorn as well, and I realized that the storage setting was different. In production/release, storage was 'pickle', in local development, it was 'memory'.
> 
> I switched my development storage mode to 'pickle'.
> I started the app... I committed an entry and a link
> I shut down the holochain instance. I restarted it.
> state_dump suggested that the entries were still there, but I wasn't getting anything back in query calls.
> Anyways. I have switched to LMDB and ITS WORKING, and not panicing.

Hai @Connoropolous are you able to give me the log with the state dump? I tried and could not reproduce this.

PR #2083, opened 2020-01-24
WIP: Sim2h Demo IM State

neonphog:
A demo of using IM readonly cloning + a channel for mutate messages.

(Also note how the `get_agents_missing_aspects` function returns references)


crates/sim2h/src/demo_im_state.rs
```diff
@@ -0,0 +1,440 @@
...
+                                        Err(tokio::sync::mpsc::error::TryRecvError::Closed) => {
+                                            should_end_task = true;
+                                            break;
+                                        }
+                                        Ok(msg) => messages.push(msg),
+                                    }
+                                }
+
+                                // now get the lock and perform the mutations
+                                let mut store = strong_ref.lock().await;
```
struktured: Why do we have to lock the store on get clone? Correct me if I'm wrong but I thought that reads would just use a possibly stale copy.

crates/sim2h/src/demo_im_state.rs
```diff
@@ -0,0 +1,440 @@
...
+        con_incr: Arc<AtomicU64>,
+    ) -> Self {
+        Self {
+            inner,
+            send_mut,
+            con_incr,
+        }
+    }
+
+    pub async fn get_clone(&self) -> StoreRef {
```
struktured: Is this the "non blocking" version of get_clone, then?

crates/sim2h/src/demo_im_state.rs
neonphog: So, I used an Arc<Mutex<>> for ref counting purposes - so the task processing the mutations would just end if there were no references pointing to it.

The only thing that ever acquires the mutex lock is this mutation task, so it should always acquire instantly... I could also use a dummy ref counter and have the state be wholly owned by the mutation task... that's a good idea, I like that.

crates/sim2h/src/demo_im_state.rs
neonphog: yeah, exactly - in the wild you should only ever have `StoreHandle`s which let you push mutation requests across the channel and `StoreRef`s which are the immutable clones of the state...

Would be nice to have a more ergonomic single struct that could serve both purposes, but I haven't yet managed to accomplish that.

crates/sim2h/src/demo_im_state.rs
neonphog: implemented :+1:

neonphog:
closing - will pull this code into my new branch

PR #2087, opened 2020-01-28
Sim2h Connection Preparation for IM State

neonphog:
- Removes the last remnants of the old thread pool stuff.
- Creates a Connection Manager which:
  - polls connections for incoming data
  - manages a switchboard for sending outgoing data
  - manages disconnecting connections either for socket closes/errors or from the sim2h side
- Removes the old non-MultiSend handling code && drops connections if WIRE_VERSION doesn't match
  - (since there's been a holoscape push we shouldn't need the non-MultiSend code anymore)

With these changes we should be ready for implementing the IM state.


crates/sim2h/src/connection_mgr.rs
```diff
@@ -0,0 +1,319 @@
...
+        let mut did_work = false;
+
+        // first, process a batch of control commands
+        for _ in 0..100 {
+            match cmd_recv.try_recv() {
+                Ok(cmd) => {
+                    did_work = true;
+                    match cmd {
+                        ConMgrCommand::SendData(_uri, frame) => {
+                            if let Err(e) = wss.write(frame) {
```
AshantiMutinta: anyway to get an async write in here? Just wondering if it is possible

crates/sim2h/src/connection_mgr.rs
```diff
@@ -0,0 +1,319 @@
...
+                                error!("socket write error {} {:?}", uri, e);
+                                let _ = evt_send
+                                    .send(ConMgrEvent::Disconnect(uri.clone(), Some(e.into())));
+                                // end task
+                                return;
+                            }
+                        }
+                        ConMgrCommand::Disconnect(_uri) => {
+                            debug!("disconnecting socket {}", uri);
+                            let _ = evt_send.send(ConMgrEvent::Disconnect(uri.clone(), None));
```
AshantiMutinta: same here

crates/sim2h/src/connection_mgr.rs
```diff
@@ -0,0 +1,319 @@
...
+            wss_map: std::collections::HashMap::new(),
+        };
+
+        tokio::task::spawn(con_mgr_task(con_mgr, weak_ref_dummy));
+
+        (ConnectionMgrHandle::new(ref_dummy, cmd_send), evt_p_recv)
+    }
+
+    /// internal check our channels
+    fn process(&mut self) -> ConMgrResult {
```
AshantiMutinta: should we look into having process async as well?

crates/sim2h/src/connection_mgr.rs
```diff
@@ -0,0 +1,319 @@
...
+}
+
+/// messages for controlling the connection manager
+#[allow(clippy::large_enum_variant)]
+#[derive(Debug)]
+enum ConMgrCommand {
+    Connect(Lib3hUri, TcpWss),
+    SendData(Lib3hUri, WsFrame),
+    Disconnect(Lib3hUri),
+}
```
maackle: Just remarking that this is pretty much exactly the enum of Walkman events that get recorded and played back, except for using a WireMessage instead of a WsFrame for data. This will make the walkman integration even smoother.

crates/sim2h/src/connection_mgr.rs
neonphog: process is called from an async function - it performs one loop and doesn't need to await anything.

If we ever need to await something here it would be easy to add the async keyword :+1:

crates/sim2h/src/connection_mgr.rs
neonphog: Would very much love to switch to [tokio-tungstenite](https://crates.io/crates/tokio-tungstenite) - but alas, the currently published version isn't compatible with the version of tokio we're using - They're working on it.

The other [tokio websocket library](https://crates.io/crates/websocket) says it's not being maintained any more and to use tokio-tungstenite.

For now, at least, the read / write sockets we're using are non-blocking so it won't block the thread if they're not ready to return data immediately.

crates/sim2h/src/connection_mgr.rs
neonphog: see above comment :+1:

crates/sim2h/src/connection_mgr.rs
```diff
@@ -0,0 +1,319 @@
...
+                    debug!("socket cmd channel closed {}", uri);
+                    let _ = evt_send.send(ConMgrEvent::Disconnect(uri.clone(), None));
+                    // channel broken, end task
+                    return;
+                }
+            }
+        }
+
+        // next process a batch of incoming websocket frames
+        for _ in 0..100 {
```
freesig: Are these 0..100 special numbers?

crates/sim2h/src/connection_mgr.rs
```diff
@@ -0,0 +1,319 @@
...
+    evt_send_from_children: EvtSend,
+    evt_recv_from_children: EvtRecv,
+    wss_map: std::collections::HashMap<Lib3hUri, CmdSend>,
+}
+
+impl ConnectionMgr {
+    /// spawn a new connection manager task, returning a handle for controlling it
+    /// and a receiving channel for any incoming data
+    pub fn new() -> (ConnectionMgrHandle, ConnectionMgrEventRecv) {
+        let (evt_p_send, evt_p_recv) = tokio::sync::mpsc::unbounded_channel();
```
freesig: One thing to keep in mind with unbounded channels is they could just keep growing. It doesn't allow any back pressure on the network and can be easy to miss.

crates/sim2h/src/connection_mgr.rs
AshantiMutinta: nice!

crates/sim2h/src/connection_mgr.rs
AshantiMutinta: Awesome!

crates/sim2h/src/connection_mgr.rs
neonphog: Just to do enough work to be useful / avoid thrashing while also not blocking other tasks - if anyone has a better idea for handling stuff like this, I'm all ears - time base solutions are heavy handed because querying the system clock is costly.

crates/sim2h/src/connection_mgr.rs
neonphog: Totally agree, but bounded channels have their own problems until we've fully adopted futures because you have to await on the send side as well, lest you end up doing blocking inside your task and halting processing.

PR #2089, opened 2020-01-29
Sim2h IM State

neonphog:
```
develop

+-------------------------------+---------+-----+------+--------+
|                          STAT | COUNT   | MIN | MAX  | MEAN   |
+-------------------------------+---------+-----+------+--------+
|               dm_result_in_ms | 1331    | 42  | 1592 | 482.90 |
|                 dm_send_count | 1943    | 1   | 1    | 1.00   |
|          ping_recv_pong_in_ms | 1743    | 11  | 730  | 180.85 |
|               ping_send_count | 1943    | 1   | 1    | 1.00   |
| publish_received_aspect_in_ms | 149733  | 8   | 974  | 310.83 |
|            publish_send_count | 1943    | 1   | 1    | 1.00   |
|           tick_job_elapsed_ms | 6593018 | 0   | 20   | 0.00   |
|         tick_sim2h_elapsed_ms | 2346    | 0   | 0    | 0.00   |
+-------------------------------+---------+-----+------+--------+

sim2h-im2

+-------------------------------+---------+-----+-----+-------+
|                          STAT | COUNT   | MIN | MAX | MEAN  |
+-------------------------------+---------+-----+-----+-------+
|               dm_result_in_ms | 2278    | 42  | 164 | 77.74 |
|                 dm_send_count | 2316    | 1   | 1   | 1.00  |
|          ping_recv_pong_in_ms | 2308    | 6   | 71  | 46.29 |
|               ping_send_count | 2316    | 1   | 1   | 1.00  |
| publish_received_aspect_in_ms | 222293  | 6   | 115 | 39.85 |
|            publish_send_count | 2316    | 1   | 1   | 1.00  |
|           tick_job_elapsed_ms | 6971370 | 0   | 18  | 0.00  |
|         tick_sim2h_elapsed_ms | 2803    | 0   | 0   | 0.00  |
+-------------------------------+---------+-----+-----+-------+
```


crates/sim2h/src/mono_ref.rs
```diff
@@ -0,0 +1,110 @@
+use crate::*;
+
```
zippy: Can you please add a bit of documentation to describe the intent of the MonoRef and why it's needed.  I think I get it but it would be useful to have a couple lines at the top.

crates/sim2h/src/lib.rs
```diff
@@ -197,56 +182,42 @@ pub enum DhtAlgorithm {
...
+#[allow(dead_code)]
+mod sim2h_im_state;
 
 #[derive(Clone)]
 /// A clonable reference to our Sim2h instance that can be passed
 /// into `'static` async blocks && still be able to make sim2h calls
 struct Sim2hHandle {
-    state: Arc<tokio::sync::Mutex<Sim2hState>>,
-    send_com: tokio::sync::mpsc::UnboundedSender<Sim2hCom>,
+    im_state: sim2h_im_state::StoreHandle,
```
zippy: is there a reason to call this `im_state` instead of just `state`?

crates/sim2h/src/lib.rs
```diff
@@ -260,49 +231,565 @@ impl Sim2hHandle {
...
-                uri,
-                space_address,
-                agent_id,
-                message,
-            })))
-            .expect("can send");
+        // these message types are allowed before joining
+        let message = match message {
+            WireMessage::Lib3hToClient(_) | WireMessage::ClientToLib3hResponse(_) =>
+                panic!("This is soo wrong. Clients should never send a message that only servers can send."),
```
zippy: I think we should just log this as an error rather than panicing.  I don't want us to create a way that people can bring down a sim2h server, or, more likely, that we bring one down by miscoding something in the sim2h_worker side.

crates/sim2h/src/lib.rs
```diff
@@ -260,49 +231,565 @@ impl Sim2hHandle {
...
+                DhtAlgorithm::NaiveSharding { redundant_count } => *redundant_count,
+            },
+            version: WIRE_VERSION,
+            extra: None,
+        }),
+    );
+    // versions do not match - disconnect them
+    if version != WIRE_VERSION {
+        warn!(
+            "Disconnecting client for bad version this WIRE_VERSIO = {}, client WIRE_VERSION = {}",
```
zippy:
```suggestion
            "Disconnecting client for bad version this WIRE_VERSION = {}, client WIRE_VERSION = {}",
```

crates/sim2h/src/lib.rs
```diff
@@ -260,49 +231,565 @@ impl Sim2hHandle {
...
+        uri,
+        &WireMessage::Lib3hToClient(Lib3hToClient::HandleGetAuthoringEntryList(GetListData {
+            request_id: "".into(),
+            space_address: data.space_address.clone(),
+            provider_agent_id: data.agent_id,
+        })),
+    );
+}
+
+fn spawn_handle_message_send_dm(
```
zippy: The only difference between `spawn_handle_message_send_dm` and `spawn_handle_message_send_dm_result` is the enum flavor `SendDirectMessageResult` or `SendDirectMessage`  
This seems worth refactoring?

crates/sim2h/src/lib.rs
```diff
@@ -903,369 +1145,73 @@ impl Sim2h {
...
-                }
-            }
-            WireMessage::Lib3hToClientResponse(Lib3hToClientResponse::HandleQueryEntryResult(query_result)) => {
-                if (query_result.responder_agent_id != agent_id) || (query_result.space_address != space_address)
-                {
-                    return Err(SPACE_MISMATCH_ERR_STR.into());
+                let query_agents = state.get_agents_for_query(&space_hash, &entry_hash, None);
+
+                if query_agents.is_empty() {
+                    continue;
```
zippy: Shouldn't we be logging this as a warning at least?  I think this means that we know that an agent needs a particular aspect, but there's nobody to ask for it about.  That's a problem that should be logged?

crates/sim2h/src/lib.rs
```diff
@@ -903,369 +1145,73 @@ impl Sim2h {
...
+                let wire_message =
+                    WireMessage::Lib3hToClient(Lib3hToClient::HandleFetchEntry(FetchEntryData {
+                        request_id: "".to_string(),
+                        space_address: (&**space_hash).clone(),
+                        provider_agent_id: (&*query_agent).clone(),
+                        entry_address: (&**entry_hash).clone(),
+                        aspect_address_list: Some(aspects.iter().map(|a| (&**a).clone()).collect()),
+                    }));
+
+                sim2h_handle.send((&*query_agent).clone(), (&*uri).clone(), &wire_message);
```
zippy: I'm wondering if we shouldn't also be building this up into a multi-message.  It seems likely to me that after going through this many of the HandleFetchEntries will actually be requests to the same agent.  If so they shouldn't go out as a multiple sends.

crates/sim2h/src/sim2h_im_state.rs
```diff
@@ -0,0 +1,1177 @@
...
+    // - mark connection as disconnected (tombstone)
+    // - clear all `holding` aspects (to prepare for another connection
+    // - remove the uri_to_connection entry
+    DropConnection {
+        aol_idx: u64,
+        space_hash: SpaceHash,
+        agent_id: AgentId,
+    },
+
+    // we need to be able to drop all connectios across spaces based on
```
zippy:
```suggestion
    // we need to be able to drop all connections across spaces based on
```

crates/sim2h/src/sim2h_im_state.rs
```diff
@@ -0,0 +1,1177 @@
...
+pub type MonoEntryHash = MonoRef<EntryHash>;
+pub type MonoAspectHash = MonoRef<AspectHash>;
+pub type MonoUri = MonoRef<Lib3hUri>;
+
+/// so we know who's holding what
+pub type Holding = im::HashMap<MonoAspectHash, im::HashSet<MonoAgentId>>;
+
+/// so we cache entry locations as well
+#[derive(Debug, Clone)]
+pub struct Entry {
```
zippy: calling the `Entry` feels confusing to me.  It seems like it's really `EntryWhereabouts` or `EntryLocationAndHolders` or something like that.  Any better name?

crates/sim2h/src/lib.rs
```diff
@@ -260,49 +231,565 @@ impl Sim2hHandle {
...
+                        uri,
+                        signer,
+                        space_hash,
+                        data,
+                    );
+                }
+                WireMessage::Lib3hToClientResponse(
+                    Lib3hToClientResponse::HandleGetAuthoringEntryListResult(list_data),
+                ) => {
+                    return spawn_handle_message_list_data(
```
neonphog: this isn't sufficient - depending on how the sharding works out, gossip won't know to go here for fetches - despite the inefficiency we should go ahead and send fetch requests for all these entries and they can be ignore on receipt if everybody already has them.

neonphog: fixed

crates/sim2h/src/mono_ref.rs
```diff
@@ -0,0 +1,110 @@
...
+}
+
+/*
+// WTF - why can't we make a generic one too?
+impl<T: std::fmt::Debug> std::fmt::Debug for MonoRef<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self.0)
+    }
+}
+*/
```
AshantiMutinta: It's possible todo this but sadly you have to turn a feature flag on 

#![feature(specialization)]

Which would also require you to put a 'default' before fn fmt

neonphog: ah, gottcha - well, it's not strictly necessary at the moment. Hopefully that'll become stable some day : )

crates/sim2h/src/lib.rs
neonphog: lol, while doing the transition I had both of them in there... I can rename it now.

crates/sim2h/src/lib.rs
neonphog: The aspects at least are already aggregated into a single fetch entry request.

The `get_agents_for_query` function, however, randomizes agents for querying against in an effort to distribute the workload. It feels like pre-optimization complexity to me to aggregate messages if the happen to randomly be assigned to the same agent.

crates/sim2h/src/lib.rs
neonphog: :+1: added `warn!` we may find this is too chatty when running real apps and people are going offline, etc - but better too verbose to start : )

crates/sim2h/src/sim2h_im_state.rs
neonphog: Updated to `EntryInfo`

crates/sim2h/src/mono_ref.rs
neonphog: added/updated the comment to include a note about the feature :+1:

PR #2094, opened 2020-02-04
[NO_MERGE] db-ci-test

neonphog:


neonphog:
found a fix - https://github.com/holochain/holochain-rust/pull/2095

PR #2099, opened 2020-02-05
Trycp update hc

freesig:
## PR summary
Adds the ability to replace the conductor through trycp_server using a build linux holochain binary.
This should allow us to easy update conductors on final-exam to add traces
## testing/benchmarking notes
I'm still waiting for the docker to build to test this.

## followups
I have just got the command pointing at releases on [holochain tracing](https://github.com/holochain/holochain-tracing/releases/tag/v0.0.16) because I wasn't sure if it was ok to add another asset to the holochain-rust release but we could do that aswell.

Should I do the same for the cli?

## changelog

```markdown
```

## documentation
To update the conductor: 
1. Build a conductor on linux.
2. Put it as an asset on the holochain tracing github under a release.
3. call `node admin replace path/to/config.json <release_tag> <file_name>` e.g. (``node admin replace pool_config.json v0.0.16  holochain-tracing-x86_64-generic-linux-gnu.tar.gz`


.circleci/config.yml
```diff
@@ -302,7 +302,16 @@ workflows:
          - 2020-01-23-docker-sim2h-fix
          - develop
          - final-exam
+         - trycp-update-hc
    - docker-build-trycp-server:
+      filters:
+       branches:
+        only:
+         - 2019-12-23-docker-aws-push
+         - 2020-01-23-docker-sim2h-fix
```
neonphog: are we keeping random things like this in here?

crates/trycp_server/src/main.rs
```diff
@@ -245,6 +252,25 @@ fn get_info_as_json() -> String {
     format!("{{{}}}", result)
 }
 
+/// very dangerous, runs whatever strings come in from the internet directly in bash
```
neonphog: lol - trycp only runs on servers that are only internally vpc accessible though, right?

crates/trycp_server/src/main.rs
```diff
@@ -453,6 +479,18 @@ fn main() {
         Ok(Value::String(response))
     });
 
+    let allow_replace_conductor = args.allow_replace_conductor;
+    io.add_method("replace_conductor", move |params: Params| {
+        if allow_replace_conductor {
+            Ok(Value::String(os_eval(&format!("curl -L https://github.com/holochain/holochain-tracing/releases/download/{}/{} -o holochain.tar.gz && tar -xzvf holochain.tar.gz && mv holochain /holochain/.cargo/bin/holochain && rm holochain.tar.gz",
```
neonphog: ... why is the holochain binary coming from the `holochain-tracing` repository releases??

freesig: I just wasn't sure if I could put binaries up on the holochain-rust/release page. But I can change it to that no probs. I didn't want to make it a parameter because then if someone did get hold of this command they could just point it to any url and put in any binary

crates/trycp_server/src/main.rs
freesig: Yeh and the only way to call this function is through the other pub functions, I just kept zippys comment so that no-one makes this pub

crates/trycp_server/src/main.rs
freesig: Ok I added a repo param, so it's still hardcoded to the holochain org but you can set the repo

PR #2100, opened 2020-02-05
Sort option for links

AshantiMutinta:
## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation


app_spec/zomes/simple/code/src/lib.rs
```diff
@@ -51,10 +51,11 @@ pub fn handle_delete_my_link_with_tag(base: Address,target : String,tag:String)
 }
 
 
-pub fn handle_get_my_links(agent : Address,status_request:Option<LinksStatusRequestKind>) ->ZomeApiResult<GetLinksResult>
+pub fn handle_get_my_links(agent : Address,status_request:Option<LinksStatusRequestKind>,sort_order:Option<SortOrder>) ->ZomeApiResult<GetLinksResult>
```
neonphog: Looks like you're just following the existing paradigm, so no need to change, but I find this pretty hard to parse without spaces - maybe we should add these rust files to our `fmt` check as a separate code hygiene effort?

AshantiMutinta: I 100% agree! i dont know if it is because they are zomes but we should definitely look into this

PR #2101, opened 2020-02-07
remove sim1h and n3h

neonphog:
## PR summary

- Removed networking support for `sim1h`
- Removed networking support for `n3h`

There are still some awkwardly integrated references, this was the majority broad stroke.

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation


AshantiMutinta:
This is awesome, does it mean rust sodium has been taken out too? I could get a running conductor in windows easily now hehe

neonphog:
@AshantiMutinta 

> This is awesome, does it mean rust sodium has been taken out too? I could get a running conductor in windows easily now hehe

haha, no, sodium is a core part of our key management/dpki - it's here to stay - it totally builds on windows if we have time to take a breath, we should figure out where the problem is.

crates/conductor_lib/src/conductor/base.rs
```diff
@@ -312,9 +302,7 @@ impl Conductor {
     }
 
     pub fn p2p_bindings(&self) -> Option<Vec<String>> {
-        self.network_spawn
-            .as_ref()
-            .map(|spawn| spawn.p2p_bindings.clone())
+        None
```
zippy: I think we need a comment here to explain why this is `None` and how something might be addable back in here in the future?

neonphog: actually, turns out this function isn't used anywhere - removed

PR #2105, opened 2020-02-11
Code coverage with kcov and codecov.io

neonphog:
looks like both `kcov` and `tarpaulin` need special docker flags to run:

```
--security-opt seccomp=unconfined
```

this makes the image download much slower and breaks the incremental build caching stuff...


.circleci/config.yml
```diff
@@ -30,6 +30,20 @@ jobs:
       command: nix-shell --run hc-rust-test
       no_output_timeout: 20m
 
+ cov-kcov:
+  machine: true
+  resource_class: xlarge
+  steps:
+   - checkout
+   - run: docker run --rm --security-opt seccomp=unconfined -v $(pwd):/project holochain/holochain-rust:circle.build.develop sh -c "cd /project && nix-shell --run hc-rust-coverage-kcov"
```
neonphog: both tarpaulin and kcov need this docker `--security-opt seccomp=unconfined` - I don't know another way of doing this with circleci - but this makes downloading the docker image slower.

codecov.yml
```diff
@@ -1,4 +1,6 @@
 # Repository Yaml
+codecov:
+  token: 684bc22e-c6ca-4ac5-b5be-d1dd250a4567
```
neonphog: ... committing this seems vaguely questionable, but I'm not a circle/codecov guru - don't know how else to do it.

rust/test/default.nix
```diff
@@ -1,12 +1,70 @@
...
-  script = pkgs.writeShellScriptBin name
+  hc-rust-test = pkgs.writeShellScriptBin "hc-rust-test"
   ''
   hc-rust-wasm-compile && HC_SIMPLE_LOGGER_MUTE=1 RUST_BACKTRACE=1 cargo test --all "$1" -- --test-threads=${holonix.rust.test.threads};
   '';
+
+  hc-rust-coverage-kcov = pkgs.writeShellScriptBin "hc-rust-coverage-kcov"
+  ''
+  # we need the kcov tool and curl for fetching the codecov.io uploader
+  nix-env -f https://github.com/NixOS/nixpkgs-channels/archive/nixos-19.09.tar.gz -iA kcov curl
```
neonphog: This is a bit of a sledgehammer - but I keep failing at being able to put in dependencies in any other way.

rust/test/default.nix
```diff
@@ -1,12 +1,70 @@
...
+
+  # using cargo-make for executing kcov
+  cargo install cargo-make
```
neonphog: again... seems like maybe this should be part of the nix environment some magical way?

rust/test/default.nix
```diff
@@ -1,12 +1,70 @@
...
+    # remove any pre-existing target dir
+    # symlink to shared target dir
+    # build the test binaries
+    # run the code coverage
+
+    ( \
+      cd $i && \
+      rm -rf target || true && \
+      ln -s ../../target
+      export CARGO_TARGET_DIR=$(readlink -f ./target) && \
```
neonphog: cargo make / kcov are very particular about the placement of the target directory - this breaks incremental builds so the coverage run takes 46 minutes - but I wasn't able to get it to work in any other way.

rust/test/default.nix
```diff
@@ -1,12 +1,70 @@
...
+
+  hc-rust-coverage-tarpaulin = pkgs.writeShellScriptBin "hc-rust-coverage-tarpaulin"
+  ''
+  unset CARGO_TARGET_DIR
+  nix-env -f https://github.com/NixOS/nixpkgs-channels/archive/nixos-19.09.tar.gz -iA curl
+  cargo install cargo-tarpaulin
+  hc-rust-wasm-compile
+  export CARGO_TARGET_DIR=$(readlink -f ./target)
+  cargo tarpaulin -v -o Xml --exclude-files "*/.cargo/*"
+  bash <(curl -s https://codecov.io/bash)
```
neonphog: The tarpaulin binary installs and runs just fine - but how they are doing the instrumenting does not seem to be compatible with our wasm and ffi binding - so many of our crates to not succeed in getting covered.

neonphog:
@thedavidmeister - would love your review / help here - there are many nix-y / circle things that aren't ideal - or if we don't want to spend time on ideal this technically works and we can always refactor later.

.circleci/config.yml
```diff
@@ -30,6 +30,20 @@ jobs:
       command: nix-shell --run hc-rust-test
       no_output_timeout: 20m
 
+ cov-kcov:
```
AshantiMutinta: Only comment I have is maybe we can change this to "Code Coverage" when it shows up? but definitely not necessarily at all

codecov.yml
thedavidmeister: added to environment

.circleci/config.yml
neonphog: :+1: - done

rust/test/default.nix
neonphog: we're fine sticking with this for ci task for now

PR #2108, opened 2020-02-12
don't run .bashrc for any commands fed from .hcbuild

### LINKED ISSUE #2097: writing to STDOUT in `.bashrc` causes `hc package` to fail
Because `hc package` spawns a new `bash` process, it runs `.bashrc`. If the user has anything that writes to `STDOUT` in their `.bashrc` file, that output gets captured by `hc package`. AFAICT this doesn't matter until it comes time to [check for the existence of the build artifact](https://github.com/holochain/holochain-rust/blob/06b1d29d04aa4c508d2a2b3da5972aee1c53a143/crates/cli/src/config_files/build.rs#L44), when the output gets incorporated into the expected artifact's path. (It might cause issues elsewhere too, but I'm not sure.)

Steps to reproduce:

1. Add the line `echo "***hello***"` somewhere in your `.bashrc`
2. Enter the nix shell: `nix-shell https://holochain.love`
3. Generate a stub project:
    ```bash
    hc init stub-project
    cd stub project
    hc generate zomes/stub-zome
    ```
4. Try to build: `hc package`
5. Witness error message:
    ```
    Error: Couldn't traverse DNA in directory "/home/alice/Holochain/stub-project": artifact path /home/alice/Holochain/stub-project/zomes/stub-zome/code/***hello***
/home/alice/target/wasm32-unknown-unknown/release/stub-zome.wasm either doesn't point to a file or doesn't exist
    ```

pdaoust:
## PR summary

Resolves #2097

## testing/benchmarking notes

Profound apologies; I can't get the silly thing to compile (missing build deps for newrelic-sys and it's late). But the means for testing would be as follows:

1. Add the line `echo "hello"` somewhere in your `.bashrc` file.
2. Compile a zome with `hc package`.

On current `develop` this should fail with a message like:

```
Error: Couldn't traverse DNA in directory "/home/alice/Holochain/stub-project": artifact path /home/alice/Holochain/stub-project/zomes/stub-zome/code/***hello***
/home/alice/target/wasm32-unknown-unknown/release/stub-zome.wasm either doesn't point to a file or doesn't exist
```

With this fix it should succeed.

## changelog

```markdown
```

### DIFF
--- a/crates/cli/src/util.rs
+++ b/crates/cli/src/util.rs
@@ -47,12 +47,17 @@ pub fn run_cmd(base_path: &PathBuf, bin: String, args: &[&str]) -> DefaultResult
     // e.g. we assume `cargo`, `wasm-gc`, `wasm-opt`, `wasm2wat`, `wat2wasm` all exist in the
     // default template (which we can't assume outside nix-shell in a portable way).
     //
+    // Note the use of --norc. This is to prevent the shell from loading up its usual startup
+    // script, which may echo things to STDOUT. The reason this matters is that some of the
+    // commands in a .hcbuild file (which are fed to this function) capture from STDOUT and do
+    // something meaningful with the value. Do we really want a MOTD in our build artifact path?
+    //
     // @TODO - does it make more sense to push "execute arbitrary bash" style features down to the
     // `nix-shell` layer where we have a better toolkit to handle environments/dependencies?
     // e.g. @see `hn-release-cut` from holonix that implements conventions/hooks to standardise
     // bash processes in an extensible way
     let status = Command::new("bash")
-        .args(&["-c", &command_string])
+        .args(&["--norc", "-c", &command_string])
         .current_dir(base_path)
         .status()?;
 


pdaoust:
@samrose Thanks for taking the time to review and recmmend! We certainly could tell devs to run `nix-shell --pure` in the install guide (although it is nice to have most of your comfy environment still available to you). But the concern here is that a new `bash -c` is spawned for every step of the zome build process, and at least one of the steps captures `STDOUT` and tries to do meaningful things with it. If any command in `.bashrc` outputs to `STDOUT` it borks the build process. So it's less about the host environment leaking into those build steps.

Do you know if `--pure` causes 'inner' invocations of bash to be pure as well, or is it just the immediate shell created by `nix-shell`?

pdaoust:
I presume the app-spec-tests-sim2h failure is a red herring... looks like it all happens after app-spec is already built.

pdaoust:
Hey @samrose just checking in on your change request -- given the extra details I shared, does this seem like a decent solution?

pdaoust:
@zippy bumping this PR -- I think it was languishing for a while because it was blocked by some requested changes. I don't know if those requested changes pertain to the issue this is trying to fix though.

I've updated this branch to `master` and the diff is still clean. It should be a quick merge. The only thing I don't know is whether this causes unintended consequences. Are there times when the inner `bash` that `hc package` spawns _should_ actually re-run the `.bashrc`? Are there times when _not_ running it could break things for the dev? Nothing comes to mind, but hey, these are computers we're talking about -- you never know.

PR #2109, opened 2020-02-12
Run respond_validation_package_request() in an async task on thread-pool

lucksus:
## PR summary

*Potentially fixes problems with direct messages seen in HoloFuel stress tests.*

The network handlers spawn asynchronous tasks according to the network messages that are received. Up until November last year, those async tasks were executed in newly spawned threads. [I switched that over to using futures and an M:N thread-pool](https://github.com/holochain/holochain-rust/pull/1915), which made the whole system much more performant and resilient under stress.

I wrapped all async network handling functions in `async` blocks and spawned those as task on the thread pool. All but `respond_validation_package_request()` because this is the only one that actually is not implemented as `async fn` (and doesn't need to). Wrapping this in an async block *should* have been no problem, but it was a problem (the non-async function was never called / the async block was not executed) because of a bug in the implementation of async/await which was still unstable at that point.

async/await got [stabilized in Rust 1.39](https://blog.rust-lang.org/2019/11/07/Rust-1.39.0.html) which also happened in November but we didn't have that new version in Holonix yet when I finished the future thread-pool implementation.

We now have Rust 1.41 in Holonix with stable async/await which makes it possible to just apply the changes in this PR to render validation package responds async (**and thus not blocking the network thread!**) by wrapping `respond_validation_package_request()` in an `async` block just like the other functions.

## testing/benchmarking notes

I got HoloFuel stress tests to pass locally with 80 nodes once with these changes.
```
1..243
# tests 243
# pass  243

# ok
```

but also got one failure
```
1..243
# tests 243
# pass  239
# fail  4
```

caused by one timeout (out of 80 - 79 were fine):
```
    "Err": {
      "Internal": "{\"kind\":\"Timeout\",\"file\":\"crates/core/src/wasm_engine/runtime.rs\",\"line\":\"221\"}"
    }
```

This is definitely a huge improvement for the *HoloFuel direct message stress issue*, but we'll have to investigate what's left causing the failure (which could totally be limited resources on my local device). 

## followups

Run more tests, esp. on final exam.

## changelog

```markdown
```

## documentation


crates/core/src/network/handler/send.rs
```diff
@@ -63,19 +63,18 @@ pub fn handle_send_message(message_data: DirectMessageData, context: Arc<Context
...
-            // rust 1.39 where async/await is part of stable (currently we are using unstable
-            // feature for this).
-            respond_validation_package_request(
-                message_data.from_agent_id.into(),
-                message_data.request_id,
-                address,
-                context,
-                vec![],
-            );
+            let c = context.clone();
```
freesig:
```suggestion
            let closure = {
                let context = context.clone();
                async move || {
                    respond_validation_package_request(
                        message_data.from_agent_id.into(),
                        message_data.request_id,
                        address,
                        context,
                        vec![],
                    );
                }
            };
            let future = closure();
            context.spawn_task(future);
```
Just a more idiomatic way to write this if you want

PR #2117, opened 2020-02-17
hc-dpki cli for DeepKey

zo-el:
## PR summary
Adds some new commands to the CLI tool for managing DPKI keys.

- `hc dpki genroot` creates a new random root seed
- `hc dpki keygen` is similar to `hc keygen` but will use a dpki root seed mnemonic so that keys can be recovered
- `hc dpki genrevoke/genauth` produce revocation/auth seeds from which keys can be derived. 
- `hc dpki sign` use a revocation/auth seed to produce a key then sign a message (which can be an agent key). This is the process by which keys can be authorized and revoked

Each time a mnemonic is produced it is optionally encrypted with a passphrase

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation

### DIFF
--- a/crates/cli/src/cli/keygen.rs
+++ b/crates/cli/src/cli/keygen.rs
@@ -1,48 +1,106 @@
-use crate::NEW_RELIC_LICENSE_KEY;
-use error::DefaultResult;
 use holochain_common::paths::keys_directory;
-use holochain_conductor_lib::{key_loaders::mock_passphrase_manager, keystore::Keystore};
-use rpassword;
+use holochain_conductor_lib::{
+    key_loaders::mock_passphrase_manager,
+    keystore::{Keystore, PRIMARY_KEYBUNDLE_ID},
+};
+use holochain_core_types::error::HcResult;
+use holochain_locksmith::Mutex;
+use holochain_dpki::seed::{SeedType, TypedSeed};
 use std::{
     fs::create_dir_all,
     io::{self, Write},
     path::PathBuf,
+    sync::Arc,
 };
+use util::{get_secure_string_double_check, get_seed, user_prompt};
 
-#[holochain_tracing_macros::newrelic_autotrace(HOLOCHAIN_CLI)]
-pub fn keygen(path: Option<PathBuf>, passphrase: Option<String>, quiet: bool) -> DefaultResult<()> {
-    let passphrase = passphrase.unwrap_or_else(|| {
-        if !quiet {
-            println!(
-                "
-This will create a new agent keystore and populate it with an agent keybundle
+pub fn keygen(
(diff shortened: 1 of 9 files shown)


zippy:
@zo-el I notice a bunch of tests are breaking, would like to review this after tests pass...

zo-el:
> @zo-el I notice a bunch of tests are breaking, would like to review this after tests pass...

Yes @zippy, this is still work in progress. I apologise.

zo-el:
@zippy this PR is ready for review

crates/cli/src/cli/keygen.rs
```diff
@@ -1,48 +1,106 @@
...
+                "Invalid number of words in mnemonic. Must be 24 (unencrypted) or 48 (encrypted)"
+            ),
+        }
+    } else {
+        user_prompt(
+            "Generating keystore (this will take a few moments)...",
+            quiet,
+        );
+        keygen_standalone(keystore_passphrase)?
+    };
```
AshantiMutinta: Noticed a few panics in here even though we are returning an HCResult, is this because we cannot recover from them if they happen?

crates/cli/src/util.rs
```diff
@@ -1,13 +1,78 @@
...
+}
+
+pub fn user_prompt_yes_no(message: &str, quiet: bool) -> bool {
+    user_prompt(format!("{} (Y/n)", message).as_str(), quiet);
+    let mut input = String::new();
+    stdin()
+        .read_line(&mut input)
+        .expect("Could not read from stdin");
+    match input.as_str() {
+        "Y\n" => true,
```
AshantiMutinta: do we care about lowercase/uppercase?

crates/dpki/src/seed.rs
```diff
@@ -278,6 +296,91 @@ impl DevicePinSeed {
...
+        let context = SeedContext::new(DEVICE_CTX);
+        let mut context = context.to_sec_buf();
+        kdf::derive(
+            &mut ref_seed_buf,
+            derivation_index,
+            &mut context,
+            &mut self.inner.buf,
+        )?;
+        Ok(KeyBundle::new_from_seed_buf(&mut ref_seed_buf)?)
+    }
```
AshantiMutinta: You could probably generalize these two structs with an enum.

enum Seed
{ Auth,Revocation}

crates/cli/src/cli/keygen.rs
maackle: Agreed, these expects should be replaced with `?`

crates/cli/src/util.rs
zo-el: Not Really

PR #2119, opened 2020-02-18
Make Holochain work offline again

lucksus:
## PR summary

Treat disconnected state the same as full-sync DHT

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation


crates/net/src/sim2h_worker.rs
```diff
@@ -532,6 +532,12 @@ impl Sim2hWorker {
         self.is_full_sync_DHT = full_sync;
     }
 
+    /// Tells us if Sim2hWorker should render the local core instance to be self-suficient,
+    /// i.e. storing everything locally and answering to queries locally.
+    fn is_autonomous_node(&mut self) -> bool {
+        self.is_full_sync_DHT || !self.connection_ready()
+    }
```
maackle: I wonder about race conditions.

- connection_ready() could flip in between when it is checked and when the message is actually handled
- there is a delay between send_wire_message putting a message on the queue and when it actually gets sent

if these are valid concerns, i wonder if there is a unified way to deal with the online and the offline case that doesn't require a condition?

thedavidmeister: @maackle if valid concerns, would they also apply to the existing code in the boundary between full sync and sharded?

lucksus: Well, currently, if you were connected and then you loose connection, it will just stop working.

With these changes, a query that is send just when loosing connection might not work, but if the user continues to use a hApp UI which then makes the DNA issue more queries, those would then get routed back to the local instance which can then respond.

If the node then reconnects to sim2h again, the server will request authoring and gossip lists and make sure that new data gets shared in the DHT. So, it *should* be all fine.. But we also should add tests for all of this. That would have prevented the regression for this is a fix.

PR #2120, opened 2020-02-18
Sim2h WireMessage Receipts

lucksus:
## PR summary

This intends to make the channel between nodes and the Sim2h server resilient against data/connection loss on the level of Lib3hMessages. (in the direction: Sim2hWorker -> Sim2h server)

This resilience is actually an assumption that code in core holds, but that wasn't met by the Sim2h/Sim2hWorker implementation yet. A connection dropped just after a message was sent would leave the disconnected node without track of that lost message. It would continue to function after a reconnect, but the message sent before would be lost, if it wasn't received by the server before the disconnect.

With this PR we add a receipt message type `WireMessage::Ack(u64)` that the Sim2h server sends back, including a hash of the processed message. The Sim2hWorker now only sends one message at a time and waits for the receipt before sending the next one. If a receipt is not received within 10 seconds, it will resend the message.

## testing/benchmarking notes

## followups
* Apply receipts also for the other direction: Sim2h -> Sim2hWorker

## changelog

```markdown
```

## documentation


neonphog:
See for getting u8 to work: https://github.com/holochain/holochain-rust/compare/low-level-receipts

neonphog:
Similarly - if you want the hashing to be consistent, I believe the problem lies in the level at which you are hashing.

On the worker side, if you hash the already serialized binary data (optimally changing the outgoing queue to have already serialized data and respective hashes), then on the server side, do the hashing before deserializing the data, then obviously they should match.

crates/net/src/sim2h_worker.rs
```diff
@@ -65,7 +89,8 @@ pub struct Sim2hWorker {
     connection_timeout_backoff: u64,
     reconnect_interval: Duration,
     metric_publisher: std::sync::Arc<std::sync::RwLock<dyn MetricPublisher>>,
-    outgoing_message_buffer: Vec<WireMessage>,
+    outgoing_message_buffer: Vec<BufferedMessage>,
+    outgoing_message_next_serial: u16,
```
lucksus:
```suggestion
```

crates/net/src/sim2h_worker.rs
```diff
@@ -107,6 +132,7 @@ impl Sim2hWorker {
                 DefaultMetricPublisher::default(),
             )),
             outgoing_message_buffer: Vec::new(),
+            outgoing_message_next_serial: 0,
```
lucksus:
```suggestion
```

crates/net/src/sim2h_worker.rs
```diff
@@ -269,7 +302,8 @@ impl Sim2hWorker {
         // we always put messages in the outgoing buffer,
         // they'll be sent when the connection is ready
         debug!("WireMessage: queueing {:?}", message);
-        self.outgoing_message_buffer.push(message);
+        self.outgoing_message_next_serial += 1;
```
lucksus:
```suggestion
```

crates/sim2h/src/wire_message.rs
```diff
@@ -1,7 +1,23 @@
...
+//use std::hash::SipHasher;
+//use hashers::builtin::DefaultHasher;
+
+fn sdbm_hash(bytes: Vec<u8>) -> u64 {
+    let mut hash: u64 = 0;
+    for byte in bytes {
+        hash = byte as u64 + (hash << 6) + (hash << 16) - hash;
+    }
+    hash
+}
```
lucksus: remove

crates/sim2h/src/wire_message.rs
```diff
@@ -106,8 +123,45 @@ impl WireMessage {
...
+                let opaque: Opaque = serde_json::to_string(&span_wrap.data)
+                    .expect("wiremessage should serialize")
+                    .into();
+                opaque.into()
+            }
+            _ => Opaque::from(self).into(),
+        };
+
+        sdbm_hash(bytes)
+    }
```
lucksus: remove

lucksus:
@neonphog, @zippy, implemented both your change requests.

crates/net/src/sim2h_worker.rs
```diff
@@ -209,58 +232,65 @@ impl Sim2hWorker {
...
+        buffered_message.last_sent = Some(Instant::now());
+        true
     }
 
     /// if we re-connected, we may need to send a join first,
     /// before other queued messages
     fn prepend_wire_message(&mut self, message: WireMessage) -> NetResult<()> {
         debug!("WireMessage: queueing {:?}", message);
-        self.outgoing_message_buffer.insert(0, message);
+        self.outgoing_message_buffer.insert(0, message.into());
```
neonphog: Do we need to set the `last_sent` for the head of the queue to `None` here - so we don't wait on it after this prepended join is processed?

lucksus: That happens here: https://github.com/holochain/holochain-rust/pull/2120/files#diff-03e64f941c57272f7ef1b84b07ed2005R66

lucksus:
```rust
impl From<WireMessage> for BufferedMessage {
    fn from(wire_message: WireMessage) -> BufferedMessage {
        BufferedMessage {
            wire_message,
            hash: 0,
            last_sent: None,
        }
    }
}
```

crates/net/src/sim2h_worker.rs
```diff
@@ -209,58 +232,65 @@ impl Sim2hWorker {
...
-            debug!("WireMessage: dequeuing sent message {:?}", message);
-            // if we made it here, we successfully sent the first message
-            // we can remove it from the outgoing buffer queue
-            self.outgoing_message_buffer.remove(0);
+            self.connection = None;
+            self.check_reconnect();
+            return true;
         }
+        #[allow(deprecated)]
+        let mut hasher = SipHasher::new();
```
neonphog: vote for going with the siphasher crate https://crates.io/crates/siphasher to avoid the deprecation

alternately https://crates.io/crates/twox-hash because it is faster for larger byte-count messages

crates/net/src/sim2h_worker.rs
neonphog: no - I mean the previous head of the outgoing queue (item index 0) which may still be pending an Ack - if we prepend a JoinSpace message and that join space is successful, the next message may still have a not-None `last_sent` in it that should be cleared, right?

crates/net/src/sim2h_worker.rs
neonphog: xxhash starts being faster at ~16 or 32 bytes - the tracer context by itself is 37 bytes.

crates/net/src/sim2h_worker.rs
lucksus: Right! Good catch!

lucksus:
@neonphog, implemented those suggestions as well :)

crates/sim2h/src/lib.rs
```diff
@@ -71,6 +74,7 @@ lazy_static! {
 
 /// if we can't acquire a lock in 20 seconds, panic!
 const MAX_LOCK_TIMEOUT: u64 = 20000;
+pub const RECEIPT_HASH_SEED: u64 = 0;
```
neonphog: > [seed your seed value, even zero is a valid seed and e.g. used by LZ4](https://create.stephan-brumme.com/xxhash/)

:+1: - just had to check

PR #2123, opened 2020-02-21
trycp_manager

zippy:
## PR summary

Adds a `register` and `request` command to trycp_server protocol.  This allows us to use a trycp_server binary as a manager of available endpoints for final-exam.

Adds the following options to the trycp_server cli:
```
    -m, --manager                    Activate ability to runs as a manager
    -e, --register <register>               Register with a manager (url + port, e.g. ws://final-exam:9000)
    -h, --host <host>                       The host name to use when registering with a manager [default: localhost]
```

The idea is that "client" trycp_servers pick one to register with when they boot up.  When final-exam wants to run a queued job, it simply asks the manager for `count` available nodes needed for that test.  The manager confirms and resets the nodes on its list and if enough are available hands the list back to final-exam for use.  When the nodes are done, they just re-register their availability with the manager.

## changelog

## documentation


crates/trycp_server/src/main.rs
```diff
@@ -291,6 +334,55 @@ fn main() {
...
+                for (url,info) {
+                    state.registered.insedrt(url,info);
+                }
+                Ok(json!({ "error": "insufficient endpoints available" }))
+            }
+            else {
+                Ok(json!({ "endpoints": endpoints }))
+            }
+        });
+    }
```
maackle: Seems like it would be nearly as easy to separate this out into its own binary. Is there a benefit to having this in the same binary as trycp_server? Seems totally orthogonal. But in any case if it's going to stay in the same binary, I would at least put an `else` block here and put all the other methods inside it, so that the thing acts as a manager XOR a server, never both.

zippy: The main reason is not to have to add any other tooling/nix-shell commands etc.  I was also thinking that the manager could be one of the available trycp endpoints at the same time, which is why I didn't do that.  But maybe that's a bad idea?

crates/trycp_server/src/main.rs
```diff
@@ -364,19 +363,26 @@ fn main() {
...
             let url_str = get_as_string("url", &params_map)?;
-            let url = Url::parse(&url_str).map_err(|e| {
+            let _url = Url::parse(&url_str).map_err(|e| {
                 invalid_request(format!("unable to parse url:{} got error: {}", url_str, e))
             })?;
-            let ram = get_as_int("ram", &params_map)?;
+            let ram = match get_as_int("ram", &params_map) {
+                Ok(val) => val,
+                _ => 0,
+            };
```
maackle:
```suggestion
            let ram = get_as_int("ram", &params_map).unwrap_or_default();
```

crates/trycp_server/Cargo.toml
```diff
@@ -12,15 +12,21 @@ documentation = "https://github.com/holochain/holochain-rust"
 repository = "https://github.com/holochain/holochain-rust"
 
 [dependencies]
+in_stream = { version = "=0.0.44-alpha3", path = "../in_stream" }
 log = "0.4.8"
 env_logger = "0.7.0"
 structopt = "0.2.15"
 jsonrpc-core = "14.0"
 jsonrpc-ws-server = "14.0"
+jsonrpc-lite = "=0.5.0"
```
neonphog: is this used, or leftover from iterating?

CHANGELOG-UNRELEASED.md
```diff
@@ -6,6 +6,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
+- Adds the notion of a manager to trycp_server so that we can dynamical manage pools available nodes for test runs in final-exam  [PR#2123](https://github.com/holochain/holochain-rust/pull/2123)
```
maackle:
```suggestion
- Adds the notion of a manager to trycp_server so that we can dynamically manage pools of available nodes for test runs in final-exam  [PR#2123](https://github.com/holochain/holochain-rust/pull/2123)
```

crates/trycp_server/Cargo.toml
maackle: Also, why was it not used, instead implementing our own JsonRPC structs?

crates/trycp_server/src/main.rs
```diff
@@ -237,19 +319,23 @@ fn save_file(file_path: PathBuf, content: &[u8]) -> Result<(), jsonrpc_core::typ
...
-        .expect("failed to execute process");
+fn get_info_as_json() -> Value {
+    let output = Command::new("holochain").args(&["-i"]).output();
+    if output.is_err() {
+        return Value::String("failed to execute holochain".into());
+    }
+    let output = output.unwrap();
     let info_str = String::from_utf8(output.stdout).unwrap();
 
-    // poor mans JSON convert
```
maackle: I would maintain that this is still a poor man's JSON convert, the comment was helpful to me in figuring out what's going on here and I think should stay

crates/trycp_server/src/main.rs
```diff
@@ -328,6 +528,7 @@ fn main() {
     });
 
     io.add_method("reset", move |params: Params| {
+        println!("reset");
```
maackle: still want this?

crates/trycp_server/test/test.js
```diff
@@ -124,3 +123,42 @@ sim2h_url = "wss://localhost:9001"
     })
     })
 }
+
+//doTestManager("ws://localhost:9000")
```
maackle: Looks like some use of commenting to manually change the way the test runs. Would it make sense to put this behind an if block? Or is this not needed anymore?

crates/trycp_server/test/test.js
```diff
@@ -124,3 +123,42 @@ sim2h_url = "wss://localhost:9001"
...
+      // call an RPC method with parameters
+      await ws.call('request', {"count": 1}).then(function(result) {
+        console.log(result.error)
+      })
+
+        // close a websocket connection
+        ws.close()
+
+        resolve()
+    })
```
maackle: Might as well indent this

crates/trycp_server/src/main.rs
```diff
@@ -161,6 +226,23 @@ fn get_as_string<T: Into<String>>(
...
+    let key = key.into();
+    Ok(params_map
+        .get(&key)
+        .ok_or_else(|| {
+            jsonrpc_core::Error::invalid_params(format!("`{}` param not provided", &key))
+        })?
+        .as_i64()
+        .ok_or_else(|| {
+            jsonrpc_core::Error::invalid_params(format!("`{}` has to be an integer", &key))
+        })?)
```
AshantiMutinta: I like this function a lot

crates/trycp_server/src/main.rs
```diff
@@ -271,6 +357,47 @@ fn os_eval(arbitrary_command: &str) -> String {
...
+    let mut connection: InStreamWss<InStreamTcp> =
+        InStreamWss::connect(&connection_uri, config).map_err(|e| format!("{}", e))?;
+
+    connection
+        .write(
+            serde_json::to_vec(&JsonRpcRequest::new("1", method, params))
+                .unwrap()
+                .into(),
+        )
+        .unwrap();
```
AshantiMutinta: This will panic is this fine?

crates/trycp_server/src/main.rs
```diff
@@ -271,6 +357,47 @@ fn os_eval(arbitrary_command: &str) -> String {
...
+    connection.flush().map_err(|e| format!("{}", e))?;
+
+    let mut res = WsFrame::default();
+    loop {
```
AshantiMutinta: this could probably be replaced with a 

while let Err(e) =connection.read()
{
//process errors 
}

this would get rid of the loop break but not entirely necessary to refactor

crates/trycp_server/Cargo.toml
zippy: I tried to use jsonrpc-lite and it didn't work over websockets, only over http, so that's why I didn't follow through.  Then I just took the sample work @neonphog did with jsonrpc for in_stream

PR #2124, opened 2020-02-21
Added update_entry with invalid content test

guillemcordoba:
Added a test to app_spec that fails proving `update_entry` inconsistency when updating with invalid content.

The failure scenario looks like this:

* Alice creates a post with valid content
* Alice updates that post with invalid content (in app_spec, more than 280 chars). This returns okey.
* Alice does a `get_entry` with the original address. The newer invalid content is returned.
* Bob does a `get_entry` with the original address. The older valid content is returned.


zippy:
@guillemcordoba hmmmm it looks like app_spec passed above!  Any thoughts on why?

guillemcordoba:
Yeah @zippy you're right! I saw that, I thought that it may be fixed on the 0.0.43-alpha3 version but I couldn't test that locally, I tested on 0.0.42-alpha5 with `build_and_test.sh` and it failed the test. Could this be on hold until I can test version 0.0.43-alpha3 locally and see if it fails?

guillemcordoba:
Hi @zippy, confirmed, bug appeared on `0.0.42-alpha5` and was fixed on `0.0.43-alpha3`, so does not show up anymore. Closing this now! Ping me if you want to add this test nonetheless for future regression tests.

PR #2133, opened 2020-02-25
Remove regex

AshantiMutinta:
## PR summary
Removed regex backend
## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation

### DIFF
--- a/crates/core_types/src/link/mod.rs
+++ b/crates/core_types/src/link/mod.rs
@@ -75,7 +75,6 @@ pub enum LinkActionKind {
 pub enum LinkMatch<S: Into<String>> {
     Any,
     Exactly(S),
-    Regex(S),
 }
 
 impl<S: Into<String>> LinkMatch<S> {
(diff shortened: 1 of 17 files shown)


crates/core_types/src/link/mod.rs
```diff
@@ -94,6 +92,15 @@ impl<S: Into<String>> LinkMatch<S> {
     }
 }
 
+impl<S: Into<String>> Into<Option<String>> for LinkMatch<S> {
+    fn into(self) -> Option<String> {
+        match self {
+            LinkMatch::Exactly(s) => Some(s.into()),
+            _ => None,
```
neonphog: Should we add the `Any` arm here?

crates/core_types/src/link/mod.rs
```diff
@@ -84,7 +83,6 @@ impl<S: Into<String>> LinkMatch<S> {
         let re_string: String = match self {
             LinkMatch::Any => ".*".into(),
             LinkMatch::Exactly(s) => "^".to_owned() + &regex::escape(&s.into()) + "$",
```
neonphog: do we still even need this function? - this is a pretty inefficient way to match a full string (similarly for the Any above).

crates/core_types/src/link/mod.rs
AshantiMutinta: Good catch! Added it

crates/core_types/src/link/mod.rs
AshantiMutinta: Yeah we don't need this anymore, wasnt using it anywhere

PR #2137, opened 2020-02-28
await on join

freesig:
## PR summary
This awaits on the new connection future instead of just spawning it because otherwise the `Lib3hToClient::HandleGetGossipingEntryList` and `Lib3hToClient::HandleGetAuthoringEntryList` messages get sent and can be responded to before the join is complete. This leads to the responses to those messages being thrown away. Then the agents id's never actually get into the DHT.
## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation

### DIFF
--- a/crates/sim2h/src/lib.rs
+++ b/crates/sim2h/src/lib.rs
@@ -311,18 +311,19 @@ impl Sim2hHandle {
         tokio::task::spawn(async move {
             // -- right now each agent can only be part of a single space :/ --
 
-            let (agent_id, space_hash) = 'got_info: {
-                // TODO might need to up this amount of tries if we get the below error
-                for _ in 0_usize..10 {
-                    // await consistency of new connection
-                    let state = sim2h_handle.state().get_clone().await;
-                    if let Some(info) = state.get_space_info_from_uri(&uri) {
-                        break 'got_info info;
-                    }
-                    tokio::time::delay_for(std::time::Duration::from_millis(100)).await;
+            let (agent_id, space_hash) = {
+                let state = sim2h_handle.state().get_clone().await;
+                if let Some(info) = state.get_space_info_from_uri(&uri) {
+                    info
+                } else {
+                    error!(
+                        "uri has not joined space, cannot proceed {} {}",
+                        uri,
+                        message.message_type()
+                    );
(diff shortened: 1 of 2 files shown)


crates/sim2h/src/sim2h_im_state.rs
```diff
@@ -882,9 +882,14 @@ impl StoreHandle {
...
-    pub fn spawn_new_connection(&self, space_hash: SpaceHash, agent_id: AgentId, uri: Lib3hUri) {
+    pub async fn spawn_new_connection(
+        &self,
+        space_hash: SpaceHash,
+        agent_id: AgentId,
+        uri: Lib3hUri,
+    ) {
         let f = self.new_connection(space_hash, agent_id, uri);
-        tokio::task::spawn(f);
+        f.await
```
neonphog: My intention was that functions prefixed with `spawn_` would not return futures (or anything else) because they spawn the task.

If we don't need this, we can just delete this whole function - and lib.rs can call `new_connection()` directly : )

crates/sim2h/src/lib.rs
```diff
@@ -522,17 +531,20 @@ fn spawn_handle_message_hello(
     }
 }
 
-fn spawn_handle_message_join_space(
+async fn spawn_handle_message_join_space(
```
neonphog: similarly, remove the `spawn_` prefix here if that's not what we're doing.

crates/sim2h/src/lib.rs
```diff
@@ -308,6 +316,7 @@ impl Sim2hHandle {
             // -- right now each agent can only be part of a single space :/ --
 
             let (agent_id, space_hash) = 'got_info: {
+                // TODO might need to up this amount of tries if we get the bellow error
```
Connoropolous:
```suggestion
                // TODO might need to up this amount of tries if we get the below error
```

:)

crates/sim2h/src/lib.rs
```diff
@@ -308,6 +312,7 @@ impl Sim2hHandle {
             // -- right now each agent can only be part of a single space :/ --
 
             let (agent_id, space_hash) = 'got_info: {
+                // TODO might need to up this amount of tries if we get the below error
```
maackle: can probably just remove this loop now

PR #2139, opened 2020-03-02
Sim2h URL property override

zippy:
## PR summary

Adds an environment variable for overriding the sim2h_url value if set in the DNA property.  We need this for running final exam tests without changing DNA.

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation

### DIFF
--- a/crates/core/src/network/reducers/init.rs
+++ b/crates/core/src/network/reducers/init.rs
@@ -203,6 +203,7 @@ pub mod test {
 
         assert_eq!(result, ());
 
+        std::env::remove_var("HC_IGNORE_SIM2H_URL_PROPERTY");
         let network = network_state.network.expect("No network connection set");
         assert_eq!(network.p2p_endpoint().as_str(), "wss://localhost:9000/");
     }
--- a/crates/core/src/network/reducers/init.rs
+++ b/crates/core/src/network/reducers/init.rs
@@ -199,11 +199,11 @@ pub mod test {
 
         root_state = root_state.reduce(ActionWrapper::new(Action::InitializeChain(dna)));
 
+        std::env::remove_var("HC_IGNORE_SIM2H_URL_PROPERTY");
         let result = reduce_init(&mut network_state, &root_state, &action_wrapper);
 
         assert_eq!(result, ());
 
-        std::env::remove_var("HC_IGNORE_SIM2H_URL_PROPERTY");
         let network = network_state.network.expect("No network connection set");
         assert_eq!(network.p2p_endpoint().as_str(), "wss://localhost:9000/");
     }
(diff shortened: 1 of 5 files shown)


crates/core/src/network/reducers/init.rs
```diff
@@ -41,13 +41,18 @@ pub fn reduce_init(state: &mut NetworkState, root_state: &State, action_wrapper:
...
+                info!(
+                    "Found property 'sim2h_url' in DNA {} - overriding conductor wide sim2h URL with: {}",
+                    dna.address(),
+                    sim2h_url,
+                );
+                // ..override the conductor wide setting.
+                sim2h_config.sim2h_url = sim2h_url;
+            } else {
+                debug!(
+                    "DNA has 'sim2h_url' override property set, but it's ignored because HC_INGORE_SIM2H_URL_PROPERTY is set");
```
maackle:
```suggestion
                    "DNA has 'sim2h_url' override property set, but it's ignored because HC_IGNORE_SIM2H_URL_PROPERTY is set");
```

crates/core/src/network/reducers/init.rs
```diff
@@ -41,13 +41,18 @@ pub fn reduce_init(state: &mut NetworkState, root_state: &State, action_wrapper:
...
         // ..and we're configured to use sim2h...
         if let BackendConfig::Sim2h(sim2h_config) = &mut p2p_config.backend_config {
-            info!(
-                "Found property 'sim2h_url' in DNA {} - overriding conductor wide sim2h URL with: {}",
-                dna.address(),
-                sim2h_url,
-            );
-            // ..override the conductor wide setting.
-            sim2h_config.sim2h_url = sim2h_url;
+            if (std::env::var("HC_IGNORE_SIM2H_URL_PROPERTY").is_err()) {
```
AshantiMutinta:
```suggestion
            if std::env::var("HC_IGNORE_SIM2H_URL_PROPERTY").is_err() {
```

crates/core/src/network/reducers/init.rs
```diff
@@ -41,13 +41,18 @@ pub fn reduce_init(state: &mut NetworkState, root_state: &State, action_wrapper:
...
+                    "Found property 'sim2h_url' in DNA {} - overriding conductor wide sim2h URL with: {}",
+                    dna.address(),
+                    sim2h_url,
+                );
+                // ..override the conductor wide setting.
+                sim2h_config.sim2h_url = sim2h_url;
+            } else {
+                debug!(
+                    "DNA has 'sim2h_url' override property set, but it's ignored because HC_IGNORE_SIM2H_URL_PROPERTY is set");
+            }
```
AshantiMutinta: Does this mean we don't really care what value HC_IGNORE_SIM2H_URL_PROPERTY is set to, only as long as it exists in our env variables?

zippy: good point, I was being lazy and sloppy.  Fixed.

PR #2143, opened 2020-03-06
remove newrelic from holochain

zippy:
## PR summary

Removes new-relic building completely from conductor for now.

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation


Connoropolous:
what was newrelic there for, and why take it out?

zippy:
> what was newrelic there for, and why take it out?

We added it for debugging purposes to try and track down some cross-process problems with the tracing it offeres, however it's causing problems in our holoport build chain so we need to clean it out for now.  In the "production" context rewrelic doesn't make any sense for holochain at all because it's centralized monitoring.  It may make sense to add back into sim2h_server which is a centralized service so that it can be monitored.

freesig:
This will need to be removed from holochain tracing aswell.
Let's just put it the deps under a conditional compile and make the new relic a pass through so you don't need to comment all these out

zippy:
obsolete

PR #2148, opened 2020-03-11
IOEerror from handle_fetch_entry of remove link aspect

zippy:
## PR summary

This PR fixes the IOError we saw in live testing and which was reliably reproduceable by [this holofuel test](https://github.com/Holo-Host/holofuel/pull/67)

The error was caused by there being unreachable code in `network/handler/mod.rs` which caused the network thread to panic silently leaving the conductor in a bad state.

Additionally looking at this code revealed a bunch of weaknesses in the code in the function `get_meta_aspects_from_dht_eav`, especially assumptions that the use of `get_entry_with_meta_workflow` as part of servicing the `handle_fetch_entry`.  `get_entry_with_meta_workflow` will try to get data locally first, but in the case of `handle_fetch_entry` you can't ask the network to get data for you that the network is asking you for. Thus, this PR also includes a refactor of that function to `get_entry_with_meta_workflow_local`.

## testing/benchmarking notes

This PR was tested against  [this holofuel test](https://github.com/Holo-Host/holofuel/pull/67), and also against final-exam.  Final exam holofuel-stress test is still failing above 20 nodes with timeouts, so unfortunately this fix doesn't address what's going on there as I originally hoped it would.

## followups
- [ ] **Fix removed link source:** Looking at this code also revealed an incorrect implementation of remove links, namely that the LinkAdd entries are retrieved by [a query](https://github.com/holochain/holochain-rust/blob/391c42362a5632388b16b8bcf8ec3ad98331109d/crates/core/src/wasm_engine/api/remove_link.rs#L77).  @artbrock says that not only is this incorrect but that there should only be 1 LinkAdd removed per call, not many, but it should not be looked up but passed in.  This is an hdk level change that needs to be made, and app_spec tests also need to be fixed as they test this assumption.

## changelog

```markdown
```

## documentation


crates/core/src/network/handler/mod.rs
```diff
@@ -488,22 +488,49 @@ fn get_meta_aspects_from_dht_eav(
...
+                    match value_entry.entry_with_meta.entry {
+                        Entry::LinkRemove((link_data, removed_link_entries)) => {
+                            Ok(EntryAspect::LinkRemove(
+                                (link_data.clone(), removed_link_entries.clone()),
+                                header,
+                            ))
+                        },
+                        Entry::LinkAdd(link_data) => {
+                            //bad = true;
+//                            debug!("didn't get an expected LinkRemove, got: {:?}",value_entry.clone());
```
AshantiMutinta: We have to do a get_links here using entry adress, link tag, link type and then match on the target to get this array of values

crates/core/src/network/handler/mod.rs
```diff
@@ -483,27 +483,49 @@ fn get_meta_aspects_from_dht_eav(
...
-                        unwrap_to!(value_entry.entry_with_meta.entry => Entry::LinkRemove);
-                    Ok(EntryAspect::LinkRemove(
-                        (link_data.clone(), removed_link_entries.clone()),
-                        header,
-                    ))
+                Attribute::LinkTag(_, _) => match value_entry.entry_with_meta.entry {
+                    Entry::LinkAdd(link_data) | Entry::LinkRemove((link_data, _)) => {
+                        Ok(EntryAspect::LinkAdd(link_data, header))
+                    }
+                    _ => Err(HolochainError::from("Invalid Entry Value for Link Add")),
```
maackle: this error should probably include the invalid data

crates/core/src/network/handler/query.rs
```diff
@@ -26,7 +26,7 @@ use std::{convert::TryInto, sync::Arc};
 
 pub type LinkTag = String;
 #[holochain_tracing_macros::newrelic_autotrace(HOLOCHAIN_CORE)]
-fn get_links(
+pub fn get_links(
```
maackle: this can be `pub(crate)` right?

crates/core/src/network/handler/mod.rs
maackle:
```suggestion
                    _ => Err(HolochainError::from(format!("Invalid Entry Value for Link Add: {:?}", link_data))),
```

crates/core/src/network/handler/mod.rs
```diff
@@ -483,27 +483,49 @@ fn get_meta_aspects_from_dht_eav(
...
+                            //not very efficient but nice to fall back on
+                            let filtered_links = links
+                                .into_iter()
+                                .filter(|link_for_filter| {
+                                    &link_for_filter.target == link_data.link().target()
+                                })
+                                .map(|s| s.address)
+                                .collect::<Vec<_>>();
+
+                            Ok(EntryAspect::LinkRemove((link_data, filtered_links), header))
```
maackle: There is so much going on here, I know it's logic to deal with tombstones, but it's so unintuitive that a Entry::LinkAdd would lead to a EntryAspect::LinkRemove, can you add some comments as to what's going on here?

crates/core/src/network/handler/mod.rs
zippy: We can't do this.  It doesn't make sense to ask the network for the links when the network is asking us for data, remember this is part of a handle_fetch_entry call

crates/core/src/network/handler/query.rs
zippy: removed

thedavidmeister:
i agree with @neonphog i approved before i realised that was an option :)

zippy:
Added back!

PR #2153, opened 2020-03-13
Fix update aspects

maackle:
## PR summary

Extending the work done in `localize-handle-query-workflows`, updates the logic for `EntryAspect::entry_address()` to accurately match the `entry_to_meta_aspect`, which it is basically the inverse function of.

## changelog

```markdown
```

## documentation

### DIFF
--- a/crates/core/src/network/handler/mod.rs
+++ b/crates/core/src/network/handler/mod.rs
@@ -407,7 +407,7 @@ fn entry_to_meta_aspect(entry: Entry, header: ChainHeader) -> Option<(Address, E
         Entry::Deletion(_) => Some(EntryAspect::Deletion(header)),
         _ => None,
     }
-    .map(|aspect| (aspect.entry_address().clone(), aspect))
+    .map(|aspect| (aspect.entry_address(), aspect))
 }
 
 #[holochain_tracing_macros::newrelic_autotrace(HOLOCHAIN_CORE)]
--- a/crates/core/src/network/handler/mod.rs
+++ b/crates/core/src/network/handler/mod.rs
@@ -395,8 +395,11 @@ fn get_content_aspect(
 /// NB: this is the inverse function of EntryAspect::entry_address(), so it is very important
 /// that they agree!
 #[holochain_tracing_macros::newrelic_autotrace(HOLOCHAIN_CORE)]
-fn entry_to_meta_aspect(entry: Entry, header: ChainHeader) -> Option<(Address, EntryAspect)> {
-    match entry {
+fn entry_to_meta_aspect(
+    entry: Entry,
+    header: ChainHeader,
+) -> Result<Option<(Address, EntryAspect)>, HolochainError> {
+    let maybe_aspect = match entry {
         Entry::App(app_type, app_value) => header
             .link_update_delete()
             .map(|_| EntryAspect::Update(Entry::App(app_type, app_value), header)),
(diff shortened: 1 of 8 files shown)


crates/core/src/network/handler/mod.rs
```diff
@@ -395,27 +395,25 @@ fn get_content_aspect(
...
-        )),
-        Entry::LinkRemove((link_data, addresses)) => Some((
-            link_data.link.base().clone(),
-            EntryAspect::LinkRemove((link_data, addresses), header),
-        )),
-        Entry::Deletion(_) => Some((
-            header.link_update_delete().expect(""),
-            EntryAspect::Deletion(header),
-        )),
+fn entry_to_meta_aspect(
```
maackle: this is the entry -> meta aspect map

crates/core_types/src/network/entry_aspect.rs
```diff
@@ -95,12 +97,24 @@ impl EntryAspect {
     }
     /// NB: this is the inverse function of entry_to_meta_aspect,
     /// so it is very important that they agree!
-    pub fn entry_address(&self) -> &Address {
-        match self {
-            EntryAspect::LinkAdd(link_data, _) => link_data.link.base(),
-            EntryAspect::LinkRemove((link_data, _), _) => link_data.link.base(),
-            _ => self.header().entry_address(),
-        }
+    pub fn entry_address(&self) -> Result<Address, HolochainError> {
```
maackle: and here is the aspect -> entry map, which the inverse map now uses to maintain agreement between the two

crates/core_types/src/network/entry_aspect.rs
```diff
@@ -95,12 +97,24 @@ impl EntryAspect {
...
+            EntryAspect::Content(_, header) => match header.link_update_delete() {
+                Some(ref updated_entry) => updated_entry.clone(),
+                None => header.entry_address().clone(),
+            },
+            EntryAspect::LinkAdd(link_data, _) => link_data.link.base().clone(),
+            EntryAspect::LinkRemove((link_data, _), _) => link_data.link.base().clone(),
+            EntryAspect::Update(_, header) | EntryAspect::Deletion(header) => {
+                header.link_update_delete().ok_or_else(|| {
+                    HolochainError::ErrorGeneric(format!(
+                        "no link_update_delete on Deletion entry header. Header: {:?}",
```
neonphog: error could be confusing if this is an `Update` type

crates/core_types/src/network/entry_aspect.rs
```diff
@@ -95,12 +97,24 @@ impl EntryAspect {
...
+            EntryAspect::LinkRemove((link_data, _), _) => link_data.link.base().clone(),
+            EntryAspect::Update(_, header) | EntryAspect::Deletion(header) => {
+                header.link_update_delete().ok_or_else(|| {
+                    HolochainError::ErrorGeneric(format!(
+                        "no link_update_delete on Update/Deletion entry header. Header: {:?}",
+                        header
+                    ))
+                })?
+            }
+            EntryAspect::Header(header) => header.entry_address().clone(),
```
zippy: Is this correct?  Is entry address of a header aspect supposed to be the address of the entry?  Also I note that there isn't an inverse for the Header aspect in `entry_to_meta_aspect` so this worries me that we have it here.

maackle: You're right, it's not correct. There's no inverse I guess because Header is not a "meta" aspect. And EntryAspect::Header is unused, anyway. I fixed it here though.

PR #2164, opened 2020-03-19
Bounded

freesig:
## PR summary
Adds bounded channels to prevent memory building up in unbounded channels
## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation

### DIFF
--- a/crates/sim2h/src/lib.rs
+++ b/crates/sim2h/src/lib.rs
@@ -52,6 +52,7 @@ use tracing_futures::Instrument;
 
 /// use the default 0 seed for xxHash
 pub const RECEIPT_HASH_SEED: u64 = 0;
+pub const CHANNEL_SIZE: usize = 1000;
 
 /// Generates a u64 hash response for an `Ack` message given input bytes
 pub fn generate_ack_receipt_hash(payload: &Opaque) -> u64 {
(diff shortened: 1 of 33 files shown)


crates/sim2h/src/lib.rs
```diff
@@ -626,46 +628,51 @@ async fn handle_message_join_space(
 ) {
     sim2h_handle
         .state()
+        .clone()
```
neonphog: uhhh - if we clone the state here the new connection will get added to the clone and not the actual state - then get dropped.

neonphog: scratch that - I see this is cloning the handle not the state : )

neonphog:
The sim2h side looks awesome! nice work async-ifying everything!

zippy:
When I test manually I see lots of problems with not being able to get a lock on state in the action loop.  Also you can't ctrl-c out of the conductor any more.

neonphog:
Using `spawn_blocking` for the for the crossbeam_channel blocking sends actually makes this playground work - although I at least get three would-block thread panics - so maybe 1000 is more than it can keep up with?

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=579de08c026ae6005e1c0ffdb319be54

freesig:
Oh yeh I haven't figured out how to get the shutdown to happen on this properly yet. Maybe test it if and then if it's helpful I can look at how to do that.

PR #2183, opened 2020-04-24
don't get links from network if we are the authority

zippy:
## PR summary

This PR adds:

- [x] an optimization of not querying the DHT for get links requests when we know that we are the DHT authority for the base of the query.
- [x] two conductor fixes during error cases (like base hasn't arrived yet) in hold_aspects request:
    - stop incorrectly recording aspects as held
    - stop locking up the future
- [x] because sim2h doesn't get error messages from a hold_aspect request it must also no longer records aspects sent in those messages as held, and thus the conductor must explicitly send back a list of aspects held after a hold_aspect request
- [x] get authoring list doesn't really work because correct items won't be gossiped and the optimization now breaks this.  Fix this by not responding with the list, and instead re-publishing.

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation

### DIFF
--- a/crates/core/src/dht/actions/hold_aspect.rs
+++ b/crates/core/src/dht/actions/hold_aspect.rs
@@ -1,7 +1,7 @@
 use crate::{
     action::{Action, ActionWrapper},
     context::Context,
-    dht::aspect_map::AspectMap,
+    dht::{aspect_map::AspectMap, dht_store::HoldAspectAttemptId},
     instance::dispatch_action,
 };
 use futures::{future::Future, task::Poll};
@@ -11,9 +11,13 @@ use lib3h_protocol::data_types::EntryListData;
 use snowflake::ProcessUniqueId;
 use std::{pin::Pin, sync::Arc};
 
-pub async fn hold_aspect(aspect: EntryAspect, context: Arc<Context>) -> Result<(), HolochainError> {
-    let id = ProcessUniqueId::new();
-    let action_wrapper = ActionWrapper::new(Action::HoldAspect((aspect.clone(), id)));
+pub async fn hold_aspect(
+    pending_id: &ProcessUniqueId,
+    aspect: EntryAspect,
+    context: Arc<Context>,
+) -> Result<(), HolochainError> {
+    let id = (pending_id.clone(), ProcessUniqueId::new());
+    let action_wrapper = ActionWrapper::new(Action::HoldAspect((aspect.clone(), id.clone())));
     dispatch_action(context.action_channel(), action_wrapper.clone());
     let r = HoldAspectFuture {
         context: context.clone(),
@@ -51,7 +55,7 @@ pub async fn hold_aspect(aspect: EntryAspect, context: Arc<Context>) -> Result<(
 pub struct HoldAspectFuture {
     context: Arc<Context>,
     //    aspect: EntryAspect,
-    id: ProcessUniqueId,
+    id: HoldAspectAttemptId,
 }
 
 #[holochain_tracing_macros::newrelic_autotrace(HOLOCHAIN_CORE)]
@@ -63,11 +67,11 @@ impl Future for HoldAspectFuture {
             return Poll::Ready(Err(err));
         }
         self.context
-            .register_waker(self.id.clone(), cx.waker().clone());
+            .register_waker(self.id.1.clone(), cx.waker().clone());
         if let Some(state) = self.context.try_state() {
             // wait for the request to complete
             if let Some(result) = state.dht().hold_aspec_request_complete(&self.id) {
-                self.context.unregister_waker(self.id.clone());
+                self.context.unregister_waker(self.id.1.clone());
                 Poll::Ready(result.clone())
             } else {
                 Poll::Pending
--- a/crates/core/src/dht/actions/hold_aspect.rs
+++ b/crates/core/src/dht/actions/hold_aspect.rs
@@ -5,6 +5,8 @@ use crate::{
 };
 use futures::{future::Future, task::Poll};
 use holochain_core_types::{error::HolochainError, network::entry_aspect::EntryAspect};
+use holochain_persistence_api::cas::content::AddressableContent;
+use lib3h_protocol::data_types::EntryListData;
 use snowflake::ProcessUniqueId;
 use std::{pin::Pin, sync::Arc};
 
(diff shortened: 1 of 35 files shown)


crates/core/src/workflows/get_link_result.rs
```diff
@@ -10,11 +11,20 @@ use crate::{
...
     GetLinksArgs, GetLinksResult, LinksResult,
 };
 use std::sync::Arc;
 
+// check to see if we are an authority on the DHT for this base, if so no need
+// to go out to request this from anybody else.  We know for sure that we are an authority
+// for anything that is linked on our own agent hash.
+pub fn am_i_dht_authority_for_base<'a>(context: &'a Arc<Context>, base: &Address) -> bool {
+    let me: Address = context.agent_id.address();
+    *base == me
```
lucksus, line 25: Hm, this actually does not check if we are a DHT authority for the base, which we easily could by testing if base is included in the holding list. Shouldn‘t we do that here?

zippy: added

crates/core/src/dht/actions/hold_aspect.rs
```diff
@@ -1,28 +1,56 @@
...
+    } else {
+        // send a gossip list with this aspect in it back to sim2h so it know we are holding it
+        let c = context.clone();
+        let closure = async move || {
+            let state = context
+                .state()
+                .expect("No state present when trying to respond with gossip list");
+
+            let mut address_map = AspectMap::new();
+            address_map.add(&aspect);
```
lucksus, line 35: Huh, this sends back a GossipList with just this one item in it? Isn't this causing sim2h to be confused about the gossip list of this node?

I really think we shouldn't do this (managing results of holding workflows) here at all and instead have sim2h properly (if it isn't) poll for gossip lists.

crates/core/src/dht/dht_reducers.rs
```diff
@@ -138,14 +143,25 @@ pub(crate) fn reduce_hold_aspect(
...
             None
         }
+    };
+    if let Some(ref mut store) = r {
+        store.mark_aspect_as_held(&aspect);
+        store.mark_hold_aspect_complete(id.clone(), hold_result);
+        r
+    } else {
+        let mut store = (*old_store).clone();
+        store.mark_hold_aspect_complete(id.clone(), hold_result);
```
lucksus, line 163: Despite my last comment, I think it is a good thing to store the result of the holding workflow in the state as done here. (So we don't [always] rerun an already failed validation - and for debugging). I still think introducing the complexity of getting this result back to sim2h as a response for the holding request is too brittle and having sim2h (re-)poll for the gossip list (instead of marking something as hold because the request was sent) is the real fix here.

crates/core/src/dht/dht_store.rs
```diff
@@ -49,6 +50,9 @@ pub struct DhtStore {
     /// All the entry aspects that the network has told us to hold
     holding_map: AspectMap,
 
+    /// Hold aspect requests from the network
+    queued_hold_aspect_requests: HashMap<ProcessUniqueId, Result<(), HolochainError>>,
```
lucksus, line 54: Confused about the name. Should this be something like `holding_attempt_results`?

crates/core/src/network/handler/lists.rs
```diff
@@ -18,26 +19,53 @@ use lib3h_protocol::{
...
     let c = context.clone();
     let closure = async move || {
-        let address_map = create_authoring_map(context.clone());
-
-        let action = Action::RespondAuthoringList(EntryListData {
+        // let address_map = create_authoring_map(context.clone());
+
+        // currently sim2h asks for the authoring list and treats it just the same
+        //as the gossiping list, i.e. as data that you hold.  This is a problem because
+        //it means that gossiping isn't actually working right.  So instead of fixing that in
```
lucksus, line 31: Woah, if sim2h is doing, that is a problem and I would consider it a defect.

Last time I looked, sim2h was using the authoring list to check if this node has something new that the DHT does not store yet, such that sim2h would request that data from the node and publish it to the according authorities.

crates/core/src/network/handler/lists.rs
```diff
@@ -18,26 +19,53 @@ use lib3h_protocol::{
...
+                }
+            }
+            let result = publish_header_entry(address, &context).await;
+            if result.is_err() {
+                error!(
+                    "Error publishing header during get authoring list: {:?}",
+                    result
+                );
+            }
+        }
```
lucksus, line 61: I would expect this hack to cause much more strain on the whole network since sim2h handles publishes different to how it (used to) handle(s) authoring lists. Doesn't publishes always result in store requests send to the according DHT authorities? Which would now happen every time a node comes online...

I don't think this is the right solution. If sim2h does not treat authoring lists as authoring lists anymore, that should get back in.

crates/sim2h/src/lib.rs
```diff
@@ -801,12 +801,13 @@ fn spawn_handle_message_publish_entry(
...
                 sim2h_handle.send((&*agent_id).clone(), uri.clone(), &multi_message);
             }
+            /* send not guaranteed to work so we can't mark as held.
             sim2h_handle.state().spawn_agent_holds_aspects(
                 (&*space_hash).clone(),
                 (&*agent_id).clone(),
                 data.entry.entry_address.clone(),
                 aspect_list.clone(),
-            );
+            );*/
```
lucksus, line 810: Yes, I regard this as the real fix for handling the case of failed holding requests.

crates/core/src/dht/actions/hold_aspect.rs
zippy: Currently the gossiplist is additive on the sim2h side so this really acts as an acknowledgement of the hold_aspect request.

neonphog: Sim2h's state knowledge is append-only.

We were previously setting "agent-holding" when we sent the hold request, but that could fail - it's much more accurate for core to send a sparse holding list that results in an additional hold record.

crates/core/src/network/handler/lists.rs
neonphog: We cannot have authors marked as holding the data - because they are not authoritative - the meta info will be incomplete.

Because Sim2h maintains central knowledge of who holds what, a publish should become a no-op if the item has already saturated a neighborhood.

crates/core/src/dht/dht_reducers.rs
zippy: No, that can't work in the end.  Remember that sim2h is where the intelligence lies of who should hold what.  It has the knowledge of who is in the sharding neighborhood.  The only reason it asks for a gossip list is to update it's understanding of who is actually holding what data so it can correctly forward queries.  But at the same time it should be making the decisions about who SHOULD hold what.  These fixes are about getting that right.

crates/core/src/network/handler/lists.rs
zippy: Yes, that was incorrect. We could, in the sim2h case actually treat it that way, and we could add to sim2h the tracking of who authored what so that it does treat authors as authoritative on the DHT.   The hack of having the conductor simply republish everything it authored when it receives the authoring list request serves the exact same purpose as what say above as long as sim2h also checks to not send the data again to nodes that it knows it's already sent the publish data too.  I will be adding that.

crates/core/src/dht/dht_reducers.rs
lucksus: How is this affecting the network code's reasoning about who should hold what?

zippy:
This work has been split into multiple PRs: #2184 #2189

PR #2199, opened 2020-06-26
Validation timeout fixes

zippy:
## PR summary

This PR fixes some long-standing problems in validation:
- Increases timeout getting headers when building a validation package from the dht (1 second isn't enough)
- Pending items were being de-queued when processing, which meant that if another request arrived from sim2h (which could happen often under some circumstances), then the same item would be queued multiple times, which could snowball.  The solution was to move in-process items to another queue for checking.

And one bug in DHT queries:
- Fixed DHT queries for ChainHeader entries which were failing because there are no headers for headers stored in the DHT which the code was expecting.

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation

### DIFF
--- a/crates/core/src/dht/actions/remove_queued_holding_workflow.rs
+++ b/crates/core/src/dht/actions/remove_queued_holding_workflow.rs
@@ -6,16 +6,31 @@ use crate::{
 };
 use futures::{future::Future, task::Poll};
 use snowflake::ProcessUniqueId;
-use std::{pin::Pin, sync::Arc};
+use std::{pin::Pin, sync::Arc, time::Duration};
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub enum HoldingWorkflowQueueing {
+    Processing,
+    Waiting(Duration),
+    Done,
+}
 
 #[holochain_tracing_macros::newrelic_autotrace(HOLOCHAIN_CORE)]
-pub async fn remove_queued_holding_workflow(pending: PendingValidation, context: Arc<Context>) {
-    let action_wrapper = ActionWrapper::new(Action::RemoveQueuedHoldingWorkflow(pending.clone()));
+pub async fn remove_queued_holding_workflow(
+    state: HoldingWorkflowQueueing,
+    pending: PendingValidation,
+    context: Arc<Context>,
+) {
+    let action_wrapper = ActionWrapper::new(Action::RemoveQueuedHoldingWorkflow((
+        state.clone(),
+        pending.clone(),
+    )));
     dispatch_action(context.action_channel(), action_wrapper.clone());
     let id = ProcessUniqueId::new();
     RemoveQueuedHoldingWorkflowFuture {
         context,
         pending,
+        state,
         id,
     }
     .await
@@ -24,6 +39,7 @@ pub async fn remove_queued_holding_workflow(pending: PendingValidation, context:
 pub struct RemoveQueuedHoldingWorkflowFuture {
     context: Arc<Context>,
     pending: PendingValidation,
+    state: HoldingWorkflowQueueing,
     id: ProcessUniqueId,
 }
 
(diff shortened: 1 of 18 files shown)


crates/core/src/dht/actions/remove_queued_holding_workflow.rs
```diff
@@ -36,11 +52,36 @@ impl Future for RemoveQueuedHoldingWorkflowFuture {
...
+                HoldingWorkflowQueueing::Waiting(_) => {
+                    if !store.has_exact_queued_holding_workflow(&self.pending)
+                        || store.has_exact_in_process_holding_workflow(&self.pending)
+                    {
+                        Poll::Pending
+                    } else {
+                        self.context.unregister_waker(self.id.clone());
+                        Poll::Ready(())
+                    }
+                }
```
maackle, line 76: Looks like these branches are identical?

```suggestion
                HoldingWorkflowQueueing::Processing | HoldingWorkflowQueueing::Waiting(_) => {
                    if store.has_exact_queued_holding_workflow(&self.pending)
                        || !store.has_exact_in_process_holding_workflow(&self.pending)
                    {
                        Poll::Pending
                    } else {
                        self.context.unregister_waker(self.id.clone());
                        Poll::Ready(())
                    }
                }
```

crates/core/src/dht/dht_store.rs
```diff
@@ -76,13 +78,15 @@ impl PartialEq for DhtStore {
...
     queued_holding_workflows: VecDeque<PendingValidationWithTimeout>,
+    in_process_holding_workflows: VecDeque<PendingValidationWithTimeout>,
 }
 
 impl From<&StateWrapper> for DhtStoreSnapshot {
     fn from(state: &StateWrapper) -> Self {
         DhtStoreSnapshot {
             holding_map: state.dht().get_holding_map().bare().clone(),
             queued_holding_workflows: state.dht().queued_holding_workflows.clone(),
+            in_process_holding_workflows: state.dht().queued_holding_workflows.clone(),
```
maackle, line 89:
```suggestion
            in_process_holding_workflows: state.dht().in_process_holding_workflows.clone(),
```

?

crates/core/src/instance.rs
```diff
@@ -41,9 +40,7 @@ use std::{
     time::{Duration, Instant},
 };
 
-pub const RECV_DEFAULT_TIMEOUT_MS: Duration = Duration::from_millis(10000);
-
-pub const RETRY_VALIDATION_DURATION_MIN: Duration = Duration::from_millis(500);
+pub const RETRY_VALIDATION_DURATION_MIN: Duration = Duration::from_millis(15000);
```
maackle, line 43: Is it significant that now the MIN is much larger than the MAX? (15000 > 3600)

crates/core/src/dht/dht_store.rs
zippy: Whoa!  Good catch.

crates/core/src/instance.rs
zippy: Oh.  max is in seconds, not millis.  Min is much less that max!

crates/core/src/dht/actions/remove_queued_holding_workflow.rs
```diff
@@ -6,16 +6,31 @@ use crate::{
 };
 use futures::{future::Future, task::Poll};
 use snowflake::ProcessUniqueId;
-use std::{pin::Pin, sync::Arc};
+use std::{pin::Pin, sync::Arc, time::Duration};
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub enum HoldingWorkflowQueueing {
```
maackle, line 12: Comments on enums and functions would be tremendously helpful in understanding this stuff and conveying intent. I know most stuff is undocumented, but at least documenting new stuff would help

crates/core/src/dht/dht_store.rs
zippy: fixed in e67919d

crates/core/src/dht/actions/remove_queued_holding_workflow.rs
zippy: Nope, just not well refactored, the ! are on opposite queues because the operation is done when the item moves from one queue to the other.

maackle: Whoops :)

crates/core/src/dht/actions/remove_queued_holding_workflow.rs
```diff
@@ -36,11 +57,36 @@ impl Future for RemoveQueuedHoldingWorkflowFuture {
...
+                        || store.has_exact_in_process_holding_workflow(&self.pending)
+                    {
+                        Poll::Pending
+                    } else {
+                        self.context.unregister_waker(self.id.clone());
+                        Poll::Ready(())
+                    }
+                }
+                HoldingWorkflowQueueing::Done => {
+                    if store.has_exact_in_process_holding_workflow(&self.pending) {
```
maackle, line 83: Can a PendingValidation ever go straight from Waiting (in the holding_queue) to Done? If not, this is fine, but if so, this isn't checking that.

crates/core/src/instance.rs
maackle: whoops again!

PR #2202, opened 2020-07-07
Cache validation packages

zippy:
## PR summary

This PR adds caching of FullChain ValidationPackages received from agents so that on subsequent attempts get data from the same agent we can skip the network roud-trip.  This is especially important when nodes first come on any may have to hold lots of data from a given node.  This can drastically reduce the number of network round-trips.

- [x] validation data gets cached from direct messages from authors
- [ ] validation data gets cached when rebuilt from dht (may not implement)
- [ ] works for header only validation_package type (may not implement)

### N.B. App-spec failing
This PR includes changing the sim2h RETRY_FETCH_MISSING_ASPECTS_INTERVAL_MS from .5s to 30 seconds. This is necessary for systemic health (especially as new nodes come online), but causes a number of app-spec tests to fail due to timing issues.

## followups

## changelog

```markdown
```

## documentation

### DIFF
--- a/crates/core/src/network/state.rs
+++ b/crates/core/src/network/state.rs
@@ -3,7 +3,13 @@ use crate::{
     network::{actions::Response, direct_message::DirectMessage, query::NetworkQueryResult},
 };
 use boolinator::*;
-use holochain_core_types::{error::HolochainError, validation::ValidationPackage};
+use holochain_core_types::{
+    chain_header::ChainHeader,
+    entry::Entry,
+    error::HolochainError,
+    validation::{ValidationPackage, ValidationPackageDefinition},
+};
+use holochain_json_api::{error::JsonError, json::JsonString};
 use holochain_net::p2p_network::P2pNetwork;
 use holochain_persistence_api::cas::content::Address;
 use im::HashMap;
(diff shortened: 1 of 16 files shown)


crates/core/src/workflows/mod.rs
```diff
@@ -134,10 +134,36 @@ async fn validation_package(
         return Ok(Some(package));
     }
 
+    log_debug!(
```
neonphog, line 137:
```suggestion
    // 1.5. Fetch a cached version of the validation package if cache version is warm
    // This mitigates hitting the author too frequently for the same validation package.
    log_debug!(
```

crates/core/src/network/state.rs
```diff
@@ -85,4 +108,96 @@ impl NetworkState {
...
+            debug!(
+                "Vcache: initial insert {}:{:?} {:#?}",
+                &agent, validation.chain_header, &validation
+            );
+            self.validation_cache.insert(
+                agent,
+                ValidationCacheData {
+                    entries: validation.source_chain_entries.clone(),
+                    headers: validation.source_chain_headers.as_ref().unwrap().clone(),
+                    cached_at: SystemTime::now(),
```
neonphog, line 145: Are we cool keeping these things cached forever? `cached_at` doesn't seem to ever be read.

zippy: I'm not persisting this data, so it will be cleared on reboot, which is fine I think because in that case the chain in the cache is probably pretty stale and will need to be fetched anyways.  I think he only real reason to dump the cache is because memory gets tight...

PR #2204, opened 2020-07-29
Fix holding-list/eavi mismatch bug for Update entries

zippy:
## PR summary

fixed bug in calculating entry address of a Content EntryAspect

## testing/benchmarking notes

## followups

## changelog

```markdown
```

## documentation

### DIFF
--- a/crates/core_types/src/network/entry_aspect.rs
+++ b/crates/core_types/src/network/entry_aspect.rs
@@ -97,12 +97,11 @@ impl EntryAspect {
     }
     /// NB: this is the inverse function of entry_to_meta_aspect,
     /// so it is very important that they agree!
+    /// NOTE: the ContentAspect address is always the entry address and this
+    /// is not used by entry_to_meta_aspect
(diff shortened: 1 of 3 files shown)


crates/core_types/src/network/entry_aspect.rs
```diff
@@ -97,12 +97,11 @@ impl EntryAspect {
...
     /// so it is very important that they agree!
+    /// NOTE: the ContentAspect address is always the entry address and this
+    /// is not used by entry_to_meta_aspect
     pub fn entry_address(&self) -> Result<Address, HolochainError> {
         Ok(match self {
-            EntryAspect::Content(_, header) => match header.link_update_delete() {
-                Some(ref updated_entry) => updated_entry.clone(),
-                None => header.entry_address().clone(),
-            },
+            EntryAspect::Content(_, header) => header.entry_address().clone(),
```
lucksus, line 104: Yes, for a `Content` aspect, the *base* (=`entry_address`) is always exactly the address of that content - even in the case of this entry being an update to another one.

